PR #2, opened 2026-04-07
fix: gate biome autofix on project config

ricardoraposo:
## Problem

   Biome autofix (and auto-install) runs on any JS/TS file regardless of whether the project actually uses Biome. This means projects that use ESLint or Prettier exclusively still
 get Biome installed and run on their files.

   The root cause is that the autofix path in `pipeline.ts` only checks `biomeClient.isSupportedFile()` (file extension match) before calling `ensureAvailable()`, which
 auto-installs Biome if missing. Unlike:

   - **ESLint autofix** — gated behind `hasEslintConfig()` (checks for config files)
   - **Biome autoformat** — gated behind `biomeFormatter.detect()` in `formatters.ts` (checks for `biome.json` / `devDependencies`)

   ## Fix

   Added `hasBiomeConfig(cwd)` that checks for:
   - `biome.json` / `biome.jsonc` in the project root
   - `@biomejs/biome` in `devDependencies` of `package.json`

   This mirrors the detection logic already used by `biomeFormatter.detect()` in `formatters.ts`.

   Gated both:
   1. The Biome autofix path (step 4) — prevents `ensureAvailable()` and `fixFileAsync()` from running
   2. The LSP sync deferral (step 3) — no point deferring LSP sync for a Biome autofix that won't run

### DIFF
--- a/clients/pipeline.ts
+++ b/clients/pipeline.ts
@@ -18,24 +18,30 @@ import type { BiomeClient } from "./biome-client.js";
 import { getDiagnosticLogger } from "./diagnostic-logger.js";
 import { getDiagnosticTracker } from "./diagnostic-tracker.js";
 import { dispatchLintWithResult } from "./dispatch/integration.js";
+import {
+	resolveRunnerPath,
+	toRunnerDisplayPath,
+} from "./dispatch/runner-context.js";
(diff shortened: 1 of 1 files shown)


apmantza:
@ricardoraposo many thanks for this. merged.

PR #3, opened 2026-04-08
perf: run startup scans asynchronously in background

ricardoraposo:
handleSessionStart was running 5 heavy scans sequentially, blocking the session from starting until all of them completed:

 1. TODO scan (walks entire source tree)
 2. Knip analysis (npx knip — dead code detection, 30s timeout)
 3. jscpd scan (npx jscpd — duplicate detection)
 4. ast-grep export scan (spawns sg scan across the project)
 5. Project index build (parses up to 500 TS files with the TypeScript compiler API)

 On a non-trivial project with cold caches, this could easily add 15-20 seconds to startup.

 All five scans now launch as fire-and-forget background promises. session_start returns immediately and the scans populate runtime state as they resolve. This is safe because
 every consumer already handles the "not ready yet" case:

 - cachedExports — guarded by runtime.cachedExports.size > 0
 - cachedProjectIndex — guarded by null + size check
 - Knip/jscpd caches — re-run at turn_end anyway; cache miss is the normal path
 - TODO baseline — only used for delta comparison on later turns

 Each background task has its own .catch() so failures are logged without crashing the session.

### DIFF
--- a/clients/runtime-session.ts
+++ b/clients/runtime-session.ts
@@ -6,6 +6,7 @@ import type { BiomeClient } from "./biome-client.js";
 import type { CacheManager } from "./cache-manager.js";
 import type { DependencyChecker } from "./dependency-checker.js";
 import { getDiagnosticTracker } from "./diagnostic-tracker.js";
+import { getKnipIgnorePatterns } from "./file-utils.js";
 import type { GoClient } from "./go-client.js";
 import type { JscpdClient } from "./jscpd-client.js";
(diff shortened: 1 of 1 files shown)


apmantza:
thanks again @ricardoraposo. really appreciate it.

I’ve opened a follow-up PR that keeps your change and adds:

-session-generation guards
-turn_end skip guards for overlapping knip/jscpd runs

Superseding with: https://github.com/apmantza/pi-lens/pull/5

PR #5, opened 2026-04-08
perf: async startup scans with session/turn-end safety guards

apmantza:
- keeps the async startup scan behavior from #3 but adds session-generation guards
- tracks startup scan in-flight status and skips overlapping knip/jscpd runs at turn_end
- centralizes background task lifecycle (mark in-flight, catch errors, clear in finally) in handleSessionStart


apmantza:
Credit to @ricardoraposo for the original async startup refactor — this follow-up only adds safety guardrails around that work

PR #11, opened 2026-04-09
fix(context): append injected messages instead of replacing event.messages

radge:
## Summary
- preserve existing `context` messages and append pi-lens guidance/findings instead of replacing the message list
- prevent the first user prompt from being dropped before provider requests
- document the failure mode in a comment near the handler

## Problem
The `context` handler was returning only injected messages:

```ts
return { messages };
```

In pi core, `context` is a transform hook over the full message list. Returning only the injected messages replaces `event.messages`, which can drop the user's first prompt. With OpenAI Responses models this can surface as:

```text
Error: One of "input" or "previous_response_id" or "prompt" or "conversation_id" must be provided.
```

## Fix
This change reads `event.messages` and returns:

```ts
{ messages: [...existingMessages, ...injectedMessages] }
```

## Validation
- `npm install`
- `npm run build`

## Note
This fix was generated using pi.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -734,20 +734,36 @@ pi.on("turn_end", async (_event, ctx) => {
 // --- Inject turn-end findings into next agent turn ---
 // jscpd, madge, and turn-end delta results are cached at turn_end and consumed here
 // via the context event, which fires before each provider request.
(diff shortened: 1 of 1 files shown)


apmantza:
thanks @radge. merged

PR #14, opened 2026-04-12
feat: lazy-load bootstrap clients

anh-chu:
Lazy-load heavy pi-lens clients via shared bootstrap promise to reduce startup work. Includes fallback for architect client dep resolution and keeps behavior same after first use.


apmantza:
thanks @anh-chu. appreciate the PR. merged.

anh-chu:
it's working, but I am still ironing out some bugs

PR #19, opened 2026-04-19
fix: unfreeze session_start by running tool probes asynchronously

Roman-Galeev:
## Problem

`session_start` blocks the Node event loop for ~8–10 s on every pi startup, making pi visibly unresponsive for several seconds while the extension boots.

The root cause is the `setImmediate` warm-up block near the top of `handleSessionStart` (`clients/runtime-session.ts`):

```ts
setImmediate(() => {
  const b = biomeClient.isAvailable();
  const a = astGrepClient.isAvailable();
  const r = ruffClient.isAvailable();
  knipClient.isAvailable();
  depChecker.isAvailable();
  jscpdClient.isAvailable();
  typeCoverageClient.isAvailable();
  ...
});
```

Each of those `isAvailable()` calls is backed by `spawnSync` (the `@deprecated` `safeSpawn`), and six of the seven probes fall back to `npx <tool> --version`. Measured per probe in a TS project with no local linters installed:

| Probe | Command it actually runs | Time |
|---|---|---|
| `biomeClient.isAvailable()` | `npx @biomejs/biome --version` | ~1.5 s |
| `astGrepClient.isAvailable()` | `sg --version` (ENOENT) → `npx --no sg --version` | ~0.25 s |
| `ruffClient.isAvailable()` | `ruff --version` | ~0.05 s |
| `knipClient.isAvailable()` | `npx knip --version` | ~1.8 s |
| `depChecker.isAvailable()` | `npx madge --version` | ~2.1 s |
| `jscpdClient.isAvailable()` | `npx jscpd --version` | ~1.8 s |
| `typeCoverageClient.isAvailable()` | `npx type-coverage --version` | ~2.2 s |
| **Total sync blocking** | | **≈ 9.7 s** |

The existing inline comment already notes that these were "blocking the critical startup path with no functional benefit" — but `setImmediate` callbacks run on the main thread, so the `spawnSync` chain still blocks the loop a tick later. `session_start total` in `~/.pi-lens/sessionstart.log` matches: 8302 ms before this patch.

## Fix

1. **`clients/runtime-session.ts`** — replace the blocking `setImmediate` block with a fire-and-forget async IIFE that calls each client's existing async `ensureAvailable()` in parallel via `Promise.all` / `Promise.allSettled`. No main-thread blocking. `typeCoverageClient` is dropped from the warm-up (no async probe; only `/lens-booboo` uses it, which probes lazily).

2. **`clients/biome-client.ts`** — extend `getBiomeBinary()` to also check `~/.pi-lens/tools/node_modules/.bin/biome` (the path `ensureTool("biome")` writes to). Without this, biome's own `ensureAvailable` still sync-spawned `npx @biomejs/biome --version` (~1.5 s) even when the binary was already installed.

3. **Test mocks** — `tests/clients/runtime-session.test.ts` and `tests/clients/runtime-event-flow.test.ts` now include `ensureAvailable` stubs on `biomeClient` / `ruffClient` / `depChecker` to match the new call surface.

## Results

`session_start total` on macOS / pi-pipe workspace, three back-to-back runs:

| | Before | After |
|---|---|---|
| run 1 | — | 183 ms |
| run 2 | — | 172 ms |
| run 3 | — | 174 ms |
| **Earlier baseline** | **8302 ms** | — |

~**47×** faster on the blocking path. Probes still run (now in the background, ~1.9 s total) and populate each client's cache, so the first real file write incurs no extra sync probe cost.

## Behaviour note worth calling out

`ensureAvailable()` on `knip` / `jscpd` / `madge` triggers auto-install into `~/.pi-lens/tools` when the tool is missing in `PATH`. This matches the existing behaviour of `firePreinstallDefaults` (which already auto-installs `typescript-language-server` and `biome` at session_start). If a more conservative probe is preferred, an alternative is to add a dedicated `isAvailableAsync()` on each client that only probes without the install fallback — happy to switch to that if you'd rather.

## CI / gate

- `npx tsc --noEmit` — clean.
- `npx vitest run tests/clients/runtime-session.test.ts tests/clients/runtime-event-flow.test.ts` — 3/3 pass.
- `npx vitest run` — 479 pass, 2 skip, 4 pre-existing failures in `tests/clients/lsp/server-policy.test.ts` (reproduce on unmodified `upstream/master`; unrelated to this PR).
- `node scripts/check-extensions.mjs` — all 609 imports resolve.


apmantza:
thanks @Roman-Galeev. appreciate it. merged.

PR #41, opened 2026-04-30
Refactor tsconfig fix type errors

silvanshade:
This PR adjusts the typescript configuration to allow proper import resolution and type-checking while editing the tests.

For this, I split the `tsconfig.json` into a general purpose one and one specifically for the build with `tsconfig.build.json`.

Then I fixed surfaced type-errors and lints.


apmantza:
appreciate it @silvanshade! thanks for the cleanup!

PR #42, opened 2026-04-30
Add missing clangd extensions and warmFiles LSP config option

silvanshade:
This PR adds missing extensions for `clangd`, along with some extension-related cleanup and refactoring.

Added a new "warmFiles" LSP config option: this allows to specify files which should be loaded at session start.

This option provides a solution to the problem of interfacing with language-servers that only perform background or on-demand indexing of translation-units, since they may return incomplete or partial results depending on whether they have encountered the correct TUs already.

For example, with `clangd`, you would specify the entry point to your application or library this way:

```json5
// .pi-lens/lsp.json
{
    warmFiles: ["src/main.cpp", "src/lib.cpp"]
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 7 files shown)


apmantza:
thanks @silvanshade; well thought. merged. small concerns which will be adressed later.

PR #44, opened 2026-05-01
Localize lens diagnostic status commands

jerryfan:
I found a small UX issue in the diagnostic/status path: pi-lens has dense command output exactly where users decide whether blockers are real, whether tools are installed, and whether read-guard is armed. If those messages are misread, users can chase the wrong diagnostic state or bypass the wrong guard.

This PR makes that command/status path optionally localizable while preserving the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-TW, de — common developer locales for diagnostics/config UI
- Covers LSP status, /lens-tdi, /lens-health, /lens-tools, and /lens-allow-edit copy

Validation:
- npm install --ignore-scripts
- npm run build
- npm pack --dry-run

Note: I accidentally tried 
> pi-lens@3.8.33 test
> vitest run --runInBand; Vitest does not support that Jest flag, so that command failed before running tests. I did not change test code.


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #47, opened 2026-05-02
Respect PILENS_DATA_DIR for rule cache

tifandotme:
## Summary

- Store tree-sitter rule cache files under `getProjectDataDir(rootDir)` instead of `<cwd>/.pi-lens/cache`.
- Add a regression test that confirms `RuleCache` respects `PILENS_DATA_DIR` and does not create a project-local `.pi-lens` directory.

## Tests

- `npm run lint`
- `npm test -- tests/clients/file-utils.test.ts`


apmantza:
thanks @tifandotme, merged.

PR #48, opened 2026-05-02
fix: preserve user prompt after context injection

tifandotme:
## Why

Some provider bridges treat the final context message as the active user action. When pi-lens appends session guidance after the user's prompt, those bridges can receive the guidance as the prompt and demote the real request to earlier context.

## Context

This keeps pi-lens guidance in the provider context but prepends it before the existing messages. The user's submitted prompt stays last, which matches bridge expectations while preserving the guidance.

## Testing

- `npm run lint`
- `npm run build`
- `npm test -- tests/index-integration.test.ts`
- `npm test`


apmantza:
thanks again @tifandotme.

PR #65, opened 2026-05-11
fix(knip): async analyze, in-flight guard, and bail when no project root

rfxlamia:
## Summary

Knip was spawned **synchronously** from `handleTurnEnd` via `safeSpawn` (a `spawnSync` wrapper the file itself flags as deprecated and event-loop-blocking). On top of that, nothing guarded against concurrent runs, and `resolveProjectRoot()` silently fell back to the caller's `cwd` when no project marker was found. Combined, this reliably caused:

- Two concurrent `npx knip --reporter=json --include files,exports,types,dependencies,unlisted` processes each pegging a CPU core to 100%, load avg >5, TUI frozen for the duration of the scan.
- Zombie `knip` processes reparented to `systemd --user` after pi exited mid-scan.
- From a bare `cwd` like `$HOME`, knip recursing through every sibling project under home before timing out at 30s — the original symptom the user reported (`htop` showing two top processes being `node .../.bin/knip`, load 5.14, both CPUs saturated).

Reproducible by running pi from a directory without a `package.json` while pi-lens is enabled: `session_start` spawns a knip background task, then the first `turn_end` fires a second one before the first finishes.

## Root causes (all three fixed)

1. `KnipClient.analyze()` used `safeSpawn` (sync). See [`safe-spawn.ts` itself](./clients/safe-spawn.ts): *"⚠️ DEPRECATED: Use safeSpawnAsync instead. This blocks the entire Node.js event loop until the process exits. If the process hangs, pi will freeze."*
2. `handleTurnEnd` is async and re-entrant. Without an in-flight guard in `KnipClient`, two back-to-back turn_ends, or a turn_end during a session_start scan, spawned two parallel `npx knip` processes over the same tree.
3. `resolveProjectRoot()` returned `startDir` as a fallback when no `package.json` / `knip.*` marker was found, so knip ran from arbitrary directories and walked huge unrelated trees.

## Fix

- Migrate `KnipClient.analyze()` and `findUnusedExports()` to `async` and `safeSpawnAsync`.
- Extract the spawn into `runAnalyze()` and add `inFlight: Map<projectRoot, Promise<KnipResult>>`. Concurrent `analyze()` calls against the same root share a single process; the slot clears when the run settles. A subsequent call after completion spawns fresh as expected.
- Return `null` from `resolveProjectRoot()` when no marker is found, and short-circuit `analyze()` with a `{ success: true, summary: "No project root found; knip skipped" }` result in that case. Added a safety-bound loop cap (64) to guard against symlink loops.
- Remove the unused legacy sync `isAvailable()` method (only caller was `analyze()` itself; `ensureAvailable()` is already the supported async path with auto-install).

## Call-site updates

- `clients/runtime-turn.ts`: `await knipClient.analyze(...)`. Refreshed `cacheManager.readCache<T>()` generic to use `KnipResult` directly (was `ReturnType<KnipClient['analyze']>`, now `Promise<KnipResult>`).
- `clients/runtime-session.ts`: same treatment for the session_start background task.
- `commands/booboo.ts`: `await clients.knip.analyze(...)` in the dead-code runner.

## Tests

`tests/clients/knip-client.test.ts`:

- **Existing** `resolveProjectRoot` test updated for the nullable return type.
- **New** `resolveProjectRoot` returns something other than `startDir` when no markers exist up the tree (catches the old buggy fallback).
- **New** `analyze()` short-circuits with no spawn when no project root is found (regression guard for `$HOME` scans).
- **New** `analyze()` returns a Promise at runtime — the TS signature alone doesn't catch someone regressing to `spawnSync`.
- **New** Concurrent `analyze()` for the same root runs exactly one spawn and both callers receive the same result; a later call after completion spawns fresh.

## Verification

```bash
npx tsc --project tsconfig.json --noEmit   # clean
npx tsc --project tsconfig.build.json       # clean
npx vitest run tests/clients/knip-client.test.ts  # 6/6 pass
npx vitest run tests/clients/session-start-parallelism.test.ts \
               tests/clients/runtime-turn-session.test.ts \
               tests/clients/runtime-session.test.ts  # 22/22 pass
```

Full suite on `master` already had 3 pre-existing failures (`tests/clients/formatters.test.ts` venv-fallback, two `tests/clients/installer/tool-discovery.test.ts` timeouts) — unrelated to this change; confirmed by stashing the diff and re-running on untouched `master`.

## Notes for reviewers

- API is now async: `analyze()` and `findUnusedExports()` return Promises. All callers in this repo are updated.
- The in-flight map is unbounded in theory but keyed by canonical project root (pi-lens analyses at most a handful of roots per session), and entries always clear via `.finally()`.

### DIFF
--- a/clients/knip-client.ts
+++ b/clients/knip-client.ts
@@ -11,7 +11,7 @@
 
 import * as fs from "node:fs";
 import * as path from "node:path";
-import { safeSpawn, safeSpawnAsync } from "./safe-spawn.js";
+import { safeSpawnAsync } from "./safe-spawn.js";
 
 // --- Types ---
 
@@ -33,11 +33,37 @@ export interface KnipResult {
 	summary: string;
 }
 
+const EMPTY_RESULT: Omit<KnipResult, "summary"> = {
+	success: false,
+	issues: [],
+	unusedExports: [],
+	unusedFiles: [],
+	unusedDeps: [],
+	unlistedDeps: [],
+};
+
+const ANALYSIS_TIMEOUT_MS = 30_000;
+
 // --- Client ---
 
 export class KnipClient {
 	private knipAvailable: boolean | null = null;
 	private log: (msg: string) => void;
+
+	/**
+	 * De-dupe concurrent `analyze()` calls against the same project root.
+	 *
+	 * Without this guard, two back-to-back turn_end events (or a turn_end
+	 * firing while the session_start scan is still in flight) can each spawn
+	 * a fresh `npx knip` process over the same tree. Two concurrent knip
+	 * runs are CPU-bound and cause the exact pathology we're fixing: load
+	 * averages >5, TUI freezes, and zombie processes reparented to init
+	 * after pi exits mid-scan.
+	 *
+	 * Key: canonicalised project root (not the caller's cwd). Value is the
+	 * in-flight promise; completing clears the slot.
+	 */
+	private inFlight = new Map<string, Promise<KnipResult>>();
 
 	constructor(verbose = false) {
 		this.log = verbose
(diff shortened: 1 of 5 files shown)


apmantza:
many thanks @rfxlamia. merged.

Issue #78, opened 2026-05-14
False positive: regex with /g flag flagged as unused-symbol when used via .exec / .lastIndex


av1155:
pi-lens v3.8.44 reports an unused-symbol diagnostic on regex literals declared at module scope and consumed only through `.exec()` and `.lastIndex`, even when there are clear references in the same file.

## Reproducer

In a `.ts` file at any project root that pi-lens scans:

```ts
const SAME_LINE_SKILL_TOKEN = /\/(?:skill:)?[a-z0-9](?:[a-z0-9-]*[a-z0-9])?/g;

function rewrite(line: string): string {
  SAME_LINE_SKILL_TOKEN.lastIndex = 0;
  let match: RegExpExecArray | null;
  while ((match = SAME_LINE_SKILL_TOKEN.exec(line))) {
    // ...
  }
  return line;
}
```

`SAME_LINE_SKILL_TOKEN` is referenced at lines 4 (`.lastIndex`) and 6 (`.exec`), so it is plainly used. pi-lens still surfaces it as unused.

## Root cause guess

The static-analysis path appears to look for direct identifier references in expression contexts but not for property accesses on the regex object. `.exec(input)` and `.lastIndex = 0` both bind the identifier through a property-access node, which the current rule pattern probably does not match.

A focused tree-sitter query that matches `member_expression` with the regex identifier as the object, in addition to plain identifier references, should close the gap. The dispatch point appears to be `clients/knip-client.ts`; the underlying rule likely lives in `clients/dispatch/runners/` or the ast-grep rule set.

## Repro environment

- pi-lens 3.8.44
- node 24.15.0 on macOS 14
- tree-sitter and ast-grep versions per the `^0.42.1` ranges in `package.json`

## Audit context

This was surfaced during a routine `semgrep` + `opengrep` + `madge` + `knip` audit of the third-party Pi extensions installed on this machine. Other findings worth flagging while I have your attention:

1. `osv-scanner` reports a malicious-package alert `MAL-2026-3432` against `@mistralai/mistralai@2.2.1`. It reaches pi-lens transitively via `@earendil-works/pi-coding-agent -> @earendil-works/pi-ai`. Fix needs to land upstream in `@earendil-works/pi-ai`, not pi-lens.

[...]

PR #79, opened 2026-05-14
test(dispatch): regression test for #78 (unused-symbol on regex literal)

av1155:
Refs #78.

The issue report describes pi-lens flagging a module-scope regex literal (`SAME_LINE_SKILL_TOKEN`) as unused even though the symbol is referenced later via property-access expressions on the regex (matcher invocations and `.lastIndex` resets).

I could not pin down which runner emits the false-positive `Diagnostic` against pi-lens 3.8.44 in the time I had. The classifier that consumes the diagnostics for promotion to blocker (`isUnusedValueDiagnostic` in `clients/dispatch/dispatcher.ts`, used by `promoteDeltaUnusedToBlockers`) keys off the diagnostic message, id, and rule string. It has no view into whether the actual source file uses the symbol via member access.

This PR adds a regression test (`tests/clients/dispatch/inline-skill-token-regression.test.ts`) that pins the classifier's behaviour for four shapes:

- TS6133 (`is declared but its value is never read`) is recognised as unused.
- biome/eslint `noUnusedVariables` is recognised.
- A benign info diagnostic that mentions a regex symbol but does not declare it unused is **not** classified as unused.
- Rule ids that contain the word `use` (for example `unsafe-regex`) are **not** misclassified as unused.

The test does not yet assert a behaviour change; it captures the current classifier surface so a future fix that filters property-access usage out of unused-symbol findings can be added with a concrete before/after pair.

## Next step for the maintainer

A complete fix likely needs the dispatcher to read the file at the diagnostic's `filePath` and look for `<symbol>.<prop>` references in the source AST. If any exist, skip the promotion in `promoteDeltaUnusedToBlockers`. Happy to follow up once the upstream view on the right scope (file-local guard vs. configurable allow-list) is settled.

## Test run

```
$ npx vitest run tests/clients/dispatch/inline-skill-token-regression.test.ts
 Test Files  1 passed (1)
      Tests  4 passed (4)
```

### DIFF
--- a/clients/dispatch/dispatcher.ts
+++ b/clients/dispatch/dispatcher.ts
@@ -287,7 +287,7 @@ function suppressLintOverlapsWithLsp(diagnostics: Diagnostic[]): Diagnostic[] {
 	});
 }
 
-function isUnusedValueDiagnostic(d: Diagnostic): boolean {
+export function isUnusedValueDiagnostic(d: Diagnostic): boolean {
 	const raw = `${d.id ?? ""} ${d.rule ?? ""} ${d.message ?? ""}`.toLowerCase();
 	if (raw.includes("no-unused")) return true;
 	if (/\b(6133|6192|6196)\b/.test(raw)) return true;
--- a/tests/clients/dispatch/inline-skill-token-regression.test.ts
+++ b/tests/clients/dispatch/inline-skill-token-regression.test.ts
@@ -1,0 +1,81 @@
+/**
+ * Regression test for apmantza/pi-lens#78.
+ *
+ * The original report describes a false positive where pi-lens classifies a
+ * module-scope regex literal as unused, even though the symbol is referenced
+ * later in the same file through property-access expressions on the regex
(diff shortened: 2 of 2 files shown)


av1155:
Closing this on round-2 adversarial review.

The PR pins the `isUnusedValueDiagnostic` classifier in `dispatcher.ts` but the actual #78 regression sits upstream of that classifier (whichever runner is emitting the TS6133-shaped `Diagnostic` for a regex that has property-access usage). The classifier correctly identifies a *true* TS6133 as unused; the bug is that the producer should not be emitting one in the first place.

A useful regression test for #78 would either:
- patch the producing runner (`knip-client.ts` or `type-safety.ts` or wherever TS6133 is collected) to recognise `<symbol>.<prop>` usage on regex literals; or
- be folded into a fix PR that does so.

The test in this PR plus the export of `isUnusedValueDiagnostic` would widen the public-API surface for no behaviour change, so I am withdrawing the PR. Issue #78 stays open with the reproducer.

PR #81, opened 2026-05-14
Fix Knip scans escaping non-JS repo roots

fabio-dee:
## Summary
- stop Knip project-root discovery at VCS boundaries so nested non-JS repos do not inherit unrelated parent `package.json` files
- skip Knip probing/installing until `KnipClient.analyze()` confirms there is a real Knip/JS project root
- run the resolved `knip` binary directly with local/managed tool paths instead of invoking `npx knip`
- add turn-end backoff after recent Knip timeout/kill failures to avoid repeatedly spending 30s on the same broken scan

## Context
I hit this in a Unity/C# repo with no root `package.json`. The old upward search walked past the repo's `.git`, found a `~/package.json`, and repeatedly launched `npm exec knip ...` from the wrong root. pi-lens logged repeated 30s Knip timeouts, and macOS recorded a Node `SIGABRT` crash for `npm exec knip --reporter=json ...`.

## Tests
- `npm run lint -- --pretty false`
- `npm test -- tests/clients/knip-client.test.ts tests/clients/runtime-session.test.ts tests/clients/runtime-turn-session.test.ts tests/clients/cascade-turn-merge.test.ts tests/clients/runtime-event-flow.test.ts`
- `npm run build && npm test`

### DIFF
--- a/clients/knip-client.ts
+++ b/clients/knip-client.ts
@@ -10,6 +10,7 @@
  */
 
 import * as fs from "node:fs";
+import * as os from "node:os";
 import * as path from "node:path";
 import { safeSpawnAsync } from "./safe-spawn.js";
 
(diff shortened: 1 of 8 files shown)


apmantza:
thanks @fabio-dee. merged.

PR #90, opened 2026-05-20
tool-policy: stop applying prettier to .md/.mdx without project config

JayceFreeman:
Prettier's markdown defaults reflow lines, normalize emphasis markers
(* -> _), and restyle lists. Users without an explicit prettier config
were getting noisy formatter diffs on every doc-only write.

Flip .md and .mdx defaultWhenUnconfigured from true to false on the
smart-default gate. The gate still honours an explicit project prettier
config (.prettierrc, prettier field in package.json, etc.) — the flip
only affects the no-config path.

To restore prior behaviour, add an empty .prettierrc to the project root.

Tests:
- tests/clients/tool-policy.test.ts: move .md out of the smart-default
  assertion; add a focused negative test plus policy-shape assertion
  for .md and .mdx.
- tests/clients/formatters.test.ts: replace
  "uses prettier as the smart default for unconfigured Markdown" with
  a negative case plus a positive case asserting that an explicit
  .prettierrc re-enables prettier (opt-in path).

CHANGELOG.md updated under Unreleased / Changed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


apmantza:
thanks @JayceFreeman. merged.

PR #105, opened 2026-05-26
fix: use workspace root for turn state in monorepos

JulienEllie:
## Problem

In monorepos with nested language roots (Go modules, Rust crates, Java subprojects), `actionable-warnings.json` is never generated.

### Root cause

`handleToolResult` in `runtime-tool-result.ts` uses `resolveLanguageRootForFile()` to resolve `cwd`, which walks up to find language markers like `go.mod`. This resolved cwd is used for **both** pipeline dispatch (correct) and turn state tracking via `addModifiedRange()` (incorrect).

At `turn_end`, the turn state is read from `ctxCwd` — the **workspace root**. Because the two cwds differ, `turn_end` reads a different `turn-state.json`, sees an empty `files` map, returns early at line 116, and never reaches the actionable-warnings phase.

| Phase | cwd source | Resolves to |
|---|---|---|
| `tool_result` | `resolveLanguageRootForFile()` | e.g. `.../platform/svc/go/daemon/` |
| `turn_end` | `ctxCwd ?? process.cwd()` | workspace root |

### Fix

Split the single `cwd` into two variables:
- **`dispatchCwd`** — language root, used for `runPipeline()` and `deferFormat()`
- **`turnStateCwd`** — workspace root, used for `addModifiedRange()` and `recordProjectChange()`

### Tests

Two regression tests added:
1. Verifies turn state is written under the workspace root (not the nested language root)
2. Verifies the pipeline still receives the language root as cwd for linting

All 1122 existing tests continue to pass.


apmantza:
thanks @JulienEllie. appreciate it. extended this fix a bit further.

PR #146, opened 2026-06-03
fix(widget): stop warning storm churn

trvon:
I found this while using pi-lens on a large codebase. When an edit kicked off a warning-heavy pass, the widget below the chat box would churn through transient states and the pi harness rendering looked broken.

## Problem

Warning-only analysis could briefly render the widget as `✓ clean` before the final warning totals landed.

### Root cause

`widget-state.ts` requested a render on every `recordRunner()` update, but files with runner activity and no final `recordDiagnostics()` snapshot were still treated as clean.

That meant a warning-heavy pass could render a sequence like:
- `LSP↑`
- header/file rows with no diagnostics yet
- transient `✓ clean`
- final `!40W`

The earlier per-file diagnostic cap kept stored warning payloads smaller, but it did not address this transient clean-frame churn.

## Fix

- track whether a file has received a final diagnostics snapshot
- suppress `✓ clean` while any file is still pending final diagnostics
- keep runner-only pending files out of the visible file rows until diagnostics arrive
- preserve capped per-file stored diagnostics so large warning bursts stay compact without losing counts

## Tests

Added/updated widget regressions to cover both:
1. capped per-file stored diagnostics with preserved warning counts
2. the real warning-storm C++ frame churn case (`✓ clean` must not appear before final `!40W`)

### DIFF
--- a/clients/widget-state.ts
+++ b/clients/widget-state.ts
@@ -32,6 +32,12 @@ interface FileRecord {
 	runners: Map<string, { status: string; count: number; durationMs?: number }>;
 	formatters: Map<string, { changed: boolean; success: boolean }>;
 	diagnostics: WidgetDiagnostic[];
+	diagnosticCounts: {
+		blocking: number;
+		errors: number;
+		warnings: number;
+	};
+	hasFinalDiagnosticsSnapshot: boolean;
 	touchedAt: number;
 }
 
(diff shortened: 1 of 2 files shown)


apmantza:
thanks @trvon. merged with a minor patch.

PR #161, opened 2026-06-05
Add all supported file extensions for oxfmt

shaDmx:
##  Problem

oxfmt extension list in clients/formatters.ts was JS/TS-only (8 extensions). Projects using oxfmt for CSS, HTML, JSON, Markdown, GraphQL etc. got no formatting benefit.

###  Cause
The original oxfmtFormatter missed some supported file types. 

 ### Fix
 - Added missing file types supported by oxfmt. (reference https://oxc.rs/compatibility.html)
 - Updated tool-policy.ts entries for all those extensions to include "oxfmt" (behind biome/prettier/etc.)
 - Added tests verifying oxfmt is selected for every new extension when oxfmt.toml or .oxfmtrc.json is present

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


apmantza:
thanks @shaDmx. merged after some small cleanup commits.

PR #168, opened 2026-06-07
fix(read-guard): register reads from bash/grep/find/sed/cat commands

loss-and-quick:
Parses tool_call input for bash, grep, and find tools to extract file paths that were explicitly read. Registers them with ReadGuard so subsequent edit calls on the same files are not blocked.


apmantza:
thanks @loss-and-quick. PR is valid.

some cleanup+consistency commits from me to continue fitting the read-guard contract. merged.


+ new follow-up issue on how to implement a grep/ast_grep/lsp_nav routine: https://github.com/apmantza/pi-lens/issues/169

PR #188, opened 2026-06-09
perf(session_start): eliminate TUI input freeze on cold boot and /new

amit-gshe:
Symptom: after `pi` boot and after `/new`, the input box rejects keystrokes for 3-6s on a 2k-file project. Cold boot is worst because both the `pi-lens` jiti transpile (~3.5s) and the full session_start scan run back-to-back on the same event-loop tick.

Root cause: the synchronous portion of `handleSessionStart` runs three O(N) walks (resolveStartupScanContext, detectProjectLanguageProfile, scheduleStartupScans → todo / call-graph) without yielding, all of which call into ignoreMatcher.isIgnored() in their hot loops. Each session_start (boot, /new, --print) repeats them from scratch, and the stdin macrotask queue is starved the entire time.

Fix (four files, all required):

  1. clients/file-utils.ts Per-matcher path → boolean memo inside the ignoreMatcher closure. The matcher itself is already mtime-cached, so the memo is dropped automatically when .gitignore changes. Cuts the bg-scan ignore cost from O(ancestors × patterns) per file to O(1) on repeat. Verified impact: warmup language-profile drops from ~3000ms to 19ms because it reuses the matcher memo populated by warmup scan-context.

  2. clients/startup-scan.ts
     - Process-lifetime memo on resolveStartupScanContext keyed by (cwd, homeDir, maxSourceFiles). Subsequent session_starts in the same process see a hit.
     - countSourceFilesWithinLimitAsync + resolveStartupScanContextAsync: chunked-yield variants for the warmup pipeline. Same memo as the sync siblings, so whichever runs first warms the cache.

  3. clients/language-profile.ts - Process-lifetime memo on detectProjectLanguageProfile keyed by projectRoot. Skipped when caller passes explicit sourceFiles so the no-arg call site keeps a single coherent cache. - detectProjectLanguageProfileAsync + collectSourceFilesForWarmup: async variant for the warmup pipeline. Reuses the sync detector once files are pre-collected (constant-time file-marker probes only at that point).

  4. clients/runtime-session.ts
     - Defer table for background scans (`call-graph`, `codebase-model`, `ast-grep exports`, `project index`): 5-5.4s setTimeout instead of setImmediate, pushing CPU-heavy work past the typing window into LLM-streaming idle. Without this, those tasks start on the next macrotask and block stdin for 3-4s each.
     - todo task rewritten to enumerate files via getSourceFiles and call todoScanner.scanFile per file with a setImmediate yield every 30 files. Falls back to the blocking scanDirectory if the scanner shape changes. Without this, the post-/new todo scan blocks stdin for ~3s. - Cold-start mitigation in handleSessionStart: the very first session_start of the process is forced to "quick" mode, then a 2-second-delayed background warmup runs the async memo-filling walks. The user's first /new sees both caches hit and finishes in <50ms.

Env knobs:
  PI_LENS_COLD_START_QUICK=0  disable cold-start quick + warmup
  PI_LENS_WARMUP_DELAY_MS=N   override the 2000ms warmup delay
  PI_LENS_STARTUP_MODE=...    explicit user override wins over auto-quick

Verified on a 1832-file project (claude-code-hub):
  - session_start total: 3ms (was 3000-6000ms)
  - warmup: scan-context done in 3683ms (background, async)
  - warmup: language-profile done in 19ms (reuses ignoreMatcher memo)
  - subsequent /new: cache hit on both walks
  - tsc --project tsconfig.json: 0 errors
  - npm run build:dist: passes

### DIFF
--- a/clients/file-utils.ts
+++ b/clients/file-utils.ts
@@ -279,8 +279,8 @@ function buildProjectIgnoreMatcher(
 	// `getProjectIgnoreMatcher` keyed on `.gitignore` mtime, so this Map's
 	// lifetime is bounded to a single set of ignore rules — when any
 	// `.gitignore` changes, the matcher is rebuilt and the memo is dropped
-	// with it. Without this memo, every background scan (todo, knip, jscpd,
-	// call-graph, source-filter, pipeline) recomputes O(ancestorDirs ×
+	// with it. Without this memo, every background scan (comment scan, knip,
+	// jscpd, call-graph, source-filter, pipeline) recomputes O(ancestorDirs ×
 	// patterns) per file, multiplying into 2-3s of pure CPU on a 2k-file
 	// project. With it, the second visitor of the same path is O(1).
 	const isIgnoredMemo = new Map<string, boolean>();
--- a/clients/file-utils.ts
+++ b/clients/file-utils.ts
@@ -275,17 +275,33 @@ function buildProjectIgnoreMatcher(
 		return nextPatterns;
 	};
 
+	// Per-matcher path → boolean memo. The matcher itself is cached by
+	// `getProjectIgnoreMatcher` keyed on `.gitignore` mtime, so this Map's
+	// lifetime is bounded to a single set of ignore rules — when any
+	// `.gitignore` changes, the matcher is rebuilt and the memo is dropped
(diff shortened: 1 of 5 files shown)


apmantza:
thanks @amit-gshe. merged after some small cleanup.

PR #246, opened 2026-06-18
feat(config): honor project-level `.pi-lens.json` `ignore` and `rules` fields

greg-hass:
## Problem

pi-lens already walks up from cwd looking for a `.pi-lens.json` to read LSP server config (see `lsp/config.ts` — `loadLSPConfig`), but the **same file's `ignore` and `rules` fields** were parsed and discarded. They documented the right shape in user-facing config samples, but no scanner actually read them. Result: a project that wanted per-scope control over what got scanned, or per-project thresholds for fact-rules, had to either patch `node_modules/` or live with the global skip lists.

This PR closes the gap by wiring the two existing-but-unhonored fields into the actual scanners, with no new schema surface and zero behavioral change for projects without the file.

### `ignore` — gitignore-style glob patterns

Added to every diagnostic scan: LSP walk, fact-rules, tree-sitter, jscpd, knip, review graph, source-filter. Useful for vendored code, generated files, or per-project noise you want to silence without editing `.gitignore` (which would also affect git itself).

```json
{ "ignore": ["**/__tests__/**", "fixtures/**", "vendor/**"] }
```

### `rules` — per-rule threshold overrides

The two fact-rules whose thresholds were hardcoded `const`s now read from config:

| Rule | Default | Config |
|---|---|---|
| `high-complexity` (cyclomatic complexity) | 15 | `rules["high-complexity"].threshold` |
| `high-fan-out` (distinct function calls) | 20 | `rules["high-fan-out"].threshold` |

```json
{
  "rules": {
    "high-complexity": { "threshold": 25 },
    "high-fan-out": { "threshold": 30 }
  }
}
```

## How it works

- **New `clients/project-lens-config.ts`** — single-purpose loader. mtime-cached parsing, walks up to find `.pi-lens.json` / `pi-lens.json`, tolerates malformed JSON (silently treated as "no config" — your diagnostics never get blocked by a syntax error in your own file), ignores unknown top-level keys and unknown rule ids for forward compat.
- **`getProjectIgnoreMatcher`** (`clients/file-utils.ts`) now passes the loaded `ignore` patterns into the existing `createProjectIgnoreMatcher` extension point. The cache is invalidated by `.pi-lens.json` mtime as well as `.gitignore` mtime — editing the file takes effect on the next scan without restarting the agent.
- **`high-complexity` and `high-fan-out`** (`clients/dispatch/rules/`) had hardcoded `const CC_THRESHOLD = 15` / `const FAN_OUT_THRESHOLD = 20`. Converted to `let` with `setHighComplexityThresholds(cc, depth)` / `setHighFanOutThreshold(n)` setters, plus `reset…` helpers for tests.
- **New `applyProjectLensConfig(cwd)`** export in `dispatch/integration.ts` loads the config and applies the setters. `resetDispatchBaselines` now takes an optional `cwd` and invokes it (backward-compatible with the two existing tests that pass no cwd).
- **`runtime-session.ts` and `mcp/session.ts`** both pass `cwd` through on session start so the config is applied automatically — no agent changes needed.

## Why it's safe

- **Zero behavior change for projects without the file** — defaults match the historical hardcoded values exactly.
- **Backward-compatible** — `resetDispatchBaselines`'s `cwd` param is optional.
- **Mtime-cached** — re-applying within a session is essentially free (one stat + one map lookup).
- **Tolerant parser** — malformed JSON, non-object roots, non-string ignore entries, non-finite thresholds, missing-threshold rule entries, and unknown rule ids are all handled without throwing.

## Test plan

5 new test files, 39 new tests, all green:

- `tests/clients/project-lens-config.test.ts` — loader unit tests (15 tests): empty config, dotfile vs no-dotfile preference, parent walk, mtime invalidation, malformed JSON, non-object root, non-string ignore entries, non-finite thresholds, fs-root walk doesn't infinite-loop, raw JSON exposed for forward-compat, etc.
- `tests/clients/project-lens-ignore.test.ts` — integration via `collectSourceFiles` (6 tests): `getProjectIgnoreMatcher` honors `.pi-lens.json`, still combines with `.gitignore`, returns clean matcher when no config exists, cache invalidates on mtime change, end-to-end with `collectSourceFiles` skipping a project-ignored dir.
- `tests/clients/dispatch/rules/high-complexity-threshold.test.ts` (6 tests) + `high-fan-out-threshold.test.ts` (6 tests) — rule threshold setters: default, override lower, override higher, depth override independent of CC, noise filters still apply, reset restores default.
- `tests/clients/dispatch/project-lens-config-integration.test.ts` (6 tests) — `applyProjectLensConfig` end-to-end: no-op when no file, applies both thresholds from a single file, ignores unknown rule ids, reset restores defaults.

**Full suite**: 1999/2001 pass. The 2 failures are pre-existing timing-sensitive tests in `runner-timeout.test.ts` unrelated to this change (verified against master `df312d4`).

## Docs

- **README.md** — new "Project Config" section next to "Global Config" with schema, behavior, and forward-compat notes.
- **CHANGELOG.md** — entry under [Unreleased] / Added.
- **AGENTS.md** — note that `getProjectIgnoreMatcher` cache now also keys on `.pi-lens.json` mtime.


greg-hass:
Heads up: the **CI** check is showing `action_required` — that's GitHub's first-time-fork-PR workflow approval gate, not a failure. Maintainer needs to click **"Approve and run"** on the workflow run to let it proceed: https://github.com/apmantza/pi-lens/actions/runs/27764707846

SonarCloud Code Analysis already passed (25s ✅). Locally I ran the full vitest suite end-to-end (with grammars downloaded): **1999/2001 pass** — the 2 failures are pre-existing timing-sensitive tests in `runner-timeout.test.ts`, verified against `df312d4` on master. All 39 new tests in the 5 new test files are green.

greg-hass:
## Why this PR is small even though the gap is big

The temptation with a "dead code" fix is to rip it out and replace it with a new system. I deliberately did **not** do that here. Why:

1. **The file already exists in real projects.** `pi-agent-config/extensions/.pi-lens.json` (and presumably a few others) has been shipping with the `ignore` and `rules` sections since the LSP-config walker was added. The config *looks* right; the docs *imply* it works. Removing the fields would be a breaking change for those projects with no visible benefit.

2. **The integration points were almost all there.** `createProjectIgnoreMatcher` already accepted `extraPatterns`. The two rules already had `const` thresholds. The session-start hook already had `cwd`. The work was wiring, not design.

3. **`.pi-lens.json` mtime belongs in the cache key.** The previous `getProjectIgnoreMatcher` cache was keyed on `.gitignore` mtime only. If I hadn't added `.pi-lens.json` to the invalidation, editing the file would have required a session restart to take effect — which would have made the feature feel broken even though it technically works.

## Design choices worth flagging in review

- **`ignore` uses gitignore semantics**, not picomatch globs. This was the cheaper call (reuses `parseGitignoreContent` → `matchesGitignorePattern` plumbing) and the matching ergonomics are better for path-heavy ignore lists. Easy to swap to picomatch later via a separate `ignoreGlobs` field if a real use case shows up.
- **`high-complexity` exposes `threshold` only**, not the `depth` sub-knob. Kept the schema tight. If anyone asks for the depth override, it's a 2-line addition.
- **Unknown rule ids are silently ignored**, not errored. Forward-compat — a project that writes `{ "rules": { "future-rule": { "threshold": 5 } } }` gets nothing today and works automatically once that rule lands.
- **`resetDispatchBaselines` `cwd` is optional**, not required. The two existing test callers don't pass it; making it required would have been a needless break.

## Not addressed (intentionally out of scope)

- **`lsp.json` schema still lives in `lsp/config.ts`, not in the new `project-lens-config.ts` loader.** Moving it would unify the two config surfaces but is a much larger diff and a behavior change for projects whose `servers` field is currently read. Worth a separate PR if desired.
- **No CLI flag to print the resolved config.** The natural follow-up is a `pi-lens config show` subcommand or an `lens_diagnostics` mode that includes the resolved values. Easy to add; not blocking.

The end-to-end check I ran: with a project `.pi-lens.json` containing `"ignore": ["fixtures/**"]`, `collectSourceFiles()` correctly skips a `fixtures/noise.ts` file (proves the wiring all the way through to the source-filter). With `"rules": { "high-complexity": { "threshold": 5 } }`, a function with CC=7 that the default 15 silently ignored now correctly surfaces a `high-complexity` warning.

greg-hass:
Closing this — I've installed the feature branch locally in my own agent config instead. The change is on greg-hass/pi-lens#feature/project-lens-config and works end-to-end here. If the upstream owner of apmantza/pi-lens wants this PR re-opened, the commit hash is 30d359d — happy to re-target.

PR #254, opened 2026-06-19
fix(widget): support native truncateToWidth Ellipsis enum signature

bloodf:
Under \`@oh-my-pi\` >=16 (current: 16.1.3) the \`@earendil-works/pi-tui\` import resolves to a **native** binding. Its \`truncateToWidth\` signature is:

\`\`\`
truncateToWidth(text, maxWidth, ellipsisKind: Ellipsis, pad, tabWidth: i32)
\`\`\`

not the pure-JS \`(text, maxWidth, ellipsis: string, pad?)\` form \`fitLine\` was written against. Passing the string ellipsis (\`"…"\` / \`"..."\`) crashes the status-widget renderer on every frame:

\`\`\`
Error: Failed to convert napi value into enum \`Ellipsis\`. NumberExpected,
Failed to convert napi value String into rust type \`i32\`
    at truncateToWidth (unknown)
    at renderWidget (.../pi-lens/clients/widget-state.ts:237)
\`\`\`

(`pi-tui` CHANGELOG: `EllipsisKind` renamed to `Ellipsis`; `truncateToWidth`/`sliceWithWidth`/`extractSegments` removed from the public JS API and re-exported from `@oh-my-pi/pi-natives`.)

## Fix

\`fitLine\` probes the legacy string signature once. On napi failure it switches to the native enum signature (\`Unicode=0\` / \`Ascii=1\` / \`Omit=2\`) and a default \`tabWidth\` for the rest of the process. Result is cached in a module-level flag so there is no per-render try/catch overhead. Works unchanged on both the pure-JS pi-tui and the native @oh-my-pi runtime.

Repro: run pi-lens under \`@oh-my-pi/pi-coding-agent@16.1.3\` with any LSP diagnostics present (triggers the \`"…"\` path) — the widget crashes before and renders after.


apmantza:
thanks @bloodf for the oh-my-pi compat contribution.

Issue #121, opened 2026-05-30
BiomeClient.getBiomeBinary() anchors to process.cwd() instead of project root — breaks biome detection in monorepos


apmantza:
## Background

`BiomeClient.getBiomeBinary()` at `clients/biome-client.ts:67-74` resolves the local biome binary by walking `path.join(process.cwd(), "node_modules", ".bin", ...)`. The 2026-05-30 audit (item 8) flagged this as inconsistent with every other tool-resolution path in pi-lens, which threads the project / file cwd through.

## The problem

In a monorepo where pi-lens is invoked from a sub-package directory, `process.cwd()` is the sub-package — but the edited file's project root may be a parent. The reverse also happens: pi is invoked from the workspace root but the file lives in a sub-package whose `node_modules/.bin/biome` is the right binary.

The fallback (npx biome) still works but is much slower (~5-10x) than the local `.bin` binary. Most users will not notice except as latency.

## Proposed change

Accept an optional `cwd` parameter on `getBiomeBinary()` and thread it from every caller. Default to `process.cwd()` only when no cwd is supplied, matching the existing pattern in `runner-helpers.ts:resolveNodeToolCommand(cwd, toolName)`.

```ts
private getBiomeBinary(cwd?: string): string | null {
  const resolveCwd = cwd ?? process.cwd();
  // ... existing logic with resolveCwd in place of process.cwd()
}
```

## Acceptance criteria

- `getBiomeBinary(cwd)` resolves local `.bin/biome` from the given cwd, falling back to `npx biome` only when neither the cwd's `node_modules` nor the managed-tools-dir has it.
- Existing call sites pass the per-edit project cwd; no signature change is required at the public surface beyond adding the optional arg.
- A new test fixtures a two-package monorepo with biome installed only at the sub-package's `node_modules` — assert the right binary is selected when the edit is in that sub-package.

## Refs

- 2026-05-30 audit item 8
- Companion to the audit-driven issues filed for the wider `--version` probe migration (item 1-7) and the walk-up dedup work

PR #264, opened 2026-06-20
fix(dispatch): go-vet vets the file's package from the module root (closes #263)

### LINKED ISSUE #263: go-vet runner false-positives same-package symbols (+ "go.mod not found") in multi-file modules
## Summary

The `go-vet` dispatch runner spawns `go vet <ctx.filePath>` **with no `cwd`** (`clients/dispatch/runners/go-vet.ts`). In any multi-file Go package this produces **false `undefined: X` diagnostics** for same-package symbols defined in sibling files, and when the spawn inherits a cwd with no `go.mod` (e.g. a monorepo root) it also surfaces **`go: go.mod file not found`**.

## Reproduction

Two-file package, both files in `package mypkg`, `go.mod` at the package/module root:

```go
// a.go
package mypkg
func New() int { return 42 }

// b.go
package mypkg
func UseNew() int { return New() }
```

pi-lens reports on `b.go`:
```
undefined: New
go: go.mod file not found in current directory or any parent directory
```

Reality (run from the module root): `go vet ./...` and `go build ./...` are both clean. Verified on macOS, go 1.26.4.

Empirical matrix (`sub/b.go` uses `sub/a.go`'s symbol):

| Invocation | Result |
| --- | --- |
| `go vet <abs>/sub/b.go`, no cwd (current pi-lens shape) | ❌ `undefined: Make` + `go: go.mod file not found` |
| `go vet ./sub`, `cwd` = module root | ✅ clean |
| `go vet -C <modroot> ./sub` (Go ≥1.20) | ✅ clean |
| `go vet .` for a root-package file | ✅ clean |

## Root cause

`go vet <one-file.go>` compiles that file **in isolation**. Same-package symbols defined in sibling files are not visible, so they read as `undefined: X`; and without a module-root cwd the run also reports `go.mod file not found`. Meanwhile `ctx.cwd` is **already** the `go.mod` module root (resolved by `resolveLanguageRootForFile`, Go markers `["go.mod"]`), but the runner never passes it.

## Why this isn't a duplicate of #214

#214 used a single **self-contained** fixture (`package main` + `func main`), where single-file `go vet bad.go` succeeds even without a cwd. The multi-file / same-package case is a distinct variant that #214 never exercised, and #214 did not add a cwd (last touch on `go-vet.ts` was the review-graph commit, unrelated). The #214 "Next step" already floated *"set `cwd` to the file's package dir"* — this issue is that direction applied to the multi-file case.

## Proposed fix

Vet the **package** containing the file from the module root (`go vet ./<relpkg>` with `cwd: ctx.cwd`), and **filter** parsed output to the edited file. The filter is required because `createLineParser` (`utils/diagnostic-parsers.ts`) discards the output filename and attributes every diagnostic to `ctx.filePath` — without filtering, package vetting would pin sibling-file issues onto the edited file.

This mirrors the cwd-threading precedent from #121 (BiomeClient `process.cwd()` → accept/thread optional `cwd`) and the sibling runner `golangci-lint.ts` (`const cwd = ctx.cwd || process.cwd()`), which avoids the single-file isolation problem because golangci-lint analyzes the package of a passed file whereas `go vet <file>` does not.

## Behavior note

Previously a non-zero `go vet` with no parseable output returned `status: "failed"` + `rawOutput`. With package vetting + per-file filtering, an edited file with 0 own diagnostics now returns `status: "succeeded"` — so an unrelated sibling-file error no longer flags the edited file's turn. Diagnostics genuinely belonging to the edited file are unchanged (still `failed` + diagnostics).

## Acceptance

- Multi-file package: edited file using a sibling symbol → no false `undefined: X`.
- No `go.mod file not found` when pi-lens' cwd is a monorepo root without `go.mod`.
- Diagnostics from sibling files are not attributed to the edited file.
- Existing single-file `tests/fixtures/tool-smoke/go/` still produces a parseable diagnostic (`go vet .` → `bad.go`).

Labels: `bug`, `area:dispatch`.

cjunxiang:
> Draft pending your review of the scope change below — happy to reshape. Closes #263.

## Problem

The `go-vet` runner spawns `go vet <ctx.filePath>` with **no `cwd`**. In a multi-file Go package this produces:

- **false `undefined: X`** — `go vet <one-file.go>` compiles that file in isolation, so any symbol defined in a sibling file is reported undefined.
- **`go: go.mod file not found`** — when the spawn inherits a cwd with no `go.mod` (e.g. a monorepo root).

…even though `go vet ./...` / `go build ./...` from the module root are clean. Full repro + empirical matrix in #263.

## Fix

Vet the **package** containing the file from the module root — `ctx.cwd` is already the `go.mod` module root via `resolveLanguageRootForFile` (markers `["go.mod"]`):

```ts
const cwd = ctx.cwd || process.cwd();
const fileRel = relative(cwd, ctx.filePath).split(sep).join(posix.sep);
const pkgPath = fileRel.includes("/")
  ? "./" + fileRel.slice(0, fileRel.lastIndexOf("/"))
  : ".";
const result = await safeSpawnAsync(goExe, ["vet", pkgPath], { timeout: 30000, cwd });
```

Because `createLineParser` discards the output filename and attributes every diagnostic to `ctx.filePath`, the parsed output is **filtered to the edited file** so sibling-file diagnostics aren't mis-attributed.

Mirrors the cwd-threading precedent of #121 and the sibling runner `golangci-lint.ts` (`const cwd = ctx.cwd || process.cwd()`). golangci-lint doesn't have this bug because it analyzes the package of a passed file, whereas `go vet <file>` isolates it.

## Scope change to flag

This widens go-vet from single-file to **package** scope (the only correct way to vet multi-file packages), and changes the tail behavior:

- **Before:** non-zero `go vet` with no parseable output → `failed` + `rawOutput`.
- **After:** edited file with 0 own diagnostics → `succeeded`; a sibling-file error no longer flags the edited file's turn (it surfaces when that file is edited).

If you'd prefer to keep emitting `rawOutput` for genuine go-vet failures, or want sibling issues surfaced differently, say so and I'll adjust.

## Verification

- `npm run build` ✅ (in-place `.js`, which is gitignored)
- `npm run lint` (`tsc -p tsconfig.json`, strict CI gate) ✅
- new `tests/clients/dispatch/runners/go-vet.test.ts` — 6/6 ✅ (spawn args `["vet","./sub"]` + `cwd`, `"."` for root pkg, sibling-filtering, clean→succeeded, skip-when-unavailable)
- full dispatch suite — 378/378 ✅ (`tests/clients/dispatch/`)
- existing single-file smoke fixture still resolves to `go vet .` → `bad.go` parseable

### DIFF
--- a/clients/dispatch/runners/go-vet.ts
+++ b/clients/dispatch/runners/go-vet.ts
@@ -4,6 +4,8 @@
  * Runs `go vet` for Go files to catch common mistakes.
  */
 
+import { relative, sep, posix } from "node:path";
+
 import { GoClient } from "../../go-client.js";
 import { safeSpawnAsync } from "../../safe-spawn.js";
 import { stripAnsi } from "../../sanitize.js";
@@ -31,9 +33,23 @@ const goVetRunner: RunnerDefinition = {
 			return { status: "skipped", diagnostics: [], semantic: "none" };
 		}
 
-		// Run go vet on the file
-		const result = await safeSpawnAsync(goExe, ["vet", ctx.filePath], {
+		// Vet the package containing the file from the module root.
+		//
+		// `go vet <one-file.go>` compiles that file in isolation, so same-package
+		// symbols defined in sibling files read as false `undefined: X`, and when
+		// the spawn cwd is outside the module the run also reports
+		// `go.mod file not found` (#263). ctx.cwd is the go.mod module root
+		// (resolveLanguageRootForFile, markers ["go.mod"]), so vet the file's
+		// package from there.
+		const cwd = ctx.cwd || process.cwd();
+		const fileRel = relative(cwd, ctx.filePath).split(sep).join(posix.sep);
+		const pkgPath = fileRel.includes("/")
+			? "./" + fileRel.slice(0, fileRel.lastIndexOf("/"))
+			: ".";
+
+		const result = await safeSpawnAsync(goExe, ["vet", pkgPath], {
 			timeout: 30000,
+			cwd,
 		});
 
 		const raw = stripAnsi(result.stdout + result.stderr);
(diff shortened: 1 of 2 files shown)


cjunxiang:
Some comments from sonarcloud. (prefer regex over match), lmk the preference. happy to swap.


https://sonarcloud.io/project/issues?id=apmantza_pi-lens&issues=AZ7kUPAaupWwgW3YDYlu&open=AZ7kUPAaupWwgW3YDYlu&pullRequest=264

apmantza:
pushed a follow up commit - normalizing the file filter to what is used in similar cases + some more tests. hope thats ok. will merge - the small sonar issues flagged are insignificant.

PR #349, opened 2026-06-29
WIP: bound lens_diagnostics pending flush

jrimmer:
## Summary

WIP: `lens_diagnostics` no longer waits forever for pending debounced tool-result pipelines before it can return cached diagnostics. If the pre-read flush stalls, the diagnostics tool now times out that wait, returns the last completed cache/widget state, and marks the result as potentially stale instead of leaving the caller hung.

The timeout also flows into flushed post-write pipelines so the major async phases can stop waiting when `lens_diagnostics` gives up on the flush. This is intended as a safety valve around rare stuck pipeline/LSP/runner cases while preserving the existing “flush before reporting” behavior when the pipeline is healthy.

## Notes

- Default flush wait is 3s.
- The timeout is configurable with `PI_LENS_DIAGNOSTICS_FLUSH_TIMEOUT_MS`.
- Timed-out results include `details.flushTimedOut` and a visible warning.
- This PR is intentionally marked WIP/draft while testing continues in real Pi sessions, because the original failure is intermittent.

## Test plan

- `npm run lint`
- `npm run build`
- `npm test -- tests/tools/lens-diagnostics.test.ts`

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex


apmantza:
many thanks for this @jrimmer. implemented #364 to fix the root cause of this (unbounded pull diagnostics). keeping this PR open in case this was misdiagnosed.

apmantza:
Thanks again for this, @jrimmer — closing because the underlying problem has since been fixed.

- **#383** — per-file wall-clock budget + overall wall-clock ceiling (`PI_LENS_LENS_DIAGNOSTICS_FULL_TIMEOUT_MS`) + `ctx.signal` cancellation (Escape) for `lens_diagnostics`.
- **#386** — bounded the pre-dispatch LSP sync (a `didChange` write that backpressured forever on a wedged server).
- **#388** — bounded `notify.open` inside `touchFile` at the source (the dispatch LSP runner was riding to the 30s ceiling) + per-server serialization so the full sweep can't flood a single-threaded server.

With those in place, an infinite hang should no longer possible. The residual idea here (a fast return with a `flushTimedOut` staleness marker) is an idea i'd be happy to revisit.

PR #350, opened 2026-06-29
fix(lsp): decouple tool discovery from install gate

moj02090:
# Bug: TypeScript/pyright LSP never starts when `PI_LENS_DISABLE_LSP_INSTALL=1`, even though the server is installed on PATH

**Area:** area:installer (also area:lsp)
**Version:** 3.8.62 (reproduced on 3.8.61; same code in `master` @ cc824cc)

## What happened?

In an environment where LSP auto-install is disabled (`PI_LENS_DISABLE_LSP_INSTALL=1`)
but all language servers are **pre-installed globally** (e.g. `npm i -g typescript
typescript-language-server pyright`), `.ts`/`.tsx` files get **no interactive LSP**
(`ready=0/N` — hover/definition/references/rename all dead). `tsc` dispatch diagnostics
still work; only the LSP is missing. `marksman`/`json` etc. work fine.

`/usr/bin/typescript-language-server` exists and is on `PATH`, but pi-lens never finds it.

### Root cause

`ensureTool()` already does **discovery-first, install-second**: it calls
`getToolPath()` (which probes PATH / npm-global / managed-bin, offline-safe, no
download) and only falls through to `installTool()` if nothing was discovered.

But the **bespoke** spawn sites gate the *entire* `ensureTool()` call behind
`canInstall()`, so when install is disabled, **discovery is disabled too**:

`clients/lsp/server.ts` — `TypeScriptServer.spawn`:
```js
if (!lspPath) {
  if (canInstall(options?.allowInstall)) {        // ← blocks the ONLY global-discovery path
    lspPath = await ensureTool("typescript-language-server");
  }
  if (!lspPath) return undefined;                 // ← server never spawns → ready=0
}
```

Same pattern in `findTsserverPath()` (`ensureTool("typescript")` gated by
`canInstall`) and `PythonServer.spawn` (`if (!canInstall(...)) return undefined;`
before `ensureTool("pyright")`).

Generic servers don't hit this because `resolveAndLaunch()` tries bare-command
candidates via `launchLSP()` first (PATH lookup) **unconditionally** — only the
TypeScript/tsserver/pyright spawns special-case their resolution and skip that.

Net effect: `PI_LENS_DISABLE_LSP_INSTALL` conflates two separate concerns —
**"may download a new tool"** vs **"may discover an already-installed one."**

## Steps to reproduce

1. Global install, no per-project `node_modules`:
   `npm i -g typescript typescript-language-server` → `/usr/bin/typescript-language-server` on PATH.
2. `export PI_LENS_DISABLE_LSP_INSTALL=1`
3. Open any `.ts` file in a project that has a `package.json` but **no** local
   `node_modules/.bin/typescript-language-server`.
4. Observe `ready=0/N` for `typescript`; hover/definition/references/rename unavailable.
   Log shows the spawn returning `undefined` without ever calling `ensureTool`.

## Expected behavior

With install disabled, pi-lens should **still discover and launch** a
language server that is already present on PATH / npm-global. Only the actual
download/install should be suppressed.

## Proposed fix (PR-ready, attached)

Decouple discovery from install at the single chokepoint:

- Add `allowInstall?: boolean` to `ensureTool(toolId, opts)`. Discovery
  (`getToolPath`) always runs; when `allowInstall === false`, skip **only** the
  `installTool()` step and return whatever discovery found (or `undefined`).
- Change the three bespoke gated sites to always call `ensureTool(id, {
  allowInstall: canInstall(...) })` instead of wrapping the call in
  `if (canInstall(...))`: `TypeScriptServer.spawn`, `findTsserverPath`, `PythonServer.spawn`.

This is general (fixes TS, tsserver, pyright uniformly) and minimal. Generic
`resolveAndLaunch` servers are unchanged.

Branch: `fix/decouple-lsp-discovery-from-install` (4 files, +92/−39).
- Regression test: `TypeScriptServer` resolves a global LSP with
  `allowInstall:false` (`tests/clients/lsp/runtime-install-discovery.test.ts`).
- Updated two `server-policy` tests to the corrected contract (ensureTool is now
  *called* with `{ allowInstall: false }` but still must not install).

`npm run build`, `npm run lint`, `npm run build:dist`, `npm run check:lockfile`
all pass; the four LSP/installer suites pass (66 tests). The only full-suite
failures on my box are pre-existing/load-flaky (ast-grep binary timeout, codegen
`--check`, MCP smoke) and reproduce identically on unmodified `master`.


apmantza:
thanks @moj02090. merged after a hardening follow-up commit.

PR #363, opened 2026-07-01
Fix POSIX LSP process-tree cleanup (refs #362)

### LINKED ISSUE #362: POSIX LSP teardown can leave launcher descendants alive
## What happened?

On POSIX systems, LSP servers launched through wrapper processes can leave descendants behind after pi-lens shuts down or resets its LSP service.

I hit this most visibly with the HTML language server while working inside zellij: after leaving/restarting sessions, `vscode-html-language-server` processes accumulated instead of being cleaned up. Eventually there were enough long-lived HTML LSP processes/diagnostics to put real pressure on memory.

The underlying issue does not appear to be zellij-specific. zellij just makes it easy to keep sessions around long enough to notice. The lifecycle gap is that pi-lens tears down the direct child process on POSIX, but a language-server launch can involve wrappers or descendants (npm shims, shell launchers, node children, etc.). Those descendants are not guaranteed to die when only the immediate child is signaled.

Windows already uses `taskkill /T` for normal LSP teardown, so it gets process-tree semantics there. POSIX currently lacks the equivalent process-group cleanup.

## Steps to reproduce

One way to observe it:

1. Start pi-lens/pi in a terminal multiplexer such as zellij.
2. Open or touch files that cause the HTML LSP to start.
3. Leave/restart sessions or otherwise trigger LSP service reset/shutdown.
4. Inspect live processes:

```sh
ps -eo pid,ppid,pgid,sid,stat,rss,etime,cmd | rg 'vscode-html-language-server|html-languageserver|pi|zellij'
```

Expected cleanup does not always happen for the full process tree; descendant HTML LSP processes can remain alive.

This is easiest to notice with HTML LSP, but the same class of bug can apply to any LSP started through a wrapper or launcher process.

## Expected behavior

When pi-lens resets or shuts down an LSP service, the whole LSP process tree should be cleaned up, not just the direct child process.

On POSIX, a typical fix is:

- spawn LSP servers detached into their own process group, and
- signal the process group during teardown with `process.kill(-pid, signal)`, with direct-child fallback if group signaling fails.

That would bring POSIX behavior closer to the existing Windows `taskkill /T` behavior.

## pi-lens version

Observed while running from source around `3.8.61`.

## Relevant logs / evidence

Useful commands for confirming the issue:

```sh
pgrep -af 'vscode-html-language-server|html-languageserver|pi|zellij'
ps -eo pid,ppid,pgid,sid,stat,rss,etime,cmd | rg 'vscode-html-language-server|html-languageserver|pi|zellij'
rg -n 'html|vscode-html|session_start|shutdown|diagnostic' ~/.pi-lens -S
```

I have a small local patch with regression tests that makes POSIX LSP launches process-group scoped and tears them down by signaling the group while preserving the existing Windows behavior.

cmptr:
## Summary

Fixes POSIX LSP teardown so launcher/wrapper descendants are cleaned up when pi-lens resets or shuts down an LSP service.

On POSIX, LSP servers are now launched in their own process group and teardown signals that group before falling back to direct-child kill. This brings POSIX cleanup closer to the existing Windows `taskkill /T` behavior.

Closes #362

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [x] area:perf
- [ ] area:observability
- [x] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

## What changed and why

Some LSP launches go through wrapper processes: npm shims, shell launchers, node wrapper scripts, or language-server bootstrap scripts. On POSIX, pi-lens previously signaled only the direct child process during teardown. If that child had spawned descendants, those descendants could survive session reset/shutdown and accumulate as orphaned language-server processes.

This PR changes POSIX LSP lifecycle handling to:

- spawn LSP servers detached, creating a dedicated process group;
- signal the process group with `process.kill(-pid, signal)` during teardown;
- fall back to direct-child `proc.kill(signal)` if process-group signaling fails.

Windows behavior is intentionally unchanged: normal mid-session teardown still uses `taskkill /T`, and `processExiting` still avoids spawning `taskkill`.

This does not try to reduce the number of LSP servers while multiple pi sessions are legitimately still alive. It only fixes cleanup for descendant processes that should die when an LSP service is reset or shut down.


apmantza:
thanks @cmptr. appreciate it. merged.

Issue #255, opened 2026-06-19
bug(tree-sitter): lua parses to ERROR trees once a 2nd grammar loads (shared WASM Module) — breaks lua symbols+imports


apmantza:
## Summary
Lua structural analysis (symbols **and** imports) is silently broken in any multi-language project. The shared `treeSitterClient` (one per process, used by the review graph) produces `ERROR` parse trees for lua once **any** second grammar has been loaded into web-tree-sitter's shared WASM Module.

## Minimal repro
```js
const c = new TreeSitterClient(); await c.init();
// fresh lua parses correctly:
parse(c, "lua", 'local x = require("mod.a")\n')  // → proper `call` node, OK
// load ANY other grammar:
parse(c, "python", "import os\n")
// re-parse the SAME lua source:
parse(c, "lua", 'local x = require("mod.a")\n')  // → ERROR node (corrupt)
```
Bisect result: `fresh lua: ok`, then `+python -> lua CORRUPT`, and CORRUPT for every grammar after. It is not grammar-specific — loading a *second* grammar of any kind is enough.

## Corrupt tree (warmed client)
```
local_variable_declaration
  = (anon)
  ERROR "require("
    identifier "require"
    ( (anon)
  expression_list ""mod.a""
    string ""mod.a""
ERROR ")"
```
vs. fresh:
```
expression_list
  call "require("mod.a")"
    variable (identifier "require")
    argument_list ( expression_list ( string ) )
```

## Impact
- Production review graph uses a single module-level `treeSitterClient` and lazily loads grammars per language encountered. Any repo with lua + at least one other language (i.e. essentially all of them) gets `ERROR` lua trees.
- `SYMBOL_QUERIES.lua` (defs/refs) and any lua `IMPORT_QUERIES` therefore extract nothing/garbage — lua symbol search, symbol-level impact, and module_report outlines are silently empty for lua. This is why the #249 lua import query (validated correct on a fresh tree) could not be shipped.

## Likely cause
[...]

Issue #432, opened 2026-07-06
swift grammar: load-skip on Node 24+ (real #423 fix) + make grammar-health guard a hard gate


apmantza:
## Problem

The #423 fix (vendored from-source `tree-sitter-swift.wasm`, #426) does **not** durably hold. The nightly `grammar-health` guard is RED: swift crashes the runtime on **all three OSes** under CI's **Node 24.18.0**, even though:

- the committed vendored wasm is **byte-identical** to a fresh CI build (sha256 `3df761e7…`);
- it survives locally on **Node 24.12.0** and survives the build-job's isolated guard step.

Root cause: the crash is a V8 **Turboshaft WASM OOM** (`Fatal process out of memory: Zone`), triggered under accumulated memory pressure on Node 24.18. It is **uncatchable in-process** (an abort, not a throw), so no try/catch can degrade it. Rebuilding the wasm from source does not reliably dodge it, and pinning web-tree-sitter / tree-sitter-cli can't help (the fault is downstream in V8, not in tree-sitter). Only Node 22 or `--liftoff-only` avoid it, neither of which we can impose on users.

## Fix

The only place graceful degradation can live is **before load**. Add a runtime blocklist:

1. `BLOCKED_GRAMMARS` + `isGrammarBlocked(filename, { nodeVersion, platform })` in `clients/grammar-source.ts` — swift blocked when Node major >= 24 (the documented #423 repro boundary; Node 20/22 keep swift).
2. Wire into `TreeSitterClient.loadLanguage` — a blocked grammar returns `null` (feature-degrades: no swift structural symbols) instead of loading + crashing.
3. `scripts/check-grammar-load.mjs` treats a blocked grammar as **intentionally skipped**, not a crash — so the nightly goes **green**, and any *new* crasher still fails it.
4. Make the `grammar-health` guard a **hard gate** (it stays nightly, but a non-blocked crasher = red = must be triaged/blocked).
5. Tests: `isGrammarBlocked` predicate + `loadLanguage` skips a blocked grammar.

The blocklist is **guard-driven**, not hand-maintained: the nightly tells us exactly what/where crashes. Vendoring stays as the mechanism to lift the block once a build is proven safe.

Keep #423 open until the reporter (bisko) confirms swift no longer crashes their pi session (it now degrades instead).

Issue #427, opened 2026-07-06
Shipped tree-sitter-yaml.wasm is unloadable (ABI-incompatible with web-tree-sitter 0.25) and unused — drop or vendor a working one


apmantza:
### Context
Surfaced while fixing #423 (the Swift grammar crash). The `grammar-load` guard (`scripts/check-grammar-load.mjs`) reported `yaml` as **"did not load"** on Node 24 — `Language.load(tree-sitter-yaml.wasm)` **fails** (returns null) against the pinned `web-tree-sitter@0.25.10`. Unlike Swift this fails **gracefully** (no crash), so it's a cleanup, not a bug taking anything down.

### The problem
- `tree-sitter-yaml.wasm` (from `tree-sitter-wasms@0.1.13`) is **ABI-incompatible** with web-tree-sitter 0.25 → it never loads.
- It's in the **`CORE`** set (`download-grammars.ts`), so it's downloaded at `prepare` and **shipped in the tarball** (~183 KB) as dead weight.
- Nothing actually tree-sitter-parses YAML anyway: `yaml` is **not** in the tree-sitter runner's `appliesTo`, nor in `resolveTreeSitterLanguage`'s ext→lang map (YAML is handled by yamllint + the YAML LSP, not tree-sitter). So the grammar is loaded by nothing on the real path.

### Options
1. **Drop it** (leanest) — remove `tree-sitter-yaml.wasm` from `CORE` + `GRAMMARS` + the manifest, since nothing parses YAML via tree-sitter. Saves tarball weight, removes a grammar that can't load.
2. **Vendor a working one** — `@tree-sitter-grammars/tree-sitter-yaml@0.7.1` ships a `tree-sitter-yaml.wasm` that **loads + parses cleanly** on 0.25 (verified locally through the guard). This is a tidy second use of the new **`VENDORED`** mechanism from #423 (build/obtain-from-source, commit, per-grammar provenance override, no CDN download) — but only worth it if we intend to add YAML to the tree-sitter runner/ext-map.

Recommendation: **drop** unless we have a near-term reason to tree-sitter-parse YAML, in which case **vendor** the 0.7.1 grammar and wire it into `appliesTo` + the ext map.

### Guard
Whichever way: the `grammar-health` nightly (#425) will keep asserting the shipped grammar either loads or is absent, so this can't silently regress.

Refs #423.

Issue #249, opened 2026-06-19
feat(review-graph): extract imports for tree-sitter languages (python/go/rust/java first)


apmantza:
## Summary

The review graph **parses every supported language** via the tree-sitter WASMs, but it only **extracts import statements** for two: jsts (`addJsTsFile` → `importFactProvider`) and cxx (`addCxxIncludeEdges` → `#include` text parse). For the other ~15 tree-sitter languages, `addTreeSitterFile` adds `file`/`symbol`/`contains`/`defines`/`references` edges but **no `imports` edges** — the tree-sitter extractor has `defs` and `refs` queries, no import query.

Consequence: a Python/Go/Rust/… file has full symbol + reference nodes in the graph but **zero import edges**, so:
- `module_report.imports` is empty for those languages at every depth (surfaced via #245).
- `pilens_impact` / cascade **under-count cross-file dependencies** for non-jsts/cxx code — their dependency edges come only from `references`, never `imports`.

## Plan

**This pass (uniform capture):**
- Add an `IMPORT_QUERIES` map (sibling to `SYMBOL_QUERIES`) in `tree-sitter-symbol-extractor.ts`: a per-language tree-sitter query capturing the import source (`@importSource`). Optional per language — a missing query degrades to "no imports" (no crash).
- Extend `ExtractedSymbols` with `imports: ImportRef[]` ({ source, line }); compile + run the import query in the extractor.
- Wire `addTreeSitterFile` to emit `imports` edges to `external:`/`module:` source nodes (relative-marker heuristic for the external/internal split), mirroring `addJsTsFile`'s unresolved branch.
- First verified batch: **python, go, rust, java** (clear import-statement grammars), each validated against a fixture via the real extractor. Others added incrementally.
- `module_report`: relativize internal import paths (currently absolute).

**Follow-up (separate):**
- Resolve internal imports to project files per language (Python dotted modules, Go package paths, Rust `use` crates) — `addJsTsFile`'s JS-relative `localImportToFile` doesn't transfer. Until then, non-jsts internal imports show as raw source strings.
- Remaining languages' import queries (ruby/php = require-call semantics; kotlin/dart/elixir/csharp/swift/lua/ocaml/zig).

## Acceptance (this pass)
- [ ] `IMPORT_QUERIES` + extractor emits `imports` for python/go/rust/java; per-language extractor test.
- [ ] `addTreeSitterFile` emits `imports` edges; review-graph test asserts import edges for a non-jsts fixture.
[...]

PR #436, opened 2026-07-07
fix(tree-sitter): source-override lua (#255) + yaml (#427) to maintained grammars

apmantza:
## What & why

The aggregator's `tree-sitter-lua.wasm` (`tree-sitter-wasms@0.1.13`) parses lua correctly **only as the sole grammar**. The instant any second grammar loads into web-tree-sitter's process-global WASM `Module`, every subsequent lua parse becomes an `ERROR` tree — so `SYMBOL_QUERIES.lua` / `IMPORT_QUERIES.lua` extract nothing, and **lua symbol search, symbol-level impact, and `module_report` outlines are silently empty in essentially every multi-language repo** (this is why the #249 lua import query couldn't ship).

I verified it's **lua-specific** with a per-grammar isolated-child sweep: `bash/ruby/python/go/js/…` (23 grammars) all re-parse cleanly after a second grammar loads; **only lua** corrupts. So it's a bad grammar *build*, not a runtime bug — and fixable by swapping the source.

## Fix — a general per-grammar source override

- **`GRAMMAR_SOURCE_OVERRIDES`** (`grammar-source.ts`) + **`SOURCE_OVERRIDES`** mirror (`download-grammars.ts`/`.js`): map a wasm → `{package, version, url}`. Both the **runtime lazy-fetch** (`grammarSourceUrl`) and the **build-time downloader** pull the override URL instead of the frozen aggregator.
- lua → **`@tree-sitter-grammars/tree-sitter-lua@0.4.1`** (the maintained org's prebuilt wasm), which parses cleanly in a multi-grammar process. Lock hash updated + new `overrides` provenance section; `expectedVersion`/`expectedPackage` keep `needsDownload` + the provenance guard from false-tripping on the different package/version.
- lua **defs/refs/import queries rewritten** for the new grammar's node types (`function_declaration` / `function_call` / `dot_index_expression`).

## Tests

- lua added to the symbol + import per-grammar smokes (was omitted because of this bug).
- **New dedicated shared-Module regression test**: load python/go/js first, then assert lua parses with **zero ERROR nodes** and extracts symbols (`M.run`, `util`) + the `require` import (`mod.helper`).
- Override unit tests (lua routes off the aggregator; other grammars don't; `.js` mirror stays in sync).
- `npm run lint` clean; full `npm test` → 2791 passed, 5 skipped (swift-blocked).

## Scope notes

- The override mechanism is **reusable for yaml (#427)** — `@tree-sitter-grammars/tree-sitter-yaml@0.7.1` ships a wasm, so #427 becomes near a one-liner.
- **swift stays a load-block (#432)** — no `@tree-sitter-grammars/tree-sitter-swift` exists, and `tree-sitter-swift` ships C source only (which we built in #426 and it still crashed). Its failure is a V8 crash, a different class.

## Changelog

Added an `[Unreleased] Fixed` entry.


apmantza:
Extended to also fix **#427** (yaml) on the same override mechanism, per request:

- yaml's aggregator wasm is ABI-incompatible with web-tree-sitter 0.25 (fails `Language.load`); overridden to **`@tree-sitter-grammars/tree-sitter-yaml@0.7.1`**, which loads + parses cleanly. Source-only (yaml has no symbol/import queries) + a load regression test.
- With both fixes, the full per-grammar corruption/load sweep is **clean for all 26 grammars** (flagged: none — was lua + yaml).

Now closes #255 **and** #427.

PR #501, opened 2026-07-10
fix(ast-grep): align generated config with project-first same-ID precedence (closes #497)

### LINKED ISSUE #497: fix(ast-grep): align generated config with project-first same-ID precedence
## Problem

Pi-lens has two ast-grep rule-discovery surfaces with different same-ID precedence:

- the in-process NAPI runner deduplicates project and bundled rules project-first;
- the generated raw ast-grep/LSP configuration can include both files when a project rule and a bundled rule share an ID.

Raw `sg` then rejects the generated configuration because rule IDs must be unique, so the LSP path does not honor the same override contract as NAPI.

## Expected behavior

1. A project rule with ID `X` overrides a bundled rule with ID `X`.
2. Generated raw ast-grep/LSP configuration includes only the winning project rule.
3. NAPI and LSP select the same winner.
4. Unique project and bundled rules remain available in deterministic order.
5. Same-layer duplicates remain visible as errors rather than being silently hidden.
6. Secondary bundled rule directories retain their documented precedence.
7. Fresh and cached diagnostics agree on the winner.

## Non-goals

- Renaming project rules.
- Weakening duplicate-ID validation globally.
- Suppressing `sg` errors.
- Broad runner architecture changes.

apmantza:
## Summary
- The in-process NAPI ast-grep runner (`evaluateAstGrepRules`) has always deduped same-id rules project-first (project rule dirs win a same-`id` collision against bundled rule dirs). The generated raw sgconfig used by the ast-grep LSP (`clients/sgconfig.ts`'s `resolveBaselineSgconfig`) instead listed the project and bundled rule dirs side-by-side in `ruleDirs` — verified with a real `sg scan` repro that raw ast-grep hard-errors ("Duplicate rule id") the moment two `ruleDirs` entries share an id, rejecting the whole config.
- `resolveBaselineSgconfig` now materializes a single merged rule directory (doc-level dedup, so multi-rule YAML files are handled correctly) built from a single project-first-ordered dir list (`shippedRuleDirsInPrecedenceOrder`), which the NAPI runner now also derives from — so the two surfaces can never drift and pick different winners.
- Same-layer duplicates (two files in the *same* source dir sharing an id) are copied through unmodified, so `sg` still hard-errors on those exactly as before — only the cross-layer (project vs. bundled) collision is resolved.
- The merged directory is fingerprinted against source-dir mtimes, so a mid-session project rule addition/change invalidates the cached config instead of serving a stale winner set.
- Files are copied (never symlinked/junctioned) — Windows has no unprivileged file symlinks, and junctions are directory-level only, so copying is the cross-platform-safe mechanism for filtering individual rule files out of a merged view.

Reported by @anasalsbey-glitch.

## Test plan
- [x] Reproduced the bug directly against the real `sg` CLI (a project rule dir shadowing a bundled rule id → `sg scan` exits 8 with "Duplicate rule id")
- [x] Confirmed the fix's generated config passes `sg scan` cleanly, with the project rule winning
- [x] `tests/clients/sgconfig.test.ts` — new cases: project override wins (points 1/2/3), unique rules from both layers survive (point 4), same-layer duplicates still produce two files for `sg` to error on (point 5), and a mid-session project rule change invalidates + re-resolves the cache to the same winner a fresh resolution would produce (point 7)
- [x] `npm run build`
- [x] `npm test` (full suite: 3136 passed, 9 skipped)
- [x] `npm run lint`
- [x] CHANGELOG.md `[Unreleased]` entry added


anasalsbey-glitch:
Thanks for fixing and merging #497.

While validating the project-rule precedence behavior, I found a few follow-up cases that may still be worth covering:

- the config materializer and NAPI YAML loader appear to scan only immediate `.yml` files, while secondary rule catalogs such as CodeRabbit are nested under language directories;
- cache invalidation based on source-directory mtimes may miss edits to an existing rule file, including equal-size changes or changes with a preserved mtime;
- generated config/rule artifacts are process-scoped and resolution still depends on `process.cwd`, which may affect nested workspaces or multiple workspace roots in one process;
- same-layer duplicates remain raw `sg` errors, while the NAPI path appears to select the first definition instead of returning an equivalent blocking configuration diagnostic.

I have generic regression coverage available for nested rule discovery, equal-size/preserved-mtime edits, workspace-root isolation, and raw LSP/NAPI parity.

Would you welcome a narrowly scoped follow-up PR for any of these cases? I can split the work into smaller PRs if preferred.

apmantza:
a PR would be great!

PR #516, opened 2026-07-10
fix(ast-grep): harden project rule precedence follow-ups

anasalsbey-glitch:
## Summary

- complete the remaining project-rule precedence edge cases identified after #501
- keep raw ast-grep/LSP and NAPI discovery behavior aligned
- add focused regression coverage for recursive rule discovery, content-aware caches, workspace-root isolation, and same-layer duplicate diagnostics

## Context

PR #501 fixed the core project-versus-bundled same-ID collision. This follow-up covers only the additional cases discussed with the maintainer after that merge.

The shared discovery order is project primary → project secondary/CodeRabbit → bundled native → bundled CodeRabbit. Project rule trees are discovered recursively and fingerprinted by relative path plus content, generated artifacts are keyed per workspace root, and same-layer duplicates remain raw `sg` errors while NAPI now reports an equivalent blocking configuration diagnostic.

## Validation

- `npx vitest run tests/clients/sgconfig.test.ts tests/clients/ast-grep-rule-precedence-followups.test.ts --reporter=dot` — 16 passed
- physical no-space ast-grep fixture/runner/tool suite — 203 passed, 2 skipped
- physical no-space `npm test` — 3183 passed, 12 skipped
- `npx vitest run tests/clients/runtime-session-warm.test.ts --reporter=verbose` in normal/full mode — 5 passed
- `npm run build`
- `npm run build:dist`
- `npm run lint`
- `npm run check:lockfile`
- `npm run changelog:check`
- `npm run docs:rule-catalogs -- --check`
- `git diff --check`

Cold synthesized-config generation measured 57.78 ms; warm cache resolution measured 0.12 ms median; resolving a second workspace root measured 37.48 ms with distinct artifacts.

## Scope

No project-specific exception, TS/JS twin-deduplication change, path-harness cleanup, installed-package change, or application behavior is included.

Refs #497


apmantza:
many thanks @anasalsbey-glitch. merged!

Issue #144, opened 2026-06-02
feat(internal-edits): seq/hash/range-validated edit substrate for partial-apply, LSP workspace edits, and actionable autofix (implementation.md Phase 6)


apmantza:
## Background

`implementation.md` Phase 6 (\"Internal edit substrate to reduce failed edits\") is the second unfiled item from the codedb-inspired implementation plan. Status in the doc: **not implemented**.

The phase addresses concrete failures we already see in production telemetry. From recent read-guard log analysis (this cycle):

- `oldtext_duplicate` — exact `oldText` matches multiple locations; the edit ambiguates globally instead of targeting a specific resolved range.
- `oldtext_not_found` — file drift or copied-text mismatch; we retry fragile text instead of failing fast with a clear stale-file reason.
- `file_modified` — file changed after read or partial apply; rejection is vague.
- `out_of_range` — line drift outside recorded read ranges; no explicit range-coverage assertion.

Each of these failure modes has a clean fix at the *internal mutation* layer: validate `expectedFileSeq` / `expectedHash` / `range` before writing, fail with structured reasons + current state when stale.

This phase explicitly is **NOT** a new public `lens_edit` tool. The native `edit` tool path stays intact — read expansion, read-guard preflight, oldText autopatch, and exact-text matching remain the first line of defense for model-authored edits. The new substrate is used by pi-lens-owned mutations (partial apply, applied LSP workspace edits, actionable autofix) which already have stronger guarantees than agent text-matching.

## Proposed API

New helper module: `clients/lens-edit.ts` (or similar).

```ts
export async function applyInternalEdit(args: {
  cwd: string;
  runtime: RuntimeCoordinator;
  cacheManager: CacheManager;
  filePath: string;
  source: \"partial-apply\" | \"lsp-edit\" | \"autofix\";
  expectedFileSeq?: number;
  expectedHash?: string;
  edits: Array<{
    startLine: number;
    endLine: number;
    newText: string;
    expectedText?: string;     // optional: exact-span guard
  }>;
  runPostEditPipeline?: boolean;
}): Promise<ApplyInternalEditResult>;

type ApplyInternalEditResult =
  | { ok: true; changed: boolean; projectSeq: number; fileSeq: number }
  | {
      ok: false;
      reason:
        | \"stale_seq\"
        | \"hash_mismatch\"
        | \"overlap\"
        | \"expected_text_mismatch\"
        | \"out_of_read_range\";
      currentFileSeq?: number;
[...]

PR #639, opened 2026-07-14
fix(read-guard): preserve atomic mixed multi-edits by default

### LINKED ISSUE #637: bug(read-guard): mixed-validity multi-edit partially writes before reporting failure
## What happened

On pi-lens 3.8.68 (and still in 3.8.69 source), a native multi-edit call with mixed preflight results can mutate the file before returning a blocking result:

1. one `edits[i].oldText` matches;
2. another `oldText` is missing or ambiguous;
3. the read-guard calls `applyPartiallyApplicableEdits` for the matching subset;
4. pi-lens then blocks the original tool call.

The caller therefore receives an edit failure even though disk state already changed. This bypasses the host edit tool's all-target preflight and single-write semantics.

## Minimal reproduction

Given:

```ts
const a = 1;
const b = 2;
```

Call native `edit` with:

```json
{
  "edits": [
    { "oldText": "const a = 1;", "newText": "const a = 10;" },
    { "oldText": "const missing = true;", "newText": "const missing = false;" }
  ]
}
```

Current result: the call is blocked, but `const a = 10;` may already be on disk.

## Expected behavior

Mixed-validity native multi-edits should be atomic by default: if any target fails preflight, no block from the batch is written.

The existing partial-apply behavior can remain available only as an explicit opt-in for users who want it.

## Proposed implementation

- default partial apply to off;
- add `--lens-partial-edit-apply` and global config `edit.partialApply: true` as explicit opt-ins;
- remove the preflight helper's unconditional "was applied" wording;
- report atomic blocks as leaving the file unchanged;
- if an opted-in partial-apply/post-edit pipeline throws, report disk state as uncertain and require a re-read;
- add integration coverage asserting the file bytes in both default and opt-in modes.

This is compatible with the internal mutation-substrate direction in #144: the substrate may make an opted-in partial apply safer, but it should not silently change the native agent edit contract by default.

LovelyLoong:
## Summary

Preserve the native edit tool's all-or-nothing contract when pi-lens preflight finds a mixed-validity multi-edit:

- default to withholding every matching block if any `oldText` target fails;
- retain legacy partial application only behind explicit `--lens-partial-edit-apply` or `edit.partialApply: true` opt-in;
- remove preflight wording that claimed matching edits had already been applied;
- report opted-in post-write failures conservatively, requiring a re-read because disk state may have changed;
- cover default and opt-in disk bytes in integration tests, plus config/flag/message wiring;
- document the new invariant and opt-in.

Closes #637
Refs #144

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [x] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [x] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes (`npm test -- --maxWorkers=2`: 3,649 passed, 3 skipped)

## What changed and why

The gate intentionally sits at the `tool_call` site that invokes `applyPartiallyApplicableEdits`. This keeps the partial-apply substrate available for explicit opt-in without coupling consent to read-guard mode, changing lower-level mutation primitives, or weakening the host edit tool's default preflight semantics.

If the opted-in mutation path throws, the handler no longer falls through to wording that implies no write occurred. It reports the disk state as uncertain and directs the caller to re-read.

## Verification

- `npm run build`
- `npm run lint`
- `npm test -- --maxWorkers=2` — 343 files passed, 1 skipped; 3,649 tests passed, 3 skipped
- `npm run build:dist`
- `npm run check:lockfile`
- `npx vitest run tests/index-integration.test.ts tests/index-wiring.test.ts tests/clients/lens-config.test.ts tests/clients/read-guard-tool-lines.test.ts` — 116 passed
- Fresh-process smoke test confirmed the local package exposes `--lens-partial-edit-apply`
- Real Pi host regression: mixed-validity native multi-edit returned `🔒 ATOMIC EDIT` and left file bytes unchanged; an all-valid two-block multi-edit still applied both blocks

On this Windows host, Vitest's default high parallelism exposed unrelated timing/resource-contention failures in existing LSP, MCP, source-walk, formatter, and pipeline tests. Bounding workers to 2 produced a clean full-suite run; the previously flaky LSP and ast-grep files also passed independently.


LovelyLoong:
Withdrawing this contribution. Apologies for the noise.

Issue #694, opened 2026-07-17
Review graph materializes gitignored compiled .js twins as file nodes (198 on pi-lens) — import resolution should prefer source twins


apmantza:
## Symptom (dogfooding /lens-map on pi-lens itself, 2026-07-17)

Of the 500 file nodes the #679 map rendered for pi-lens, **216 were compiled `.js` files — 198 of them in-place build twins of `.ts` sources** (`clients/actionable-warnings.js` alongside `clients/actionable-warnings.ts`, …). These are `.gitignore`d build artifacts (`*.js` is ignored at repo root), yet they exist as first-class review-graph file nodes carrying import edges.

## Why it matters beyond the map

#679 now merges the twins at its own aggregation layer (map-level canonicalization), but the graph itself is what `module_report`, `blastRadius` (#304), the impact cascade, and word-index symbol ranking priors consume. Compiled twins mean:
- transitive-dependents/blast-radius counts inflated or split across the twin pair,
- `module_report` who-uses-this lists polluted with artifact paths,
- centrality/ranking priors diluted,
- the graph violating the scan-exclusion invariant (#243/#250/#252/#253: every file-collection path routes through the shared ignore matcher) for any project that builds in place.

## Suspected mechanism (unverified — first investigation step)

Source files use ESM `./foo.js` specifiers. `resolveImportToFiles` (clients/review-graph/import-resolvers.ts) appears to resolve the specifier to the compiled `foo.js` **when it exists on disk**, instead of preferring the `foo.ts` source twin — so on a repo with in-place builds every import edge materializes an artifact node. On repos without in-place output the same code lands on `.ts`, which is why this stayed invisible. (#301 was supposed to map `.js` specifiers to TS sources for module_report's cold path — verify whether the graph builder path diverged from that.)

Note git semantics for the fix: a **tracked** `.js` file (vendored deps, pure-JS projects) is NOT ignored even when a `.gitignore` pattern matches it — any ignore-matcher-based exclusion must not drop tracked files. Twin-preference (`.js`→`.ts`/`.tsx`, `.mjs`→`.mts`, `.cjs`→`.cts` when the source twin exists) is the safer primary rule; ignore-matcher routing is the belt-and-braces.

## Acceptance criteria
- [ ] Import resolution prefers the source twin when both `X.ts(x)` and compiled `X.js` exist (language-uniform: mjs/cjs families too)
[...]

Issue #504, opened 2026-07-10
Phase 3: user-summoned annotated-diff review window via glimpseui (conditional GO from 2026-07-10 spike)


apmantza:
## What

An explicitly user-summoned native window rendering pi-lens's findings as an annotated diff — diagnostics as clickable annotations over the changed code, accepted feedback staged back into pi's editor. The rich-surface complement to the transcript summary (#484/PR #500) and its renderDiff expansion (#503); the data feed is shared with #502.

## Spike verdict (2026-07-10, full report in session): CONDITIONAL GO

**glimpseui** (MIT, 257KB, zero npm deps, github.com/hazat/glimpse) — native WebView child process, JSON Lines over stdin/stdout.

What checked out:
- **Zero network surface**: no HTTP server, no localhost port — pipes only. Materially better security posture than any browser/Electron-class option.
- **Orphan story TESTED clean on Win11**: clean open/message/close with no residual process; AND on simulated ungraceful parent exit the native host self-terminates on stdin EOF — no orphan (#472-class concern pre-answered). Hard-kill (SIGKILL ancestor) untested — if we want a hard guarantee, attach the child to our process-reaper/Job-Object infra.
- **Prior art proves the pi shape**: @diegopetrucci/pi-annotate-git-diff — `registerCommand` + `ctx.mode !== "tui"` gate + `session_shutdown` close hook + `ctx.ui.pasteToEditor` (stage, never auto-send) + content through Monaco text models (never innerHTML — the XSS-from-repo-content mitigation to copy verbatim).

The conditions:
1. **Install-time native compilation** — glimpseui ships NO prebuilt binaries; postinstall compiles the host (.NET 8 SDK on Windows / swiftc on macOS / Rust+GTK or Chromium-CDP fallback on Linux) and silently no-ops without the toolchain; `open()` then throws (catchably) at first use. ⇒ MUST be a lazily-probed optional capability, never required; missing-toolchain message must be clear (upstream `buildHint` strings usable).
2. **No built-in availability preflight** — build our own probe: native-host binary exists (`getNativeHostInfo().path`) + `ctx.mode === "tui"` + env heuristics (`SSH_CONNECTION`, missing `DISPLAY`/`WAYLAND_DISPLAY` on Linux). Linux/Chromium fallback does zero DISPLAY checking itself.
[...]

Issue #306, opened 2026-06-24
Repo-calibrated thresholds for structural-metric checks (avoid the absolute-threshold death-spiral)


apmantza:
## Problem
Any structural-smell check built on an **absolute constant threshold** tends to either flood with idiom false-positives or go silent — there's rarely a single global number that works across repos. We've already lived this: the old similarity runner died exactly here (see #128 — `0.98` silences everything, `0.95` floods). Hard-coded structural thresholds (function length, branch count, nesting depth, params, returns, methods-per-class, imports-per-file, …) would inherit the same failure mode.

## What we already have
- Review-graph cyclomatic complexity; `module_report` flags (`high complexity` ≥8, `high fanout` ≥4).
- Session-start **complexity baselines** + error-debt baseline.

So the calibration concept already exists narrowly — this generalizes it.

## Proposal
When adding structural-metric checks, **calibrate each metric's threshold to the repo's own distribution** (e.g. clamp to ~95th percentile of the current codebase) computed once at session_start, instead of a single global constant. Persist alongside the existing complexity baselines; allow explicit per-metric overrides via config.

This makes a structural-metrics family viable without the tuning death-spiral, and is the calibration discipline any future structural runner (including #128's similarity thresholds) should adopt.

## Acceptance
- [ ] Calibration helper: derive a per-metric threshold from the repo's metric distribution (percentile-clamp), pure + unit-tested
- [ ] Session-start integration; persisted with existing baselines under the project cache
- [ ] Config override path per metric
- [ ] Any new structural check consumes the calibrated threshold, not a constant
- [ ] Referenced by #128 as the threshold-setting approach for the similarity runner

apmantza:
Folded into #1456 (Umbrella: consolidate pi-lens's tech-debt signal work). Your scope — the calibration helper, session-start integration, and per-metric config override — is now Slice 1 there, sequenced first since #172's own design thread put calibration ahead of the baseline and aggregator work. Nothing dropped.

— consolidation by AI, maintainer-directed

Issue #260, opened 2026-06-20
Review-graph persistence is O(total graph size) per turn → host OOM; move to delta persistence


apmantza:
## Problem

The review graph is a **monolithic blob** — one object in memory, persisted as one `JSON.stringify` to one `review-graph.json`. So every lifecycle operation is **O(total graph size)**, not O(changed), and the graph has grown to **13.5 MB** on the pi-lens repo. The result is a host-killing `Fatal process out of memory: Zone` (V8 zone allocator) during active editing, with `DiyFp::Multiply` frames in the trace — the signature of number-heavy `JSON.stringify`/`parse` (every line number / complexity / fanout in the blob).

This persists even after module_report was made read-only (#256), because the OOM was never the report path — it's the **build/persist cycle**.

### Evidence
- `.pi-lens/cache/review-graph.json` = **13.5 MB**.
- `persistGraph` (`builder.ts:491`) does a **synchronous `JSON.stringify` of the entire graph** on the main thread.
- Called from 5 sites on the build/update hot path (`builder.ts:1089, 1104, 1156, 1171, 1201`) — i.e. **every file-changing turn**.
- Crash dump shows a concurrent host `tsc` type-check (`typescript:2322`) — the edit pipeline, overlapping the stringify.
- (Ruled out: orphan 49 MB `.pi-lens/index.json` written by `project-index.js`, which has **zero importers** — dead code, safe to delete, not the bug. `project-snapshot.json` is only 542 KB — `symbolSearch`'s stateless re-parse is wasteful but not an OOM driver.)

## Caller audit (`buildOrUpdateGraph`)

| Caller | `changedFiles` | FactStore | Verdict |
|---|---|---|---|
| `dispatch/integration.ts:590` (per write/edit cascade) | `[file]` | session | Incremental parse ✓ |
| `dispatch/runners/tree-sitter.ts:55` | `[file]` | session | Incremental parse ✓ |
| `runtime-session.ts:652` (session_start call-graph) | `[]` | fresh | Full build, guarded (cache-stale only), once/session |
| `lens-engine.ts:184` — `symbolImpact` / `pilens_impact` | `[]` | **fresh** | **read path that triggers a BUILD — bug** |

[...]

PR #696, opened 2026-07-17
feat: /lens-map — interactive HTML project map from the review graph (closes #679)

### LINKED ISSUE #679: feat: /lens-map — render the review graph as an interactive HTML project map
## What

A new human-facing command, `/lens-map` (name open to bikeshedding), that renders the existing review graph as a self-contained, interactive HTML file the user can open in a browser — a visual map of the project rather than a text report.

## Why now

Came out of a design discussion (2026-07-16) about UX for the debt/complexity work (#172/#306/#329): those are all text/TUI-oriented, but there's real value in a visual artifact too, separate from the debt/metrics report itself since it's a different data source (the review graph, not the debt/metric time series).

## Data source

The existing review graph (`clients/review-graph/builder.ts`, already collision-safe IDs + qualified names + resolution tiers per #659/#666) already has everything needed:
- Nodes: files (function-level granularity would likely explode visually at project scale — start file-level, consider drill-down as a v2).
- Edges: `imports`/`calls` relationships already computed.
- `module_report`'s existing `blastRadius` (transitive dependents, #304) is a free "how load-bearing is this file" signal — candidate for node sizing.
- Once #306 (repo-calibrated thresholds) lands, node color could reflect a file's calibrated complexity/fan-out standing (repo-relative, not an absolute threshold) — so the map visually agrees with whatever the debt/complexity commands are flagging elsewhere. Not a hard dependency — the map is useful with just structural edges even before #306 exists — but color-coding should adopt the calibrated scale once available rather than inventing its own absolute thresholds.

## UX reference

The sibling `pi-visual-recap` extension (a separate, already-shipped extension on this machine, not part of pi-lens) has a proven, reusable HTML-artifact pattern worth forking rather than inventing fresh: self-contained single file, `:root` CSS custom properties, real `@media (prefers-color-scheme: dark)` support, `--accent: #2563eb` light / `#60a5fa` dark (already matches pi-lens's blue brand convention), a hero section + embedded JSON payload for client-side interactivity, strict CSP meta tag. pi-lens has zero HTML-output commands today (`/lens-tdi`/`/lens-health`/`/lens-booboo` are all markdown/TUI text) — this and the #172 `format=html` addition would be the first.

## Open questions (for design, not blocking filing)
- Force-directed layout vs. a simpler grouped/clustered layout (force-directed graph libraries exist via CDN — this command isn't bound by the Artifact-publishing tool's strict self-contained/no-external-request CSP, since the file is written to disk and opened locally, not published — but still worth considering a zero-dependency layout if feasible for offline use).
- Where the file gets written (a `.pi-lens/reports/` or similar convention, analogous to `visual-recap`'s `.visual-recaps/` output dir).
- Whether this needs an agent-facing tool counterpart (mirroring `/lens-debt`'s planned human-command + agent-tool pairing) or stays human-only, since a visual map isn't obviously useful to an agent the way a score+delta is.

apmantza:
## What

`/lens-map` (#679): a human-summoned command rendering the review graph as a self-contained, interactive HTML project map written to `<project-data>/reports/lens-map.html`.

## Design (the issue's three open questions, settled)

- **Zero-dependency layout** — deterministic force simulation (repulsion + edge springs + center gravity, FNV-1a path-hash-seeded) computed in Node at generation time; the client is an embedded JSON payload + vanilla-JS SVG with pan/zoom/hover-neighbor-highlight/click-details. No CDN, works fully offline.
- **Output location** — `getProjectDataDir(cwd)/reports/lens-map.html`; the command notifies the path, never auto-opens a browser.
- **Human-only** — no agent tool, no MCP mirror. Node color is a neutral brand-blue dependents scale; calibrated complexity coloring explicitly deferred to #306.

## Aggregation (symbol graph → file map)

- File-level nodes sized by transitive dependents (same aggregation family as `module_report`'s blastRadius); weighted file→file edges; externals counted in the header, not drawn.
- **Compiled-twin canonicalization**: `X.js` merges into `X.ts`/`X.tsx` (`.mjs`→`.mts`, `.cjs`→`.cts`) when both exist — edges remap, never drop. On pi-lens itself this took the map from 500 nodes (truncation cap engaged) to **369 nodes / 1,370 edges, no truncation** — 300 twins merged. Graph-level fix for the underlying artifact pollution is tracked as #694.
- **Test files excluded** belt-and-braces (the graph already drops them by role since #260); both exclusion counters render only when nonzero.
- 500-node degree-ranked truncation cap (env `PI_LENS_MAP_MAX_NODES`), ranked pre-cut so the cut can't bias itself, with a visible in-page note.

## Security

Repo-derived strings (file paths) are attacker-influenceable in cloned repos: the payload is JSON-embedded with unicode escaping of `<`/`>`/`&` (no substring can close the script tag), rendered client-side exclusively via `textContent`/`createElementNS`, under a strict CSP meta tag — the #504-spike XSS mitigation. A test asserts a malicious `</script>` filename cannot break out.

Also in this PR: `review-graph/builder.ts`'s literal-NUL key separators rewritten as `` escape sequences (same semantics; the raw bytes made grep/ripgrep treat the file as binary — lens-map.ts uses the escaped form from the start).

## Test plan
- [x] 12 tests: aggregation (twin merge + edge remap + self-edge drop, test-file exclusion + ranking isolation, external counting, dependents BFS, degree-ranked truncation), layout (determinism, finite coords, empty graph), rendering (script-breakout escaping, CSP + both theme accents, no external hosts), end-to-end file write under a faked data dir
- [x] `tests/index-wiring.test.ts` updated for the new command
- [x] build + lint clean; full `npm test` on the rebased branch: 3900 passed, 2 known-flaky (biome-client-cwd, lsp/lifecycle) — both pass in isolation (7/7)
- [x] Dogfooded on pi-lens itself (369 nodes / 1,370 edges, ~400 KB HTML, interactivity verified)
- [x] CHANGELOG entry included; AGENTS.md documents the command

Closes #679. Refs #694, #306, #260, #504.


apmantza:
Third commit `53d1e005`: untracked-gitignored files are now excluded from the map (git semantics honored — tracked files matching an ignore pattern, e.g. vendored `clients/deps/*.js`, stay). Twin-merge takes precedence over exclusion so an ignored compiled twin merges (edges preserved) instead of vanishing; only twin-less ignored orphans drop, surfaced as a hidden-when-zero `ignoredFileCount`. Degrades gracefully to no-filter when git is unavailable. Dogfood: 369 → 367 files (the 2 orphans), all 7 tracked vendored deps retained, `ignoredFileCount: 2`. Full suite re-run: 3902 passed, 3 known-flaky (pass in isolation).

Issue #298, opened 2026-06-22
bug: Lens diagnostics mode=full not honoring ignore and rules config


carolitascl:
Running pi-lens diagnostics manually, it runs in mode=full, and  does not honor ignore patterns in pi-lens.json or ~/.pi-lens/config.json files

Config file
```
{
  "ignore": [
    "**/__tests__/**",
    "**/*.test.ts",
    "fixtures/**",
    "vendor/**",
    "**/.*/**",
    "**/.*",
    "**/.history/**",
    "pi-session-*.html"
  ],
  "rules": {
    "high-complexity": { "threshold": 25 },
    "high-fan-out": { "threshold": 30 }
  }
}
```

```
run lens diag
```

```
 lens_diagnostics
 pi-session-2026-06-20T19-43-05-630Z_019ee68e-e79e-7809-9518-643e1111eb19.html  🔴 7 blocking  1W
   🔴 L1177: Expression expected. [javascript]
   🔴 L1241: Declaration or statement expected. [javascript]
   🔴 L2420: ',' expected. [javascript]
   🔴 L2430: ';' expected. [javascript]
   🔴 L2431: ',' expected. [javascript]
   🔴 L2454: ',' expected. [javascript]
   🔴 L2456: Declaration or statement expected. [javascript]
   L626: Do not use empty rulesets [css:emptyRules]
 .history/plugin_20260614161438.ts  🔴 4 blocking
   🔴 L2: Cannot find module './reader.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L8: Cannot find module './service.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L9: Cannot find module './selector.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L10: Cannot find module './hint.ts' or its corresponding type declarations. [typescript:2307]
 .history/plugin_20260614162357.ts  🔴 4 blocking
   🔴 L2: Cannot find module './reader.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L8: Cannot find module './service.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L9: Cannot find module './selector.ts' or its corresponding type declarations. [typescript:2307]
   🔴 L10: Cannot find module './hint.ts' or its corresponding type declarations. [typescript:2307]
 .history/plugin_20260614162410.ts  🔴 4 blocking
   🔴 L2: Cannot find module './reader.ts' or its corresponding type declarations. [typescript:2307]
[...]

PR #720, opened 2026-07-20
fix(widget): reconcile stale entries on the TUI render path (closes #719)

### LINKED ISSUE #719: Bug: Widget/TUI shows stale diagnostics - fixed errors persist
## Symptom

The pi-lens widget (the line under the editor) keeps showing diagnostics for files the agent has **already fixed**. The agent sees the stale error in the TUI and wastes turns re-investigating ("but wait, those are explicitly fixed…").

The error only clears if the agent manually runs `lens_diagnostics` (any mode) or re-touches the exact file through pi-lens's per-edit pipeline.

## Root cause

The widget's diagnostic store is the `files` map in `clients/widget-state.ts`. It is only ever *corrected* in two places:

1. The per-edit dispatch pipeline calls `recordDiagnostics(...)`
   (`clients/pipeline.ts`, after `dispatchLintWithResult`) with a fresh (possibly empty) result.
2. `reconcileStaleWidgetFiles()` drops any entry whose underlying file changed on disk **after** it was last recorded (`st.mtimeMs > rec.touchedAt + 1`) - i.e. diagnostics the agent already fixed/edited away.

`reconcileStaleWidgetFiles` is the *designed* self-correction for exactly this bug, but it is **stranded**: it is only invoked from the `lens_diagnostics` tool (`tools/lens-diagnostics.ts:317`). The widget render path (`renderWidget` → mounted via `mountLensWidget` in `index.ts`) **never calls it**, so `renderWidget` renders the cached `files` map verbatim. Any entry whose file was fixed/rewritten without a corresponding re-confirmation (a cross-file fix, an external edit, or a missed write event) stays on screen indefinitely.

Empirically confirmed: feeding a "fixed" file (mtime newer than the record's `touchedAt`) into `reconcileStaleWidgetFiles` drops the stale entry correctly — the function works; it just isn't wired into the render path.

## Impact

- Agent over-thinks / re-fixes already-fixed errors → wasted turns.
- Trust in the widget erodes; users start ignoring it (then miss real errors).

## Workaround

Run `lens_diagnostics` (any mode) to force a reconcile, or ensure the fixed file is re-touched through pi-lens's per-edit pipeline.

### Steps to reproduce

1. Open a source file (e.g. `src/app.ts`) that has a real diagnostic; the widget shows it.
2. Fix the error by editing/saving the file — but the fix is a *cross-file* fix, an external edit, or the per-edit write event was missed, so pi-lens's per-edit pipeline never re-confirms this exact file.
3. Observe the widget: it still shows the original error.
4. Run `lens_diagnostics` (any mode) → the error disappears.

(Minimal repro is a file path + the `lens_diagnostics` tool invocation in step 4. The fix in step 2 need not go through pi-lens.)

### Expected behavior

When a file's content changes on disk after its diagnostics were recorded (`mtimeMs > touchedAt`), the widget should re-confirm and drop the stale entry on its own — without the agent having to run `lens_diagnostics` by hand.

### pi-lens version

3.8.70 (any `>=` #298)

### Relevant logs

  ```log
  # ~/.pi-lens/bus-events.log — the widget's diagnostic feed (pilens:diagnostics bus)
  {"ts":"2026-07-14T20:50:40.066Z","event":"pilens:diagnostics","outcome":"emitted","cwd":"/Users/ymek/.openclaw/workspace/job-hunt","fileCount":1,"seq":1}
  {"ts":"2026-07-14T20:51:45.911Z","event":"pilens:diagnostics","outcome":"emitted","cwd":"/Users/ymek/.openclaw/workspace/job-hunt","fileCount":1,"seq":2}
  ... (bus keeps emitting the diagnostics the widget renders)
  {"ts":"2026-07-14T20:54:38.682Z","event":"pilens:files:touched","outcome":"emitted","cwd":"/Users/ymek/.openclaw/workspace/job-hunt","reason":"autofix","fileCount":1}

  # ~/.pi-lens/sessionstart.log — widget state survives a reload with NO render-path reconcile
  [2026-07-19T16:03:20.265Z] session_start: reload — keeping widget state

  # ~/.pi-lens/sessionstart.log — the actual reproduction (repro-widget.mjs dispatched + linted; render path never reconciles)
  [2026-07-19T16:17:21.126Z] tool_result: tracking turn state for write on /Users/ymek/.pi/repro-widget.mjs
  [2026-07-19T16:17:21.128Z] autofix: policy for /Users/ymek/.pi/repro-widget.mjs -> biome (preferred: biome, gate: smart-default, safe: yes)
  [2026-07-19T16:17:21.182Z] autofix: attempted biome for /Users/ymek/.pi/repro-widget.mjs, but no fixes were applied
  [2026-07-19T16:17:21.182Z] dispatch: running lint tools for /Users/ymek/.pi/repro-widget.mjs

  # ~/.pi-lens/sessionstart.log — contrast: the persisted snapshot IS reconciled for staleness (#298); the widget render path is not
  [2026-07-19T16:03:19.597Z] project_snapshot: miss reason=stale(seq=1471, current=1483)
  ```

### Area

area:lsp

ymek:
## Summary

Closes #719. `reconcileStaleWidgetFiles()` (introduced in #298) drops widget entries whose file changed on disk after they were last recorded (diagnostics the agent already fixed), but it was only ever wired into the `lens_diagnostics` tool — never the widget render path. renderWidget therefore rendered cached diagnostics verbatim and kept showing fixed errors until `lens_diagnostics` was run by hand, causing the agent to over-think and re-fix already-fixed code.

This PR adds a debounced `scheduleStaleReconcile()` and call it from mountLensWidget's render callback + render wrapper so stale entries self-correct within ~1.5s of a fix. The debounce is unref-ed so it never holds the process open.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [ ] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

### `clients/widget-state.ts`
Add a debounced scheduler next to `reconcileStaleWidgetFiles`:

```ts
let staleReconcileTimer: ReturnType<typeof setTimeout> | null = null;
export const STALE_RECONCILE_DEBOUNCE_MS = 1500;
export function scheduleStaleReconcile(): void {
  if (staleReconcileTimer !== null) return;
  staleReconcileTimer = setTimeout(() => {
    staleReconcileTimer = null;
    void reconcileStaleWidgetFiles().catch(() => {});
  }, STALE_RECONCILE_DEBOUNCE_MS);
  staleReconcileTimer?.unref?.();
}
```

### `index.ts`
Import it and call it from `mountLensWidget`'s render callback + `render`
wrapper:

```ts
setRenderCallback(() => {
  scheduleStaleReconcile();
  tui.requestRender();
});
return {
  render: (width: number) => {
    scheduleStaleReconcile();
    return renderWidget(width, theme);
  },
  invalidate: () => setRenderCallback(() => {}),
};
```

The debounce (1.5 s, `unref`-ed so it never holds the process open) collapses the burst of renders that accompany a save into a single sweep. Dropped entries re-trigger a render via `reconcileStaleWidgetFiles` → `requestRenderFn`.

## Verification

- **Unit (new regression test):** `tests/clients/widget-state.test.ts` records an error, bumps the file's mtime past `touchedAt` (simulating a fix the pipeline didn't re-confirm), calls `scheduleStaleReconcile()`, advances fake timers, and asserts the entry is gone.
- **Source compile + behavior:** the edited `clients/widget-state.ts` was bundled with esbuild and exercised with real timers — `scheduleStaleReconcile` drops the stale entry after the debounce. `node --check` on the bundled runtime passes.
- `reconcileStaleWidgetFiles` itself was already covered by `tests/clients/session-state-store.test.ts` (lens_diagnostics path); this PR adds coverage for the render-path wiring.

## Residual caveat (not addressed here)

This fixes the dominant "file changed on disk but the widget wasn't re-confirmed" case. If a fixed error *still* persists after this lands, the remaining cause is the LSP path returning pre-fix diagnostics as "confirmed" on the fix-edit (global `diagnosticsVersion` counter early-return in `clients/lsp/client.js`; version-less servers; reopen-on-resync servers that skip the cache-clear). Those record `touchedAt > mtime`, so they are out of scope for a staleness-based reconcile and need a separate fix.

## Files changed

- `clients/widget-state.ts` — add `scheduleStaleReconcile()` + `STALE_RECONCILE_DEBOUNCE_MS`.
- `index.ts` — import + wire into `mountLensWidget` render path.
- `tests/clients/widget-state.test.ts` — regression test.

### DIFF
--- a/clients/widget-state.ts
+++ b/clients/widget-state.ts
@@ -382,6 +382,28 @@ export async function reconcileStaleWidgetFiles(): Promise<number> {
 	return dropped;
 }
 
+/**
+ * Keep the TUI honest (#298 follow-up). `reconcileStaleWidgetFiles` drops
+ * widget entries whose file changed on disk after they were last recorded
+ * (i.e. diagnostics the agent already fixed) — but it was only ever wired
+ * into the `lens_diagnostics` tool, so the widget rendered cached diagnostics
+ * verbatim and kept showing fixed errors until `lens_diagnostics` was run by
+ * hand. This debounced scheduler fires it from the widget render path (see
+ * `mountLensWidget` in index.ts) so stale entries self-correct. The debounce
+ * collapses the burst of renders that accompany a save into a single sweep.
+ */
+let staleReconcileTimer: ReturnType<typeof setTimeout> | null = null;
+export const STALE_RECONCILE_DEBOUNCE_MS = 1500;
+export function scheduleStaleReconcile(): void {
+	if (staleReconcileTimer !== null) return;
+	staleReconcileTimer = setTimeout(() => {
+		staleReconcileTimer = null;
+		void reconcileStaleWidgetFiles().catch(() => {});
+	}, STALE_RECONCILE_DEBOUNCE_MS);
+	// Don't keep the process alive solely for this background sweep.
+	staleReconcileTimer?.unref?.();
+}
+
 /** Summary of current diagnostic counts across all files in the widget. */
 export interface FileDiagnosticSummary {
 	filePath: string;
--- a/index.ts
+++ b/index.ts
@@ -14,6 +14,7 @@ import {
 	importWidgetState,
 	type PersistedWidgetState,
 	renderWidget,
+	scheduleStaleReconcile,
 	setRenderCallback,
 } from "./clients/widget-state.js";
 import { selectLspStatus } from "./clients/lsp-status.js";
@@ -475,9 +476,15 @@ export default function (pi: ExtensionAPI) {
 		setWidget(
 			"pi-lens",
 			(tui: LensWidgetTui, theme: LensWidgetTheme) => {
-				setRenderCallback(() => tui.requestRender());
+				setRenderCallback(() => {
+					scheduleStaleReconcile();
+					tui.requestRender();
+				});
 				return {
-					render: (width: number) => renderWidget(width, theme),
+					render: (width: number) => {
+						scheduleStaleReconcile();
+						return renderWidget(width, theme);
+					},
 					invalidate: () => setRenderCallback(() => {}),
 				};
 			},
--- a/tests/clients/widget-state.test.ts
+++ b/tests/clients/widget-state.test.ts
@@ -1,7 +1,8 @@
 import * as fs from "node:fs/promises";
+import * as os from "node:os";
 import * as path from "node:path";
 import { visibleWidth } from "@earendil-works/pi-tui";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import {
 	__testing,
 	clearWidgetState,
@@ -767,7 +768,8 @@ describe("reconcileScanDiagnostics — full-scan/on-demand footer reconciliation
 describe("scheduleStaleReconcile — widget self-corrects fixed files (#298 follow-up)", () => {
 	it("drops a widget entry once its file is edited on disk after the last record", async () => {
 		vi.useFakeTimers();
-		const filePath = path.join(process.cwd(), `stale-reconcile-${Date.now()}.ts`);
+		const tmpDir = await fs.mkdtemp(path.join(os.tmpdir(), "stale-reconcile-"));
+		const filePath = path.join(tmpDir, `stale-reconcile-${Date.now()}.ts`);
 		try {
 			await fs.writeFile(filePath, "const x = 1;\n");
 			// Pipeline records a real error for the file.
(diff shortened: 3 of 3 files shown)


apmantza:
many thanks for taking the time @ymek. merged

Issue #699, opened 2026-07-17
perf: startup scan-context re-walks the entire monorepo on every session start, even though the result (canWarmCaches=false) is never usable


aaronkyriesenbach:
## What happened?

`resolveStartupScanContextAsync()` re-walks the entire resolved project root on every `pi` process start, even when the outcome (`canWarmCaches=false`, reason `too-many-source-files`) is discarded immediately and never persisted. In a large monorepo this walk costs ~17s on every single startup, with zero benefit since the result can't be used.

Bisected by toggling installed packages one at a time in `settings.json`: with only pi-lens enabled, `pi -p "..." --no-session` took ~20s; every other installed package took 2-4s; with none, 2.2s. `~/.pi-lens/sessionstart.log` confirms:
```
warmup: scan-context done in 17086ms (canWarm=false)
warmup: skipping language-profile (canWarm=false, reason=too-many-source-files)
```
`--no-lens` does not suppress this scan.

Unlike the `canWarmCaches=true` path (which persists a `project-snapshot.json` under `~/.pi-lens/projects/<hash>/cache/`), the `too-many-source-files` verdict has no on-disk cache, so it's recomputed from scratch every process start. This is distinct from #622/#250, which are about the walk escaping to `$HOME`/an unrelated tree — here the resolved root is a legitimate project root, just a large one, so that guard doesn't apply.

## Steps to reproduce

1. Open a large monorepo (single `.git` at the repo root, well over `MAX_STARTUP_SOURCE_FILES` (2000) tracked source files, normal `.gitignore`).
2. `cd` into a subdirectory several levels deep that has no `.git`/`package.json` of its own (so `findNearestProjectRoot` walks up to the monorepo's `.git`).
3. Run `pi -p "hi" --no-session` and time it.
4. Repeat — the same ~17s scan-context cost recurs every time.

## Expected behavior

The `too-many-source-files` verdict should be cached per resolved `projectRoot` (e.g. keyed on root path + a cheap freshness check) so repeat session starts in the same large repo skip the walk instead of recomputing and discarding it every time, mirroring how the `canWarmCaches=true` case already persists a snapshot.

## pi-lens version

3.8.69 (`pi` 0.80.6, Node 22.22.0, macOS arm64)

## Relevant logs

```
[warmup] warmup: starting background warmup
[warmup] warmup: scan-context done in 17086ms (canWarm=false)
[warmup] warmup: skipping language-profile (canWarm=false, reason=too-many-source-files)
```

[...]

PR #759, opened 2026-07-22
perf(startup-scan): bound the source-count walk by total entries (closes #758)

### LINKED ISSUE #758: PERF: Performance Drag on Large Repos with a mix of source files + non-source files
### What happened?

I run a mod for Hearts of Iron IV and I use pi-lens for my other repos. The issue I ran into was when I was using it with smaller repos or that I had a large amount of lag time when trying to start pi with pi-lens due to the way the file walk works. 

It early skips after 2000 (.py/.ts etc) for those so the pi startup is not super laggy. I only have 300 scripts in my repo but I have a large sum of other files due to the mods nature. 

Claude Report:
>   - The culprit is pi-lens's scan-context phase: a full-tree source-file count walk
  (resolveStartupScanContext → countSourceFilesWithinLimit), which runs synchronously on the 
  blocking session_start path (runtime-session.js:1009).
  - The trap: that walk early-exits only when it finds >2000 source files
  (.ts/.js/.py/.go/.rs/.rb). Millennium-Dawn has only 301, so it never exits early and walks
  all 84k entries, dominated by ignoreMatcher.isIgnored() per file.

Adding a .pi-lens.json works around the issue but this might be something worth exploring for someone with more experience to try to fix this. Please let me know if there is anything I can do differently or expound on here.

### Steps to reproduce

1. Have a large source repo with a mix of code types (in my case .txt files with python scripts
2. Start your choice of harness (pi-coding-agent)
3. You can visibly see the slowdown with the files 

### Expected behavior

_No response_

### pi-lens version

3.8.71

### Relevant logs

_No response_

### Area

None

JSap0914:
## Summary

Bounds the startup source-file count walk by total directory entries, not just by source-file count, so a repo with few source files but a huge pile of non-source files can't drag `session_start`.

Closes #758

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:perf
- [x] area:session

## What changed and why

`computeStartupScanContext` / `resolveStartupScanContextAsync` (`clients/startup-scan.ts`) count source files under the project root to decide `canWarmCaches`. The walk early-exited **only** once it had seen more than `MAX_STARTUP_SOURCE_FILES` (2000) `.ts/.js/.py/.go/.rs/.rb` files.

A repo with **few** source files but a huge pile of non-source files never trips that exit — the reporter's case was a Hearts of Iron IV mod: ~300 scripts among ~84k data/`.txt` entries. So the walk enumerated the **entire tree**, dominated by one `ignoreMatcher.isIgnored()` call per entry, blocking the cold-start `session_start` path for seconds. (This matches the reporter's own root-cause note pointing at `countSourceFilesWithinLimit`.)

Fix — add a second, independent bound to the walk:

- New `MAX_STARTUP_SCAN_ENTRIES` (50000, env-tunable via `PI_LENS_STARTUP_SCAN_MAX_ENTRIES`) total-entry ceiling. This is the same finite-walk-cap pattern already guarding `jscpd-client.ts`'s `hasSourceFilesRecursive` (`MAX_ENTRIES = 6000`).
- When a walk hits the ceiling before crossing the source-file limit, the tree is large and dominated by non-source files, so it returns a new `too-many-entries` verdict (`canWarmCaches: false`) rather than trusting the partial count and warming caches on — then re-walking — an enormous tree.
- `too-many-entries` is TTL'd by `isStartupScanVerdictFresh` exactly like `too-many-source-files` (#699), so a shrunk repo auto-recovers on the next session.

Kept minimal and non-regressive:

- The public `countSourceFilesWithinLimit` / `countSourceFilesWithinLimitAsync` wrappers keep their **pre-existing** contract (only the source-file limit bounds them). The entry ceiling applies solely to the startup-scan verdict path, via the shared `walkSourceCount` / `walkSourceCountAsync` cores that both the wrappers and `computeStartupScanContext*` now call.
- A generous default (50000 = 25× the source limit) means healthy warm-eligible projects (≤2000 source files) are unaffected; only pathological mixed/large trees trip the new bound.

## Tests

New `tests/clients/startup-scan-entry-budget.test.ts`:

- Drives the `too-many-entries` verdict deterministically with a tiny mixed-repo fixture (`.git` root + 3 `.ts` + a pile of `.txt`) and a small `maxScanEntries` option — sync and async.
- Confirms the same tree **warms** (`canWarmCaches: true`) under a generous budget, so the cap is proven to be the mechanism (no false positives).
- Covers the env-tunable `getStartupScanMaxEntries` getter (default / override / non-finite fallback / memoization + reset).
- Pins the TTL behavior for `too-many-entries`.

## Verification

- `npm run lint` ✓
- `npm run build` / `npm run build:dist` ✓
- `npm run check:lockfile` ✓
- New suite + the adjacent startup-scan / source-walk suites green (`startup-scan-entry-budget`, `startup-scan-home-ceiling`, `startup-scan-verdict-cache`, `source-walker-equivalence`, `runtime-session-scan-cache`, `source-walk-occupancy`).


apmantza:
many thanks again @JSap0914!

Issue #775, opened 2026-07-23
Monorepo support: root-assumption audit findings (tracking)


apmantza:
Tracking issue for monorepo behavior across pi-lens, from a full static audit of root-resolution assumptions (every claim verified against code; file:line refs below are as of current master).

## Top risks (ranked)

1. **Review graph disables itself entirely above 1,000 files.** `RUNTIME_CONFIG.reviewGraph.maxFiles = 1_000` (`clients/runtime-config.ts:70`); the builder walks to `maxGraphFiles+1` then bails whole-cloth (`clients/review-graph/builder.ts:478-520`). Above the cap: zero usedBy/blastRadius/fan-in for the whole session, silently. The tightest and most consequential of the budgets.
2. **Workspace-package imports never resolve to file edges (JS/TS/Python).** `resolveJsTs` treats every bare specifier as external (`clients/review-graph/import-resolvers.ts:153-156`) — `import {x} from "@scope/pkg"` pointing at a sibling workspace package is indistinguishable from an npm dep. The workspace-manifest detection that could fix this **already exists** (`clients/review-graph/workspace-modules.ts` reads pnpm-workspace.yaml, package.json#workspaces, Cargo [workspace], go.work) but is wired only into the coarse module-level impact-cascade fallback (`query.ts:189-199`, capped 20 files/module, depth ≤4). Go and Java resolvers already handle their cross-package cases.
3. **Warm-pipeline shutdown is silent.** On a `too-many-source-files` verdict (cliff: 2,000 source files, `clients/startup-scan.ts:30`), heavy scans, TODO scan, and dominant-language LSP pre-warm all skip with only a debug-log line (`clients/runtime-session.ts:1360-1541,1573-1591`) — unlike the slow-FS probe which fires a visible notify (`:1413-1418`).
4. **Five uncoordinated size budgets** — 500 (diagnostics scanner), 1,000 (review graph), 2,000 (startup scan), 6,000 (jscpd dir entries), 6,000 (word index) — each independently re-walks and independently degrades; a large repo pays for several doomed walks before every feature has quietly turned itself off. No shared "large project" signal.
5. **`startup-scan`'s counting walk follows symlinks unconditionally with no cycle detection** (`clients/startup-scan.ts:170`), while `source-filter.ts` defaults to not following (`:315`). node_modules/.pnpm are defused by name-based exclusion (`clients/file-utils.ts:85-127`), but any other symlinked dir (or loop) is a plausible hang vector.

[...]

PR #805, opened 2026-07-25
feat(review-graph): resolve tsconfig path aliases

apmantza:
## What

- Resolve TypeScript `compilerOptions.baseUrl` + `paths` aliases to real workspace files in the review graph.
- Honor relative `tsconfig.json` `extends` chains and tolerate comments/trailing commas.
- Preserve the existing workspace-package → `external:` fallback when an alias does not resolve.
- Add regression coverage and an Unreleased changelog entry for #775.

## Approach

- Add a session-cached, per-importer-directory `parseTsconfigPaths` helper using the shared `findNearestContaining` primitive and `isAtOrAboveHomeDir` ceiling.
- Apply longest-prefix alias matching, single-`*` substitution, and mapped-target ordering.
- Reuse the existing JS/TS source-twin and extension/index probing in both the cold import resolver and warm graph builder before workspace-package matching.
- Bail gracefully on malformed configs and package-name `extends`.

## Testing

- `npm run build` — passed.
- `npm run lint` — passed.
- Focused review-graph tests — 32 passed, 0 failed.
- Full `npm test` — 4,181 passed, 6 skipped, 24 failed under parallel load (4,211 total). Failures matched documented load/Windows flakes: walk and sgconfig/ast-grep/tree-sitter timeouts, LSP timing drift, and temp-directory EPERM.
- Isolated failed-file rerun with one worker — 218 passed; remaining failures were two Windows temp cleanup EPERMs plus one CSS-root fixture polluted by a marker at `C:\Users\apman` (and one suite-level EPERM cleanup). No review-graph alias test failed.

## Follow-ups intentionally out of scope

- TypeScript project references as cross-package graph edges.
- Review-graph cap/budget changes (R1); `clients/project-scale.ts` is untouched.

References #775.


apmantza:
Review note (non-blocking): `clearTsconfigPathsCache` is only invoked from tests — the per-directory matcher cache is session-lived, so a `tsconfig.json` edited mid-session won't refresh aliases until process restart. Acceptable for this slice (configs rarely change mid-session and the graph has its own staleness handling), but worth wiring into the review-graph reset path in the R2 follow-up (project references). Also noted: JSONC trailing-comma stripping is regex-based post-comment-removal and TS-5 array-form `extends` is ignored — both degrade to the documented fall-through, no action needed.

Issue #620, opened 2026-07-13
feature: measure total pi-lens resource footprint (CPU/RAM) across host, LSP children, and spawned analyzers


apmantza:
## Motivation

Dogfooding on pi-drykiss (2026-07-13) surfaced a `lens_diagnostics`/`lsp_diagnostics` mode=full sweep whose per-file confirmation rate collapsed from a reliable 0% timeout rate (10+ full sweeps, 2026-07-04 through 2026-07-12 17:14 UTC) to near-100% timeouts on every full sweep since 2026-07-12 ~20:19 UTC — with no pi-lens or project code change in that window that plausibly explains it.

The leading theory, checked by hand via `tasklist` (found ~25 concurrent `node.exe` processes, several at 600MB-2GB RSS): sustained CPU contention from multiple concurrent Node/LSP processes on the machine (other worktrees/sessions) is starving tsserver of cycles during the sweep. `touchFile`'s per-file wait is serialized within a server (#387) but that doesn't bound *how long* a single request takes under CPU contention — so a request that's normally well inside the 3.8s notify+diagnostics budget can blow past it when the box is busy.

Right now this is entirely inferred after the fact from `tasklist` and log timing correlation — pi-lens has no first-class way to answer "how much CPU/RAM is pi-lens itself (and everything it spawned) actually using right now, and was it starved or was it the one doing the starving?"

## Proposal

Extend the existing #449 multi-agent instance registry (`clients/instance-registry.ts`, `~/.pi-lens/instances.json`) which already tracks, per pi-lens host process: `pid`, `projectRoot`, `lspChildren[]` (each with its own `pid`/`serverId`), `rssBytes` (host only, via `process.memoryUsage().rss`), and `heartbeatAt`. Two gaps to close for full-footprint visibility:

1. **Host + long-lived LSP children**: add CPU% (in addition to the RSS already captured for the host) and RSS for each `lspChildren[].pid`, sampled at the existing heartbeat cadence via a small cross-platform sampler (e.g. `pidusage`, since Windows/macOS/Linux each expose this differently at the OS level and hand-rolling `/proc` vs `wmic`/perf-counters isn't worth it).
[...]

Issue #214, opened 2026-06-14
go-vet runner returns 0 diagnostics in dispatch though 'go vet' reports them in identical manual runs


apmantza:
The #209 tool-smoke harness found the `go-vet` runner **executes** but returns **0 diagnostics** (status `failed`, no `failureKind`, empty `diagnostics`, takes the lines 48-56 "non-zero but no parseable output" path) on a fixture that `go vet` clearly flags.

## Evidence

Fixture: `tests/fixtures/tool-smoke/go/{go.mod, bad.go}` where `bad.go` has `fmt.Printf("%d\n", "not a number")`.

`go vet` reports the bug in **every manual reproduction**, including the runner's exact spawn shape:

- `cd fixture && go vet .` → `bad.go:7:14: fmt.Printf format %d has arg ... wrong type string` (exit 1)
- `cd fixture && go vet bad.go` → same (exit 1)
- from repo root: `go vet <abs>/bad.go` → `C:\...\bad.go:7:14: ...` (exit 1)
- from repo root on a fresh temp copy (mimics `safeSpawnAsync(go, ["vet", ctx.filePath])` with no cwd): `go vet <temp-abs>/bad.go` → reports it (exit 1)

The parser regex (`parseGoVetOutput`, `/^(.+?):(\d+):(\d+):\s*(.+)/`, non-greedy) parses the Windows `C:\...:line:col` line correctly. So neither the invocation nor the parser explains the 0-result.

## Hypothesis

The discrepancy is `safeSpawnAsync`-env-specific: the runner (`clients/dispatch/runners/go-vet.ts:35`) spawns `go vet <file>` with **no cwd and the spawn's env** — if that env drops `GOCACHE`/`GOPATH`/`GOMODCACHE`, `go vet` emits a *build* error (which `parseGoVetOutput` doesn't match) instead of the vet diagnostic, yielding `failed` + 0 diags.

## Next step

Instrument the runner to log `rawOutput` (it already captures `raw`) on the 0-diags path, run via the harness (`node scripts/smoke-tools.mjs go --step2 --verbose`), and inspect what `go vet` actually emitted. Likely fixes: set `cwd` to the file's package dir and/or ensure the Go env (GOCACHE/GOPATH) is inherited.

Found by the #209 tool-smoke harness.

Issue #18, opened 2026-04-18
Slow startup even with quick mode


Artawower:
I’ve noticed a significant slowdown when using pi-lens, even with the quick startup mode enabled.

With pi-lens (PI_LENS_STARTUP_MODE=quick), the app appears to start faster, but the time until the first input becomes available is still around ~20s, whereas without pi-lens the app is fully usable in ~7s.

apmantza:
thanks @Artawower . should be fixed in the new version.

LeonardoRick:
hey @apmantza, when are you planning to release this new version? I'm getting the same problem in 3.8.32

PR #818, opened 2026-07-25
fix(safe-spawn): eliminate cmd.exe command-line construction on Windows (#817)

### LINKED ISSUE #817: safe-spawn: eliminate cmd.exe command-line construction for Windows spawns (CodeQL #17/#18/#19, CWE-78)
## Problem

After #815 replaced `shell: true` with an explicit cmd.exe invocation, CodeQL re-flagged the same spawn (`clients/safe-spawn.ts:192`, now 3 open alerts):

- **#17 + #19** `js/command-line-injection` (error): `buildWindowsShellCommand(command, args)` string-concatenates caller-tainted values into the `/c` payload. To CodeQL this is equivalent to `shell: true` — it has no sanitizer model for hand-rolled cmd.exe quoting.
- **#18** `js/shell-command-injection-from-environment` (warning, ~60 flows): `ComSpec`/`COMSPEC` env feeding `spawnCmd` plus env-derived paths in the payload.

And the escaping is genuinely unsound, not just a taint FP:

1. **`%` expansion** — cmd expands `%VAR%` even inside double quotes, and `%` cannot be escaped on a `/c` command line (`%%` only works in batch files). `%APPDATA%.md` is a legal NTFS filename; args include file paths from whatever repo is open (untrusted-repo threat model).
2. **Quote-state toggling** — cmd has no official escape for `"`; the `""` doubling in `cmdEscapeArg` is the MSVCRT convention, and with `/s` + `windowsVerbatimArguments` crafted embedded quotes can turn an escaped `&` back into a command separator.

## Fix

- Resolve `command` ourselves via a PATH + PATHEXT walk (a `where`-equivalent helper, cached).
- Resolves to `.exe`/`.com` → `spawn(resolvedPath, args, { shell: false })` directly — args as a real array, no cmd.exe, injection-proof and the pattern CodeQL recognizes. This covers the overwhelming majority of spawns (sg, biome, ruff, git, node...).
- Keep the cmd.exe wrapper ONLY for `.cmd`/`.bat` shims (they cannot run without it), and there **validate** — reject args containing `"`, `%`, `!`, CR/LF — instead of trying to escape them.
- Pin the interpreter to `%SystemRoot%\System32\cmd.exe` (like `killTree` already pins taskkill) instead of trusting `ComSpec` (addresses #18; dismiss the remaining env-flow instances with justification if any survive).

## Constraint: chcp 65001 / output encoding

The current wrapper's `chcp 65001` prefix works by mutating the inherited console's code page so sg/biome/ruff output isn't decoded as CP850/CP936/CP932. Dropping the wrapper for .exe tools loses that. The fix must preserve UTF-8 output decoding — options: one-shot `chcp 65001` spawn before the tool (same console mutation as today), and/or per-tool env (`PYTHONIOENCODING`, etc.), plus keeping existing decoding fallbacks. Behavior must be verified, not assumed.

## Acceptance

- No cmd.exe command-line string is built from unvalidated caller input.
- `.cmd`/`.bat` shims still work (npm-installed wrappers are the common case); dangerous args to shims fail loudly with a clear error.
- UTF-8 output decoding behavior preserved for the tools that needed chcp.
- CodeQL alerts 17/19 close on the next master analysis; 18 fixed or dismissed-with-comment.

apmantza:
## What

CodeQL re-flagged `clients/safe-spawn.ts:192` after #815/#816 replaced `shell:true` with an explicit `cmd.exe /d /s /c <hand-built string>` invocation: `js/command-line-injection` ×2 (alerts #17, #19) and `js/shell-command-injection-from-environment` (alert #18, ~60 flows via `ComSpec`/`COMSPEC`). The escaping was genuinely unsound, not just a CodeQL taint false positive:

- `%VAR%` expands even inside double quotes and cannot be escaped on a `/c` command line (only `%%` works, and only in batch files).
- cmd.exe has no official escape for `"` — the `""` doubling convention can be defeated by quote-state toggling.

## Approach

`safeSpawnAsync` now resolves `command` itself instead of always routing through cmd.exe:

- **New `resolveWindowsCommand`** — a cached `where`-equivalent PATH + PATHEXT walk, keyed by `(command, PATH, cwd)`. Respects `options.cwd` for relative commands.
- **`.exe`/`.com` resolutions** (the overwhelming majority of spawns — sg, biome, ruff, git, node, `where.exe`...) → `spawn(resolvedPath, args, { shell: false })` directly. Args stay a real array, no cmd.exe involved at all — the pattern CodeQL recognizes as safe.
- **`.cmd`/`.bat` resolutions** (npm-installed shims, e.g. `npx.cmd`) → keep the cmd.exe wrapper since these can't run without it, but:
  - Pin the interpreter to `%SystemRoot%\System32\cmd.exe` — never trust `ComSpec`/`COMSPEC` (fixes alert #18), same pin `killTree` in this file already uses for `taskkill`.
  - **Validate instead of escape**: if `command` or any arg contains `"`, `%`, `!`, CR, or LF, refuse to spawn and return a `SpawnResult.error` explaining which value was rejected — fail loudly, never silently strip. Normal quoting for spaces is unchanged (`cmdEscapeArg`/`buildWindowsShellCommand`, kept and still exported for the existing #214 regression tests).
- **Unresolvable command** → synthesize an `ENOENT`-shaped `Error` (`message`/`code`/`syscall`/`path`), matching Node's native shape, so existing `err.message.includes("ENOENT")` / `err.code === "ENOENT"` tool-not-installed detection (e.g. `clients/sg-runner.ts`) keeps working — previously this happened *inside* cmd.exe as a non-zero exit, not a spawn error event.
- **`chcp 65001` / UTF-8 output decoding preserved**: the `.cmd`/`.bat` wrapper path keeps its `chcp 65001 >nul 2>&1 &&` prefix unchanged. For the new direct `.exe` path, a new `ensureUtf8ConsoleCodePageOnce()` runs the *same* static, hardcoded `chcp 65001` via one pinned-cmd.exe spawn, memoized to run exactly once per process — the console code-page mutation persists for the console's lifetime (documented in a comment), so today's per-spawn re-run was always redundant; running it once is a strict improvement, not a behavior change.
- **Cache reset hook**: `resetSafeSpawnWindowsCommandCache()`, wired into `runtime-session.ts`'s `handleSessionStart` alongside this repo's other session-lived caches (`resetWorkspaceTopology`, `clearTsconfigPathsCache`) — a PATH change is picked up on the next session/reset.

### Out of scope (noted, not fixed here)

Grepped `clients/` and `scripts/` for other `ComSpec`/`cmd.exe`/`/c` call sites building a command line from tainted data. Two similar patterns exist outside `safe-spawn.ts` but are **not** part of this issue (#817 is scoped to `clients/safe-spawn.ts:192`) and are left for a follow-up:
- `clients/lsp/client.ts` — the `.ps1`→PowerShell-execution-policy-bypass fallback path hand-escapes a cmd.exe command line for `.cmd`-wrapped LSP binaries.
- `clients/sg-runner.ts` — the non-bash Windows fallback (`isWindows && !hasBash`) still uses `spawn(..., { shell: true })` with its own `escapeWindowsArg`.

`scripts/playground-verify-rule.mjs` also references `cmd.exe`/`ComSpec` but is dev-tooling, not runtime code.

## Testing

- `npm run build` — clean.
- `npx vitest run tests/clients/safe-spawn-windows-command.test.ts` — **13/13 passed**, including 8 new Windows-conditional tests exercising real spawns on this machine (Windows 11):
  - (a) bare command PATH+PATHEXT resolution → direct `.exe` spawn (no cmd.exe)
  - (b) `.cmd` shim fixture still executes through the pinned cmd.exe wrapper, args with spaces round-trip verbatim
  - (c) `%`- and `"`-bearing args targeting a `.cmd` shim are rejected loudly, nothing spawned (two tests)
  - (d) the same dangerous args pass through **literally** to a direct `.exe` spawn (array args are safe)
  - (e) an unresolvable command returns an ENOENT-shaped error
  - (f) resolver cache reuse across repeated calls, and the reset hook not breaking subsequent resolution
- `npx vitest run tests/clients/safe-spawn-resource-usage.test.ts` — **8/8 passed** (updated one #620 test: an unresolvable command now short-circuits *before* any process spawns, so the CPU/RSS sampler is correctly never started/stopped on Windows — previously it wrapped the always-succeeds-then-fails-inside cmd.exe wrapper spawn).
- Full suite: `npx vitest run` → **4257 passed, 5 failed, 10 skipped** (394 files). All 5 failures are pre-existing timing/load flakes in unrelated files (`source-filter-async.test.ts`'s event-loop-budget test, two `lsp/workspace-diagnostics-sweep-*` flush-count tests, `ast-grep-rule-precedence-followups.test.ts`) — confirmed by re-running each in isolation, all pass cleanly:
  - `tests/clients/source-filter-async.test.ts` — 1/1 passed isolated (was: timeout under load)
  - `tests/clients/lsp/workspace-diagnostics-sweep-batch-open.test.ts` + `-preopen-chunk.test.ts` — passed isolated (were: off-by-one flush counts under load)
  - `tests/clients/ast-grep-rule-precedence-followups.test.ts` — 11/11 passed isolated
  - None touch `clients/safe-spawn.ts`, `clients/runtime-session.ts`, or spawn behavior.

### Real smoke test (built output, this Windows machine)

```
$ node smoke-817.mjs   # imports clients/safe-spawn.js (npm run build output)

=== git --version (.exe, direct spawn) ===
{ "stdout": "git version 2.55.0.windows.3\n", "stderr": "", "status": 0 }

=== npx.cmd --version (.cmd shim, pinned cmd.exe wrapper) ===
{ "stdout": "11.16.0\n", "stderr": "", "status": 0 }

=== npx.cmd with dangerous %-arg (should be rejected, nothing spawned) ===
{ "stdout": "", "stderr": "", "status": null, "error": {} }   // rejected, no process spawned

=== unresolvable command (should be ENOENT-shaped) ===
{ "stdout": "", "stderr": "", "status": null,
  "error": { "code": "ENOENT", "syscall": "spawn", "path": "definitely-not-a-real-tool-817" } }
```

Closes #817


apmantza:
**Follow-up commit** (988bb213) — a review pass found the deprecated synchronous `safeSpawn` still had the identical `buildWindowsCommand` + `spawnSync(..., { shell: true })` sink as the async version had before this PR's first commit (same CodeQL alert class, kept the flows alive; the only current caller is `clients/test-runner-client.ts:367`'s sync pytest-on-PATH probe).

Fixed by reimplementing `safeSpawn`'s Windows branch against the exact same seams `safeSpawnAsync` already uses:
- `resolveWindowsCommand` (cached PATH+PATHEXT walk) → direct `spawnSync(resolvedPath, args, { shell: false })` for `.exe`/`.com`
- pinned `%SystemRoot%\System32\cmd.exe` wrapper + `findCmdUnsafeValue` validation for `.cmd`/`.bat`
- synthesized `ENOENT`-shaped error when unresolvable

No `shell: true` anywhere in `clients/safe-spawn.ts` anymore. The now-dead `escapeWindowsArg`/`buildWindowsCommand` helpers were removed.

`test-runner-client.ts`'s call site is unchanged — `detectRunner`/`findTestFile` are synchronous with many sync callers and sync tests across the codebase, so migrating that call to `safeSpawnAsync` wasn't a trivial, in-scope change here; the sync `safeSpawn` it calls is now hardened instead.

Also hardened the **async** path per review: `findCmdUnsafeValue` now validates the *resolved* path (what's actually interpolated into the `/c` line via `buildWindowsShellCommand`), not just the caller's original `command` string — a resolved path itself containing `%`/`!` would otherwise reach the shell unvalidated.

New Windows-conditional tests mirror the async coverage for the sync entry point (direct `.exe` spawn, `.cmd` wrapper round-trip, dangerous-arg rejection, literal pass-through to `.exe`, ENOENT synthesis).

**Testing**: `npm run build` clean. `tests/clients/safe-spawn-windows-command.test.ts` + `safe-spawn-resource-usage.test.ts` + `safe-spawn-ambient-signal.test.ts` + `test-runner-client.test.ts` → **73/73 passed**. Full suite: 4256 passed / 9 failed / 12 skipped — all 9 failures are pre-existing timing/load-budget flakes in unrelated files (`cascade-graph-occupancy`, `sgconfig`, `source-filter-async`, `source-walk-occupancy`, `startup-overhead`, one `lsp/workspace-diagnostics-sweep-batch-open` flush-count test), confirmed by isolated re-run: all 6 files / 30 tests pass cleanly alone. None touch `safe-spawn.ts` or `test-runner-client.ts`.

Real smoke test against the built sync `safeSpawn`:
```
=== sync: git --version (.exe, direct spawnSync) ===
{"stdout":"git version 2.55.0.windows.3\n","stderr":"","status":0}
=== sync: npx.cmd --version (.cmd shim, pinned cmd.exe wrapper) ===
{"stdout":"11.16.0\n","stderr":"","status":0}
=== sync: npx.cmd with dangerous %-arg (should be rejected) ===
{"status":null,"errorMessage":"Refusing to spawn \"C:\Program Files\nodejs\npx.cmd\" via cmd.exe: \"%APPDATA%.md\" contains a character (\", %, !, or CR/LF) that cannot be safely escaped on a cmd.exe /c command line (CWE-78, #817). ..."}
=== sync: unresolvable command (should be ENOENT-shaped) ===
{"status":null,"code":"ENOENT","message":"spawn definitely-not-a-real-tool-817-sync ENOENT"}
```

Issue #822, opened 2026-07-25
Same-workspace warm attach: route new same-root sessions to the incumbent warm process (slice 3 of #449)


apmantza:
Split from #449 (its slice 3 — the central unimplemented result). Prereq #242 (per-server deadlines) is done, so sequencing is satisfied.

## Problem

A second pi session opened on the SAME workspace root spins up its own full LSP fleet even when a warm incumbent process already has everything running. The per-workspace IPC seam exists but exposes only `requestWarmAnalyze` (`clients/mcp/ipc.ts` — schema has just `file` and `cwd`; `mcp/server.ts` ~251 invokes `analyzeFile` directly). No session-level attachment.

## Design sketch

- Extend the IPC surface with a diagnostics-query/dispatch route: a new same-root session discovers the incumbent via the instance registry (#474), and routes LSP-backed diagnostic needs to it instead of spawning its own servers.
- Ownership rules: the incumbent owns the fleet; the attaching session is a client. Define takeover on incumbent exit (registry heartbeat staleness) → attacher promotes to primary and spawns normally.
- Timeout/freshness/fallback: every attached call carries a deadline; on timeout, IPC error, schema mismatch, or version skew the attacher falls back to local spawn (fail-open — warm attach is an optimization, never a correctness dependency).
- Relationship to #478 (`pilens:rpc:*` in-process query surface): related plumbing, different boundary — cross-PROCESS here. Coordinate schemas where cheap, don't block on it.

## Follow-ups gated on this landing (from #449, tracked here)

- [ ] Relocate capability-inventory/wait-strategy modules behind the sharing boundary once it exists (architectural consolidation, no user-visible behavior).
- [ ] Optional: persist subagent marker/run ID/agent name in `InstanceEntry` for concurrency-profile observability.

## Likely touch points

`clients/mcp/ipc.ts`, `mcp/server.ts`, `clients/lens-engine.ts`, a host-neutral diagnostics facade, `clients/dispatch/runners/lsp.ts`, `clients/lsp/index.ts`, `clients/runtime-session.ts`, `index.ts`; tests under `tests/clients/mcp/`, `tests/mcp/`, `tests/clients/lsp/`.

## Acceptance

- Two same-root sessions: second one attaches (observable in latency/cascade logs), total LSP child count does not double, diagnostics still correct in both.
- Incumbent killed hard: attacher recovers to local spawn within one touch, no orphans (interacts with the #661 reaper backstop).
[...]

PR #824, opened 2026-07-25
feat(lsp): attach same-workspace sessions to warm incumbent

apmantza:
## What

- add opt-in same-workspace warm LSP attach behind `PI_LENS_WARM_ATTACH=1`
- add a PID-scoped, versioned diagnostics IPC route carrying file content, SHA-256 content hash, and an absolute deadline
- discover the oldest PID-confirmed, heartbeat-fresh same-root incumbent from the instance registry
- skip local LSP pre-warm while attached and route per-edit LSP diagnostics through the incumbent
- permanently promote to the unchanged local-spawn path after any timeout, IPC error, schema mismatch, stale/inconclusive response, or incumbent loss
- log disabled/local/attach/takeover/listener events through the `lsp_warm_attach` latency phase
- document the soak switch and update AGENTS.md plus CHANGELOG

Part of #822 (first slice: core protocol + per-edit routing; see review notes for what remains).

## Approach

The new `clients/warm-attach.ts` coordinator owns session-level election and the one-way attached-to-local state transition. Every opted-in pi process exposes a PID-scoped diagnostics listener, while an attaching process selects a different live same-root PID before registering itself. The incumbent executes the existing `LSPService.touchFile` document-diagnostics path against the supplied current content.

Freshness is checked on both sides: requests carry a content hash and deadline; responses echo the hash, schema version, served timestamp, freshness, and inconclusive state. Before every attached touch, the client rechecks the incumbent against the registry and the existing conservative PID/heartbeat semantics. Any failure promotes locally for the remainder of the session, preventing flapping.

Attached sessions never register or claim the incumbent's LSP children. The #661/#525 reaper contract remains unchanged: heartbeat staleness can remove an entry but cannot kill children, and process killing still requires a PID-confirmed-dead owner plus child identity safeguards.

With the environment variable unset or set to `0`, no IPC listener is opened and the existing local LSP path is unchanged.

## Testing

- `npm run build` — passed
- `npm run lint` — passed
- focused warm-attach suite: 2 files, 11 tests passed
- full `npm test`: 395 files total; 381 passed, 13 failed, 1 skipped; 4,289 tests total; 4,260 passed, 20 failed, 9 skipped. Failures were in pre-existing load/filesystem/process timing suites; the CSS root-policy failure also reproduces in isolation. The new warm-attach suites passed in the full run and in isolation.


apmantza:
Review notes (merging as opt-in soak): two honest scope asterisks vs #822's full acceptance. (1) Routing covers the dispatch LSP runner (per-edit hot path) only — `lens_diagnostics`/`lsp_diagnostics` still spawn a local fleet even when attached, so a project sweep on an attached session temporarily defeats the no-duplicate-fleet goal. (2) When attached, code-action/quickfix lookups are skipped (no local client to query). Both acceptable for the `PI_LENS_WARM_ATTACH=1` soak; keeping #822's remaining checklist items open means these can ride along there. Core protocol verified: versioned/content-hash/deadline-bound IPC with strict client-side rejection and permanent promote-to-local; incumbent selection PID-confirmed + heartbeat-fresh + same-root; flag off is byte-for-byte today's behavior; #661 reaper identity guards untouched.

Issue #851, opened 2026-07-28
Bound multi-client LSP shutdown latency before generation handoff


neumie:
## What should improve?

After #850 serializes replacement-generation spawning behind prior `LSPService.shutdown()`, the existing multi-client shutdown tail becomes user-visible rather than being hidden by an overlapping duplicate fleet.

`LSPService.shutdown()` currently awaits each client in a `for` loop. On POSIX, one non-fast `killProcessTree()` may wait 1.5s before SIGKILL escalation, so teardown is O(client count × grace). A captured incident showed an idle reset with 17 clients taking roughly 27s client-by-client.

## Proposed scope

Make a multi-client service teardown bounded/concurrent (or use a safe per-key generation handoff) while preserving:

- #850's no-spawn-before-retiring-teardown ownership invariant,
- #706's in-flight spawn guards,
- graceful vs `fast` / `processExiting` semantics,
- process-group tree killing and instance-registry deregistration,
- a deterministic test proving the tail is bounded independently of fleet size.

This is intentionally separate from #850's narrow correctness fix; parallel process/registry teardown deserves its own review and soak.

apmantza:
Thanks @neumie — agreed on keeping this separate from #850's correctness fix. Confirmed the O(clients × grace) sequential tail in `LSPService.shutdown()` (per-client `await client.shutdown(options)` loop), which becomes user-visible once #852 gates the replacement generation's first spawn on full prior teardown. Will pick this up after #852 lands and soaks.

Issue #706, opened 2026-07-19
fix(lsp): shutdown/reset racing an in-flight spawnClient leaks a live LSP server process


apmantza:
## Bug

`LSPService.shutdown()` (`clients/lsp/index.ts:3034`) clears `state.inFlight` but neither awaits nor cancels the in-flight spawn promises — and `spawnClient()` (`clients/lsp/index.ts:1094`) never re-checks `this.isDestroyed` after its awaits:

- `await server.spawn(root, ...)` (line 1108) can take 1–15s cold (jdtls/ruby-lsp/clojure-lsp ~5s+; INITIALIZE_TIMEOUT_MS is 15s).
- `await createLSPClient(...)` (line 1150) completes the initialize handshake regardless of service state.
- The live client is then registered into `state.clients` (line 1167) of the already-destroyed instance.

`resetLSPService()` (`clients/lsp/index.ts:3108-3113`) nulls the singleton synchronously with a fire-and-forget `shutdown()`, so the leaked client lives in a discarded instance nothing can reach: no idle reaper, no later shutdown will touch it. The cross-process orphan reaper (#472/#449) doesn't help either — it only kills children of **dead** parents, and the host process is alive. The server process leaks for the lifetime of the host.

## Triggers (all routine)

- Session-start resets (`index.ts:1272`, `:1384`, `:1425`) racing a cold spawn from a just-dispatched edit.
- Idle reset (`index.ts:1553`).
- Nested session_start from in-process subagent binds — this is very likely the mechanism behind the unverified 'each subagent bind may tear down the parent's LSP fleet mid-flight' note in #449's research comment.
- `processExiting` reset (`index.ts:1736`) — narrower window, but same shape.

## Fix direction

1. In `spawnClient`, re-check `this.isDestroyed` after each await:
   - after `server.spawn(...)`: if destroyed, kill the freshly spawned process (no client exists yet) and return `undefined`;
   - after `createLSPClient(...)`: if destroyed, `client.shutdown({ fast: true })` (best-effort) and return `undefined` — never register into `state.clients`.
2. Optionally: `shutdown()` snapshots `inFlight` promises and `Promise.allSettled`s them, shutting down any client that slipped through — belt-and-braces for future call sites.

## Acceptance

- [ ] A spawn resolving after `shutdown()` never leaves a live child process or an entry in `state.clients`
[...]

PR #852, opened 2026-07-28
fix(lsp): serialize singleton generation handoff

### LINKED ISSUE #850: LSP reset can spawn a replacement generation before prior teardown finishes
## What happened?

A long-running pi process on 3.8.71 accumulated multiple live LSP processes for the same server/root while the parent pi process remained alive (56 direct LSP children, 14 duplicate groups). The dead-parent reaper therefore cannot help.

Within one `LSPService`, `state.clients` + `state.inFlight` correctly dedupe `${server.id}:${normalizeMapKey(root)}`. The gap is cross-generation: `resetLSPService()` starts `globalLSPService.shutdown()` fire-and-forget, immediately sets the singleton to `null`, and lets the next request create/spawn through a new service while the old generation is still tearing down clients.

Current `master` (`b058213e`) retains that lifecycle. This is distinct from #706 (shutdown racing one in-flight spawn), #473 (concurrent secondary session guard), and #449 (cross-process fleet sharing/reaping).

## Steps to reproduce

Deterministically with mocked clients:

1. Spawn one client through `getLSPService()`.
2. Make that client's `shutdown()` return a controlled unresolved promise.
3. Call `resetLSPService({ reason: "idle" })`.
4. Immediately call `getLSPService().getClientForFile()` for the same server/root.
5. Observe a second `server.spawn()` before the first client's shutdown promise resolves.

Idle teardown makes the production window substantial because clients are shut down asynchronously and may wait for process-tree escalation.

## Expected behavior

A replacement generation may be allocated synchronously, but its spawn path must wait for all prior-generation teardown promises to settle. Warm reuse within a generation and the concurrent-secondary session ownership guard must remain unchanged.

## pi-lens version

Observed on 3.8.71; reproducible by inspection/test on current `master` (`b058213e`, 3.8.72).

neumie:
## Summary

- keep `resetLSPService()` synchronous while carrying every still-pending singleton teardown into a replacement service's one-shot generation handoff
- block the replacement's first spawn until that handoff settles, and re-check destruction after async root lookup and dead-client cleanup so retired generations cannot start late processes
- preserve within-generation client/in-flight deduplication, concurrent-secondary ownership, and #706 raw-process cleanup; add deterministic regression coverage plus lifecycle documentation

Closes #850.

## Root cause

`resetLSPService()` discarded the published singleton and fire-and-forgot its asynchronous `shutdown()`. A new singleton could therefore be allocated and spawn the same server/root while the previous generation still owned live clients. Existing `state.clients` / `state.inFlight` deduplication only applies inside one `LSPService`, so it could not prevent overlap across generations.

## Verification

- `npm run build`
- focused #850 regression tests: 2/2 passed
- combined #706/#850 race tests: 5/5 passed
- related session lifecycle, idle-reset, and teardown-logging tests passed
- `npx vitest run --maxWorkers=2`: 4,298 passed, 23 skipped, 0 failed (all 4,321 tests; no retries or timeout changes)
- `npm run lint`
- `npm run changelog:check`
- `npm run check:lockfile`
- `npm run build:dist`
- independent read-only standards/spec review: no actionable findings

The host reached a load average of 165 during default-worker runs, producing unrelated 5-second timeout/occupancy failures across MCP, Biome, graph, scanner, and LSP tests. A clean `upstream/master` run passed at default concurrency; the branch's bounded-concurrency full run above passed completely.

## Follow-up

#851 separately tracks bounding the current sequential multi-client shutdown tail before the generation handoff. This PR intentionally keeps the correctness fix narrow and preserves existing shutdown/process-tree semantics.


apmantza:
many thanks again @neumie. merged.

Issue #282, opened 2026-06-21
fix(ast-grep): napi fallback runner drops utils/TSX/Python rules


apmantza:
Split out from #239 (Phase 2 residue). Now that the **ast-grep LSP is the primary rule engine** and the in-process napi runner (`clients/dispatch/runners/ast-grep-napi.ts`) only runs as the **Gate-B fallback** (when the `ast-grep` CLI/LSP binary can't spawn — sandboxed/spawn-restricted env, broken platform binary), these gaps affect *only* that fallback path:

- **`utils:` cross-rule refs** — rules using `utils:` + `matches:` (e.g. the Rust `rust-2024-let-chain-candidate` catalog port) aren't resolved by the napi runner's invocation, so they silently no-op.
- **TSX** — the runner's language selection doesn't invoke the tsx grammar, so `.tsx` rules don't fire.
- **Python** — same: Python rules in the shipped baseline aren't invoked.

Net effect: ~15 shipped rules silently drop **when running through the napi fallback**. On the normal path (LSP present) these all run correctly — the LSP is grammar-strict and honors `utils:`/TSX/Python — so this is a fallback-only correctness gap, low urgency.

### Acceptance
- napi fallback runner invokes the tsx + python grammars for the matching file kinds
- `utils:`/`matches:` cross-rule refs resolve in the napi invocation (or the runner explicitly skips + logs rules it can't honor, rather than silently dropping)
- a fallback-path test asserts a `utils:`-based rule and a `.tsx`/`.py` rule fire when the LSP is forced unavailable

Context: engine equivalence (napi ≡ CLI ≡ LSP, same Rust core via `findAll`) holds for *pattern* evaluation; the gap is in the napi **runner's invocation/language-selection layer**, not the engine. See #239 final comment ("Phase 2 status — core complete", item under "Remaining → independent").

apmantza:
**Rescope after code assessment (2026-07-28, master `845c6476`):** two of the three gaps are already fixed, so narrowing this issue to the residue.

**Fixed since filing:**
- `utils:`/`matches:` — fixed by #663: the runner passes `{rule, constraints, utils}` natively to `findAll` (`ast-grep-napi.ts:388-392`; `NapiConfig.utils` supported in @ast-grep/napi 0.44.1, verified by live probe against `no-dupe-keys`). Covered by `ast-grep-utils-block.test.ts`.
- `.tsx` grammar loading — the runner accepts `.tsx` and parses with `sgModule.tsx`.

[...]

PR #867, opened 2026-07-28
fix(ast-grep): aggregate napi unsupported-rule skips into latency.log

apmantza:
Follow-up to #857 (#282): the honest-skip logging was too loud in practice — the shipped rules tree has ~30 non-jsts (mostly Python) ast-grep rules, not just the 4 in the curated catalog, so a single napi-fallback evaluation emitted ~30 `[dispatch]` lines straight into the pi session.

- Per-rule terminal lines removed. Skips are collected during evaluation and flushed as ONE `logLatency` phase entry (`astgrep_napi_unsupported_rules_skipped`) per evaluation, with `{language: {count, ruleIds}}` metadata — same file-based logger the rest of dispatch uses (`~/.pi-lens/latency.log`), zero terminal output.
- The cross-run dedup set is unchanged, so a rule is still recorded at most once per session; a second evaluation sharing the set emits nothing.
- Both flush points covered (normal completion and the duplicate-overflow early return).
- Tests updated: assert zero terminal lines, one aggregated entry containing `no-compile-call` under `python`, and no second entry across two evaluations sharing the dedup set (latency-logger mocked).

Verified: `npm run build`, `npm run lint`, 21/21 across the napi runner suites.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Fixed
 
+- **NAPI fallback unsupported-rule skips no longer spam the session log** (refs #282) — the per-rule terminal lines (one for each of the ~30 non-jsts catalog rules) are replaced by a single aggregated `astgrep_napi_unsupported_rules_skipped` entry in `latency.log` carrying per-language counts and rule ids.
+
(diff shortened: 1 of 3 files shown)


apmantza:
Fixed on master in a7e816c7 (pushed directly, no PR): per-rule terminal lines removed; skips aggregate into a single `astgrep_napi_unsupported_rules_skipped` latency.log entry with per-language counts and rule ids. Verified: build, lint, 21/21 napi runner tests. Note #868 tracks the underlying reason the napi fallback was running at all alongside a healthy LSP.

Issue #440, opened 2026-07-07
python-assert-production fires on every assert in tests/** (no test-file exclusion)


xificurC:
**What happened?**

The `python-assert-production` tree-sitter rule fires on every `assert` statement, including in test files. `assert` is the idiomatic test assertion; there is no test-path carve-out, so a normal test suite produces ~20 advisory warnings and trains users to ignore the rule.

**Steps to reproduce**

```python
# tests/test_repro.py
def test_basic():
    assert 1 + 1 == 2   # flagged: python-assert-production
```

Run pi-lens diagnostics on `tests/test_repro.py`; observe a `python-assert-production` finding on the assert line.

**Expected behavior**

Exclude `tests/**` by default (a test-path glob), matching ruff's `S101` per-file-ignore convention. `assert` in tests is correct usage.

---

**pi-lens version:** `npm ls pi-lens` (via `@earendil-works/pi-coding-agent`, observed 2026-07-07)

apmantza:
Fixed in #447 (merged) — tree-sitter rules gained the opt-in `skip_test_files` field and `python-assert-production` sets it; exercised through the real runner (prod file fires, tests/ skips). This stayed open only because the commit's "closes #439, #440, #441" phrasing auto-closes just the first issue — GitHub needs a keyword per issue.

Issue #439, opened 2026-07-07
no-init-return over-matches: flags return in factory fns whose body contains def __init__


xificurC:
**What happened?**

The `no-init-return` ast-grep rule flags `return` statements that are *not* inside an `__init__`. It matches any `function_definition` whose **body text** regex-contains `def __init__`, so a factory function that returns a class with an `__init__` trips it — the factory's own `return` (and sibling methods' `return`s) get flagged.

**Steps to reproduce**

```python
# repro.py
def make_thing():
    class Thing:
        def __init__(self):
            self.x = 1
    return Thing   # flagged: no-init-return (but this return is in make_thing, not __init__)
```

Run pi-lens diagnostics on `repro.py`; observe a `no-init-return` finding on the `return Thing` line.

**Expected behavior**

The rule should constrain to a function *named* `__init__` (e.g. pattern `def __init__($$$): $$$`), not a body-content regex. `return` in `make_thing` is correct and should not flag.

---

**pi-lens version:** `npm ls pi-lens` (via `@earendil-works/pi-coding-agent`, observed 2026-07-07)

PR #871, opened 2026-07-29
test: real-runner harness for rule/dispatch behavior (closes #448)

### LINKED ISSUE #448: test: real-runner harness for rule/dispatch behavior — reduce behavior-mocking
## Motivation

The #439/#440 rule-quality bugs existed *and shipped* despite tests, because the tree-sitter runner's tests (`tree-sitter-runner.test.ts`) mock the client, the query loader, AND the review graph — they verify plumbing (availability gates, skip-paths, error handling) but can never see a real rule fire. A false-positive is invisible to them by construction. The #440 fix was only provable via a **real-runner** test (real client + real query loader + real fixture), and #439 via ast-grep's real `ast-grep test` fixture harness.

## Principle

- **Mock the environment** — LSP-server availability, tool presence, network, abort/error injection. Legitimate + fast.
- **Never mock the behavior under test** — tree-sitter parsing, ast-grep matching, rule dispatch, suppression. Use the real engine or the test proves nothing.

Note the existing asymmetry: **ast-grep rules** have a real-engine fixture harness (`ast-grep test`); **tree-sitter rules** only have query-level `runQueryOnFile` (bypasses the runner's dispatch/skip logic) or fully-mocked runner tests. That gap is exactly where #440 hid.

## Proposal

1. Generalize the ad-hoc ctx factory in `tests/clients/dispatch/runners/tree-sitter-skip-test-files.test.ts` (#447) into a shared **`makeRealRunnerCtx()`** helper (temp dir, real FactStore, python/js/etc. kind).
2. Adopt real-runner tests as the standard for tree-sitter **rule behavior + dispatch** (skip_test_files, delta, blockingOnly, inline-tier filtering) — complementing, not replacing, the mocked control-flow tests.
3. Make it a rule: **a rule-bug fix ships a real-runner regression test** (fixture in → assert rule fires/doesn't).
4. Audit other behavior-mocked suites for the same gap.

## Risks to handle

Real-runner tests are slower + can have side effects (review-graph building writes to the data dir) — sandbox via temp cwd + confirm no writes escape; grammars must be present (CI already downloads them).

Origin: fell out of the #439/#440/#441 batch (#447).

AngriestBird:
The #439/#440 postmortem: the tree-sitter runner's tests mocked the client, the query loader, and the review graph, s...
## Harness
`tests/support/real-runner-ctx.ts` generalizes the ad-hoc `ctxFor()` from the #440 skip-test-files suite:
- `makeRealRunnerEnv(defaults?)`: a temp project root that can hold several fixtures sharing one cwd (one rule-cache/...
- `makeRealRunnerCtx(relPath, content, overrides?)`: one-shot convenience over `makeRealRunnerEnv().addFile(...)`.
- `assertGrammarAvailable(languageId)`: call from `beforeAll`. A missing grammar degrades silently to zero matches, w...
- `firedRuleIds(result)`: the `d.rule` projection every suite was hand-rolling.
- `napiFallbackHasTool`: simulates the ast-grep binary absent so the napi runner's matching path actually executes (t...
`tree-sitter-skip-test-files.test.ts`, `ast-grep-sonar-rules.test.ts`, and `ast-grep-utils-block.test.ts` migrated on...
## Bugs found by the harness
**Rule cache dropped `skip_test_files`/`fix_action` on persist.** The `cache.set` projection in the tree-sitter runne...
**Tree-sitter diagnostics were off by one line.** The client's `searchFileWithQuery` already returns 1-indexed positi...
## New coverage
- `tree-sitter-rule-cache-warm.test.ts`: runs the runner twice against the same cwd so run 2 rehydrates from run 1's ...
- `tree-sitter-dispatch-behavior.test.ts`: dispatch filtering through the real engine. `blockingOnly` review-tier fil...
- CI guard: `describe.skip` on a missing ast-grep CLI silently vanishes the whole real-engine fixture harness with a ...
- `smoke-fixture-coverage.test.ts`'s EXEMPT entry for `ast-grep-napi` now cites the suites that actually exercise rea...
## Docs
- CONTRIBUTING.md: a rule-bug fix ships a real-runner regression test through `treeSitterRunner.run()` via `makeRealR...
- AGENTS.md: new "Real-runner rule/dispatch tests (#448)" section. Mock the environment, never the behavior under tes...
- `skills/pi-lens-write-tree-sitter-rule/SKILL.md`: points rule-bug fixes at the same template.
## Audit results (follow-ups, not filed as issues)
- 184 vendored coderabbit ast-grep rules have zero fixtures. They ship in prod via `clients/sgconfig.ts`, but `ast-gr...
- `clients/dispatch/runners/lsp.ts` is only covered by fully-mocked tests; real LSP coverage is nightly-smoke only.
- Suppression is never tested end-to-end through the dispatcher against a diagnostic a real rule produced.
- `ast-grep-napi.test.ts` mocks `yaml-rule-parser`, so that suite never runs the runner against a real shipped rule.
- ~15 tests construct `new TreeSitterClient()` directly, which `clients/tree-sitter-shared.ts` forbids (separate clie...
## Review pass

Adversarial + perf review of the branch, applied as follow-up commits:

- **`RuleCache.set()` now prunes orphaned prior-version cache files.** The v3→v4 bump (and every bump before it) left `<language>-rules-v3.json` on disk forever; nothing ever read or deleted old versions. Unit test included.
- **CI verifies the ast-grep CLI explicitly** (`npx ast-grep --version` after install). The CI-loud guard otherwise rested on `npm test` putting `node_modules/.bin` on PATH implicitly.
- **Helper cleanup.** Removed the `flushReviewGraphPersistsForTests()` call from `cleanup()` (a no-op: test setup zeroes the persist debounce, and persists land under `PI_LENS_HOME`, not the fixture root). Doc comments now state the two invariants the helper can't enforce itself: tree-sitter suites must mock `review-graph/service.js` (module-global 5s blast cooldown; unmocked suites go order- and wall-clock-dependent), and `RunnerCtxOverrides` replace rather than merge (`facts` is the one stale-content path).
- Shared deduped `linesFor` in `real-runner-ctx.ts` replaces the two divergent per-suite copies; `afterAll(() => env?.cleanup())` so a failed `beforeAll` isn't masked by a TypeError; the warm-cache suite documents why `[false, true]` holds (per-language cache, python-only then typescript-only tests).
- CHANGELOG: strict/false-positive dispositions recorded against the old off-by-one line numbers are anchored to the wrong line's content and resurface once after upgrading. Expected, one-time.

Perf was measured before touching anything: the full suite is ~44s wall clock and `tests/clients/lsp/integration.test.ts` (41s) is the whole critical path. The suites on this branch run in 0.1-1.7s each, so no test-speed changes were warranted.

## Perf pass

A measured optimization round on top of the review pass (16-core host, full suite, 2 runs per config). Suite wall clock 44-48s → **~40s**, same aggregate peak RSS, all 4314 tests green; `PI_LENS_TEST_MAX_WORKERS=6` trades back to ~44s / 8 GB for memory-constrained machines (CI keeps vitest defaults).

- **`killProcessTree` resolves on the child's exit event** instead of unconditionally sleeping the 1500ms SIGTERM→SIGKILL escalation window. Every graceful LSP teardown paid it — `lsp/integration.test.ts` alone went 41s → 4s, and production sessions save the same ~1.5s per client shutdown. Hung servers still escalate at 1500ms.
- **Tool probes stop spawning what cannot succeed**: the dispatcher and install-fallback pre-check the binary on disk (μs stat/PATH walk) before a `--version` spawn, and the install-fallback no longer re-verifies a probe-cache-validated managed install with a node-shim boot per analyze. A broken managed install now surfaces as the runner's own spawn error instead of a silent null.
- **Project scans reuse the query-loader singleton** (`scanner.ts` was re-reading the ~180 tree-sitter rule YAMLs on every scan).
- **Test infra: shared tool-home template.** A lockfile-keyed template is seeded once per suite run (globalSetup) and handed to workers via probe-cache, killing the per-worker cold oxlint install that made one file swing 14s vs 47s on install/network luck.
- **Real-clock waits now inject the existing env-tunable budgets** (50-250ms) instead of paying production defaults: silent-clean-confirm 12.2s → 3.2s, sweep-batch-open 7.5s → 3.6s, build-staleness.smoke 14.5s → 4.7s (new `PI_LENS_MCP_STALENESS_INTERVAL_MS`), shutdown-timeout tests ~1s → ~0.2s each. The one test whose 1500ms wait IS the assertion (max-of-budgets) keeps the real wait; one elapsed-time bound was tightened 5000 → 1000ms.
- Local vitest default `maxWorkers: "50%"`; the snapshot-occupancy test warms the dirent cache in `beforeAll` so it stops passing only via `retry: 2`.

Closes #448.


apmantza:
i am grateful @AngriestBird.  /taking a bow. merged.

Issue #884, opened 2026-07-29
32 enabled tree-sitter rules never compile and are silently dead; add compile-all-rules CI guard


apmantza:
@-

apmantza:
Completed by the five-PR #884 merge train already on master: rule repairs for TS/JS (aea614c3), Python (4d1828fa), Go/Rust/Kotlin (bc5ebf3b), Ruby (7ecc4700), Java/C++/CSS/PHP (05eabf34), plus the compile-all guard (1e4bc3ae, tests/clients/tree-sitter-rule-compile-guard.test.ts) with an empty known-broken allowlist (ffa4a0da). A worker re-audit today compiled every enabled rule against its real grammar: 61/61 pass, no remaining broken rules.

Issue #879, opened 2026-07-29
Tracking: implement or remove the ~40 unimplemented post_filters silenced by fail-closed handling (#877)


apmantza:
@-

PR #901, opened 2026-07-29
fix(rules): repair non-compiling java/cpp/css/php tree-sitter rules (refs #884)

apmantza:
@-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


apmantza:
Pushed two follow-up commits: (1) merge-conflict resolution with master's CHANGELOG; (2) the six post_filters these rules reference (no_break_or_return, is_double_checked_locking, shadows_parent_field, missing_break_between_cases, scoped_lock_empty_args, calc_missing_spaces) were unimplemented, so the repaired queries compiled but still fail-closed to zero findings in production — implemented all six (refs #879) and reworked the test to run through the shared client's runQueryOnFile instead of raw web-tree-sitter, which also fixes the lint failure (bad web-tree-sitter import typings).

PR #906, opened 2026-07-29
Add /lens-perf latency percentile report

### LINKED ISSUE #767: feature: /lens-perf (or extend /lens-health) surfacing the slowest latency.log phases in-session
Split out from #191 (its last unchecked item) so #191 could close after #761/#765 superseded its walker scope.

From the #191 audit: hot-path regressions are invisible until someone greps `~/.pi-lens/latency.log` by hand. A `/lens-perf` command (or an extension of `/lens-health`) should surface the slowest phases — e.g. top-N by p50/p99 over the current session and over the log's window — so a regression like the #758 startup walk shows up in-session instead of via a user bug report. The AGENTS.md "Performance" invariants are the written half of this; this is the tooling half.

AngriestBird:
## Summary

Adds `/lens-perf`, which reports the slowest latency phases by independent p50 and p99 rankings for the current process session and the machine-wide active log window. The bounded asynchronous reader keeps the command responsive on large logs and reports truncation or malformed data instead of silently treating it as clean.

Closes #767

## What changed and why

The command flushes pending latency writes, streams at most the newest 10MB, yields while parsing, and retains separate bounded sample rings so machine-wide traffic cannot evict current-session data. Logger-owned PID and timestamp attribution provides an honest current-process scope, while startup total and scan-context phases make the regression that motivated the issue visible in the report.

Percentiles use positive-duration phase samples, qualify generic phase names as `toolName/phase`, and rank p50 and p99 independently before rounding displayed values.


apmantza:
many thanks again @AngriestBird. merged.

Issue #888, opened 2026-07-29
Per-edit tree-sitter runner still walks the tree once per rule — port #877's batched query to the hot path


apmantza:
@-

apmantza:
Fixed by #907 (merged): the per-edit runner now dispatches all rules via one batched runQueriesOnFile call.

Issue #885, opened 2026-07-29
Entity-extraction skip threshold is inverted — small edits still pay ~500-800ms


apmantza:
@-

apmantza:
Fixed by #907 (merged) plus the follow-up on master (55765f44): the skip threshold now gates correctly and absent modifiedRanges is treated as a full-file dispatch.

PR #911, opened 2026-07-29
fix(dispatch): run entity extraction when modifiedRanges is absent (refs #885)

apmantza:
Follow-up to #907, which merged before review findings were addressed.

## Summary
- #907's `skipEntityExtraction` threshold treated absent/empty `modifiedRanges` as "0 lines changed", so full-file dispatches (e.g. the MCP review facade at `clients/mcp/analyze.ts`, fresh file writes) never ran entity extraction or blast-radius — a regression vs pre-#907 behavior. The guard now treats absent/empty ranges as non-trivial, matching `isLineInModifiedRanges` semantics.
- The batched `query_error` log now carries a `queryIds` metadata field so a batch-level failure remains triageable from `~/.pi-lens/tree-sitter.log` (the old per-rule `queryId` was lost in the #888 batching).
- Regression test added (`modifiedRanges: undefined` must still extract) — verified to fail against the unfixed guard.


apmantza:
Superseded: commit 79baa3ee pushed directly to master (cherry-picked as 55765f44) per maintainer request.

Issue #963, opened 2026-07-31
bug(security): ts-ssrf flags fixed provider endpoints as SSRF


apmantza:
## What happened

`lens_diagnostics mode=full refreshRunners=all` reports three blocking `ts-ssrf` findings in a TypeScript extension whose outbound destinations are fixed provider endpoints:

- `providers/qoder/auth.ts:148` — `fetch(USERINFO_URL)` where `USERINFO_URL` is the module constant `https://openapi.qoder.sh/api/v1/userinfo`.
- `providers/qoder/auth.ts:255` — the same fixed `USERINFO_URL` in the OAuth flow.
- `providers/cline/cline-auth.ts:234` — `fetch(authUrl.toString())`, where `authUrl` is constructed from the fixed `BASE_URL_CLINE` constant plus a locally-generated `127.0.0.1` callback URL.

No request data, redirect response, or remote input controls these destinations. The Cline callback port is selected from a local fixed range. These are not SSRF sinks.

## Steps to reproduce

1. Use pi-lens 3.8.73.
2. Run `lens_diagnostics mode=full refreshRunners=all` in https://github.com/apmantza/pi-free.
3. Observe the three blocking `ts-ssrf` findings above.

## Expected behavior

The rule should distinguish fixed/allowlisted URL constants and locally constructed URLs from attacker-controlled URL fetches, or provide a documented project-level allowlist/taint escape hatch. These findings should not be blocking.

## Relevant context

The provider URLs are intentionally compile-time constants: the extension must call those vendor APIs and does not accept an arbitrary endpoint configuration.

Issue #533, opened 2026-07-11
Honest emptiness for scoped diagnostic queries: lsp_diagnostics workspace/directory on push-only servers + lens_diagnostics cold project mode must error or fall back, never render "0 diagnostics"


apmantza:
## Observed (live, 2026-07-11)

An agent session ran `lsp_diagnostics` and rendered **"workspace — 0 diagnostics"** against classic typescript — a server with NO workspace-diagnostics support (`ws-pull: ·` in docs/servercapabilities.md) AND `silentOnClean` push behavior (publishes nothing on clean). A "0" from that combination is unverifiable: it can mean clean, still-analyzing, or never-asked. This is the #240 doctrine (a failed/absent pull is never clean) applied to the TOOL surface.

## Maintainer directive

When a requested scope cannot be answered by the serving LSP (no workspace pull), the tool must **either exit with an explicit error or automatically fall back to the per-file scan ("all files")** — never silently render an empty result for a question that was never actually asked of the server.

## The class (fix together — conformity)

1. **`lsp_diagnostics` directory/batch/workspace modes** (tools/lsp-diagnostics.ts): per-file touches on a push-only silent-on-clean server produce ambiguous zeros. Investigate what `collectFileDiagnosticResult` does with a no-publish outcome — the LSP layer already has the discriminated `found|clean|unavailable` outcome (#240); the tool must PRESERVE that discrimination to the rendered result: "N files clean, M unresolved (server cannot confirm)" — an unresolved-majority result must not compact-render as "0 diagnostics". If a workspace-pull path exists (#235-adjacent), gate it on the capability snapshot's mode and apply the directive (error or auto-fallback to per-file).
2. **`lens_diagnostics` project/full mode** (clients/dispatch extractors registry, #395): cache-only extractors surface whatever is cached; a COLD cache must render as "cold — run the heavyweight scan / pilens_rebuild-style hint", never as zero findings. Audit each extractor's cold behavior (the #511/#514 honesty-warning pattern is the house shape for this).
3. **Sweep for other members**: any tool/render path where an \"empty\" result can mean \"couldn't ask\" (grep the compact renderers for zero-count summary lines and check each source's discrimination).

## Acceptance
- `lsp_diagnostics` on a dir/workspace served by classic typescript renders unresolved counts distinctly (or errors / falls back per the directive) — a clean render requires every file to have a CONFIRMED clean/found outcome.
- `lens_diagnostics` full mode on a cold cache says cold + how to warm, not 0.
[...]

PR #1008, opened 2026-07-31
fix(ssrf): don't flag fixed new URL(literal, constBase) endpoints as SSRF sinks, keep tainted detection (closes #1000, refs #963, #533)

### LINKED ISSUE #1000: ts-ssrf: avoid false positives for fixed endpoint URLs
## Summary

The `ts-ssrf` rule reports SSRF sinks for fixed, allowlisted provider endpoints where no user-controlled value reaches the fetch target.

Observed in a fresh project-wide pi-free scan (`lens_diagnostics mode=full refreshRunners=all`):

- `providers/qoder/auth.ts:148` and `:255`
  - `fetch(USERINFO_URL, ...)`
  - `USERINFO_URL` is a module constant: `https://openapi.qoder.sh/api/v1/userinfo`.
- `providers/cline/cline-auth.ts:234`
  - `fetch(authUrl.toString(), ...)`
  - `authUrl` is built with `new URL("auth/authorize", \`${BASE_URL_CLINE}/\`)`, where `BASE_URL_CLINE` is a fixed module constant.
  - The only dynamic value (`callbackUrl`) is placed in query parameters (`callback_url` / `redirect_uri`); it is a locally generated `http://127.0.0.1:<port>` callback and does not control the fetch destination.

These are fixed outbound OAuth/profile endpoints, not SSRF-capable sinks. The current findings are therefore false positives, while the rule should continue to report cases where attacker/external input taints the URL or base URL.

## Possible rule changes

1. Add constant/data-flow awareness: do not report `fetch()` when the target is a literal or a `const`/readonly value initialized from a literal URL, unless that value is subsequently tainted.
2. Treat `new URL(relativePath, fixedConstantBase)` as safe when the base is a fixed HTTPS endpoint and the relative path is a literal. Track dynamic query parameters separately from the URL origin/path; query values should not taint the request destination.
3. Keep reporting URL targets derived from function parameters, request data, environment/config values, parsed redirect/location values, or other external input.
4. Consider an explicit, rule-specific allowlist/suppression mechanism for intentional fixed third-party endpoints, requiring a rationale in the source.
5. Add regression fixtures for:
   - fixed URL constant -> no finding;
   - fixed URL base + dynamic query parameter -> no finding;
   - user-controlled URL/base -> finding;
   - user-controlled redirect/location subsequently fetched -> finding.

The pi-free code was not changed for this report; the endpoints above were manually verified as fixed/static.

apmantza:
Closes #1000 — `ts-ssrf` (the tree-sitter `ts_ssrf_sink`, not an ast-grep rule) reported SSRF on fixed outbound endpoints. Triage found FP1 (`fetch(USERINFO_URL)`, a module-const literal) was **already fixed by #963**; the real gap was FP2 — `fetch(new URL("auth/authorize", \`${BASE_URL_CLINE}/\`).toString())` — the `new URL(literalPath, fixedBase)` shape had no exemption (the URL capture is a `.toString()` call_expression, so the identifier-const check was skipped; base is a cross-file imported const; ctor is aliased).

## Fix (`clients/tree-sitter-client.ts`)
- Refactored the clean-const gate out of `resolvesToFileLiteralConst` into a shared `resolveFileConstValueNode` (same refuse-on-ambiguity/reassignment/non-const semantics — #963 behavior preserved).
- `resolvesToFixedNewUrlConst(name)`: true only for a clean file-local `const = new URL(<literalPath>, <fixedBase>)` — literal path (string / substitution-free template); base (`isFixedUrlBaseExpr`) must be a literal, substitution-free template, an identifier resolving to a file-local literal const OR an import binding, or a template whose every `${…}` is such. URL ctor may be aliased.
- `newUrlBaseVarName` extracts the receiver from `u.toString()`/`u.href`, wired after #963's check; anything not provably fixed falls through to the unchanged broad heuristic.

## The one security judgment call (flagged for review)
An **imported** const base is treated as fixed (required to fix the real cline case). Rationale: a module-level const is initialized at import time and cannot carry request/user taint (the SSRF threat model); env-derived operator config is out of SSRF scope. **Direct env/param/request taint AT THE SINK still fires.** Trade-off documented in-code. Fail-closed alternative (drop `isImportedBinding`) would leave the exact cline FP reporting → source suppression.

## TP preservation (the point)
`tests/clients/tree-sitter-security-gap-rules.test.ts`: 3 FP-exemption fixtures (file-local const base; imported base + aliased ctor; single-arg `new URL().href`) → 0 findings; **4 must-still-fire** (base from function param; tainted path; **const initialized from `process.env`**; parsed `resp.headers.location` redirect) → all fire. Proven non-vacuous: over-broadening the base check flips the param-base + env-const guards to FAILING, then reverted. 36 + 54 tests pass; `npm run build` + full `tsc` clean.

Closes #1000. Refs #963, #533.


apmantza:
Adversarial review found two exemption bypasses; both now fixed in `1837b8f4`:

1. **Post-construction origin mutation** (a regression the exemption introduced — master fires, the PR was suppressing): `const u = new URL(literal, fixedBase); u.host = req.x; fetch(u.toString())` was exempted even though the origin was rewritten. Now fail closed on member/subscript property writes + augmented assignments to the receiver; query-only writes (`u.search`, `u.searchParams.set`) stay exempt.
2. **Scope-blind base**: a request-tainted function *parameter* base was exempted when an unrelated same-named module-level const literal existed. Now refuse the file-const path when an enclosing function binds a same-named parameter (destructuring-aware). Imported bindings stay trusted — imported-base-as-fixed was reviewed and confirmed sound.

Added 4 must-fire fixtures (verified to fail on pre-fix code) + 2 must-stay-exempt fixtures. `build` + full `tsc` clean; gap-rule suite 42 pass.

PR #1011, opened 2026-08-01
Config/single source of truth for cli flags

### LINKED ISSUE #166: refactor: make ~/.pi-lens/config.json able to express every CLI flag/toggle (single source of truth)
## Problem

Users can set persistent defaults for only a handful of pi-lens flags. Most flags are session-only (CLI) with no global-config equivalent, so anyone who always wants e.g. `--no-tests` or `--lens-guard` has to pass it on every launch.

The flag → config mapping is also hand-maintained in two disconnected places — `pi.registerFlag(...)` in `index.ts` and an if/else chain in `resolvePiLensFlag()` in `clients/lens-config.ts` — which is how the coverage gap crept in.

## Current coverage

| CLI flag | type | config-backed today? |
|---|---|---|
| `--no-lens` | bool | ❌ |
| `--no-lsp` | bool | ❌ |
| `--no-autoformat` | bool | ✅ `format.enabled` |
| `--immediate-format` | bool | ✅ `format.mode` |
| `--no-autofix` | bool | ❌ |
| `--no-tests` | bool | ❌ |
| `--no-delta` | bool | ❌ |
| `--lens-guard` | bool | ❌ |
| `--lens-semgrep` | bool | ❌ |
| `--lens-semgrep-config` | string | ❌ |
| `--no-read-guard` | bool | ❌ |
| `--no-lens-context` | bool | ✅ `contextInjection.enabled` |

Plus four config-only knobs with no CLI flag (`lens-actionable-warning*`) and config keys that aren't flags at all (`dispatch.runnerTimeoutFloorMs`, `widget.visible`). The model is inconsistent in both directions.

## Goal

Every flag/toggle should be settable in `~/.pi-lens/config.json`, with **one consistent precedence** everywhere: **env → CLI flag → config → built-in default**.

## Proposed approach

1. **Single declarative flag registry** — one table that is the source of truth for each setting: `{ flagName, type, default, configPath, envVar? }`. Drive `registerFlag()`, `resolvePiLensFlag()`, and env resolution from it instead of three hand-written lists. New flags then get CLI + config + env for free.

2. **Backfill config coverage** for the nine missing flags. Use **positive** config keys (`enabled: false` semantics) for consistency with the existing `format`/`contextInjection` style, rather than mirroring the negative `no-*` flag names:

   ```json
   {
     "lens":       { "enabled": false },   // --no-lens
     "lsp":        { "enabled": false },   // --no-lsp
     "autofix":    { "enabled": false },   // --no-autofix
     "tests":      { "enabled": false },   // --no-tests
     "delta":      { "enabled": false },   // --no-delta
     "readGuard":  { "enabled": false },   // --no-read-guard
     "guard":      { "enabled": true  },   // --lens-guard
     "semgrep":    { "enabled": true, "config": "p/ci" }  // --lens-semgrep[-config]
   }
   ```
   (Keep accepting the already-shipped `format`, `contextInjection`, `widget`, `actionableWarnings`, `dispatch` sections; this is additive — no breaking changes.)

3. **Generalize `resolvePiLensFlag`** to look the flag up in the registry (config path + type coercion + bool/string handling) instead of the per-name if/else chain.

4. **Validation/diagnostics** — optionally warn on unknown top-level config keys (currently silently dropped) and surface the effective resolved config in `/lens-health` so users can confirm what actually took effect.

## Non-goals
- No change to flag precedence semantics beyond making it uniform.
- No removal of existing config sections or CLI flags (backward compatible).

## Acceptance
- Each flag in the table above is settable via config and resolves with env → CLI → config → default.
- A single registry drives flag registration + resolution; adding a flag in one place wires all three layers.
- Tests cover the resolution matrix (config-only, CLI-override, env-override, default) for representative bool and string flags.
- README "Global Config" section documents the full key set.

AngriestBird:
## Summary

Briefly describe what this PR does and which issue it resolves.

- Every CLI flag is now able to be pinned as always on via one registry allowing you to permanently enable via pi-lens.json or via the CLI

Closes #166


apmantza:
thanks for taking the time @AngriestBird! merged.

PR #1026, opened 2026-08-01
fix(python): bare-except rule flags dotted/qualified exception types as bare

cy7372:
## Problem

The `bare-except` rule (tree-sitter query + `bare_except_only` post-filter) produces **false-positive blocking diagnostics** for every Python `except` clause that uses a dotted/qualified exception type:

```python
except asyncio.TimeoutError:   # flagged as bare-except ❌ (false positive)
except json.JSONDecodeError:   # flagged as bare-except ❌ (false positive)
```

This is the common Python idiom, so the rule is effectively broken for most real-world Python code.

## Root cause

The `bare_except_only` post-filter (`clients/tree-sitter-client.ts`) decides whether an `except` clause is bare by checking its named children for an exception-spec kind. It recognizes:

- `identifier` — `except ValueError`
- `tuple` — `except (E, F)`
- `as_pattern` — `except E as e`
- `parenthesized_expression`

But a dotted name like `asyncio.TimeoutError` parses in tree-sitter-python as an **`attribute`** node — which is not in that list. So the post-filter treats every qualified `except` as bare and fires.

> Note: the sibling ast-grep rule `rules/ast-grep-rules/rules/no-bare-except.yml` has the same blind spot, but ast-grep napi does not support Python, so that rule is not actually executed against Python files — the tree-sitter query is the live path.

## Reproduction

Fixture (`refreshRunners=cheap`):
```python
try:
    pass
except asyncio.TimeoutError:   # flagged 🔴 (wrong)
except ValueError:             # not flagged ✅
except (A, B):                 # not flagged ✅
except Exception as e:         # not flagged ✅
except:                        # flagged 🔴 (correct)
    pass
```

Python `ast` confirms **0** true bare-excepts in the affected real-world file, yet pi-lens reported 8 blocking errors.

## Fix

Add `attribute` to the recognized exception-spec kinds in `bare_except_only`:

```diff
 return (
   c.type === "identifier" ||
+  c.type === "attribute" ||
   c.type === "tuple" ||
   c.type === "as_pattern" ||
   c.type === "parenthesized_expression"
 );
```

## Verification

- New regression test in `tests/clients/tree-sitter-879-post-filters.test.ts` asserts: only a true `except:` matches; identifier/tuple/as_pattern/attribute forms do not.
- `npm run lint` ✅
- `npx vitest run tests/clients/tree-sitter-879-post-filters.test.ts` → 11/11 passed ✅

## Checklist
- [x] `npm run lint` passes
- [x] `npm test` — post-filter suite passes; remaining failures in the broader client suite are pre-existing environment issues (missing ast-grep native binary from `--ignore-scripts` install, jedi LSP, git binary) unrelated to this change
- [x] `package-lock.json` unchanged


apmantza:
many thanks @cy7372.

PR #1045, opened 2026-08-02
feat(config): per-rule enable/disable in .pi-lens.json (closes #444)

### LINKED ISSUE #444: feat(config): per-rule enable/disable in .pi-lens.json (project-level suppression)
Split from #442. Today the only ways to suppress a finding are inline `# pi-lens-ignore: <rule>` comments (now honored in both mode=all and mode=full after #442) and `.pi-lens.json` `ignore` **path globs** (which drop a whole file/dir from every scan). There's no way to disable a single **rule** project-wide without editing every call site.

`.pi-lens.json` `rules` currently only carries complexity thresholds — extend it (or add a sibling key) to express per-rule enable/disable, e.g.:

```json
{
  "rules": {
    "disable": ["no-eval", "ast-grep:some-rule"],
    "only": ["..."]
  }
}
```

Applies uniformly to mode=all and mode=full (same seam as the #442 inline filter). Normalize rule ids the same way (`ast-grep:<id>`/`<id>-js`). Complements inline suppression: inline = per-site, config = project-wide.

Acceptance: a rule listed in `.pi-lens.json` disable is dropped from both dispatch and the full sweep; documented; tested.

AngriestBird:
## What

`.pi-lens.json` can now carry `rules.<id>.disable` / `rules.<id>.select` alongside the existing complexity thresholds, so a project can silence or allowlist individual rules without touching every call site. The policy applies on the per-edit dispatch path and across `lens_diagnostics` `mode=delta`, `mode=all`, and `mode=full`, re-summarizing so blocking/error/warning counts reflect the drop.

## Why

Closes #444. Suppression was previously either inline (`# pi-lens-ignore: <rule>`, per-site) or `ignore` path globs, which drop a whole file or dir from every scan. Nothing sat in between, so there was no way to turn off one noisy rule project-wide.

## How

Deviations from the issue sketch worth calling out:

- Schema is `rules.<id>.disable` / `rules.<id>.select`, not top-level `rules.disable` / `rules.only`. This keeps the existing per-rule shape instead of adding a sibling key. The `<id>` key is a grouping label, not a filter scope: every `disable` list unions into one drop list and every `select` list into one allowlist, so `"security": { "disable": ["no-eval"] }` works without `security` being a real rule id. Disable wins over select project-wide, across keys.
- A non-empty `select` anywhere silences every rule not listed, project-wide. It's a big hammer, and it's documented as one.
- Filtering is output-only. The baseline, widget state, and dedup cache still see everything, so editing the policy never corrupts or resets delta tracking.
- Findings that carry no rule id are exempt from `select`. `Diagnostic.id` is a dedup key rather than a rule id, and measuring it against an allowlist would have silently dropped blocking findings like eslint parse errors.
- Rule-id normalization (`ast-grep:<id>`, `<id>-js`) is consolidated in `clients/dispatch/rule-id-normalize.ts` so inline suppression, dedup, and the policy matcher can't drift. That normalization conflates a rule whose real name ends in `-js` with its stem (`prefer-js` with `prefer`); pinned by tests and called out in the docs.
- Dispatch resolves the config from the project root, so a package-local `.pi-lens.json` in a monorepo can't shadow the root's policy. That matches the root `lens_diagnostics` already used.

Lists are parsed tolerantly in `clients/project-lens-config.ts`: a non-array, an empty array, or an all-whitespace array is logged once and dropped, matching the existing threshold-validation warn-once contract. Threshold-only and policy-only entries both work. #533 hygiene is preserved, so typo and unknown-key warnings still surface and foreign LSP namespaces stay silent.

Docs: `docs/settings.md` and `docs/globalconfig.md` gained the schema, precedence, the select warning, and the `-js` caveat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,51 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Added
 
+- **Project-level rule policy via `.pi-lens.json` `rules.<id>.disable` / `rules.<id>.select`** — a project's own config can now narrow what diagnostics actually surface.
+	Filtering is output-only, so the baseline, widget state, and dedup cache stay
+	authoritative and a policy edit never corrupts or resets delta tracking.
+	Matching is project-wide: the `<id>` key is a grouping label, not a filter
+	scope. Every `disable` list across every key unions into one drop list, and
+	every `select` list unions into one allowlist, so
+	`"security": { "disable": ["no-eval"] }` works without `security` being a real
+	rule id. Disable wins over select project-wide, even across different keys.
+	A non-empty `select` anywhere silences every rule not listed, project-wide, so
+	it is a big hammer (documented as such). Findings that carry no rule id at all
+	are exempt from select, since `Diagnostic.id` is a dedup key rather than a rule
+	id and measuring it against an allowlist would have silently dropped blocking
+	findings like eslint parse errors. A user lists `no-eval` once and the filter
+	covers `no-eval`, `ast-grep:no-eval`, and `no-eval-js`, reusing the rule-id
+	normalization that `inline-suppressions.ts` and `tools/lens-diagnostics.ts`'s
+	dedup already applied, now consolidated in
+	`clients/dispatch/rule-id-normalize.ts` so the three surfaces cannot drift.
+	That normalization conflates a rule whose real name ends in `-js` with its stem
+	(`prefer-js` with `prefer`), which is now pinned by tests and called out in the
+	docs. The policy applies on the per-edit dispatch path
+	(`clients/dispatch/dispatcher.ts`, resolved from the project root so a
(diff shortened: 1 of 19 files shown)


AngriestBird:
@apmantza This should be ready to go and get more project/user based configuration settled.

apmantza:
Follow-up fixes pushed to the PR branch in commits 9f940d98 and c277d792. The per-edit path now carries RuntimeCoordinator.projectRoot separately from the nested language/tool root through PipelineContext, dispatchLintWithResult, and DispatchContext; the monorepo regression exercises the distinct roots. Full-mode policy now still applies when a file cannot be reread, and code-only project diagnostics use code consistently so structured counts match rendered output. Focused policy/dispatch/pipeline tests pass (174 tests total across the touched suites), with build and lint passing.

apmantza:
many thanks @AngriestBird. minor follow-up commits. merging.

Issue #1024, opened 2026-08-01
bug(dispositions): agent's own false-positive/flagged mark silently defeated (raw-key write vs normalized read) — #1020 class


apmantza:
## Confirmed live bug (path-key divergence class — sibling of #1020/#210)

When the agent marks a diagnostic `false-positive`/`flagged` via `lens_diagnostic_mark`, the disposition is anchored under a **raw, un-normalized** path, but `applyDispositions` looks it up under a **normalized** key — so the mark is invisible and the diagnostic keeps re-firing / re-blocking on every subsequent turn.

- WRITE (raw): `tools/lens-diagnostic-mark.ts:253,292-294` passes raw `cwd` / `path.resolve(cwd, filePathArg)` into the disposition anchor; `clients/diagnostic-dispositions.ts:110-113` (`relativeFile`) takes the **un-normalized** branch whenever `path.relative` succeeds — no `normalizeMapKey`.
- READ (normalized): `clients/dispatch/dispatcher.ts:151-194` (`createDispatchContext`) resolves via `normalizeMapKey(...)` (realpath-canonicalized cwd/filePath).

**Consequence (#533-adjacent):** the agent's own explicit dismissal is silently dropped and the "resolved" diagnostic resurfaces as a phantom repeat block — a resolved state rendering as still-broken, on the exact surface meant to let the agent suppress noise. Most likely to bite on Windows (`\` vs `/`) and where `path.relative` succeeds with a different form than the realpath-canonicalized read.

## Fix
Thread the write side through the SAME normalization the read uses — anchor the disposition via `normalizeMapKey` (or route the mark tool through `createDispatchContext`'s canonicalization) so write and read keys are identical. Do not weaken the read.

## Regression test (must fail today)
Record a `false-positive` disposition under one path form (e.g. forward-slash / `path.relative` form), then run `applyDispositions` with the normalized/realpath form for the same file, and assert the diagnostic is suppressed. Today it is not (keys differ → mark not found).

Found by the #1020 class sweep. Same defect shape as #210 (read-guard raw keys) and #1020 (widget-state raw keys). Umbrella + other sweep findings: see the path-key class tracking issue.

Issue #523, opened 2026-07-11
read_symbol: did-you-mean candidates on miss + Class.method qualification


apmantza:
Dogfooding 2026-07-11 (post-#517): a `read_symbol` miss returns `found:false` plus a hint to call `module_report` — a typo or a qualified `Class.method` name costs the agent a full extra round-trip (miss → module_report → retry).

The outline already knows every symbol name at miss time, so the miss can be self-healing:

1. **Did-you-mean on miss** — embed the ~3 nearest symbol names (Levenshtein, same approach as the read-guard's did-you-mean suggestions — conformity with the existing pattern, refs #505) directly in the miss response, on both the pi tool and the MCP mirror.
2. **Accept `Class.method` qualification** — members are already extracted (line-range containment nesting, #301); a dotted name should resolve to the member instead of missing.

Deliberately NOT included (assessed and rejected): batch multi-symbol reads — saves only per-call overhead while complicating the read-guard recording contract, and ranked `recommendedReads` already sequences single calls well.

[S] effort. Sibling follow-ups in #522 (read_enclosing MCP parity + size-aware honesty).

Refs #245, #517, #505.

apmantza:
Scope additions from the 2026-07-11 read_symbol assessment (live-probed):

3. **Include the attached doc comment in the returned source** — the body currently starts at the declaration line (verified: `isAgentNudgeEnabled` returns from `export function...`, its contract comment excluded). An agent reading a symbol to edit it needs the contract above it; excluding it forces a second read. #517's `extractDocComment` already computes attachment (position-based, blank-line-gap aware) — extend the returned range to the comment block start. The read-guard's `recordSymbolRead` range must cover the comment lines too, so editing the doc comment doesn't trip a zero-read block. Probably the highest-value item in this issue.
4. **Duplicate-name disambiguation** — same-file name collisions (overloads, type+value) currently return the first match silently; the result should note the ambiguity ("2 matches — returned the function; pass kind to disambiguate") and accept an optional `kind`. Edge-case-grade, do last.

(Callback-handle reads verified still working post-#517.)

Issue #505, opened 2026-07-10
Read-guard: fuzzy did-you-mean suggestion when hash relocation fails (+ confusable-hyphen normalization)


apmantza:
## What

When the read-guard's content-verified range relocation (#144 family) truly fails — no hash match, relocation window exhausted — the block error today gives no hint where the intended text actually lives. Borrow the fallback UX from jarkkojs/pi-readseek (`hashline.ts` `findSimilarLines`): a token-Jaccard similarity scan (threshold ~0.8, bounded window ±50 lines) over the current file that appends a **did-you-mean suggestion** to the error — "closest match for your oldText is now at lines 118-121 (similarity 0.87)". SUGGESTION ONLY, never auto-apply — the fail-safe-to-block posture (#144) is untouched; this just turns a dead-end error into a one-hop recovery for the agent.

Bundle in the same PR (near-one-liner, same comparison path): **Unicode confusable-hyphen normalization** before content comparison (readseek's `confusable-hyphens.ts` idea — U+2010/2011/2012/2013/2014/2212 → ASCII hyphen), matching the existing whitespace-tier tolerance philosophy (Tier A/B autopatch) without weakening verification: normalize both sides identically, only for COMPARISON, never for the written content.

## Where
clients read-guard relocation seam (relocateEditRange / the mismatch-error construction) — extend the existing did-you-mean plumbing (the whitespace-tier suggestions already exist; this adds the similarity fallback when those tiers also miss).

## Tests
- Suggestion fires on a moved-and-edited block (hash miss + relocation miss ⇒ nearest candidate reported with line range).
- Never auto-applies; block outcome unchanged.
- Confusable hyphen in oldText vs ASCII in file ⇒ comparison tier matches (and vice versa); written bytes never normalized.
- Cross-form path keys per the #210 invariant.

_Origin: 2026-07-10 inspiration research (pi-readseek). Effort: S._

PR #1051, opened 2026-08-03
fix(read-guard): did-you-mean must quote raw bytes, not normalized text (closes #1050)

### LINKED ISSUE #1050: Read-guard did-you-mean suggests NFKC-normalized text as "current lines" (CJK punctuation)
### What happened?

The `Did you mean one of these current lines?` block (#505) prints lines taken from the **NFKC-normalized** view of the file, not the file's actual bytes. For CJK text this silently rewrites the punctuation the agent is being told to copy verbatim:

```
file on disk : - [05 执行生命周期：批量要不要脱离 HTTP 进程](issues/05.md) — **选 (d) 混合，但续跑是手动的**；
suggested    : - [05 执行生命周期:批量要不要脱离 HTTP 进程](issues/05.md) - **选 (d) 混合,但续跑是手动的**;
```

NFKC folds `：`→`:`, `，`→`,`, `；`→`;`, `（）`→`()`, and `HOST_UNICODE_DASHES` folds `—`→`-`. The suggestion is presented as "current lines", so it reads as ground truth about file content.

This contradicts #505's own stated invariant — *"normalize both sides identically, only for COMPARISON, never for the written content"* — and the comment above `normalizeContent` in this same file ("Comparison-only, same as every tier below it ... never this normalized form").

**Why it's more than cosmetic.** The agent copies the half-width suggestion into `oldText` and retries. The guard passes (it normalizes both sides), and the host's `fuzzyFindText` then misses exact match, falls into fuzzy space, and `applyReplacementsPreservingUnchangedLines` rewrites **the touched lines** from the normalized base. The full-width punctuation on the matched line is silently replaced with ASCII, and the tool still reports success. Verified on host 0.83.0:

```
before: - [05 票名：说明](issues/05.md) — **选 (d)**；结尾。
after : - [05 票名:说明](issues/05.md) - **选 (d)**;结尾。
        - [06 新票：追加行](issues/06.md) — 新增。      <- newText, intact
```

Untouched lines are safe (earendil-works/pi#5899), and touched-line normalization was closed as wontfix upstream (earendil-works/pi#3554: *"use a custom edit tool"*). So the host will not change; the suggestion is what steers the agent into that path. Worse, on the next turn the agent re-reads the file, sees changed punctuation, and concludes a markdown formatter is mangling the file — a plausible-but-wrong diagnosis it then acts on. That is how I found this.

The most common trigger for CJK markdown is the ordinary "append an item after this list entry" edit, where the anchor line *is* the line that gets rewritten.

### Steps to reproduce

1. Create a markdown file with a long CJK list item containing `：`, `，`, `；`, `—`.
2. Issue an `edit` whose `oldText` fails to match (e.g. one wrong character deep in a long line), anchored on that item.
3. The block error's `Did you mean one of these current lines?` shows that line with half-width punctuation.
4. Copy the suggested line into `oldText`, keep it as the anchor, append a line in `newText` → the anchor line lands on disk half-width, reported as success.

### Expected behavior

Suggestions quote the file's real bytes. `normalizeForGuardMatch` is line-count-preserving (its own docstring says so), so the line numbers from the normalized scan index `rawContentLf` 1:1 — score in normalized space, render from raw:

```ts
// findSimilarLines: score on normalized text, return the raw line
const rawLines = rawContentLf.split("\n");
// ... scored.push({ line: i + 1, text: rawLines[i], score });
```

`rawContentLf` (line 366) is already in scope at the call site (line 472). Alternatively have `findSimilarLines` return only `{ line, score }` and let `formatSimilarLines` look up raw text, which makes it structurally impossible to leak normalized text into an error message.

Refs: `clients/read-guard-tool-lines.ts` — `findSimilarLines` (82-102, `content.split("\n")` at 91, `text` captured at 98), `formatSimilarLines` (104-113), `const content = normalizeContent(rawContent)` (367), call site (472-476).

Suggested test: file with full-width `：，；` + `—`, force a miss, assert the suggestion block contains the full-width forms (today it contains the folded ASCII forms).

### pi-lens version

3.8.74 (host `@earendil-works/pi-coding-agent` 0.83.0)

floatGray:
## Summary

The read-guard's `Did you mean one of these current lines?` block (#505) rendered its rows from the **NFKC-normalized** match space instead of the file's real characters. For CJK markdown that folds `：`→`:`, `，`→`,`, `；`→`;`, `（）`→`()` and `—`→`-`, and presents the result as ground truth about file content — inverting #505's own stated invariant that normalization is *"only for COMPARISON, never for the written content"*.

Closes #1050

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:read-guard

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes

Notes on two of those boxes:

- **`npm test`** — 3 tests fail on this branch, and the same 3 fail on untouched `master` with my changes fully stashed (`coderabbit-ast-grep-rules`, `diagnostic-dispositions` #1024 anchor path-form, `sgconfig` #497 precedence, plus a `project-diagnostics` 5s timeout that comes and goes between runs). They're flaky/pre-existing, not from this PR. `tests/clients/read-guard-tool-lines.test.ts` is 79/79 green.
- **`AGENTS.md`** — no change needed. This restores the documented invariant rather than altering it; the file's only mention of `findSimilarLines` is the #523 note explaining why `read_symbol` deliberately does *not* reuse it, which is still accurate.
- **`package-lock.json`** — deliberately untouched. `npm install` reformats the whole file and bumps a stale `3.8.73`→`3.8.74` that `master` already carries; I reverted it to keep the diff focused. `npm run check:lockfile` passes against the committed lockfile.

## What changed and why

**Why this is worse than a misleading error message.** The suggestion steers the agent into real corruption. Copying the half-width row into `oldText` passes the guard (it normalizes both sides), then the host's `fuzzyFindText` misses exact match, falls into fuzzy space, and `applyReplacementsPreservingUnchangedLines` rewrites the **touched** line from the normalized base. Full-width punctuation becomes ASCII, and the tool reports success. Verified on host 0.83.0:

```
before: - [05 票名：说明](issues/05.md) — **选 (d)**；结尾。
after : - [05 票名:说明](issues/05.md) - **选 (d)**;结尾。
        - [06 新票：追加行](issues/06.md) — 新增。      <- newText, intact
```

Untouched lines are safe (earendil-works/pi#5899) and touched-line normalization was closed **wontfix** upstream (earendil-works/pi#3554 — *"use a custom edit tool"*), so the host won't change and the suggestion is the only place this is fixable.

It's also self-reinforcing, which is how I found it: the agent re-reads the file next turn, sees changed punctuation, and concludes a markdown formatter is mangling the file — then acts on that wrong root cause. The trigger for CJK markdown is the ordinary "append an item after this list entry" edit, where the anchor line *is* the line that gets rewritten, and full-width→half-width is nearly invisible in review.

**The fix.** `findSimilarLines` now returns only `{ line, score }`; `formatSimilarLines` takes the raw text and renders from it. Dropping `text` from the scan's return type makes leaking normalized text structurally impossible rather than merely avoided — there is no longer a normalized string in scope at the render site.

**The subtle part — which raw view.** My first attempt passed the `rawContentLf` already in scope (`raw.replace(/\r\n/g, "\n")`) and was wrong. The match space is `normalizeForGuardMatch` = `normalizeForFuzzyMatch(normalizeToLF(stripBom(text).text))`, and `normalizeToLF` also folds **lone** `\r`, which a `\r\n`-only replace does not. On a lone-CR file the two views have different line counts, so every suggestion after the first CR misindexes:

```
lone CR   normLines: 3  rawLfLines: 2  *** DESYNC ***
BOM       normLines: 2  rawLfLines: 2  OK
CRLF      normLines: 2  rawLfLines: 2  OK
```

So the renderer gets `normalizeToLF(stripBom(rawContent).text)` — the same **structural** normalization as the match space (line indices cross-index 1:1, BOM included), none of the **character** folding (full-width punctuation, smart quotes, NBSP survive). Both helpers were already exported from `clients/host-edit-normalize.ts`. The `?? ""` on the line lookup is belt-and-braces; with structural alignment the index is always in range.

**Tests.** Two regression tests in the existing `did-you-mean suggestions` describe block: full-width CJK punctuation quoted verbatim (asserting the folded forms are *absent*, so it can't pass vacuously), and correct cross-indexing on a BOM + lone-CR file. Both were confirmed to **fail on the unfixed code** (`git stash` the source change, rebuild, run) and pass with it — the second one is what caught my own `rawContentLf` mistake.


apmantza:
many thanks for fixing this @floatGray. merged.

Issue #1053, opened 2026-08-03
feat(partial-apply): apply preflight-resolved spans with stale-batch validation


apmantza:
## Purpose

Replace the current partial-apply mutation step with a small, range-safe Phase 1 that consumes the exact replacements already proven safe by read-guard preflight.

Supersedes #144's original broad internal-edit-substrate scope. This issue is intentionally limited to **single-file partial apply**.

## Reuse-first constraints

Do not introduce a parallel `clients/lens-edit.ts` abstraction or duplicate mutation bookkeeping. Extend the existing seams:

- `clients/read-guard-tool-lines.ts` — authoritative oldText resolution and host-compatible matching;
- `clients/partial-edit-apply.ts` — partial-apply integration seam;
- `clients/runtime-coordinator.ts` — project/file sequence state;
- `clients/atomic-write.ts` — atomic file replacement;
- `clients/file-time.ts` — same-file serialization via `FileTime.withLock`;
- `ReadGuard`, `CacheManager`, `appendProjectChange`, and `handleToolResult` — existing bookkeeping and post-edit pipeline.

Any new helper must be a narrow extension of one of these seams and justify why reuse is insufficient.

## Required behavior

1. Have preflight carry the original snapshot identity and exact replacement spans into apply; never re-search `oldText` against a changed buffer.
2. Validate the complete batch before writing:
   - captured file sequence/content identity is still current;
   - spans are in bounds and non-overlapping;
   - expected text, when present, matches the original snapshot.
3. Reject the whole batch before any write on stale, invalid, or overlapping input, with a structured re-read result.
4. Serialize snapshot-validation plus file commit with `FileTime.withLock`.
5. Apply disjoint edits bottom-up (or with one equivalent offset transform) using the existing host line-ending/BOM/Unicode behavior from `host-edit-normalize.ts`.
6. Commit through `writeFileAtomicAsync` or the existing shared writer; do not add another atomic-write implementation.
7. Reuse the existing post-edit route through `handleToolResult`; do not call `runPipeline` a second time.
8. Preserve file-aware mutation flags, session identity, path normalization, modified-range tracking, read-guard write stamping, and change-log attribution.
9. Distinguish a file-commit failure from a committed mutation whose bookkeeping or post-edit analysis failed; never retry the native edit after the file was committed.

Use “one sequence/change-log entry per logical mutation.” Formatting/autofix side effects may legitimately create additional mutation entries and must not be double-counted by the partial-apply layer.

[...]

PR #1066, opened 2026-08-03
fix(lsp): preserve workspace edit ordering

mjfaga:
## Summary

- preserve LSP `documentChanges` ordering across text and resource operations, including directory subtrees
- stabilize same-position insert ordering and collapse byte-identical non-empty replacement duplicates
- prevalidate every text-edit batch before the first filesystem mutation
- close an open old URI before `workspace/didRenameFiles` after a file rename

This ports the contained workspace-edit safety patterns from OMP while preserving pi-lens's safer solicited-only `workspace/applyEdit` gate and existing read-guard/change-tracking lifecycle.

Refs #1053.

## Verification

- `npm run build`
- `npx vitest run tests/clients/lsp/edits.test.ts tests/clients/lsp/service-rename-file.test.ts tests/clients/lsp/client-crash-logging.test.ts` — 21/21 pass
- `npm run lint`
- full suite — 5084 tests pass; the same two pre-existing local baseline failures remain:
  - `diagnostic-dispositions.test.ts` macOS path-case assertion
  - `coderabbit-ast-grep-rules.test.ts` temporary relative rule-directory resolution
- independent Johnny review: PASS, no findings

## Scope

No public tool/schema changes, no broad edit substrate, no unconditional server edits, and no directory-rename enumeration.


apmantza:
thanks @mjfaga. I went down through some rabbit holes on this one and added a lot of follow-up commits. i am happy of how this is supposed to work now - will be testing it more. merging.

PR #1077, opened 2026-08-04
fix(ast-grep): stop false nested-link diagnostics

### LINKED ISSUE #1076: no-nested-links flags every JSX ancestor containing an anchor
## What happened

The bundled `no-nested-links` ast-grep rule reports every outermost JSX ancestor that contains an `<a>`, even when there is only one anchor or multiple sibling anchors.

For example, this valid JSX is reported:

```tsx
<div><a href="/x">link</a></div>
```

A component containing two sibling links also reports its non-anchor wrapper as a nested-link violation.

## Root cause

The rule requires only that the matched `jsx_element` contain an anchor descendant. Its `not: inside:` clause excludes elements beneath an anchor, but never requires the matched outer element to itself be an anchor.

The behavioral fixture currently classifies `<div><a ...>` as invalid. Git history shows this was moved from valid to invalid to tolerate an ast-grep 0.42 JSX CLI limitation. The project now uses ast-grep 0.45 and documents that the JSX limitation was fixed in 0.44, but the rule and fixture were not corrected afterward.

## Expected behavior

- A single anchor inside a non-anchor wrapper does not match.
- Sibling anchors do not match.
- An anchor containing another anchor directly or through intermediate JSX elements matches.

## Proposed scope

Require the matched outer `jsx_element` to have an opening tag named `a`, and restore/add behavioral fixtures for the valid and invalid shapes above.

marvtub:
## Summary

Fixes `no-nested-links` so it reports only an anchor containing another anchor, rather than every JSX ancestor containing a link.

Closes #1076

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [x] area:dispatch
- [ ] area:installer
- [x] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] I have read `CONTRIBUTING.md` and `AGENTS.md`
- [x] The change has tests covering the regression and edge cases
- [x] `npm run lint` passes
- [ ] `npm test` passes locally (see validation note below)
- [x] Build-dist is not required; no runtime TypeScript changed
- [x] `AGENTS.md` documents the corrected rule invariant
- [x] `CHANGELOG.md` has an Unreleased entry
- [x] Commit subject includes `(closes #1076)`

## What changed and why

The previous rule required only that the matched `jsx_element` contain an anchor descendant. Its `not: inside:` clause excluded elements under an anchor but never required the matched outer element to be an anchor. As a result, wrappers and sibling-link containers were reported.

The rule now separately requires:

1. the matched outer `jsx_element` has an opening tag named `a`; and
2. it has another anchor as a recursive JSX descendant.

Fixtures cover a single wrapped anchor, sibling anchors, direct nesting, and deep nesting through an intermediate element. The static rule has no review-graph/module blast radius; both consumers load the same shipped YAML.

## Validation

- Regression fixture failed before the rule change on both valid wrapper cases.
- `npx ast-grep test ... --filter no-nested-links`: 1 rule / 7 cases passed.
- Focused CLI + NAPI Vitest suites: 2 files / 9 tests passed.
- Real scan: zero findings in the two Lindy files that originally produced six false positives; a deeply nested synthetic anchor still reports.
- `npm run lint`, `npm run check:lockfile`, and `npm run changelog:check` passed.
- Full `npm test`: 443 files / 5,119 tests passed. Two unrelated local path/filesystem tests still fail deterministically (`coderabbit-ast-grep-rules` temp-dir traversal and `diagnostic-dispositions` mis-cased path); two parallel-load timeouts from the full run passed on focused rerun. The changed rule suites are green through both engines.

---
Generated with GPT-5.6.


apmantza:
many thanks for reporting and the PR @marvtub!

Issue #459, opened 2026-07-08
perf: skip the reverse-deps index rebuild + snapshot write when the graph build reports unchanged


apmantza:
Follow-on to #451. Every cascade run does, after the graph build (`clients/dispatch/integration.ts` `computeCascadeForFile`):

```
buildReverseDependencyIndexFromGraph({ cwd, graph })   // O(graph edges), every time
writeReverseDependencyIndexToSnapshot({ ... })          // disk write, every time
```

Measured: `reverse_deps_cache refresh_from_review_graph` mean **763ms × 1,599 runs** (includes graph time, but the index rebuild + snapshot write are a fixed tax on top of even a fully-cached graph hit).

When `getLastGraphBuildInfo()` reports the graph did NOT change — mode `cached`, or a `seq-fastpath` no-op (empty update set) — the reverse-dependency index is **unchanged by construction**: it's a pure function of the graph. Rebuilding and rewriting it is wasted CPU + IO in the deferred cascade path (background contention with the next edit's dispatch, post-#450).

## Sketch
- Cache the last-built index per workspace keyed by the graph's signature (the builder already computes/stores one); rebuild + persist only when the signature moved (mode `full`/`incremental`/`seq-fastpath` with a non-empty update set).
- Keep the merge-into-cascade-selection behavior identical — only the rebuild/write is skipped.
- Log a `reverse_deps_cache` event with `action: "reused_unchanged"` so the skip rate is visible in cascade.log.
- This is also a candidate first consumer of the shared **seq-guarded cache primitive** floated in #451's "broader thread".

Issue #210, opened 2026-06-14
read-guard: zero_read false-block when a file is read and edited under different path forms (separator/casing)


apmantza:
## Symptom

In a real session (`pi-free`), the agent read `providers/model-fetcher.ts` and `tests/kilo.test.ts` multiple times, yet every edit was rejected with `zero_read` ("Edit without read"). The agent fell back to `write` and a failed `bash` approach. `read-guard.log` shows the reads logged with **forward-slash** paths (`C:/Users/.../...`) and the blocking edits with **backslash** paths (`C:\Users\...\...`), `readCount:0, reads:[], reasonKind:"zero_read"`.

## Root cause

`ReadGuard` keyed its `reads`/`edits`/`exemptions`/`pendingCreations`/`writtenThisSession` maps on the **raw file-path string** — it relied on the read-path and edit-path strings being byte-for-byte identical. `resolveToolCallFilePath` (used by both the Read and Edit tools) returns absolute paths **verbatim**, so the key is whatever separator/casing the model emitted, and models freely mix `/` and `\` on Windows.

**Trigger (what changed):** read-guard began recording reads from new sources that produce a *different* path form than the Edit tool — ast_grep_search matches (#169, slash-normalized from ast-grep output) and LSP-expanded synthetic reads (URI → forward slash). On Windows a file "read" via search/LSP got a `C:/…` key while the subsequent edit arrived `C:\…` → key miss → false `zero_read`. Pre-#169, reads and edits both flowed through the Read/Edit tools with the same form, so the latent flaw stayed hidden.

## Why we didn't catch it

Every read-guard unit test used the **same POSIX path** on both `recordRead` and `checkEdit` (e.g. `/src/api.ts`), so the raw keys always matched. No test exercised cross-separator / cross-source path agreement — the exact axis that broke.

## Fix

Canonicalize every `ReadGuard` map key through `normalizeFilePath` (folds separators + Windows casing). Added `tests/clients/read-guard-path-normalization.test.ts` pinning that record and lookup agree across separator/casing forms, plus a negative (genuinely-unread file still blocks).

## Follow-up (separate)

Wire the pi built-in `grep` tool and `lsp_navigation` as read sources for parity with bash `grep -n` / `ast_grep_search` (they reveal the exact editable span). `ls`/`glob`/`find` intentionally excluded (name-only, not content).

PR #1082, opened 2026-08-05
test(project-intelligence): cover module call graph

### LINKED ISSUE #1073: Expose the derived call graph in module_report
## Context

#1070 integrates Tree-sitter/review-graph call/reference evidence into the existing function call-graph substrate. The review graph is the canonical evidence graph; the call graph is a derived caller/callee view.

As a consumer, `module_report` should present that derived call view alongside its existing symbol structure, `usedBy`, and import/dependency blast radius. The current #1070 scope explicitly left this response surface out so the evidence adapter could be corrected first.

## Goal

Expose a bounded, read-only `callGraph` section in `module_report` backed by the cached/derived call-graph view.

## Requirements

- Do not add a second source parser or graph authority: use the review graph → call-graph projection seam.
- Keep `usedBy`/references and import blast radius semantically separate from actual calls.
- Show per-symbol callers/callees where available, with stable symbol IDs and source locations.
- Include explicit availability and coverage (`complete`, `partial`, `unavailable`), languages, evidence counts, and truncation.
- Never rebuild the graph on a module-report query path; cold cache remains read-only and honest.
- Preserve `view: "summary"` and `view: "compact"` payload contracts with bounded call data.
- Keep pi and MCP surfaces identical through the lens-engine seam.
- Add fixture-grounded end-to-end tests: Tree-Sitter fixture → review graph → call-graph projection → module_report for TS/TSX, JS-family, Python, Go, Rust, and Java/Kotlin type-only references.

## Non-goals

- Compiler-grade dynamic dispatch/reflection resolution.
- Treating ambiguous, unresolved, type-only, or unsupported evidence as concrete calls.
- Removing the separately materialized call-graph cache if it remains a performance index; it must remain derived and freshness-coupled to the review graph.

apmantza:
## Summary
- add real committed-fixture coverage for the merged #1072 module_report callGraph projection
- verify caller/callee identities, source locations, bounds, views, cache identity/partial states, cold read-only behavior, and type-only references
- keep this follow-up test-only; no production or generated files changed

## Verification
- `npm run build`
- `npm run lint`
- targeted module-report/call-graph/review-graph suites: 184 passed
- `npm test -- --reporter=dot`: 5213 passed, 15 skipped; one unrelated flaky occupancy failure in `tests/clients/lsp/edits.test.ts` (max sync block ~478-505ms vs 300ms threshold)

Closes #1073


apmantza:
Addressed the pre-merge audit findings from #1089 and #1088 with three commits (4bd90e43, 99617358, 9b865288):

**#1089 — call-graph subsystem**
- De-vacuated the two persisted-cache validator "rejects..." tests (`tests/clients/call-graph.test.ts`): they hand-wrote `version: 4`, identity-less fixtures that `loadCallGraph` rejected at the version/identity gate before `validatePersistedCallGraph` ever ran. Rewrote both as `version: 5` fixtures with a valid canonical identity, each now isolating the one validator check it claims to cover (inDegree/adjacency cross-check; `complete && unsupportedEvidence > 0`). Proved non-vacuous by temporarily disabling each check and confirming the corresponding test failed, then restoring.
- Added the first tests anywhere that pass `expectedIdentity` to `loadCallGraph`, covering the canonical-freshness invariant (mismatched `reviewGraphVersion`/`reviewGraphSignature` rejects the cache — the #210/#1020 stale-replay class; matching identity loads normally). Same fail-then-pass proof.
- Covered `module_report`'s `callGraph: true` surface for the `file-cap` and `identity-missing` unavailability reasons (previously untested) and the `provenance.callGraph` field across available/unrequested/cold states. `identity-missing` isn't reachable synchronously in a single `moduleReport` call (the divergence is exactly the #1088 race below), so it's covered in a dedicated file (`tests/clients/module-report-call-graph-identity-missing.test.ts`) that mocks `getReviewGraphCacheIdentity` to force the gap, kept isolated from the main suite's many real-identity assertions.
- Not addressed (out of scope for this pass, filing as noted below): the Java try-with-resources `\b` regex dead-code item and the 19-grammar co-load test-file split — both are real but separate from the cache/freshness/module_report gaps this pass targeted.

**#1088 — review-graph builder**
- One-line fix: `getReviewGraphCacheIdentity` (`clients/review-graph/builder.ts`) compared `cached.graph.version`, the constant schema tag (`"v8"`, identical for every live graph) — a vacuous guard that could never detect a stale graph instance. Now compares `buildGeneration`, the per-content generation stamp (#459) that actually travels with each graph instance. Added a regression test in `tests/clients/graph-cache.test.ts` that builds two distinct graph instances for the same cwd and confirms the stale one's identity lookup is rejected once the workspace cache moves on — proved non-vacuous by reverting to the `version` compare and confirming the test failed.
- `parseSymbolKey`'s LSP-fallback-kind mis-parse (#1088's other P2) is not addressed here — separate fix, not touched by this pass.

**Verification**
- `npm run build` then targeted files (`call-graph.test.ts`, `graph-cache.test.ts`, `module-report.test.ts`, the new identity-missing file) plus every other test file importing `call-graph.js`/`review-graph/builder.js` — all green.
- Full suite (`npx vitest run`): 5238 passed, 1 failed (the documented `edits.test.ts` "keeps large text/resource planning occupancy bounded" environment flake — unrelated, pre-existing), 12 skipped.

apmantza:
**Adversarial review round 2 applied** (head `14ff8037`).

The review of `9b865288` found the buildGeneration guard compared the STORED graph object's stamp — but `cloneGraph` drops it and the drift-reuse/disk-hit paths store an unstamped clone, so identity lookups failed for exactly the graphs reuse builds return (call graph silently off after warm restarts), while two unstamped instances matched vacuously (`undefined === undefined`). Both directions were reproduced against built output.

Fixes in this round:
- `be480c7e` — guard now compares the workspace-cache **entry's** `buildGeneration`; both-undefined matches only the tier-3 hydration same-instance case. New regression test drives the real pure-drift path (mtime-only touch, real project dir) — **proven to fail on `9b865288`** (`expected undefined to be defined`) and pass now.
- `14ff8037` — `tree-sitter-call-graph.test.ts` and `module-report-call-graph.test.ts` moved into the grammar-heavy vitest project (#255/#902 co-load class); CHANGELOG `closes` → `refs` for #1088/#1089 (both keep open items: parseSymbolKey LSP-fallback kinds; Java try-with-resources regex).

Incidental find while building the repro: the source walk's generated-artifact name heuristic silently drops files named `gen.ts` — the pre-existing `graph-cache.test.ts` fixtures all use `gen.ts`, so those tests exercise empty-walk builds without knowing it (their assertions still hold; noted in the new test's comment).

Full suite: 5,239 passed / 1 failed — the documented `edits.test.ts` occupancy flake only (tracked separately, pre-existing).

Issue #1093, opened 2026-08-05
widget-state staleness class: cross-file invalidation gap, unwired cascade reconcile, touchedAt re-arming from cached views (root cause of #1092)


apmantza:
Assessment of issue #1092 (stale TUI footer diagnostics surviving across windows while `tsc` is clean) traced the full widget-state lifecycle and identified a staleness class not covered by #1084-#1090. Line numbers as of current master.

## Root cause of #1092 (code-traced)
Widget diagnostics for file A caused by file B (missing declaration/changed export) are fixed by editing B — but every invalidation gate compares A's OWN mtime to a record timestamp:
- `reconcileStaleWidgetFiles` (clients/widget-state.ts:400-424): drop iff `mtimeMs > touchedAt + 1`
- `dropStaleFiles` (clients/session-state-store.ts:114-135): resume keeps entries with `mtimeMs <= savedAt + 1`
- `reconcileProjectDiagnosticsSnapshot` (clients/project-diagnostics/cache.ts:84-111)

A's mtime never advances, so the resolved finding renders forever — footer, `lens_diagnostics mode=all/delta` replay, resume, and fork adoption (index.ts:1446-1459 — every subagent fork inherits and re-persists the parent's entries).

## Findings (new; ranked)
- **P1-1 Cross-file staleness invisible to every gate** (above). Defect class: per-file freshness stamps guarding diagnostics that are a function of whole-project state.
- **P1-2 The cascade computes the correcting truth and throws it away**: `clients/dispatch/integration.ts` re-checks dependents of every edited file (`neighbors.push` ~:1194/:1427/:1453) but results are documented "ephemeral display-only, never reconciled into widget/dedup state" (:1170-1175). The one writer that runs automatically at the right moment is unwired.
- **P2-3 "Non-empty = confirmed" re-cements stale state**: `reconcileWidgetFromLspResult` (tools/lsp-diagnostics.ts:881-907) writes any non-empty result with `touchedAt = Date.now()`; the workspace-cache-hit branch (:939-963) re-reconciles CACHED diagnostics with a fresh stamp — a "fresh check" against a stale warm-server view permanently disarms the mtime gate. Matches #1092's window-1 transcript (17 diagnostics incl. ts:2339 while tsc exits 0).
- **P2-4 Cross-session/fork amplification**: persistence (#190) + fork adoption propagate one stale record to every subagent and resume.
[...]

Issue #1088, opened 2026-08-05
review-graph: vacuous cache-identity guard enables stale call-graph projection under fresh signature; parseSymbolKey breaks on LSP-fallback kinds


apmantza:
Adversarial review of the review-graph subsystem changes since v3.8.74 (#1048 thread / #1070 integration). Line numbers as of 293e7b54.

## P2 — `getReviewGraphCacheIdentity`'s graph-consistency guard is vacuous; concurrent build can stamp a stale call-graph projection with the newest signature
`clients/review-graph/builder.ts:339-357`: `if (graph && cached.graph.version !== graph.version) return undefined;` — but `version` is the constant schema tag ("v8"), identical for every live graph, so the guard can never detect that `graph` is not the graph the workspace-cache entry describes.
Concrete race (`runtime-session.ts:1093-1119`): session call-graph task builds, then a per-edit cascade build (different `_buildCache` key, same cwd) replaces `_workspaceGraphCache` in the microtask window → identity returned is the NEWER build's signature while `extractSymbolsAndRefsFromGraph(graph)` projects the OLDER graph → `saveCallGraph` persists a stale projection under a fresh signature, which `loadCallGraph`/`module_report` then validate as current. Silently defeats "freshness is owned exclusively by the review-graph snapshot."
One-line fix: compare `buildGeneration`, not `version`.

## P2 — `parseSymbolKey` mis-parses canonical ids with LSP-fallback kinds
`clients/review-graph/symbol-id.ts:64-66` whitelists 7 kinds, but `addLspFallbackSymbols` mints ids with `lspSymbolKindName` — 19 other kinds (`enum`, `constant`, `struct`, `namespace`, `field`, `constructor`, `module`, `lsp-symbol-<n>`, ...). `c:\p\a.kt:Color:enum:42` fails the kind check, falls to the legacy last-colon split → `filePath: "c:\p\a.kt:Color:enum"`, `symbolName: "42"`. Reaches `codebase-model.ts` (inferKind/entry names/allFiles) and `formatImpact` whenever the LSP fallback fires (the documented degraded-extractor case, e.g. kotlin defs query failing). Fix: accept the LSP kind vocabulary or match the trailing `:<kind>:<digits>` shape structurally.

## P3
[...]

Issue #1087, opened 2026-08-05
rules/policy: sg scan exit-1 drops real matches (P1 regression); silencing-is-not-fixing sweep incomplete; rule-id normalization asymmetries


apmantza:
Adversarial review of the rules/ast-grep/diagnostics-policy changes since v3.8.74 (#444 thread). Line numbers as of 293e7b54.

## P1 — `sg scan` exit-1-with-matches treated as CLI failure; real matches dropped (regression vs v3.8.74)
`clients/sg-runner.ts` `interpretScanResult()` (~:527) / `exec()` (~:400). **Empirically verified with the bundled CLI**: a rule with `severity: error` that matches exits 1 with valid JSON matches on stdout and stderr "Scan succeeded and found error level diagnostics". Old code parsed stdout regardless; new code's only exemption is `status===1 && !stdout && !stderr` → any other nonzero status returns `failure: "cli-failure"` with `matches: []`.
Reachable scenarios (raw-rule validation only checks NUL/length, `tools/ast-grep-search.ts:117`):
1. `ast_grep_search`/`pilens_ast_grep_search` with a raw rule containing `severity: error` (agents copy this shape from ast-grep docs and this repo's own rules) → error text instead of matches.
2. `pilens_ast_grep_replace`: preview fails; `apply: true` refused ("ast-grep preview failed") even though `--update-all` itself exits 0 (verified).
3. `validateRule` declares a valid error-severity rule invalid when it matches the validation snippet.
Fix: `status===1` + stdout parses as JSON → success (ast-grep linter-style exit); keep stderr-with-no-stdout as failure. Test gap: `tests/clients/sg-runner.test.ts` covers only status-1-empty-output.

## P2 — "silencing is not fixing" fixed only for the policy member of its class
`clients/dispatch/dispatcher.ts:813-836`: baseline is `dedupedDiagnostics`, captured BEFORE `suppressLintOverlapsWithLsp`, `applyInlineSuppressions`, `applyDispositions`; delta `after` is fully filtered. #444's fix excludes policy-dropped ids from `filtered.fixed` — but the identical failure remains for the three sibling filters: a finding persistently dropped by an inline `pi-lens-ignore`, a disposition, or LSP-overlap suppression is counted in `resolvedCount` on EVERY dispatch → `pipeline.ts:1307` `trackAgentFixed` inflates forever. Per the AGENTS.md class-sweep rule: filter the baseline through the same pipeline before `filterDelta`.

## P3
[...]

Issue #1086, opened 2026-08-05
Consistency sweep (v3.8.74..HEAD): raw path-keyed maps, raw-URI edit-merge buckets, hand-copied language lists, unversioned codebase-model cache


apmantza:
Cross-cutting consistency/reuse sweep of all source changed since v3.8.74 (+ working tree), checked against AGENTS.md conventions (PathKeyedMap for path-keyed maps, single source of truth, shared helpers, OS-agnosticism, resource bounding). Line numbers as of 293e7b54 + working tree.

## P1 — new raw path-keyed maps in `clients/runtime-tool-result.ts` (the #210/#1020/#1025 class)
`inFlightPipelines` (:132, keyed `${filePath}:${stateHash}` at :539, ADDED this window in 2cac7fb8/#1062) plus siblings `lastAnalyzedStateByFile` (:133) and `debouncedPipelines` (:162) key on `filePath` that is only `path.resolve`d from raw agent-supplied tool input (:369) — no case fold, no realpath. AGENTS.md: "reach for PathKeyedMap … whenever you add or touch a path-keyed in-memory map."
- Risk: divergent Windows spellings (`c:\p\Foo.ts` vs `C:\p\foo.ts`, realpath-divergent forms on resumed sessions — #1020's exact failure) bypass concurrent-state dedupe and debounce coalescing → duplicate pipeline runs, doubled LSP dispatch, split telemetry participant counts; `flushDebouncedToolResults(filePath)` silently misses pending entries.
- Fix: `PathKeyedMap` with `normalizeEphemeralMapKey` (hot path; turn-scoped contents).

## P1/P2 — cross-server workspace-edit merge buckets by raw URI string (`clients/lsp/edits.ts`)
`mergeWorkspaceTextEditsByPriority` (:398) and `flattenWorkspaceTextEdits` (:370), added in 386f3a92 (#1053), bucket by each server's verbatim URI. Servers spell the same file differently on Windows (`file:///c%3A/...` vs `file:///C:/...`, drive-letter case) → divergent spellings land in different buckets, the `rangesOverlap` conflict check (:412) never runs across them → conflicting edits from two servers are BOTH applied, and `droppedConflicts` under-reports. The right helper exists in the same file: `pathIndexKey(uri)` (:443), used by the planner but not the merger. Bucket by `pathIndexKey`, preserve one display URI per bucket.

[...]

PR #1094, opened 2026-08-05
fix(diagnostics): clear stale cross-file footer diagnostics (refs #1093, refs #1092)

### LINKED ISSUE #1092: Quite often the status bar grows stale and displays incorrect errors
### What happened?

For instance in one window the status bar shows - 
```
pi-lens  ●2E !8W
   ● spindle.ts 2E8W  · session-factory.test.ts  · spindle.test.ts  · types.ts
   ● L263 ts:2339  Property 'extensions' does not exist on type 'SpinUpConfig'.
   ● L264 ts:2339  Property 'toolsMode' does not exist on type 'SpinUpConfig'.
```
This hint seems to have been given to the model as well as it tried to figure out what was wrong and couldn't make sense of it  -

```
 The two blocking LSP errors look stale (tsc passes). Let me confirm with a fresh LSP check, and also remove the now-unused LoadExtensionsResult import:                                      

                                                                                                                                                                                              
 lsp_diagnostics                                                                                                                                                                              
 lsp_diagnostics spindle.ts — 17 diagnostics                                                                                                                                                  
                                                                                                                                                                                              

 The LSP's "Primary LSP (typescript): confirmed clean" but the auxiliary ast-grep finds "error[ts]" 2339 at line 263-264. This is the LSP's TypeScript diagnostics reporting that             
 extensions/toolsMode don't exist on SpinUpConfig. This contradicts tsc exit 0.                                                                                                               
```

Following this I opened a second terminal and had pi check it -

```
 in another session pi-lens reports as follows in the live status bar underneath the prompt -                                                                                                 
                                                                                                                                                                                              
  pi-lens  ●2E !8W                                                                                                                                                                            
    ● spindle.ts 2E8W  · session-factory.test.ts  · spindle.test.ts  · types.ts                                                                                                               
    ● L263 ts:2339  Property 'extensions' does not exist on type 'SpinUpConfig'.                                                                                                              
    ● L264 ts:2339  Property 'toolsMode' does not exist on type 'SpinUpConfig'.                                                                                                               
                                                                                                                                                                                              
 I am suspecting that it has fallen out of sync with the live files somehow. Can you check?                                                                                                   
 ```
Pi confirms there are no errors -

```
 Confirmed — your suspicion is right. The status-bar report is stale, and the live files are fine:                                                                                            
                                                                                                                                                                                              
 - Live LSP on spindle.ts and types.ts: both clean (no ts:2339 anywhere).                                                                                                                     
 - Project-wide tsc --noEmit: exits 0 — clean.                                                                                                                                                
 - lens_diagnostics in this session: no issues.                                                                                                                                               
                                                                                                                                                                                              
 The current types.ts really does define both properties on SpinUpConfig (lines 50 and 59): extensions?: string[] and toolsMode?: "extend" | "replace", and spindle.ts uses them at L263/L264 
 exactly where the stale errors point.                                                                                                                                                        
                                                                                                                                                                                              
 Why it went stale: those two errors come from an intermediate state of the last commit (87417f6 "fix(engine): make spin_up capability grants work end-to-end"). During that work, spindle.ts 
  was edited to reference config.extensions / config.toolsMode at those lines before types.ts had been updated to declare them. pi-lens in the other session captured the ts:2339 diagnostics 
 at that moment and has been showing the cached snapshot ever since — the LSP never re-ran on those files after the fix landed, even though git working tree is clean and both files are now  
 consistent.                                                                                                                                                                                  
                                                                                                                                                                                              
 To clear the stale status bar in that other session: force pi-lens to re-dispatch onto the files so the LSP recomputes — e.g. re-save (touch) or do a trivial edit on spindle.ts and         
 types.ts in that session, or reload/reopen the project so the language server restarts. The moment it re-runs, those two errors vanish because the source is clean.                          
                                                                                                                                                                                              
 No code change is needed — the extensions / toolsMode implementation (ADR-0003, issue 32) is correctly committed.                                                                            
```

### Steps to reproduce

Happened during a coding session, using subagents to review and fix code.

### Expected behavior

_No response_

### pi-lens version

_No response_

### Relevant logs

_No response_

### Area

None

apmantza:
## Defect class

**Per-file freshness stamps guarding diagnostics that are a function of whole-project state.** A widget finding in file A caused by file B (a missing declaration, a changed export) is fixed by editing B, but every widget staleness gate compares A's OWN mtime to the record timestamp — and A's mtime never advances, so the resolved finding renders forever (footer, `lens_diagnostics` replay, resume, fork adoption). This is the reported shape of #1092, root-caused in #1093.

Two minimal fixes; the structural seq-stamping redesign stays a follow-up under #1093 (see non-goals).

### Fix 1 — wire cascade neighbor reconciliation into widget state
The per-edit cascade (`clients/dispatch/integration.ts`) already re-checks every edited file's dependents but documented its results as "ephemeral display-only, never reconciled." Each **confirmed** neighbor result now reconciles into widget state — **including the confirmed-clean `[]` case** that clears A's now-stale footer entry.

- **Confirmed = written**: a valid passive snapshot (within `CASCADE_TTL_MS`) or a completed, **non-inconclusive** active touch (#571 semantics).
- **Inconclusive is NOT confirmed** (review round 1, Blocking 1): a touch can resolve `[]` while carrying a non-enumerable `inconclusive: true` (its notify/diagnostics wait lapsed — e.g. the 1000ms cold-snapshot budget on a slow server). That is never treated as confirmed-clean: it neither reconciles nor seeds `recentlyCleanNeighborCache` (so a later short-circuit can't make a wipe self-sustain). Mirrors `tools/lsp-diagnostics.ts`'s gate on the same flag.
- **Merge, not replace** (review round 1, Blocking 2): the cascade re-checks only the language server, only for errors (`severity === 1`), so a full replace would erase the neighbor's live findings from sources the cascade never examined (biome/ruff/ast-grep runners, and LSP *warnings*). `reconcileCascadeNeighborLspErrors` replaces only the existing **LSP-error** entries (`tool === "lsp"` && error/blocking) and preserves everything else. A stale LSP warning or biome finding survives an errors-only re-check and self-corrects on the next per-edit dispatch / `lens_diagnostics` scan (which DO re-examine every source).
- **Auxiliary findings excluded** (review round 2, P2): the cascade payload (`clientScope:"all"` / all clients) includes auxiliary-server findings (opengrep/ast-grep/zizmor/typos), and — unlike the other writers — the cascade never runs `retagAuxiliaryDiagnostics`. `cascadeReconcilableLspErrors` drops any diagnostic whose `source` matches an auxiliary profile (`findAuxiliaryProfileForSource`) before converting for the reconcile, so an aux finding is never double-counted or escalated into a blocking `tool:"lsp"` error (aux policy — semantic downgrade, native suppression, `skipTestFiles` — is not bypassed). Aux findings are per-file signals the neighbor's own per-edit runners own; the cascade's DISPLAY list is left unchanged (pre-existing behavior, out of scope).
- **Not confirmed = never written**: tier-3-silent skip, recently-clean short-circuit, within-turn neighbor-touch cache, and the rejected-touch passive fallback are all left untouched.
- **Write ordering**: keyed by the primary edit's `writeSeq` (the same monotonic `nextWriteIndex()` token the per-edit path draws), so a genuinely newer per-edit write for the neighbor still wins the `WriteOrderingGuard`.

### Fix 2 — stop `touchedAt` re-arming from non-fresh (aging) sources
`reconcileScanDiagnostics` / `recordDiagnostics` accept an **observation timestamp**. Cache/snapshot replays now stamp `touchedAt` with when the diagnostics were **observed**, not `Date.now()`:
- `lsp_diagnostics` workspace-cache-hit branch and `lens_diagnostics mode=full`'s cache-served results → the cache entry's `scannedAt`.
- The cascade's passive-snapshot reconcile → the snapshot's own publish time `entry.ts` (up to `CASCADE_TTL_MS` old) — review round 1, Blocking 3. The active-touch reconcile stays `now` (a genuinely fresh observation).

## Invariants preserved
- **#571 confirmed-only contract** — only snapshot-valid / touch-completed-and-**non-inconclusive** results reach the footer; inconclusive/timed-out/silent-tier never present as confirmed-clean.
- **#533 honesty** — an errors-only, language-server-only cascade never asserts clean over sources or severities it did not examine (merge, not replace).
- **#555 / WriteOrderingGuard** — cascade writes carry the primary edit's `writeSeq`; a higher per-edit token still supersedes.
- **#1020 path-key normalization** — reconcile keys through the same `fileMapKey` fold; no new raw-key map.
- **Footer reconcile never fails the tool** — all reconcile call sites stay inside their existing try/catch.
- Backward compatible: every new parameter/field is optional/additive (`observedAt` on the writers, `scannedAt` on `WorkspaceDiagnosticsCacheLookup`, `observedAt` on `LSPWorkspaceDiagnosticResult` + mode=full's local mirror). `tsc` confirms no caller breaks.

## Matrix cells tested (each proven to FAIL pre-fix, then pass)
`tests/clients/cascade-compute.test.ts` — `neighbor reconciliation into widget state (#1093/#1092)`:
- Confirmed-clean passive snapshot **clears** a neighbor's stale LSP footer entry.
- Confirmed active touch **writes** the neighbor's diagnostics.
- Rejected touch → passive fallback does **not** overwrite an existing record.
- Write-ordering: **older** `writeSeq` doesn't clobber a newer per-edit record; **newer** `writeSeq` wins (proves the reconcile fires and the token is `writeSeq`).
- **Inconclusive active touch** (`[]` + `inconclusive` flag) does **not** wipe a live footer finding. *(Blocking 1; pre-fix: wiped)*
- Confirmed LSP-clean cascade **clears the LSP error but preserves a live biome finding** — merge, not replace. *(Blocking 2; pre-fix: biome wiped)*
- Passive-snapshot reconcile **stamps `touchedAt` at the snapshot's publish time** (`entry.ts`), so an aging snapshot drops once the file's mtime passes it. *(Blocking 3; pre-fix: touchedAt=now → survives)*
- Confirmed cascade **does not double-count or escalate** a neighbor's auxiliary (opengrep) finding present in the payload — it stays a single `tool:"opengrep"` advisory entry. *(round 2 P2; pre-fix: 2 entries, one escalated to `tool:"lsp"`/blocking)*

`tests/clients/widget-state.test.ts` — observation-timestamp units: a cache-hit reconcile stamps `touchedAt` at observation time so the entry drops once the file mtime passes it *(pre-fix: survives)*; a fresh reconcile (no stamp) is observed now and survives an older mtime (control).

`tests/tools/lsp-diagnostics-cache.test.ts` — end-to-end: a cache-hit reconcile stamps the footer with the cache's original `scannedAt`, not now *(pre-fix: `observedAt` undefined)*.

## Blast radius
Verified by grep of actual call expressions + `tsc` (repo `module_report` MCP not wired in this checkout). The diagnostic **writer surfaces** are exactly five:
- `recordDiagnostics(...)` — `clients/pipeline.ts` (per-edit), `clients/mcp/analyze.ts` (on-demand). Added trailing optional `observedAt`; existing calls unchanged.
- `reconcileScanDiagnostics(...)` — `tools/lens-diagnostics.ts` (mode=full), `tools/lsp-diagnostics.ts` (on-demand/batch). Added trailing optional `observedAt`.
- `reconcileCascadeNeighborLspErrors(...)` — `clients/dispatch/integration.ts` (cascade), new merge-mode writer.
- `WorkspaceDiagnosticsCacheLookup.scannedAt` — additive; both `lookup()` consumers (`clients/lsp/index.ts` `runWorkspaceDiagnostics`, `tools/lsp-diagnostics.ts` batch sweep) updated to thread it.
- `LSPWorkspaceDiagnosticResult.observedAt` (+ mode=full's local `WorkspaceLspDiagnosticResult` mirror) — additive optional; set only on cache-hit results, consumed only by mode=full's reconcile.

(Correction from the initial description: `clients/code-quality-warnings.ts`, `clients/diagnostics-publish.ts`, and `index.ts` are NOT writer call sites — the earlier grep matched imports/re-exports/comments, not `recordDiagnostics(...)` calls.)

## Non-goals (explicit)
- **Structural seq-stamping (kills the class)** — adding `seq`/`staleSuspect` to `FileRecord`, deriving a changed file's reverse-dep set on dispatch completion and marking those records unverified up front (mark, don't drop — #533), a third `markDependentsUnverified` verb, and dimmed/pending render + mode=all "re-verify with mode=full" annotation. **Stays a follow-up under #1093.**
- The pre-existing `lens_diagnostics` / `mode=full` scan reconciles still whole-replace (they capture all severities + retag aux, so they are more complete than the cascade); the merge policy here is scoped to the cascade's new automatic every-edit channel. Applying merge uniformly to the scan paths is a separate, larger change (would touch #571 behavior) — noted, not done here.
- Overlaps deliberately NOT touched: widget-state hand-`fileMapKey` → `PathKeyedMap` (#1086), `rule = d.rule ?? d.id` policy (#1087 P3-5), review-graph identity guard (#1088). Cross-reference only.

## Test result
Full suite: **5216 passed**, 11 skipped, 1 failed — the failure is the known environment flake `edits.test.ts > keeps large text/resource planning occupancy bounded` (passes in isolation; timing bound under parallel-CPU contention), unrelated to this change.

Do NOT write "closes #1092" — the structural half stays open under #1093.


apmantza:
## Review round 1 — addressed (head `fd75ce03`)

Adversarial review confirmed Fix 2 (`observedAt`) and the write-ordering/`writeSeq` design as sound, and flagged three ways Fix 1's "confirmed" predicate was not #571-faithful. All three are fixed in `fd75ce03`; each has a new regression test proven to fail on the pre-fix code and pass after.

- **Blocking 1 (P1) — inconclusive touch treated as confirmed-clean.** The active-touch site only checked `if (!rawDiags)`; a resolved `[]` carrying the non-enumerable `inconclusive: true` (wait lapsed, e.g. the 1000ms cold-snapshot budget) was reconciled as clean and could **wipe a live footer finding** (#533), then self-sustain via `recentlyCleanNeighborCache`. Now both the reconcile and the recently-clean seed are gated on `!inconclusive`, mirroring `tools/lsp-diagnostics.ts`. Test: `an INCONCLUSIVE active touch ... does NOT wipe a live footer finding`.

- **Blocking 2 (P1) — whole-replace erased non-LSP findings.** The cascade re-checks only the language server, only for `severity === 1`, but `recordDiagnostics` is a full replace — so a confirmed LSP-clean cascade silently erased a neighbor's live biome/ruff/ast-grep finding or LSP warning (a new automatic every-edit false-clean channel). **Policy chosen: MERGE.** New `reconcileCascadeNeighborLspErrors` replaces only existing **LSP-error** entries (`tool === "lsp"` && error/blocking — `convertLspDiagnostics` tags every language-server diagnostic `tool:"lsp"`, and `retagAuxiliaryDiagnostics` re-tags aux LSPs to their real tool, so this uniquely identifies genuine language-server entries) and preserves everything else. Errors-only means an LSP **warning** is preserved rather than dropped (chose "preserve existing severity-2" over "widen the payload to severity 2", to keep cascade display output errors-only and avoid changing existing cascade behavior/tests). Documented in code + above. Test: `... clears the neighbor's LSP error but PRESERVES a live biome finding (merge, not replace)`.

- **Blocking 3 (P2) — passive-snapshot reconcile re-armed the mtime gate.** Snapshots up to `CASCADE_TTL_MS` (4 min) old were reconciled with `observedAt` omitted → `touchedAt = now()`, the exact defect Fix 2 corrects for cache hits. The snapshot site now passes the snapshot's publish time `entry.ts` as `observedAt`; the active-touch site stays `now` (a fresh observation). Test: `a passive-snapshot reconcile stamps touchedAt at the snapshot's publish time (entry.ts), not now`.

Also corrected the PR body's blast-radius list: `code-quality-warnings.ts` / `diagnostics-publish.ts` / `index.ts` are NOT writer call sites (the earlier grep matched imports/re-exports/comments). The five real writer surfaces are pipeline, mcp/analyze, lens-diagnostics, lsp-diagnostics, integration.

Full suite after the changes: **5215 passed**, 11 skipped, 1 failed (the known `edits.test.ts` occupancy flake — passes in isolation).

apmantza:
## Review round 2 — addressed (head `35c04b85`)

Round-2 verification confirmed the three round-1 blockers genuinely fixed and flagged one new P2 (introduced by the merge design) plus two P3s. All addressed.

- **Blocking (P2) — auxiliary findings written as `tool:"lsp"`/blocking.** Both cascade convert sites (`integration.ts` passive + active) fed the reconcile the full `getAllDiagnostics` / `touchFile({clientScope:"all"})` payload, which includes auxiliary-server findings (opengrep/ast-grep/zizmor/typos). Unlike the other three widget writers, the cascade never runs `retagAuxiliaryDiagnostics`, so a neighbor's live opengrep finding got written **again** as `tool:"lsp"`/blocking by the merge — footer **double-count** + **escalation** of an advisory/suppressed aux finding, bypassing aux policy (semantic downgrade, native `# nosemgrep`/`zizmor:ignore`, `skipTestFiles`). **Fix:** new `cascadeReconcilableLspErrors()` drops any diagnostic whose `source` matches an auxiliary profile (`findAuxiliaryProfileForSource`) before converting for the reconcile, at both sites. Aux findings are per-file signals the neighbor's own per-edit runners own (not cross-file impact), so the cascade excludes them and the merge preserves the neighbor's existing aux entries untouched. The cascade **display** list is intentionally left as-is (pre-existing behavior, out of scope). New test `does NOT double-count or escalate a neighbor's auxiliary (opengrep) finding present in the cascade payload` — pre-fix: **2 entries, one escalated to `tool:"lsp"`/blocking**; post-fix: **1 entry, `tool:"opengrep"`, non-blocking**.

- **P3(a) — passive-snapshot merge stamps the whole record's `touchedAt` with `entry.ts`.** A merged record's single `touchedAt` uses the snapshot's publish time even for preserved entries observed more recently, so if the neighbor's mtime falls between `entry.ts` and a preserved entry's real observation, `reconcileStaleWidgetFiles` can drop the whole record including newer findings. Per-entry timestamps are the real fix — out of scope. Named the edge in a code comment at the site and folded it into the structural follow-up under #1093.

- **P3(b) — inconclusive-flag test mock was enumerable.** Switched from `Object.assign([], {inconclusive:true})` to a non-enumerable `Object.defineProperty(..., {enumerable:false})`, matching the real `touchFile` return, so the test now also catches a future regression where the code copies/spreads the array (dropping a non-enumerable flag) before reading it. The gate still passes with the stricter mock (the code reads the flag off the original array before any filter).

Full suite after the changes: **5216 passed**, 11 skipped, 1 failed (the known `edits.test.ts` occupancy flake — passes in isolation). CHANGELOG updated to note the aux exclusion.

Issue #1095, opened 2026-08-05
LSP diagnostics lack content binding: staleness is inferred from mtime/TTL proxies instead of the document version diagnostics were computed against


apmantza:
Surfaced by the #1092/#1093 staleness investigation and the PR #1094 review rounds: pi-lens records LSP diagnostics as point-in-time observations with NO binding to the document content they were computed against. Every downstream staleness decision is a proxy reconstruction — and each proxy has a demonstrated failure mode.

## The gap
- **`touchFile` "completed" is not "analysis completed"**: the push model gives no completion signal, so cleanliness is inferred from silence within a 1-2s wait deadline; the `inconclusive` flag (clients/lsp/index.ts:1536-1554) is a non-enumerable, per-caller honesty tax that PR #1094 round 1 proved gets missed (a timed-out `[]` wiped a live finding as confirmed-clean).
- **Warm-server view can diverge from disk undetected**: open-document buffers (tsserver) and the last-publish fallback (tools/lsp-diagnostics.ts:748-754) can lag disk when a fix lands outside this process (another window, subagent, git op). The "non-empty result = definitionally confirmed" doctrine (:866/:889) has no check that the server's view corresponds to current content — a fresh check can faithfully report a stale world and stamp it with a fresh timestamp (#1092's window-1 transcript: 17 diagnostics incl. ts:2339 while tsc exits 0).
- **Passive snapshots have age-only validity**: cascade snapshots trust `entry.ts` for up to CASCADE_TTL_MS (240s) with no invalidation on file change, server restart, or the primary edit; right after a fix-edit the first cascade often replays the PRE-fix snapshot because the server hasn't republished (found in #1094 review round 1).
- **All consumer gates are proxies**: mtime-vs-touchedAt (widget-state), savedAt (resume), observedAt (#1094 Fix 2) — necessary patches, but none can answer "were these diagnostics computed against what's on disk now?"

## What the protocol already offers (unused)
- `publishDiagnostics` carries an optional `version` (the document version the diagnostics apply to) — most mainstream servers (tsserver wrapper, pyright, gopls, rust-analyzer) populate it.
- pi-lens already owns the didOpen/didChange version counter per document (clients/lsp/client.ts) and knows the content it sent at each version.
- Pull diagnostics (where supported) have resultIds.

## Proposed design (bind at the client boundary, decide at the consumers)
[...]

Issue #1025, opened 2026-08-01
path-key divergence class: sweep results + typed PathKeyedMap wrapper (follow-ups to #1020)


apmantza:
Results of the repo-wide sweep for the #1020 path-key divergence class (a map/cache/guard keyed on a raw path where write and read forms can diverge — `\`vs`/`, drive-case, resolved-vs-raw — causing stale replay, silent-miss, or a guard no-op). Recorded here so the sweep's coverage is auditable (per AGENTS.md "a newly root-caused bug is a class trigger").

## Confirmed
1. **dispositions** — filed separately (agent's own mark defeated; high severity, #533-adjacent). 
2. **word-index** (`clients/word-index.ts:161-198`) — build uses raw on-disk casing (`readdirSync`), per-edit update uses `path.resolve()` on raw tool-input path (`integration.ts:736-742`, `runtime-tool-result.ts:320-326`); mismatched casing → duplicate doc entry, `forward.has()` misses so the stale entry is never removed → stale postings until full rebuild. Degrades `symbol_search` relevance silently; not agent-blocking. Fix: fold both sides through `normalizeMapKey`.
3. **`AgentBehaviorTracker.getEditCount`** (`clients/agent-behavior-client.ts:150-151`) — reads raw while `:121-125` writes normalized. **Dead code** (zero callers) — guard before anyone wires a caller.

## Suspect (asymmetric/unprotected, no proven trigger)
4. `RuntimeCoordinator._pendingInlineBlockers` / `_pendingDeferredFormatFiles` / `_lspReadWarmState` (`runtime-coordinator.ts:107-115`) — raw `path.resolve` both sides, no `normalizeMapKey`; `_pendingInlineBlockers` could replay a resolved blocker (same shape as #1020). Inconsistent with the file's own documented `_fileSeq` convention.
5. `cachedExports` value-equality (`runtime-tool-call.ts:1089`, `runtime-tool-result.ts:431`) — if `ast-grep`-emitted vs raw tool-input path forms diverge, the duplicate-export STOP check could **falsely block** a legitimate write (higher-severity direction, unproven — couldn't confirm ast-grep's path form).
6. `astGrepWarnDebounceTimers` (`integration.ts:198-201`) — duplicate background scans; low.
7. `diagnostic-tracker` `shown`/`occurrenceCounts` (`:63-64`) — double-counted stats; telemetry only.
[...]

PR #1096, opened 2026-08-05
feat(project-intelligence): bind LSP diagnostics to document content (refs #1095)

apmantza:
refs #1095 (first PR — second PR remains for cascade adoption; does not close).

## Defect / capability class
LSP diagnostics were stored as point-in-time observations with **no binding to the document content they were computed against**. Every downstream staleness decision was a proxy reconstruction (mtime-vs-touchedAt, savedAt, observedAt, cascade TTL) — none could answer *"were these diagnostics computed against what's on disk NOW?"* So a fresh check could faithfully report a **stale world** and stamp it confirmed (the #1092 window-1 transcript: 17 live-looking diagnostics incl. `ts:2339` while `tsc` exits 0). This PR binds diagnostics to a content fingerprint at the client boundary and decides at the `lsp_diagnostics` consumer.

## What the protocol already offered (now used)
`publishDiagnostics` carries an optional `version`; pi-lens already owns the didOpen/didChange version counter and knows the exact text it sent at each version. This PR fingerprints that sent text and, when the server echoes the version, binds the stored diagnostics to it.

## In scope (this PR)
1. **Capture** (`clients/lsp/client.ts`): fingerprint the exact didOpen/didChange payload at **send time** (`recordSentContent`, `documentContentHashes`), and on an accepted publish record `diagnosticBindings[path] = {version, contentHash?}` — contentHash only when the echoed version matches the last-sent version. Never a disk read on the notification path.
2. **Expose** (`clients/lsp/index.ts`, `clients/lsp/workspace-diagnostics-cache.ts`): `getAllDiagnostics`/`touchFile` results and the workspace-cache lookup surface a `binding` `{version?, contentHash?, boundToCurrentDisk: boolean | "unknown"}`. `boundToCurrentDisk` is verified **lazily** at read time, **memoized per (file, mtime)** (`clients/lsp/diagnostic-binding.ts` → `createDiskBindingCache`).
3. **Adopt** (`tools/lsp-diagnostics.ts`): `confirmed` now requires `boundToCurrentDisk !== false`. A mismatched binding demotes even a **non-empty** result to inconclusive — it no longer re-cements the footer (`reconcileWidgetFromLspResult` gated on `boundMismatch`; cache-mismatch entries are not served). `"unknown"`/`true` leave the pre-#1095 doctrine untouched.
4. **Observability**: `bindingState` (`bound`/`mismatch`/`unknown`) in the `lsp_touch_file` latency metadata.

## Invariants (written before coding)
- **I1 fingerprint source**: sha256 of the exact didOpen/didChange payload, computed on in-memory content at send time — never a disk read on the notification path.
- **I2 read-transform symmetry**: disk compare reads with the identical `fs.readFile(path,"utf-8")` transform the payload was built with → CRLF/BOM files round-trip to `true`.
- **I3 version gate**: contentHash stored only when the accepted publish's version equals the last-sent version. Version-less publish → no contentHash → `"unknown"` → **exactly current behavior** (zero regression for non-binding servers).
- **I4 superseded publish**: version < last-sent is dropped before caching; no binding recorded for a dropped publish.
- **I5 honest disk compare**: `true` iff diskHash === contentHash; `false` iff they differ; `"unknown"` iff no contentHash OR disk unreadable. A read failure is **never** `false` (the #533 direction — never false-clean OR false-live).
- **I6 lazy + cheap**: stat mtime first; read+hash only on mtime change. No hashing on the hot notification path; version-less servers do **zero** extra disk I/O.
- **I7 multi-client composition**: ANY contributor `false` → `false`; all `"unknown"` (or none) → `"unknown"`; else `true` (unknowns never block a true).
- **I8 clear on resync**: `clearDiagnosticsForPath` drops the stored binding with the diagnostics.
- **I9 confirmation demotion**: `boundToCurrentDisk === false` ⇒ unconfirmed (no widget write, not cached).
- **I10 non-enumerable**: `binding` attached to the diagnostics array like `inconclusive` — no shape change for array consumers.
- **I11 observability**: binding state in latency logs so an unbinding server is diagnosable.

## Binding composition rule chosen
`composeBoundToCurrentDisk([...])`: **any `false` → `false`; else all `"unknown"` (or empty) → `"unknown"`; else `true`.** A single client whose view diverged from disk marks the whole merged (primary + auxiliaries) result mismatched; a version-less auxiliary alongside a bound primary does not erase the primary's binding.

## Test matrix (fail-then-pass — each guarding test verified RED on pre-fix logic)
| Cell | Test | 
|---|---|
| T1 publish version matches → bound | `client-internals` "binds … when the publish version matches" |
| T2 **headline**: stale binding on a NON-EMPTY result → demoted, footer not re-cemented | `lsp-diagnostics.test` "does NOT reconcile a NON-EMPTY result whose binding mismatches disk (#1092)"; client-side `client-internals` superseded/mismatch cases |
| T3 version-less → unknown → doctrine unchanged | `lsp-diagnostics.test` "DOES reconcile … when the binding is unknown"; `client-internals` version-less |
| T4 disk changed (mtime+content) → false | `diagnostic-binding.test` |
| T5 mtime changed, content identical → true (hash, not mtime) | `diagnostic-binding.test` + cache lookup test |
| T6 CRLF(+BOM) round-trip → true | `diagnostic-binding.test` |
| T7 composition [false,true]/[unknown,unknown]/[true,unknown]/[] | `diagnostic-binding.test` |
| T8 disk unreadable + contentHash → unknown, never false | `diagnostic-binding.test` |

Fail-then-pass proof: with the capture/demotion/cache-verify logic temporarily neutralized (interfaces intact), the 5 guarding tests (incl. the headline) went **RED**; restored → **GREEN**.

## Blast radius
`getAllDiagnostics` consumers (`cascade-tier.ts`, `dispatch/integration.ts`, `tools/lsp-navigation.ts`) and `touchFile` consumers (dispatch lsp runner, pipeline, runtime-session, warm-attach, tools) read `.diags`/`.ts`/array elements only — `binding` is a new **optional/non-enumerable** field, so all are additive-compatible (no shape change). Workspace-cache `record` gained an **optional** trailing `contentHash` param; the pull-sweep record site passes none (honest `"unknown"`). Cache format bumped **v1 → v2** (legacy entries rejected by the existing version guard → re-scanned, fail-open).

## Non-goals (explicit deferrals)
- **Cascade adoption** in `clients/dispatch/integration.ts` — that file is actively changing under #1094; **untouched here**. #1094 (branch `fix/1093-widget-staleness`) is **not yet merged to master** (its `LSPWorkspaceDiagnosticResult.observedAt` is absent on master); when it lands, the merged binding should feed `observedAt`/cascade validity in the SECOND PR.
- pull-diagnostics resultIds; widget per-entry timestamps (#1093); aux-LSP retag (#1094).

## Verification
`npm run build` clean. Full suite: **5258 passed / 12 skipped**, 1 failure = the known `edits.test` occupancy timing flake under full-suite load (passes in isolation: 27/27).


apmantza:
## Review round 1 addressed (rebase + P2s + P3s)

Rebased onto current `master` (#1094 now merged) and applied the adversarial-review findings.

### Rebase / merge (merge commit `176cb3f3`)
Three conflicts, all unions:
- **`reconcileWidgetFromLspResult` signature** — both PRs appended a trailing param. Resolved to `(…, boundMismatch = false, observedAt?: number)`. **Positional-arg hazard fixed**: #1094's cache-hit call passed `cached.scannedAt` positionally into the (now) `boundMismatch` slot — corrected to `(…, undefined, false, cached.scannedAt)`. The two fresh-touch calls keep `boundMismatch` and omit `observedAt` (fresh touches observe now).
- **`WorkspaceDiagnosticsCacheLookup`** — unioned my `binding` with #1094's `scannedAt`.
- **cache test assertions** — unioned (`toMatchObject` + `scannedAt` + `binding`).

### P2-1 (bug-class sweep miss) — FIXED
`runWorkspaceDiagnostics`'s service cache-hit (`clients/lsp/index.ts`) had no binding gate while the `tools/lsp-diagnostics.ts` sibling did. Since both sweeps share cache entries under the same `scopeKey`, a hash-bearing entry whose bytes diverged **without** an mtime bump could be replayed as confirmed via `mode=full`. Added the same `boundToCurrentDisk !== false` gate (mismatch → fall through to fresh collection). Regression test: seed a v2 entry with a mismatching `contentHash` at a matching mtime, run the service sweep, assert it is NOT served (fresh touch happens). Verified red→green.

### P2-2 (eager cost, zero consumers) — FIXED (lazy)
Service `getAllDiagnostics` no longer eagerly stat+hashes every bound file. `binding` is now a **lazy, non-enumerable, memoized getter** — the disk verify runs only when a consumer reads `.binding`. The current caller (cascade) doesn't read it (adoption deferred to PR 2), so eager cost is now **zero**, while the field stays surfaced for the PR-2 consumer.

### P2-3 (version-counter reuse → false-TRUE) — FIXED
`reopenOnResync` (opengrep) reset the document version to 0 on every close+reopen, so a late publish for an earlier resync's content echoed the same `0` as the current send → `isSupersededPush` (0 < 0 false) accepted it → a **stale** result bound to the **current** content's fingerprint (affirmative `boundToCurrentDisk: true`, worse than "unknown"). Fixed by carrying the version counter **forward** across reopen (LSP permits any didOpen version) — the late echo is now strictly older → dropped → never bound. Regression test simulates the late-echo-after-reopen sequence. Verified red→green. (The `closeDocument`-reopen path is analyzed to yield "unknown", not a false-TRUE, because binding requires a version match; documented at the site.)

### P3s — all done
- **(a)** `diskBindingCache` converted to `PathKeyedMap` + `normalizeEphemeralMapKey` (#1025 discipline) and **bounded** (max-size clear at 4096).
- **(b)** tsserver **sync-confirmed** results now surface binding `"unknown"` instead of the stale publish-path binding, so a fresh sync answer can't be demoted by an older publish fingerprint.
- **(c)** documented that `preserveDiagnostics:true` format-resyncs retain the prior binding (a self-healing transient "unconfirmed", the safe #533 direction — not a hazard).

### Verification
`npm run build` clean. Full suite green except the known `edits.test` occupancy flake (passes in isolation). Every new regression test proven fail-then-pass by neutralizing the specific fix.

PR #1102, opened 2026-08-06
fix(test-runner): preserve real subcommands during binary resolution (closes #1098)

### LINKED ISSUE #1098: test runner strips real subcommand for non-npx runners (cargo test → cargo)
### What happened?

The on-write/turn-end test runner executes a bare `cargo --no-fail-fast` for
Rust projects — the `test` subcommand is missing — so cargo fails with a clap
error and pi-lens reports it as a test failure:

```
1/1 failed, 0 passed — cargo
error: unexpected argument '--no-fail-fast' found
```

The agent-facing turn-end note then claims tests are failing on every edit,
even though `cargo test --no-fail-fast` passes. Same bug class affects the
`go`, `dotnet`, `maven`, and `mix` runners (their real `test` subcommand is
stripped the same way).

### Steps to reproduce

Any directory with a `Cargo.toml` and `cargo` on PATH. Edit a file (or run
the runner directly):

```js
// node --input-type=module -e "..."
import { TestRunnerClient } from '<pi-lens>/dist/clients/test-runner-client.js';
const c = new TestRunnerClient(true);
const det = c.detectRunner('<project>', '<project>/src/main.rs');
const r = await c.runTestFileAsync('<project>/src/main.rs', '<project>', det.runner, det.config);
console.log(r);
```

Observed:

```
[test-runner] Detected runner via config: cargo
[test-runner] Running (async): /usr/bin/cargo --no-fail-fast
{ passed: 0, failed: 1, failures: [ { name: "cargo failure", message: "error: unexpected argument '--no-fail-fast' found" } ] }
```

### Expected behavior

```
[test-runner] Running (async): /usr/bin/cargo test --no-fail-fast
{ passed: N, failed: 0 }
```

### pi-lens version

`npm ls pi-lens` → `pi-lens@3.8.74`

### Relevant logs

From `~/.pi-lens/sessionstart.log` (0ms = cargo fails instantly, before any
test runs):

```
[2026-08-06T00:59:58.598Z] turn_end: src/cellfile.rs → test cargo src/cellfile.rs (related)
[2026-08-06T00:59:58.743Z] turn_end: [stale] test cargo cellfile.rs → FAIL 0p/1f (0ms)
[2026-08-06T01:00:10.432Z] turn_end: src/cellfile.rs → test cargo src/cellfile.rs (failed-first)
[2026-08-06T01:00:10.577Z] turn_end: [stale] test cargo cellfile.rs → FAIL 0p/1f (0ms)
```

### Area

`area:tests`

---

### Root cause and proposed fix

`resolveExec` in `dist/clients/test-runner-client.ts` drops `args[0]`
unconditionally whenever it resolves a runner binary (local
`node_modules/.bin/<name>` or global bin via `findGlobalBinary`), assuming
the npx convention where arg 0 is the runner name (vitest/jest:
`npx vitest run ...`). For direct runners, arg 0 is a real subcommand:

```js
// current
if (fs.existsSync(localBin)) {
    return { command: localBin, args: config.args(testFile, cwd).slice(1) };
}
const globalBin = await findGlobalBinary(binName);
if (globalBin) {
    return { command: globalBin, args: config.args(testFile, cwd).slice(1) };
}
```

Fix: only drop the wrapper-convention args — npx-style runner name
(`args[0] === binName`) or python `-m` pair (`args[0] === "-m" && args[1] ===
binName`, pytest):

```js
const args = config.args(testFile, cwd);
let dropCount = 0;
if (args[0] === binName) {
    dropCount = 1;
} else if (args[0] === "-m" && args[1] === binName) {
    dropCount = 2;
}
if (fs.existsSync(localBin)) {
    return { command: localBin, args: dropCount ? args.slice(dropCount) : args };
}
const globalBin = await findGlobalBinary(binName);
if (globalBin) {
    return { command: globalBin, args: dropCount ? args.slice(dropCount) : args };
}
return { command: config.command, args };
```

Verified against a locally patched copy: cargo runs
`/usr/bin/cargo test --no-fail-fast` → `{ passed: 14, failed: 0 }`, and a fake
local `node_modules/.bin/vitest` still receives
`run <file> --reporter=json --passWithNoTests`.

Secondary observation (pre-existing, separate): `rspec`'s `binName` falls
back to `"rspec"` while its args lead with `"exec"` (bundle convention), so
its local-bin path is mismatched too — `binName: "bundle"` would be the
correct resolution.

apmantza:
## Summary

**Defect class:** `resolveExec` applied the npx-wrapper convention (arg 0
of `args()` names the binary) unconditionally to every runner's argv once
local/global binary resolution succeeded. That convention only holds for
wrapper-style runners (vitest/jest/pytest); for direct runners whose
`args()` leads with a **real subcommand**, the same unconditional
`.slice(1)` silently ate it.

**Confirmed impact:** `cargo test --no-fail-fast` resolved to the
argv-invalid `cargo --no-fail-fast` (a clap usage error) whenever a local
or global `cargo` binary was found — same class for `go`, `dotnet`,
`maven`, `mix`. The runner then misreported the usage error as a false
"1/1 failed" at turn-end, both misinforming the agent and burying real
signal.

**Secondary defect (same PR, per triage):** `rspec`'s config declares
`command: "bundle"` with `args: ["exec", "rspec", testFile]` (because the
real invocation is `bundle exec rspec <file>`), but `binName` resolution
defaulted to the runner key `"rspec"` — the wrong binary name for
local/global lookup.

## Fix

- `stripWrapperArgs(binName, args)` (new, exported) strips a leading arg
  **only** when it actually names the resolved binary:
  - `[binName, ...rest]` (vitest/jest: `npx <bin> ...`) → drop 1
  - `["-m", binName, ...rest]` (pytest: `python -m <bin> ...`) → drop 2
  - anything else (`cargo`'s `["test", "--no-fail-fast"]`, `go`'s
    `["test", "-run", …]`, `rspec`'s `["exec", "rspec", file]` once
    `binName` is `"bundle"`) is returned **unchanged**.
- Applied at both `resolveExec` resolution sites (local `node_modules/.bin`
  and `findGlobalBinary`) — no duplicated logic.
- `rspec`'s config now declares `binName: "bundle"` so local/global
  resolution looks up the right binary; `args[0]` is `"exec"` (a real
  `bundle` subcommand, not the binary name), so nothing is stripped and
  `bundle exec rspec <file>` reaches `bundle` intact.

## Invariant

`resolveExec` must never alter the semantic argv of a runner — it may only
strip the leading token(s) that named its own resolved binary.

## Test coverage (the class guard)

`tests/clients/test-runner-client.test.ts` gained a matrix
(`describe("resolveExec argv preservation matrix (#1098)")`) that iterates
the **exported** `RUNNERS` config table — not a hand-copied runner list —
so a future runner is automatically covered. For every runner it exercises:

- **local-bin hit** — a real file dropped in `node_modules/.bin` (or
  `vendor/bin` for phpunit's Composer convention)
- **global-bin hit** — `findGlobalBinary` mocked to resolve
- **fallback** — neither resolves; `config.command` used verbatim

The expected argv per case is computed by an independent local
re-statement of the strip contract (not by importing `stripWrapperArgs`),
so a regression can't trivially pass by agreeing with a broken
implementation. Dedicated cases also assert the exact reported `cargo`
argv and the `rspec`/`bundle` resolution.

**Fail-then-pass proof:** reverted just the production diff
(`git apply -R`) with the test file left in place — the whole file failed
(`RUNNERS` not exported, `TypeError: Cannot convert undefined or null to
object`), i.e. red on pre-fix code. Re-applying the fix (`git apply`) turns
all 79 tests in the file green. Also verified directly: pre-fix, `cargo`'s
resolved args were `["--no-fail-fast"]` (the `test` subcommand eaten);
post-fix, `["test", "--no-fail-fast"]`.

## Blast radius

`resolveExec` is private; its only callers are internal to
`TestRunnerClient` (the public `runTest` path). Public API surface changes
are additive only — `RUNNERS`, `RunnerConfig`, and `stripWrapperArgs` are
now exported (for the test matrix), nothing removed or renamed. Consumers
(`runtime-session.ts`, `runtime-turn.ts`, `review-graph/builder.ts`,
`project-diagnostics/runner-adapters/runner-findings.ts`, `bootstrap.ts`)
use only the pre-existing public methods (`detectRunner`, `findTestFile`,
`runTest`), whose signatures/behavior are unchanged except for the
correctness fix itself.

## Class sweep (recorded per the bug-class-sweep rule)

Grepped the repo for the same defect shape — positional argv stripping
tied to binary resolution, and command/binName-resolution name mismatches
— across every other `findGlobalBinary` caller:
`clients/biome-client.ts`, `clients/formatters.ts` (biome/prettier/rubocop/
standardrb formatters, including the Ruby `bundle exec` ones),
`clients/dispatch/runners/utils/runner-helpers.ts`, `clients/sg-runner.ts`,
`clients/lsp/launch.ts`. Every one of them builds its resolved argv as
fresh literals per resolution branch (`args: []`, or a hand-written
`["bundle", "exec", "rubocop", ...]`) rather than slicing a shared
`args()` generator — none share this defect's precondition (a config table
of `args()` functions combined with a single generic strip applied
uniformly across resolution paths). `clients/test-runner-client.ts`'s
`RUNNERS` table is the only place with that shape in the codebase
(confirmed via `grep -r "args:\s*(.*=>\s*\["` — no other matches). No
further instance to fix in this PR.

## Non-goals

- No new test runners added.
- No change to `RUNNERS` entries beyond `rspec.binName`.
- `gradle`/`maven`'s `binName` also defaults to the runner key even though
  their real commands (`gradlew`/`mvn`) differ — structurally the same
  mismatch shape as `rspec`, but currently **unreachable**: neither
  `findGlobalBinary` (npm/pnpm/yarn/bun global dirs only) nor a
  `node_modules/.bin` lookup will ever resolve a binary literally named
  `"gradle"` or `"maven"`, so both runners always fall through to
  `config.command` unchanged today. Left as-is to keep this PR scoped to
  the reported (reachable) defect; flagging here rather than silently
  leaving it for a future re-discovery.

## Test plan

- [x] `npm run build` (tsc, clean)
- [x] `npm run lint` (tsc typecheck, clean)
- [x] `npx vitest run tests/clients/test-runner-client.test.ts` — 79/79 pass
- [x] Fail-then-pass proof on the production diff (see above)
- [x] Full suite: `npx vitest run` — 5295 passed / 14 failed / 12 skipped
      across two independent runs. Zero failures involve
      `test-runner-client`. Re-ran every distinct failing file in
      isolation: all passed except `tests/clients/lsp/edits.test.ts`'s
      "keeps large text/resource planning occupancy bounded" (a
      documented, timing-sensitive flake under parallel load, unrelated to
      this change).


apmantza:
Adversarial review round 1: **merge** — no P1/P2. Independent verification: real behavioral fail-then-pass re-proven (20 argv-content assertion failures with stripWrapperArgs neutralized, zero compile failures); class sweep re-run and confirmed accurate; all three resolution paths diffed for all 12 runners. Touch-up commit addresses the P3s: CHANGELOG test-path corrected, and the full changed-set disclosed (pytest resolved-path argv was silently broken pre-fix — slice(1) ate `-m`; gradle/minitest corrected too; minitest recorded as the fourth command≠binName deferral member). P3-4 (freeze the exported RUNNERS table) left as reviewer-suggested hardening for the hygiene backlog (#1108).

Issue #1085, opened 2026-08-05
LSP workspace-edit apply paths: reversed same-position inserts, win32 casing destruction, versioned-edit rejection (post-#1053/#1066 adversarial review)


apmantza:
Adversarial review of the LSP workspace-edit / write paths reworked since v3.8.74 (#1053/#1066: `clients/lsp/edits.ts`, `clients/lsp-mutation.ts`, `clients/lsp/index.ts`, `clients/lsp/client.ts`, `tools/lsp-navigation.ts`, `clients/actionable-warnings.ts`). The three P1s were **empirically reproduced** against the built `clients/lsp/edits.js` on Windows; line numbers as of 293e7b54.

## P1 — same-position insert order REVERSED through `applyWorkspaceEdit` (the exact #1066 invariant)
`clients/lsp/edits.ts:241-265` (`sortAndValidateTextEdits`, tie-break `b.index - a.index` at :246) interacting with :768 (preflight sort) and :907 → `applyTextEditsToString` (:306, second sort).
- Reproduced: zero-width inserts `["A","B"]` at (0,1) on `"abc"` → `applyTextEditsToString` directly: `"aABbc"` (correct); `applyWorkspaceEdit({changes})` / `({documentChanges})`: **`"aBAbc"`** (reversed). LSP spec: array order defines insertion order.
- Root cause: the reverse-index tie-break is correct only for exactly ONE sort pass before reverse-order application; the apply pipeline sorts an EVEN number of times (each pass flips same-position ties). The utf-8/utf-32 path through `normalizeWorkspaceEditToUtf16` sorts a THIRD time and comes out correct — so behavior differs by server position-encoding; utf-16 servers (typescript-language-server, pyright — the majority) are the broken ones.
- Affected callers: `workspace/applyEdit` handler (`clients/lsp/client.ts:1127`, i.e. every executeCommand mutation), rename apply (`tools/lsp-navigation.ts:1494`), code-action autofix (`clients/actionable-warnings.ts:737`). Multi-import inserts at one position are a classic code-action shape.
- Fix direction: normalization returns ascending array-order edits and the write site sorts exactly ONCE, or make the sort idempotent by carrying original-array order through (not a per-call index).

## P1 — Windows case-preservation destroyed on create/rename; case-only renames rejected (invisible to Linux CI)
[...]

Issue #1084, opened 2026-08-05
Guard store: stale-record false blocks / recovery (residual of #1084 after #1395)


apmantza:
As of PR #1395, the two security-real items are FIXED and merged. One item remains.

**Remaining — stale-record false blocks / recovery.** The grant/guard store can hold stale records that cause false blocks (or fail to recover cleanly) after the underlying file/command state changes. Needs: a recovery path that revalidates or ages out stale grant/guard records so a legitimate operation isn't blocked by an outdated record. Deferred from #1395 (which was scoped to the command-detection + quote-blind-grant classes).

---
Done and merged in #1395 (for the record):
- Command-detection bypasses — root-cause hardening: whole-command IFS canonicalization (`$IFS`/`${IFS}`/`$IFS$N`/`${IFS%x}`), verb-escape normalization (`\`, backtick, `^`), `.exe/.com/.bat/.cmd`+path launcher normalization, and a fail-closed "non-leading git = indirect" backstop that needs no wrapper/flag allowlist. Substitution (`$(...)`, backticks, `<(...)`) treated as execution, not text.
- Quote-blind read-guard grants — grants tokenized/quote-aware and normalized on both record and lookup; grants minted only after successful tool_result (failed commands mint nothing).
- Hot-path cost — no perf regression introduced (item deferred as non-blocking; revisit only if measured).

Four adversarial review rounds closed six distinct bypass classes before merge.

apmantza:
**Triage verdict: KEEP OPEN.** Re-assessed against current master — the remaining items are real security work, not polish: (1) command-detection bypasses in the guard's tokenizer path, (2) quote-blind permission grants, (3) stale-record recovery on the grant store. None are covered by the merged guard PRs. Staying open until each has a fix or an explicit wont-fix disposition.

PR #1103, opened 2026-08-06
fix(diagnostics): stop dropping sg-scan exit-1 matches; sweep silencing-is-not-fixing (refs #1087)

apmantza:
Fixes the P1 + P2 (and two contained P3s) from the #1087 adversarial-review audit. Remaining P3s stay open on #1087 (this PR **refs**, does not close).

## Defect classes

1. **An exit code that means "scan succeeded with findings" classified as a CLI failure.** ast-grep is linter-style: a rule with `severity: error` that MATCHES exits **1**, writes valid JSON matches to stdout, and writes `Scan succeeded and found error level diagnostics` to stderr.
2. **"Silencing is not fixing" applied to only one member of the suppression class.** A finding persistently dropped by a filter that runs on the live set but not the delta baseline is re-counted as `fixed` on every dispatch.
3. **Rule-id normalization / config-warning asymmetries** (the two contained P3s).

## Invariant

> **An exit code that means "scan succeeded with findings" must never be classified as a CLI failure.**
> The delta's `fixed` set must be measured between two views filtered through the *same* pipeline — a persistently-suppressed finding is not a newly-fixed one.

## Empirically verified (bundled ast-grep 0.45.0, outside vitest)

A temp `severity: error` rule (`pattern: eval($ARG)`) scanned against a matching file:

| invocation | exit | stdout | stderr |
|---|---|---|---|
| `scan --config … --json target.js` | **1** | valid JSON array of matches | `Error: 1 error(s) found in code.` / `Help: Scan succeeded and found error level diagnostics` |
| `scan --config … --update-all` (rule w/ `fix:`) | 0 | (rewrites file) | as above pre-apply |

The observed shape is encoded in the fixtures.

### P1 — `clients/sg-runner.ts` (`interpretScanResult()` + `exec()`)
`status===1` + stdout that parses as valid JSON → SUCCESS (parse matches, via shared `tryParseSgMatches`). stderr-with-unparseable-stdout stays a failure; the historical status-1/no-output no-match convention is preserved.

### P2 — `clients/dispatch/dispatcher.ts`
The output-filter pipeline (docker/LSP overlap → inline `pi-lens-ignore` → dispositions → rule policy) is now a single `applyOutputFilters` closure applied identically to the live set **and** the delta baseline before `filterDelta`. Previously only the policy member was re-applied to the baseline. Stored baseline remains the unfiltered deduped set.

### P3 (contained)
- `clients/dispatch/inline-suppressions.ts`: the comment token contributes its `normalizeRuleId` form too, so `// pi-lens-ignore: no-eval-js` suppresses a normalized `no-eval` finding (symmetric with `disable: ["no-eval-js"]`).
- `clients/project-lens-config.ts`: `"disable": []` / `"select": []` are a well-formed no-op (dropped silently, no warning); a non-empty array of all-blank/non-string entries still warns.

## Seams extended / blast radius

`interpretScanResult` callers: `tempScanDetailedAsync` → `tempScanAsync` (session scanners), `tempScanWithFixAsync` (replace apply/preview), and `AstGrepClient.validateRule` / `searchWithRule` / `replace`. `exec()` is the raw `ast_grep_search` path. All now surface real matches on error-severity rules instead of a "cli-failure" / "ast-grep preview failed". The dispatcher change only affects the delta `fixed` set (and thus `resolvedCount`/`trackAgentFixed`); the rendered blockers/warnings and the stored baseline are unchanged.

## Tests (fail-then-pass verified — stashless: reverted via `git checkout`, rebuilt, ran, restored)

- `tests/clients/sg-runner.test.ts`: status-1-with-JSON-matches → success (both `interpretScanResult` and `exec` paths); status-1 unparseable-stdout stays a failure; status-1-empty no-match preserved.
- `tests/clients/ast-grep-severity-error.test.ts` (new, tool-level): `validateRule` reports a matching `severity: error` rule VALID through the real `AstGrepClient` → `SgRunner` chain (only `safeSpawnAsync` mocked).
- `tests/clients/dispatch/rule-policy-dispatcher.test.ts`: an inline-suppressed finding is absent from the first render and NOT counted in `resolvedCount` on a repeat dispatch.
- `tests/clients/dispatch/inline-suppressions.test.ts`: language-suffixed comment token suppresses a normalized id.
- `tests/clients/project-lens-config.test.ts`: `disable: []` is a silent no-op (updated from the old warn-expectation); non-empty all-blank list still warns.

All 7 new/changed assertions fail on pre-fix `master` and pass with the fix. Full suite: 5270 passed; the only failures are the documented resource/timing flakes (edits.test occupancy et al.) whose set changes run-to-run and which all pass in isolation.

## Non-goals (remain open on #1087)

P3-2 rule-id suffix design (`-c`/`-cpp`/`-csharp`/`-python`), P3-3 `source-filter.ts` twin-candidate/precedence matrices, P3-4 workspace-root threading gaps, P3-5 widget-state `rule = d.rule ?? d.id`, P3-6 `select` blocker-gate render notice, P3-7 remaining config-warning edges (mixed arrays / `only` drop), P3-8 `sg scan --update-all` lifetime-coupling. Guard (#1084), LSP apply (#1085), consistency sweep (#1086) are separate.


apmantza:
Review round 1: approved modulo the type fix — applied in 1283c291.

- `SgMatch` now declares `ruleId?: string` / `severity?: string` (real `sg scan --json` fields the type never modeled; the P1 fixture already carries them). Fixes the red **Lint & type-check** job; assertion at `sg-runner.test.ts` was NOT weakened.
- `tryParseSgMatches` hardened to accept only an array or object payload — a JSON scalar/`null` is rejected so a hypothetical error-report scalar on stdout can't become a phantom match. Added a `status 1 + stdout "null" -> cli-failure` regression test.

`npm run lint` (tsc) and `npm run build` both clean; sg-runner + tool-level test files green (24 passed). No full-suite rerun — CI covers it.

PR #1112, opened 2026-08-06
feat(test-infra): serialize full-suite test runs with a machine-wide lock (closes #1101)

### LINKED ISSUE #1101: Serialize concurrent full-suite test runs with a machine-wide lock (.install.lock pattern)
## Problem
Multiple concurrent full-suite vitest runs on one dev machine (now common: several agents working parallel branches, plus the maintainer's interactive runs) interfere destructively. Each run spawns a fork pool sized for a dedicated machine (`maxWorkers: "50%"`, `--max-old-space-size=4096` per fork — vitest.config.ts). N concurrent suites → memory pressure and CPU contention → vitest worker-crash cascades (observed repeatedly on 2026-08-06: edits.test occupancy + runtime-session + shared-extraction-ir fail together, all pass in isolation), inflated occupancy/timing-budget flakes, and wasted review rounds chasing phantom failures.

## Fix: machine-wide full-suite lock (queue, not broker)
A small cross-process mutex so full-suite runs serialize on one machine — the repo already has the exact pattern in the installer's atomic `.install.lock` (stale only after its recorded PID is confirmed dead; AGENTS.md documents the semantics).

Design:
- `scripts/with-test-lock.mjs` wrapper (or acquire in vitest globalSetup — implementer's choice, wrapper preferred: zero cost on the no-contention path and works for any command): acquires `~/.pi-lens/test-suite.lock` (or os.tmpdir-scoped per-repo lock keyed on the repo path slug — must not collide across DIFFERENT repos), runs the wrapped command, releases in a finally.
- Waiting runner polls with a visible heartbeat line ("waiting for test-suite lock held by PID <x> since <t>") so a queued run never looks hung.
- Stale-lock rule: identical to .install.lock — a lock is stale only after its PID is confirmed dead; take over then.
- Opt-out `PI_LENS_TEST_NO_LOCK=1` (CI sets it — GitHub runners are isolated; the lock is a dev-machine concern).
- Wire it into `npm test` / `test:unit` / `test:integration` scripts so both humans and agents get it for free.
- Scope: full-suite invocations. Targeted single-file runs stay unlocked (cheap, and serializing them would hurt iteration).

## Companion policy (layer 1, already adopted operationally)
Agent briefs now mandate: targeted tests freely; at most ONE full-suite run at the end; `PI_LENS_TEST_MAX_WORKERS=4` for that run; GitHub CI is the authoritative full-suite green; crash-cascade failures re-run in isolation before being treated as real. The AGENTS.md update encoding this policy should land IN the lock PR (same commit that changes the world it describes).

## Non-goals
- A broker service/daemon — the flock-style lock is sufficient.
- Fixing the load-sensitive occupancy budgets themselves (edits.test, cascade-graph-occupancy) — that belongs to the timing-sensitive-project thread (#1081-adjacent), tracked separately.

apmantza:
## Summary

- `scripts/with-test-lock.mjs` wraps a command (argv passthrough after `--`) with a machine-wide lock, so concurrent full-suite vitest runs on one dev machine queue instead of fighting over CPU/RAM (fork pools sized for a dedicated machine, per vitest.config.ts's `maxWorkers: "50%"`).
- `npm test` / `npm run test:unit` / `npm run test:integration` now go through the wrapper. CI's "Run tests" step sets `PI_LENS_TEST_NO_LOCK=1` explicitly (runners are isolated, one job per box — nothing to serialize against).
- `AGENTS.md` documents the lock (auto-acquired, heartbeat = queued not hung, opt-out) and the companion layer-1 policy: agents run touched-file tests freely, at most ONE full-suite run at the end with `PI_LENS_TEST_MAX_WORKERS=4`, GitHub CI is the authoritative full-suite green, and crash-cascade failures under load must be re-run in isolation before being treated as real.

## Design

- **Lock location**: ONE machine-wide file, `~/.pi-lens/test-suite.lock` (or `$PI_LENS_HOME/test-suite.lock`) — deliberately NOT per-repo/per-worktree, since worktrees of the same repo (or different repos entirely) still contend for the same physical CPU/RAM.
- **Pattern**: mirrors the installer's `.install.lock` (`clients/installer/index.ts`) — atomic create via `fs.open(path, "wx")`, owner body `{ pid, startedIso }`, and "stale only after the recorded PID is confirmed dead" takeover (immediate, no age wait). Re-implemented minimally in `scripts/lib/suite-lock.mjs` rather than imported, since `scripts/` runs as plain `.mjs` before `npm run build` compiles `clients/*.ts` → `clients/*.js`; the header comment points back at the original.
- **Windows correctness**: `process.kill(pid, 0)` with EPERM treated as alive (not just ESRCH-is-dead); atomic `wx` create is atomic on both POSIX and Windows filesystems; unlink retries a few times on transient EBUSY/EPERM (AV/indexer file holds) before giving up (the next waiter's stale-PID check still recovers it); `with-test-lock.mjs`'s child spawn needed its own Windows-specific fix — `shell:true` + an args array does NOT quote for you on Windows (confirmed experimentally: an arg containing a space silently split into two argv entries), so win32 now builds one CRT-quoted command-line string instead.
- **Waiting**: polls with backoff, heartbeat line at least every 15s (`waiting for test-suite lock held by PID <pid> since <iso>`), no wait timeout by default (`PI_LENS_TEST_LOCK_TIMEOUT_MS` optional).
- **Opt-out**: `PI_LENS_TEST_NO_LOCK=1`.
- **Naming note**: the lock core module is `scripts/lib/suite-lock.mjs`, not `test-lock.mjs` — the repo's `.gitignore` has a `test-*.mjs` pattern reserved for ephemeral scratch fixtures, which silently ignored the file under its natural name. Renamed (with a matching `.d.mts` sibling, following the `scripts/lib/clean-signal.mjs` + `clean-signal.d.mts` convention already in the repo, needed so cross-file JSDoc inference from a `.ts` test resolves instead of erroring as implicit `any`).

## Test plan

- [x] `tests/scripts/suite-lock.test.ts` — 9 unit tests: atomic acquire/release (incl. idempotent release, nested-dir creation), immediate takeover of a lock held by a confirmed-dead PID (no age/timeout wait), mtime-based takeover of an unreadable/corrupt lock past `staleMaxAgeMs`, and **real contention**: two concurrent `acquireTestLock()` calls against the same lock file — the second is proven blocked for several poll cycles (heartbeats observed), then resolves in-order only after the first explicitly releases; plus a timeout-error case.
- [x] `npm run build` — clean.
- [x] `npm run lint` — clean (tsc project-wide, including the new `.mjs`/`.d.mts` pair).
- [x] Wrapper smoke: `node scripts/with-test-lock.mjs -- npx vitest run tests/scripts/suite-lock.test.ts` — passthrough works end-to-end.
- [x] Real dead-PID takeover at the process level (not just unit test): a lock file with `pid: 999999999` was taken over in ~260ms by a fresh `with-test-lock.mjs` invocation.
- [x] Real two-process contention: two background `with-test-lock.mjs` invocations against the same lock — the second printed `waiting for test-suite lock held by PID <pid> since <iso>` heartbeats and only ran after the first released.
- [x] **Dogfooded on this machine**: ran the full suite in the foreground through the wrapper (`PI_LENS_TEST_MAX_WORKERS=4 node scripts/with-test-lock.mjs -- npx vitest run`) immediately after an earlier full-suite `npm test` run had *just* released the lock — the second invocation's start time was gated almost exactly to the first's release (~4s gap), i.e. the two real full-suite runs serialized on this exact machine rather than overlapping. The second run then hit a genuine crash-cascade under load (`Fatal process out of memory: Zone` + 4 unrelated timeout failures across `source-filter-async`, `shared-extraction-ir`, `cross-package-graph`, `mcp/analyze`); re-running those 4 files in isolation immediately afterward passed all 47 tests cleanly — a textbook instance of the exact flake class this issue exists to reduce, and consistent with the AGENTS.md guidance this PR adds (re-run crash-cascade failures in isolation before treating them as real).
- [x] Final clean full-suite baseline (no contention): 453 files passed, 5282 tests passed, 12 skipped, 0 failed.

Closes #1101.


apmantza:
Addressed round-2 review (commit a0204138):

**P2-2 (doc truthfulness, REQUIRED)** — fixed both directions:
- `.install.lock` ALSO ages out a lock with a still-live, readable PID past the install bound + slack (#946 F1's PID-recycle defense) — the pre-existing AGENTS.md line claiming "stale only after PID confirmed dead" was wrong for that lock too. Corrected.
- `suite-lock.mjs` does NOT do that, deliberately — a full-suite run has no bounded duration to size a timeout against, and the owner never touches the lock file's mtime while holding it. Consequence stated explicitly (both in AGENTS.md and the module's header): a PID-reuse race there means a waiter waits **forever** until a human reads the heartbeat's `pid`+`startedIso` and deletes the lock by hand. Noted owner-side periodic mtime-touch as unimplemented future hardening.

**P3-3** — added the explicit sentence: only `npm test`/`test:unit`/`test:integration` acquire the lock; `npx vitest run <file>` directly stays unlocked; `npm test -- <file>` still queues (same npm script → same wrapper).

**P3-1** — `suite-lock.mjs`'s header now states the installer divergence explicitly (point 5), matching the AGENTS.md language.

**P2-1 (STRONGLY RECOMMENDED) — went with option (a), not (b):** `with-test-lock.mjs` now resolves vitest's own `package.json` `bin` entry and spawns it via `node <entry> <args>` directly — `shell:false`, no cmd.exe, identical behavior on every OS. This is the actual dispatch path for every real caller (`npm test`/`test:unit`/`test:integration` all wrap `vitest run [...]`), so it deletes the entire quoting minefield you demonstrated live (quote-parity flip → `&` executes as a separate command; `%VAR%` expands regardless of quoting) for the path that matters. The CRT-quoting shell fallback is kept only for a hypothetical non-vitest caller (none exists today) with the over-claiming comment corrected and its real, unfixed limits pinned down by new tests (`tests/scripts/with-test-lock.test.ts`) instead of asserted away — quote-containing arg, `%VAR%`, `&`, trailing backslash are all covered, each documenting what actually breaks.

**Bonus fix found by the new tests**: a lock file just unlinked by another process's `release()` could transiently reject a fresh `open(path, "wx")` with EBUSY/EPERM before Windows fully dropped the deleted file's handle — this crashed the second waiter in the real two-process contention test instead of letting it retry. Now treated the same as EEXIST (contended → retry). Reproduced the crash pre-fix, then confirmed stable across repeated runs post-fix.

**P3-2 (OPTIONAL)** — left the stale-takeover ABA race as a documented code comment (inherited as-is from `.install.lock`, same shape, millisecond window, post-crash only) rather than fixing it here, per the bug-class-sweep discipline: a real fix needs to land in both places, which is out of scope for this round.

**Test evidence:**
- `npm run build` / `npm run lint` — clean.
- All lock unit tests green and stable across 3 repeated runs (17 tests: 9 in `suite-lock.test.ts`, 8 new in `with-test-lock.test.ts`).
- Re-verified live on Windows after the spawn-construction change: exit-code propagation (deliberately failing test → wrapper exits 1), real two-process contention (heartbeat printed, second run gated on first's release), vitest-entry resolution (`resolveVitestEntry()` returns a real file on disk and the wrapper runs real test files through it end-to-end).
- Because the spawn construction changed, re-ran the full suite in the foreground through the wrapper (`PI_LENS_TEST_MAX_WORKERS=4 npm test`): 452 files / 5281 tests passed, 1 file / 1 test failed on a worker OOM crash-cascade (`scanner.test.ts`, unrelated to this change) — re-ran that file in isolation immediately after and it passed clean (25/25), consistent with the AGENTS.md guidance this PR itself adds.

apmantza:
Round 3 (commit 04f147ca):

**Correction to my round-2 report**: I claimed the EBUSY/EPERM-on-open fix was "verified: failed intermittently pre-fix, stable post-fix" during development — that was true of an ad hoc local run, but I never committed a test that exercises it, so the claim was not backed by anything in the PR. `tests/scripts/suite-lock.test.ts` was in fact byte-identical to round 1 on that front. Fixed now, for real, in this commit — not retroactively claimed for round 2.

**1. Added the missing test** — `acquireTestLock — transient EBUSY/EPERM on the create itself`. Injection mechanism: `vi.spyOn(fsp, "open")` on the shared `node:fs/promises` module object (Node's ESM module registry is process-wide, so this is visible to `suite-lock.mjs`'s own `import fsp from "node:fs/promises"` binding — no separate DI seam needed). The mock rejects exactly once for the test's `lockPath` + `"wx"` flag with `{code: "EBUSY"}`, then calls through to the real `fsp.open` for every other call, so the underlying lock semantics are exercised for real rather than stubbed end-to-end.

Proof it actually exercises the retry path (not a tautology): I reverted the `EEXIST`-only guard back to round-1 shape locally (`if (error.code !== "EEXIST") throw error;`), re-ran the new test, and it failed with the injected `EBUSY` surfacing as an uncaught error from `acquireTestLock` — exactly the crash this fix prevents. Restored the fix, re-ran 3x, stable pass each time (10/10 in `suite-lock.test.ts`, 18/18 across both lock test files).

**2. P3 — bounded the delete-pending retry sub-path** (`suite-lock.mjs`, the `catch { continue }` when both the owner-read and the mtime `stat` fail, i.e. a release race): added the same `pollIntervalMs` sleep the other retry paths already use, so this can no longer tight-spin ignoring `timeoutMs`.

**3. P3 — case-insensitive `isMain` guard on win32** (`with-test-lock.mjs`): now folds case before comparing `process.argv[1]` against `import.meta.url` on win32, with a basename-match + loud `console.error` warning fallback for a deeper 8.3-short-name mismatch, instead of a silent no-op that would let `npm test` exit 0 having run nothing.

**Test evidence:** `npm run build` / `npm run lint` clean. `tests/scripts/suite-lock.test.ts` (10 tests) + `tests/scripts/with-test-lock.test.ts` (8 tests) = 18/18 passing, stable across repeated runs. No full-suite rerun done this round per the review's own scope note (test + bounded-sleep + guard-compare only, no spawn-construction change).

Issue #1097, opened 2026-08-06
Regression: pi --print remains alive after agent_settled (3.8.74, recurrence of #22)


aeturnal:
# Regression: pi-lens keeps completed `pi --print` process alive after `agent_settled`

## TL;DR

- A background worker ran Pi with pi-lens to check TypeScript files.
- The checks finished successfully, and Pi announced that the agent was fully done.
- The Pi process did not exit. It stayed alive doing nothing, so the worker looked frozen.
- This happened twice, which makes it unlikely to be a one-time provider or network delay.
- Stopping only the finished nested Pi process immediately let the worker continue.
- Keeping language servers alive during an active session is useful. A one-shot `pi --print` command is normally expected to exit after completion, so this behavior appears unintended.
- The likely problem is shutdown ordering: pi-lens appears to keep a background resource open while relying on Pi's shutdown event to close it.
- The exact open resource is not yet confirmed. It could be a language-server connection, timer, worker, socket, or quiet-window task.
- The current workaround is to run the command with a hard timeout and keep its completed output.
- This was exposed when a lightweight background Pi worker with Bash access launched a second Pi command and explicitly loaded pi-lens. The worker did not receive the extension from its parent; it loaded the extension only in the new process it started.

## Summary

A non-interactive Pi process launched with pi-lens can complete its requested diagnostics, emit its final assistant response, emit `agent_end`, and emit `agent_settled`, but remain alive indefinitely instead of exiting.

The process was observed waiting in Node's event loop (`do_epoll_wait`). Its parent shell therefore remained blocked even though the JSON event stream contained a complete result.

This happened twice during final verification of a TypeScript project. Manually sending `SIGTERM` to only the settled nested Pi process made the waiting parent command return and allowed the calling worker to continue.

The expected behavior is that `pi --print` exits after the prompt has settled. Persistent LSP/background resources are useful during an active session, but they should not keep a completed print-mode process alive.

## Related issues

This appears to be a regression or a new form of the lifecycle bug previously fixed in:

- #22 — Missing `session_shutdown` handler causes subagent processes to hang indefinitely: https://github.com/apmantza/pi-lens/issues/22

Issue #22 describes the same core symptom in JSON print mode, but it was reported against pi-lens `3.8.31` and closed after adding a `session_shutdown` handler and unref'ing a timer. This report reproduces the problem on pi-lens `3.8.74`, where the shutdown handler is present.

Potentially relevant lifecycle work:

[...]

Issue #1080, opened 2026-08-05
Tracking: align cascade neighbors with ignored-file and test-role display policy


apmantza:
## Assessment

The review graph already excludes files classified as tests, but the cascade display boundary does not apply the same test-role policy to every neighbor source. In particular, LSP reference expansion, module-graph downstream expansion, and the passive-diagnostics fallback can still surface an **unignored** test file as a cascade neighbor. The existing `getProjectIgnoreMatcher` filtering fixes only the ignored-file half of this gap.

This is not a request for a blanket "never diagnose tests" policy. Primary LSP diagnostics and runners whose `skipTestFiles` policy is explicitly false remain valuable on test files. This issue is specifically about collateral cascade/impact output and the persisted call-graph-impact path.

Adjacent tracking: #576 covers the broader per-edit dispatch/test-file policy assessment; #1070 covers call-graph evidence coverage. Neither tracks this cross-surface cascade/display alignment.

### Ignore policy

- `clients/file-utils.ts:getProjectIgnoreMatcher` is the canonical matcher for `.gitignore`, global config, and project `.pi-lens.json` `ignore` rules, including tracked-file semantics.
- `clients/lens-diagnostics.ts:createCurrentIgnoreFilter` is the canonical cached-diagnostic display filter. It fail-opens on matcher/config errors, retaining findings.
- `clients/dispatch/integration.ts:isIgnoredCascadeNeighbor` is the existing cascade composition point and already uses the same matcher. It also fail-opens (`false` = not ignored) on config-probe errors, as required by the existing #297 policy.
- The LSP workspace walk (`clients/lsp/index.ts:collectWorkspaceDiagnosticFiles`) and the standalone LSP walk (`tools/lsp-diagnostics.ts:projectIgnorePredicate`) use the same matcher; the latter explicitly fail-opens. Explicit `paths` entries intentionally bypass ignore matching and must remain explicit-user-input semantics.

Do not add a cascade-local ignore glob list or a second matcher.

### Test-role policy

- `clients/file-role.ts:detectFileRole` is the structural role classifier used by the review graph and is already available as `DispatchContext.fileRole`.
- `clients/review-graph/builder.ts` applies `detectFileRole(file) !== "test"` at the source-file walk and incremental/per-file graph chokepoints; review-graph versioning rejects older test-containing snapshots. This is the established graph baseline.
[...]

PR #1113, opened 2026-08-06
fix(dispatch,safe-spawn): clear two more uncleared-race-timeout members (closes #1109)

### LINKED ISSUE #1109: Uncleared Promise.race timeout in cascade references (clients/dispatch/integration.ts:~1033) — same class as #1097
## Defect shape (from the #1097 sweep)

An operation is raced against an independently-armed deadline timer, but the losing timer is not cancelled when the operation resolves first — leaving a **ref'd** pending timer that keeps the event loop alive until it fires. Harmless in a long-lived session; a keep-alive leak in a one-shot `pi --print --no-session` process.

## Site

`clients/dispatch/integration.ts` (~line 1033), in the cascade references derivation:

```ts
const refs = await Promise.race([
  lspService.references(normalizedFile, line - 1, column - 1, false),
  new Promise<never>((_, reject) =>
    setTimeout(() => reject(new Error("timeout")), 750),
  ),
]);
```

When `references()` wins (the common case), the 750ms `setTimeout` is orphaned but still pending and ref'd, so it can keep a one-shot process alive up to 750ms per invocation after `agent_settled`.

## Why not fixed in the #1097 PR

`clients/dispatch/integration.ts` was under concurrent modification by another agent when the #1097 sweep ran; per the territory convention it was flagged rather than touched to avoid a merge collision.

## Fix

Store the timer and `clearTimeout` it in a `finally` (or reuse `withDeadline`/`withTimeout` from `clients/deadline-utils.ts`, which already clears on all outcomes). Impact is bounded (750ms) but it is a genuine member of the same class fixed for `getClientForFile`, `quiet-window` heartbeat, and `format-service` in #1097.

apmantza:
## Summary

Fixes #1109: PR #1110's adversarial review flagged two more members of the #1097 uncleared-race-timeout defect class — *an operation raced/bounded by an independently-armed timer whose loser is never cancelled on the operation's resolution* — rather than fixing them in that PR.

### 1. `clients/dispatch/integration.ts` — 750ms LSP-references race

`computeCascadeForFile`'s symbol-level blast-radius upgrade races `lspService.references()` against an inline `setTimeout(() => reject(...), 750)` per changed symbol (up to 3 per file). When `references()` wins — the common case — the losing timer was never stored, so it stayed a REF'D pending timer for the remaining budget. Fixed with the same store-and-`clearTimeout`-in-`finally` pattern the #1097 fix established (`clients/lsp/index.ts`, `clients/quiet-window.ts`, `clients/format-service.ts`): a per-call `let refsTimer` holds the handle, cleared in a `finally` alongside the existing `catch`.

### 2. `clients/safe-spawn.ts:625` — non-Windows kill escalation timer

`safeSpawnAsync`'s non-Windows `killTree` else-branch arms a REF'D 1s SIGTERM→SIGKILL escalation timer that was neither cleared on child exit nor unref'd (the main per-spawn `timeoutId` right next to it was already correctly cleared on both the `close` and `error` paths — that one was untouched). Bounded to 1s and only on kill paths, so it can't reproduce #1097's hang by itself, but it's the same defect shape. Fixed by clearing (not unref'ing — matching the main timeout's existing contract) alongside `timeoutId` in both handlers, via a per-call `let escalationTimer`.

## Tests (fail-then-pass evidence)

- `tests/clients/cascade-compute.test.ts` — new test uses fake timers and mocks a synchronously-resolving `references()`, then asserts `vi.getTimerCount() === 0` after `computeCascadeForFile` resolves. **Fails on pre-fix code** (`expected 1 to be 0`) and passes post-fix. Verified via temporary revert + rebuild + rerun.
- `tests/clients/safe-spawn-kill-escalation-timer.test.ts` (new file) — mocks `process.platform` and `node:child_process`'s `spawn` (mirroring `tests/clients/lsp/kill-process-tree.test.ts`'s established technique for the analogous LSP-teardown escalation timer) so the non-Windows branch is exercised deterministically on **any** host OS, including this Windows dev machine — no vacuous platform-gated pass. Two tests: close-path and error-path. Both **fail on pre-fix code** (`expected false to be true`) and pass post-fix. Verified via temporary revert + rebuild + rerun.

## Sweep re-run (per AGENTS.md bug-class-sweep discipline)

Re-ran #1110's repo-wide `setTimeout`/`setInterval`/`Promise.race` grep, scoped to every file changed since that sweep's commit (04f48fb9) through current master — i.e. #1103, #1110's own merge, and #1111:

```
git diff 04f48fb9..a7b6bda8 -- clients/ tools/ index.ts mcp/ scripts/ \
  | grep -nE "^\+.*setTimeout|^\+.*setInterval|^\+.*Promise\.race"
```

**Verdict: clean.** No new `setTimeout`/`setInterval`/`Promise.race` sites were added in any file touched by those three merges (`clients/source-filter.ts` was the only production file in #1111's diff, and it added neither).

## Blast radius

- `clients/dispatch/integration.ts`'s `computeCascadeForFile` runs on every LSP-diagnosed file edit's cascade computation; the change only affects the rarely-taken changed-symbol LSP-references branch (up to 3 symbols/file) and is behavior-preserving for the race result — the sole observable difference is the orphaned reject-timer no longer lingers after the call resolves.
- `clients/safe-spawn.ts`'s `safeSpawnAsync` is used by ~30 call sites across every analyzer client (biome, ruff, ast-grep, jscpd, knip, security scanners, installers, etc.) — the busiest hot path in the codebase. The change only touches the non-Windows kill-escalation timer (armed only when a spawn is actually killed via abort/timeout/output-cap) and is likewise behavior-preserving: the child is still SIGTERM'd, still escalates to SIGKILL if it doesn't die within 1s, and every existing `safe-spawn-*.test.ts` file stays green.

## Full suite

`PI_LENS_TEST_MAX_WORKERS=4 npx vitest run`, foreground: **455 test files passed (5 skipped), 5257 tests passed (12 skipped) — no failures** (the known `edits.test.ts` occupancy flake did not reproduce this run).


apmantza:
Adversarial review: **merge** — both fixes verified correct instances of the #1097 pattern (member 1: per-iteration timer scoping + intact rejection propagation; member 2: clear correctly placed on child-EXIT, escalation window preserved); fail-then-pass reproduced first-hand; sweep re-run independently confirmed clean over the full diff range. Review also exposed a PRE-EXISTING defect: the SIGKILL escalation action is dead code (`!child.killed` always false after SIGTERM send) — filed as #1114 with the missing escalation-fires test; CHANGELOG amended to record it. #1080 (in progress) inserts at ~:1156, after this PR's hunk — trivial rebase, no semantic conflict.

Issue #784, opened 2026-07-23
scanTruncated reaches lens-engine but is never rendered — project-diagnostics truncation invisible to the agent/user


apmantza:
Found empirically by the #775 Track B harness (PR #780, pinned in `tests/monorepo/diagnostics-truncation.test.ts`).

The `scanTruncated` flag added in #764 (project-diagnostics scanner entry/file budget) propagates intact to the `lens-engine.ts` seam — but no downstream consumer renders it: neither the diagnostics tool output nor any notify/hint mentions truncation, so a capped scan still reads as a complete clean sweep to the calling agent, which is exactly what #764 intended to prevent.

Fix direction: surface it where the scan result is rendered (one line, e.g. "scan truncated at N files — results are partial; raise maxProjectFiles to scan fully"), consistent with #777's warm-skip notify wording. Flip the pinning test once rendered.

Refs #775.

PR #1115, opened 2026-08-06
Source-walk generated-artifact escape hatch + tool-facing surfacing (closes #1107)

### LINKED ISSUE #1107: Source walk silently excludes real files with generated-looking names (gen.ts) — no observability, no content-probe escape
Found 2026-08-06 while building the #1088 drift-path regression test (recorded in PR #1082's review comment): the source walk **silently drops real source files whose NAMES match the generated-artifact heuristic** — a file literally named `gen.ts` is excluded by `shouldSkipGeneratedOrArtifact`/`isBuildArtifact` name rules (empirically verified: `src/gen.ts` invisible to `collectSourceFilesWithBudgetAsync`; `src/alpha.ts` beside it collected fine).

## Impact
For a code-correctness/intelligence tool this is a silent coverage hole: a user's real `gen.ts` (or other generated-looking names — enumerate the heuristic's patterns) gets NO review-graph node, NO word-index entries, NO project-diagnostics scan, NO call-graph participation — and nothing anywhere says so. The pre-existing `graph-cache.test.ts` fixtures all use `gen.ts`, so several of its tests have been exercising empty-walk builds unknowingly (their assertions still hold — noted in PR #1082's round; the fixtures should still be renamed or made deliberate).

## Class
Name-based skip heuristics that lack (a) observability and (b) a content-probe escape hatch. Sweep all of them: generated-name patterns, artifact-directory names, `detectFileRole` test/vendor classification, `isBuildArtifact` sibling probes — for each, answer: can a REAL source file plausibly match, and what tells the user it was skipped?

## Fix direction
1. **Observability first** (cheap, ship soon): the walk counts and logs name-heuristic skips (per-directory rollup, existing latency/scan logging channel); `module_report`/project scans surface "N files excluded by generated-name heuristics" so the hole is at least visible.
2. **Content-probe confirmation** for name-only matches (the sibling-probe machinery exists — `createArtifactProbeCache`): a generated-looking NAME with no other artifact evidence (no sibling source twin, no generated header in first 4KB, git-tracked) should be KEPT or at minimum logged loudly.
3. Rename the `gen.ts` fixtures in graph-cache.test.ts (or make the empty-walk usage explicit) so the tests say what they exercise.

Refs #1090-era audit wave; discovered in PR #1082 verification.

apmantza:
## Summary

Phase 2 of #1107 (phase 1: #1111, file-level skip counters + latency-log rollup). Three items:

1. **Directory-level skip counting.** `shouldRecurseIntoDir`'s `isGeneratedArtifactDirectoryName` branch (`clients/source-walker.ts`) pruned whole `generated/`/`codegen/` directories with zero counting. It now takes an optional `onGeneratedDirSkip` callback, threaded through `classifyEntry` (`clients/source-filter.ts`) into a new `SourceCollectionResult.generatedDirSkips` counter — **one event per pruned DIRECTORY, never per file inside it** (enumerating the contents would defeat the point of pruning at the directory level). The `SourceCollectionResult` docblock's former "KNOWN GAP" note is now the real field doc. Folded into the existing `source_walk_skip_summary` latency-log line.

2. **Tool-facing surfacing.** `ProjectDiagnosticsSnapshot` gains `generatedFileSkips`/`generatedDirSkips` (present only on a walk that actually ran, same convention as `scanTruncated`/`entryBudgetExceeded`). `clients/lens-engine.ts`'s new `generatedSkipNotice` renders them into a one-line notice mirroring `scanTruncationNotice`'s (#784) style — wired into both `pilens_project_scan` (`mcp/server.ts`) and `lens_diagnostics` (`tools/lens-diagnostics.ts`). **`module_report` seam decision:** it has no walk-backed section of its own — it reads the cached review graph, built asynchronously elsewhere by `getGraphSourceFiles` — so threading these counts through it would mean either inventing a new walk or reaching into the graph-build/persistence schema (a materially larger, riskier change). Per the issue's explicit fallback, this PR surfaces in the project-diagnostics path only.

3. **Content-probe escape hatch (the actual behavior change).** A file matching a generated-artifact NAME pattern but with no other evidence is now **kept** instead of silently dropped. `clients/generated-artifacts.ts` splits its path check into:
   - **STRONG** evidence — a generated directory segment, a lockfile — always conclusive, never rescued.
   - **WEAK** evidence — the filename-regex patterns alone (`gen.ts`, `foo_generated.go`, `bar.min.js`, …).

   `classifyGeneratedOrArtifact` now requires a WEAK match to be corroborated, cheapest-first: (a) a higher-precedence source sibling (already checked upstream by every walk-driven caller via `findSourceSibling`/`isBuildArtifact`), then (b) a generated-code header in the first 4 KB. With **neither**, the file is kept and counted under a new `generatedNameOverrides` counter (also in the rollup log line), so the heuristic's rescues stay observable, not just its skips.

## Gating decision

**Shipped ungated.** The repo's `LENS_FLAGS`/`.pi-lens.json` toggle registry (`clients/lens-config.ts`) is sized for whole-subsystem behavior (`lsp`, `tests`, `delta`, `guard`, `opengrep`, `readGuard`) — a `{ enabled?: boolean }` section per subsystem with a `--no-X` CLI flag. Adding a registry entry for this narrow heuristic refinement would be disproportionate machinery, and `SourceCollectionOptions.includeGenerated` already gives callers a full opt-out of the entire generated-artifact filter (not just the escape hatch) if they need the old behavior back. Documented in code and CHANGELOG.

## Blast radius / invariants

This changes walk output for every consumer that funnels through `collectSourceFiles(WithBudget)(Async)` — review-graph builder, project-diagnostics scanner, word index, `file-role.ts`'s dispatch role detection, call-graph indirectly via the review graph.

- **A repo with a real, hand-written `gen.ts`** (or any WEAK-name-only match with no corroborating evidence) now **gains coverage** — intended, this is the fix.
- **A repo with a headerless generated file that has no source twin** (e.g. a minified bundle with no banner) also gains coverage — accepted per the issue: a false-KEEP (one extra file an agent must itself judge) is strictly less harmful than the prior silent false-DROP (a real file invisibly never analyzed at all).
- **Never silently skipped without evidence, still and always:** lockfiles (`LOCKFILE_NAMES`), files under a generated-artifact directory segment (STRONG path match), declaration stubs when `includeDeclarationFiles` is not requested, and anything the ignore-matcher or extension filter already excludes.
- **Can still be silently skipped** (unchanged, out of scope): a WEAK name match confirmed by either evidence check (sibling or header) — this is the intended "generated" verdict, not a hole.
- `generatedOrArtifactSkips`/`buildArtifactSkips`/`generatedDirSkips`/`generatedNameOverrides` are always additive/optional on `SourceCollectionResult` — existing manual literals (e.g. `project-diagnostics/scanner.ts`'s caller-supplied-`files` branch) keep compiling unchanged, `undefined` meaning "no walk happened."

## Test plan

- [x] `npm run build` (tsc, clean) before every test run
- [x] Directory-skip counting: one `generated/` dir with 3 files inside → 1 dir skip, 0 file skips (`tests/source-filter-skip-observability.test.ts`)
- [x] Escape-hatch matrix: gen-named + source-twin → skipped; gen-named + generated header → skipped; gen-named + neither → **kept**, counted under `generatedNameOverrides`; lockfile → always skipped
- [x] Fail-then-pass proof documented inline for the gen-named+neither case (pre-#1107-phase-2: dropped with `generatedOrArtifactSkips: 1`; post: kept with `generatedNameOverrides: 1`)
- [x] Sync/async agreement asserted for both the dir-skip and escape-hatch cases
- [x] Two pre-existing `tests/source-filter.test.ts` fixtures (`api.pb.go`, `bundle.min.js`) updated with inline rationale for the new evidence requirement
- [x] New `tests/monorepo/diagnostics-generated-skips.test.ts`: snapshot surfacing end-to-end (absent/populated/rendered/never-populated-on-explicit-files)
- [x] Phase-1 tests stay green (`tests/source-filter-skip-observability.test.ts`'s original three cases, adjusted only where item 3 deliberately changes the fixture's own scenario)
- [x] Full suite: **449 files / 5236 tests passed, 0 failed** after two flake fixes: `tests/scripts/changelog.test.ts` (my own CHANGELOG entry needed a title that closes `**` on one line — fixed) and `tests/mcp/module-report.smoke.test.ts` (WASM-zone-memory OOM crash-cascade under full-suite parallelism, a known flake per repo memory — passed clean in isolation, 3/3)


apmantza:
## Review round 2 — all three blockers addressed

Commits: `4b280fde` (the fixes) + `17757ee1` (merge origin/master, #1113 test-suite-lock — CHANGELOG conflict resolved by keeping both entries).

### Blocking 1 (P1, empirically proven) — notice fired permanently

Fixed by splitting the skip counter into evidence tiers. `generated-artifacts.ts`'s `classifyGeneratedOrArtifactDetailed` now returns `{ verdict, evidence }` for every classification, where `evidence` is `"strong"` (lockfile/generated-dir-segment/minified-bundle/declaration opt-in — no probe involved), `"content"` (a WEAK name match CONFIRMED by a real generated-code header), or `"name-only"` (a WEAK name match trusted with **no** corroborating evidence at all — only reachable when a caller opts out of the header probe; the default project-walk path always enables it, so this is rare-to-zero in practice).

`source-filter.ts` threads this into a new `generatedNameOnlySkips` counter (a subset of `generatedOrArtifactSkips`), and `generatedSkipNotice` now keys off `generatedNameOnlySkips + generatedDirSkips`, **not** the raw `generatedFileSkips` total. Directory pruning stays in the trigger condition — it has no escape hatch at all (contents are never enumerated), so it's always a genuine unverified-content signal, unlike a lockfile.

**Regression proof** (`tests/monorepo/diagnostics-generated-skips.test.ts`): a repo with `package-lock.json` + an ambient `.d.ts` now shows `generatedFileSkips: 2` (still fully observable) but `generatedNameOnlySkips: undefined` and **no notice** — the exact scenario the review called out. A companion test confirms a content-confirmed `gen.ts` match also stays silent. The directory-pruning case is kept as the "does trigger" retained case.

### Blocking 2 (P2) — unactionable advice

Chose **(a)**: threaded `includeGenerated` through `ProjectDiagnosticsScanOptions` → `scanProjectDiagnostics`'s walk call → `projectScan` → both `pilens_project_scan` (`mcp/server.ts`, new `includeGenerated` tool param) and `lens_diagnostics` mode=full (`tools/lens-diagnostics.ts`, new `includeGenerated` tool param). Default `false` (no behavior change unless requested). The notice text now names the real parameter instead of a nonexistent one. Chose (a) over (b) because the diff stayed small — five call sites, all additive/optional — and it makes the escape-hatch story actually testable end-to-end (see the new `includeGenerated: true` test in the same file).

### Blocking 3 (P2, maintainer decision) — min/bundle/chunk carve-out

`MINIFIED_BUNDLE_FILE_PATTERNS` (`.min.js`/`.min.css`, `.bundle.*`, `.chunk.*`) moved out of the WEAK escape-hatch-eligible set into `hasStrongGeneratedArtifactPath`, alongside lockfiles — always skipped, never rescued. Both dead legs are documented inline: the header probe never confirms (minifiers strip banners), and the sibling probe checks the wrong stem (`app.min.ts`, never `app.js`), so under the old WEAK-tier classification this class would have been a **permanent** override, not an occasional rescue.

**Fail-then-pass proof** (`tests/source-filter-skip-observability.test.ts`): `app.min.js` next to `app.js` — pre-fix this was kept and counted under `generatedNameOverrides` (the reviewer's empirical finding); post-fix it's skipped unconditionally, counted under the STRONG-tier `generatedOrArtifactSkips`, never as an override. A companion test covers `vendor.bundle.js`/`runtime.chunk.js`. The pre-existing `tests/source-filter.test.ts` fixture that had (incorrectly) asserted `bundle.min.js` was kept is reverted to assert it's skipped.

### Verification

- `npm run build` + `npm run lint` (tsc-based) — both clean
- Touched test files run individually — all green, including the new fail-then-pass proofs
- Full suite in the foreground (`PI_LENS_TEST_MAX_WORKERS=4 npm test`, now routed through #1112's `with-test-lock.mjs` wrapper): **456 files / 5283 tests passed, 0 failed, 12 skipped** (2 unhandled WASM-zone-memory OOM worker-crash errors during the run — the same known crash-cascade flake noted in repo memory, non-fatal to the pass/fail counts)

apmantza:
Round-2 verification: **merge** — notice tiering fixed at the real seam (lockfile+.d.ts fixture: raw counters intact, notice null; generated/ dir fixture: notice fires), includeGenerated threaded end-to-end through all five sites incl. the mcp schema and verified empirically both ways, min/bundle carve-out proven fail-then-pass (vendor.bundle.js indexed pre-fix, skipped-as-STRONG post-fix), merge hygiene clean. One P3 doc nit for a follow-up: lens_diagnostics includeGenerated only takes effect with refreshRunners cheap/all — the param description should say so.

PR #1117, opened 2026-08-06
FEAT: Added Terragrunt Linting

AngriestBird:
## Summary

Extends the existing terraform linting and configuration to also now include terragrunt when there is a terragrunt.hcl and root.hcl in the project root.

Scope grew during review: clearing the SonarCloud quality gate turned up a hand-copied filename list in six places, and the follow-up test run turned up an `npm run test:unit` script that had never worked.

### Changes

- **Terragrunt linting and formatting** — `terragrunt.hcl` and `root.hcl` now
	get a dedicated `terragrunt` file kind (filename-detected only; a plain
	`.hcl` file stays unmapped). New `terragrunt` dispatch runner wraps
	`terragrunt hcl validate --json`, and a `terragrunt-hcl` formatter wraps
	`terragrunt hcl fmt`. Terragrunt has no LSP, so it's runner/formatter only.
	The `terragrunt` binary installs the same way as tflint (GitHub release,
	bare per-platform binaries). Also added `.tf` (Terraform) to the
	`trivy-config` runner's `appliesTo`, gated behind the existing
	`trivy.enabled` opt-in — terragrunt itself stays excluded from trivy, since
	trivy evaluates the Terraform language, not terragrunt config.

- **Single source of truth for the entrypoint filenames** —
	`["terragrunt.hcl", "root.hcl"]` was hand-copied in six places: kind
	detection, the linter policy, the formatter policy, the formatter's own
	`filenames`, and both project-marker maps. Miss one and that filename
	silently loses its linting or formatting with nothing failing. They now come
	from one exported `TERRAGRUNT_FILENAMES` in `clients/file-kinds.ts`,
	following the `DOTNET_*_ROOT_MARKERS` precedent already in that file, and
	`tests/clients/terragrunt-filenames.test.ts` derives every case from that
	list so a call site that drifts back to a literal fails CI.

### SonarCloud quality gate

The gate failed on new-code duplication (16.4% against a 3% threshold) plus one
CRITICAL `typescript:S3776`. All three findings are fixed:

- `clients/installer/index.ts` — the terragrunt release entry hand-rolled the
	platform ladder, so 27 of its 29 new lines duplicated tflint's entry directly
	above it. It now uses the existing `archAssetMatch()` table helper. That
	helper didn't guard non-x64/arm64 arches (it handed back the x64 asset for
	`ia32`), so it gains that guard to preserve the entry's `ia32 → undefined`
	behavior — which also stops **gitleaks** and **trivy**, its two other users,
	selecting an unrunnable x64 binary on an exotic arch.
- `clients/dispatch/runners/terragrunt.ts` — three identical skip results
	collapse to one `SKIPPED` constant, shortening the preamble the runner shared
	verbatim with `hadolint.ts` and `htmlhint.ts` below Sonar's 10-line /
	100-token duplication minimum.
- `parseTerragruntOutput` — extracted `toRawDiagnostics()`, taking cognitive
	complexity from 16 to ~11.

That took duplication to 4.2%, still over the bar, and the remainder turned out
to be one 24-line block: tflint's `assetMatch` ladder plus the entry closers
plus the terragrunt entry's header, matching terraform-ls's ladder plus closers
plus zls's header. tflint and terraform-ls carry **byte-identical** ladders —
same asset strings, not merely the same shape — so the new entry was duplicated
purely by sitting where zls sits relative to a repeated ladder. Both now call
`archAssetMatch(OS_ARCH_ZIP_ASSETS)` against one hoisted table, removing 18
lines of verbatim-duplicated data.

Gate now passes: **new-code duplication 0.6%** (threshold 3%), zero open issues,
all four ratings at 1.

### Unrelated fix picked up along the way

`npm run test:unit` had never run. It passed `--ignore`, which is not a vitest
option, so every invocation died with `CACError: Unknown option --ignore` before
a single test loaded. CI never caught it because CI runs `npm test`. Renaming to
the real flag isn't enough either: each project in `vitest.config.ts` sets its
own `exclude`, which *replaces* the root/CLI value rather than merging, so a
command-line `--exclude` is silently dropped (confirmed with `vitest list` —
both integration files still resolved). The exclusion moved into the config,
keyed off the script name npm exports, which survives the `with-test-lock`
wrapper identically on every OS unlike an inline `FOO=1` prefix. `test:unit` is
now 457 files, `test:integration` the other 2, `npm test` all 459.


apmantza:
thanks @AngriestBird. looking at it now

apmantza:
@AngriestBird I pushed two small follow-up commits to your branch (thanks for keeping `maintainerCanModify` on) — this adds an empirical layer to the review rather than asking you to do another round. I downloaded a real terragrunt **v1.1.2** binary (the same GitHub release asset your installer targets, win-x64) and exercised the CLI in a scratch dir outside vitest. Summary:

**Verified and kept as-is**
- `terragrunt hcl fmt --file <f>` — real flag, formats in place (exit 0). Your formatter command is correct.
- Path-based diagnostic attribution (resolving `range.filename` and comparing via `pathsEqual`) — exactly right: terragrunt emits an **absolute** `range.filename`, so your parent-`find_in_parent_folders()` reasoning holds.
- The `result.error && !result.stdout` → SKIPPED guard correctly catches an older binary that predates the `hcl` command group (it errors on stderr with empty stdout).

**One real bug found + fixed** (`clients/dispatch/runners/terragrunt.ts`)
- `--filter=<basename>` turned out to be a tflint-ism. terragrunt's `--filter` takes **component** filter-syntax, not a filename — `--filter=terragrunt.hcl` matches zero components, so terragrunt validated **nothing** and returned exit 0 / empty stdout *even for a broken unit* (confirmed against a deliberately-invalid fixture). In other words the runner would never have surfaced a single real finding. I removed the flag; the unit is validated at its directory and findings are attributed to the edited file (your existing path logic). Added a regression test that pins "no `--filter` argument."

**Observed contract, now encoded in a docblock + fixtures** (captured from v1.1.2):

| case | exit | stdout |
|---|---|---|
| clean unit | 0 | **empty** (not `[]`) |
| findings | 1 | flat JSON array (+ stderr log line) |
| unknown command (old binary) | non-zero | empty (error on stderr) |

The JSON is a **flat array** with **string** `severity` ("error") and an **absolute** `range.filename` — so of the parser's two hedged branches, the flat-array/string-severity one is the real shape. I updated the parser docblock to match and relabelled the nested `invalid_files`/numeric-severity branch as a retained tolerant fallback (not observed on v1.1.2). The `hcl validate`/`hcl fmt` group requires the redesigned CLI (~v0.75+; legacy `hclvalidate`/`hclfmt` were removed in v1.0) — noted in the runner docblock.

**Two doc P3s**
- Honest one-liner about the `root.hcl` misdetection edge (a non-terragrunt `root.hcl` gets generic HCL canonicalization / validation — fails soft) in both the formatter policy and the runner docblock.
- Moved the "Bare-binary GitHub release assets skip signature siblings" CHANGELOG entry from Added to Fixed.

All affected test files pass (161 across runner/formatters/file-kinds/filenames/smoke), and I confirmed the invocation-argv tests fail on the pre-fix runner and pass on the fix. Nothing merged — your call.

AngriestBird:
Changes looked good to me. I extended the tflint for terraform and then also fixed some more minor issues with the system and then added TFLINT_CONFIG_FILE for terraform work. 

I believe this should be good to go now and should be in a much better spot overall.

Issue #1118, opened 2026-08-06
rule-cache disk cache: mtime+size signature can replay a stale compiled rule set (same size+mtime edit)


apmantza:
Deferred member from the #1105 mtime-only cache-freshness sweep.

## The gap
`clients/cache/rule-cache.ts`'s `computeRuleHash` fingerprints the compiled
tree-sitter rule query set over rule-file **metadata** — `${file}:${stat.mtimeMs}:${stat.size}` — not file **content**:

```ts
hash.update(`${file}:${stat.mtimeMs}:${stat.size}`);
```

This is the review-graph *first-filter* standard (`size:mtimeMs`) but it lacks
the content-hash CONFIRM step the gold standard pairs it with. A rule-file edit
that preserves BOTH mtime and byte size (git checkout timestamp restoration + a
same-length rule tweak; a formatter/rewrite preserving mtime) replays a STALE
compiled rule set from `<language>-rules-v6.json`. Because this is a **persisted
disk cache**, the stale set is also re-persisted under the fresh-looking
fingerprint, poisoning every future process (the exact failure mode the #878
`force` reload was added to prevent, one axis over).

The composed `ruleFiles` (via `ruleFilesForLanguage`) include **project-local,
mutable** rule files — the ones a developer actually edits — so this is not
bundled-only.

## Why it wasn't fixed inline in #1105
The naïve fix (hash file CONTENT) cannot be applied unconditionally: `RuleCache.get`
runs on the **per-edit tree-sitter runner hot path**, and `computeRuleHash`
already stats every rule file in the effective set. There are **~705** bundled
`.yml`/`.yaml` rule files; reading+hashing the per-language subset on every edit
is exactly the unconditional hot-path hashing the event-loop discipline forbids.

## Proposed approach
Mirror the split `ast-grep-napi.ts` already uses (`source.origin === "project"
? loadYamlRulesFresh : loadYamlRules`) and `yaml-rule-parser.ts` already
implements (content-hash `loadYamlRulesFresh` for project trees, dir-mtime
`getCachedRules` for immutable bundled trees):

- Content-hash ONLY the **project-origin** rule files (few) in `computeRuleHash`.
- Keep `mtimeMs:size` for the many **bundled** files (immutable within an install;
  a package upgrade lands them at a new path/mtime anyway).

This requires threading rule-file origin into `computeRuleHash` (today it takes a
flat `string[]`) — a `RuleCache` API/`ruleFilesForLanguage` change, hence a
[...]

Issue #1106, opened 2026-08-06
Class sweep: test fixtures hardcoding cache version literals go vacuous on version bumps


apmantza:
Class generalized from #1089 (found 2026-08-05): the call-graph "rejects..." validator tests hand-wrote `version: 4` cache fixtures; when CACHE_VERSION bumped to 5 the fixtures started failing the version gate BEFORE reaching the validator under test — the ~200-line validator had zero effective coverage and the tests stayed green (vacuous). The class: **any test fixture hand-writing a persisted-format version literal (or shape) that a runtime gate checks before the code-under-test runs** — a version bump silently orphans the test.

Fresh member risk created THIS WEEK: #1096 bumped the workspace-diagnostics cache v1→v2 (its own tests were updated, but any OTHER test seeding v1-shaped entries now exercises the reject-and-rescan path instead of what it claims to test).

## Sweep
1. Grep tests/ for hardcoded `version:` (and `"v<digit>"` schema tags) inside fixture objects written to caches/snapshots: call-graph cache (done, #1082), workspace-diagnostics cache v2, review-graph persisted snapshot ("v8"), project-snapshot, word-index format, session-state/PersistedWidgetState, probe-cache, rule-cache.
2. For each: does the fixture derive the version from the EXPORTED runtime constant (safe) or hardcode it (member)? Convert members to import the constant — single source of truth so bumps can't orphan them.
3. Stronger variant where cheap: a canary assertion that the fixture actually REACHES the code-under-test (e.g. assert the loader got past the version gate — the #1082 de-vacuation pattern: prove non-vacuous by neutralizing the guarded check once during development).

Refs #1089, #1082, #1096.

apmantza:
Recorded before PR #1116 closes this issue (the #1104 lesson — deferrals need a home that survives the close). **Remaining items from the sweep, verified by the PR's adversarial review:**

1. **session-state `STATE_VERSION` reject path has NO test at all** (not a vacuous fixture — a coverage gap): `loadSessionState`'s `state?.version !== STATE_VERSION` rejection (clients/session-state-store.ts:150) is unexercised; the test file round-trips only through real writers. A future format bump's migration/rejection behavior is unpinned.
2. **One unpinned same-class sibling**: tests/clients/cache/rule-cache.test.ts:255's deliberate `raw.version = "v2"` reject-path override lacks the adjacent `expect(...).not.toBe("v2")` pin the other six sites got.
[...]

Issue #1105, opened 2026-08-06
Class sweep: mtime-only cache freshness beyond diagnostics (word-index, probe-cache, rule-cache, snapshot seams)


apmantza:
Class sweep deferred from the 2026-08-06 audit wave. The #1092→#1095/#1096 arc fixed the "freshness stamp doesn't cover the data's real dependency" class for LSP diagnostics (mtime/TTL proxies → content binding). The SAME mtime-only validity assumption underpins several other persisted/derived caches — each is a candidate for stale replay (content changed, mtime not bumped: git checkout timestamp restoration, formatters preserving mtime, same-second writes) or spurious invalidation (mtime bumped, content identical — cheap but wasteful).

## Sweep targets (verify each: what invalidates it, and can content diverge without that trigger?)
- **word-index**: persisted with per-file mtimes in the snapshot (AGENTS.md); refresh keyed on stale/new/deleted by mtime. Member risk: stale identifiers served to symbol_search after mtime-preserving content changes.
- **project-snapshot / project-changes seq**: seq-stamped — verify the seq source covers all write paths (external edits bypass?).
- **installer probe-cache.json**: what invalidates a cached probe when the tool binary is replaced in place (same path, new version)?
- **rule-cache (clients/cache/rule-cache.ts)** and the sgconfig generation cache: keyed on what; can a rule file edit with preserved mtime replay a stale compiled set?
- **reverse-deps index**: derived from the review graph — verify it's generation-coupled post-#1088 (buildGeneration), not mtime-coupled.
- **TreeCache / tree-sitter parse caches**: keyed how?
- **Gold standard to copy**: review-graph's size:mtimeMs signature + content-hash CONFIRM step (confirmContentChanged) — mtime as a cheap first filter, hash as the decider; and #1096's contentHash binding.

## Deliverable
Per cache: verdict (safe / member / already-hardened) with the invalidation trace; fix contained members (add a content-confirm step on the mtime-match path, following the review-graph pattern); file follow-ups for big ones; record the sweep coverage in the closing PR per AGENTS.md.

Refs #1095, #1088, #1020 (the class's history).

Issue #890, opened 2026-07-29
TreeCache mtime check causes false misses on content-identical trees; cache is FIFO not LRU


apmantza:
@-

apmantza:
Fixed by #909 (merged): cache hits are now hash-authoritative (save-without-change refreshes metadata instead of invalidating) and eviction is true LRU via re-insert-on-hit. The PR body template wasn't filled in so GitHub didn't auto-close this.

PR #1119, opened 2026-08-06
fix(project-intelligence): mtime-only cache-freshness sweep — bind word-index + dep-checker to size (refs #1105)

apmantza:
Closes the audit wave item #1105: the "mtime unchanged ≠ content unchanged"
class beyond LSP diagnostics (the #1092→#1096 arc already fixed diagnostics).
mtime-preserving content changes come from git checkout timestamp restoration,
formatters that preserve mtime, and same-second/same-clock writes.

## Sweep verdict table (the centerpiece)

| Cache | File | Validity keyed on | Verdict | Evidence |
|---|---|---|---|---|
| **word index** (refresh) | `clients/word-index.ts` | per-file `mtimeMs` only → **now `mtime`+`size`** | **MEMBER — FIXED** | gate was `fileMtimes.get(file) !== mtimeMs` (refresh L495); mtime-preserving edit served stale identifiers to symbol_search |
| **dependency checker** | `clients/dependency-checker.ts` | `timestamp >= mtimeMs` → **now `+ size`** | **MEMBER — FIXED** | `importsChanged` fast path (L279) skipped madge circular-dep re-check on mtime-preserving import edit |
| **rule-cache (disk)** | `clients/cache/rule-cache.ts` | `file:mtimeMs:size` metadata, **no content hash** | **MEMBER — FOLLOW-UP #1118** | same-size+same-mtime rule edit replays stale compiled set & re-persists it; content-hash confirm can't be added to the ~705-file per-edit hot path unconditionally |
| project-snapshot / seq | `clients/project-snapshot.ts` | `seq` logical clock (`isProjectSnapshotFresh`) | **SAFE** | seq is a logical clock for pi-lens-observed edits; external edits reconciled by the review graph's `size:mtimeMs` + `confirmContentChanged` diff on next scan (pi-lens has no fs-watcher by design) |
| snapshot parse cache | `clients/project-snapshot.ts` | `mtimeMs` **AND** `size` | **ALREADY-HARDENED** | L425 gates on both; residual documented as fail-open/self-healing (in-process, single-writer + authoritative map) |
| installer probe-cache | `clients/installer/index.ts` | `path` exists + `mtimeMs` + 24h TTL | **SAFE (by design)** | validates PATH RESIDENCE, not tool identity/version; in-place replace bumps mtime in the common case; version identity is a separate mechanism (#589 drift, deliberately scoped) |
| sgconfig baseline | `clients/sgconfig.ts` | **content** SHA-256 digest | **ALREADY-HARDENED** | `snapshotRuleSource` hashes `fs.readFileSync(file)` content (L197-204), not metadata |
| reverse-deps index | `clients/reverse-deps.ts` | snapshot `seq` / graph generation | **SAFE** | `loadReverseDependencyIndexFromSnapshot` rejects on `seq !== currentProjectSeq`; derived from the content-hardened (#202) graph, not mtime-coupled |
| TreeCache | `clients/tree-sitter-cache.ts` | **content hash** authoritative | **ALREADY-HARDENED** | hash decides hits; mtime only refreshes metadata on a hash-confirmed hit (#890); content-less path is test-only |
| yaml-rule-parser (bundled) | `clients/dispatch/runners/yaml-rule-parser.ts` | dir `mtimeMs` (`getCachedRules`) | **SAFE (by design)** | mtime variant is **bundled-only**; mutable **project** rules use the content-hashed `loadYamlRulesFresh` (`ast-grep-napi.ts:362` origin split) |
| project-diagnostics snapshot | `clients/project-diagnostics/cache.ts` | `mtimeMs > scannedAt` drop | **OUT OF SCOPE** | diagnostics domain — the #1095/#1096 arc this sweep is explicitly "beyond"; paired with `workspace-diagnostics-cache`'s contentHash |

## Fixes (contained members)

Both members fixed with the review-graph pattern — mtime as the cheap first
filter, **byte size as a free second axis** (the stat is already run). Neither
adds any I/O to a hot path. The same-mtime-AND-same-size residual is left as the
review graph's own accepted residual.

**word-index** (`word-index.ts`): stores per-file `fileSizes` (parallel to
`fileMtimes`), re-reads on `mtime OR size` mismatch, serializes `fileSizes` as an
optional array (pre-#1105 snapshot → one self-healing full re-read, no version bump).

**dependency-checker** (`dependency-checker.ts`): `importsChanged` fast path now
also requires `cached.size === stat.size`.

### Fail-then-pass evidence
Both regression tests are NEW files (no existing fixtures touched, per the #1106
coordination). Each preserves mtime by re-pinning an identical fixed `Date`
(natural `utimesSync` round-trips lose sub-ms precision) so the mtime axis matches
byte-for-byte and **size alone** must trigger the re-read.

- `tests/clients/word-index-mtime-preserved-freshness.test.ts` (2 cases: direct + serialize→deserialize round-trip). Reverting the `size` clause → `refreshed` is `0`, stale identifier lingers (verified).
- `tests/clients/dependency-checker-imports-freshness.test.ts`. Reverting the `size` clause → `importsChanged` returns `false` on the edited imports (verified).

## Follow-up filed
- **#1118** — rule-cache disk cache (content-hash the project-origin rule files only; keep mtime+size for the ~705 bundled files to protect the per-edit hot path).

## Test result
`PI_LENS_TEST_MAX_WORKERS=4 npm test`: **5247 passed / 12 skipped**, plus one
`project-diagnostics/scanner.test.ts` timeout that is the known grammar-heavy-pool
load flake — it **passes in isolation** (9.56s) and is unrelated to these changes
(word-index/dependency-checker don't touch project-diagnostics).


apmantza:
Review round 1 (P2-1 addressed): added the backward-compat regression the reviewer asked for. The compat claim "a pre-#1105 snapshot lacking fileSizes forces one self-healing full re-read" is enforced by two independent guards (deserialize's presence+parallel-length check, and the refresh gate's `?? -1`) but was previously asserted only in comments.

`tests/clients/word-index-mtime-preserved-freshness.test.ts` gains two cases:
- v2 snapshot with `fileSizes` omitted → refresh re-reads every file (`refreshed == fileCount`) with NO content change; a second refresh is a no-op (heal happens exactly once).
- length-mismatched `fileSizes` array → rejected wholesale (empty map), same full re-read (never positionally trusted).

Red/green evidence: green as committed; neutralizing the gate's `?? -1` to `?? size` fails both new cases (`refreshed=0`) while the size-change tests stay green. Commit `c218394f`. word-index.ts is untouched on master (no #1116 conflict); build clean, word-index suites green (65 passed).

PR #1120, opened 2026-08-06
fix(lsp): workspace-edit apply paths — reversed same-position inserts, win32 casing, versioned-edit rejection, out-of-range clamping (closes #1085)

### LINKED ISSUE #1085: LSP workspace-edit apply paths: reversed same-position inserts, win32 casing destruction, versioned-edit rejection (post-#1053/#1066 adversarial review)

apmantza:
Closes #1085. Fixes the three empirically-reproduced P1s, the P2, and P3-1 in the LSP workspace-edit apply paths (`clients/lsp/edits.ts` + `clients/path-utils.ts` seams). Each fix has a fail-then-pass regression test (the issue's reproductions, encoded as tests, verified failing on the pre-fix build).

## Defect classes fixed

- **P1-1 — same-position insert order reversed** (`applyWorkspaceEdit`, all encodings). The tie-break sort is correct for exactly one pass, but the pipeline sorted an even number of times and differed by server position-encoding, producing `"aBAbc"` instead of `"aABbc"`.
- **P1-2 — win32 casing destruction on create/rename; case-only rename rejected.** Disk mutations routed through `normalizeFilePath`, which lowercases a nonexistent path tail on win32. Invisible to Linux CI.
- **P1-3 — versioned `TextDocumentEdit`s always rejected on tool apply paths.** Rename `apply:true` and code-action autofix never pass `documentVersions`, but normalization preserved the version, so it failed 100% for version-stamping servers (gopls).
- **P2 — out-of-range positions threw instead of clamping** (LSP 3.17 regression). The `(0,0)-(9999,0)` whole-document sentinel broke.
- **P3-1 — insert-at-replace-boundary spuriously rejected as overlapping** (LSP-legal; folded into the sort rework).

## Invariants

1. **Single application-ordering sort.** The application-order sort happens **exactly once**, at the string-write site (`applyTextEditsToString`). `normalizeTextEditsForContent` returns validated edits in **original array order** (never pre-sorted), so the same-position order cannot flip on an even number of passes. The tie-break is: start descending → end descending (replace before a boundary insert; order-independent → P3-1) → array index descending (reproduces array order for same-range zero-width inserts → P1-1).
2. **Confine on normalized paths, mutate on decoded paths.** URI confinement/validation and all path-keyed maps stay on `normalizeFilePath` output; `fs.*` mutations target the URI's decoded path (`uriToDiskPath`, no map-key normalization) so casing is honored. A case-only rename is recognized (decoded paths differ, normalized key matches) as a legitimate refactor. Report/key paths stay normalized (forward-slash) — unchanged.
3. **Validate-then-strip versions.** `normalizeClientWorkspaceEdit` validates `textDocument.version` against the live map (unchanged) and then strips it (spec `null` = don't check). The server-initiated `workspace/applyEdit` handler does **not** route through here — it applies `params.edit` with the live `documentVersions` — so its real preflight version check is untouched.
4. **Clamp out-of-range, throw only malformed.** Line past EOF → end of document; character past line end → line length. Negative/non-integer, `start > end`, and mid-codepoint boundary positions still throw.

## Seams extended / preserved

- `applyTextEditsToString` / `normalizeTextEditsForContent` / preflight / apply-loop in `clients/lsp/edits.ts` (all text entry paths funnel through the single sort).
- New `uriToDiskPath` primitive in `clients/path-utils.ts` (re-exported via `clients/lsp/path-utils.ts`), consumed only by the disk-mutation sites.
- `normalizeClientWorkspaceEdit` in `clients/lsp/client.ts` (the one place tool-path edits are normalized).
- Preserved: reverse-order application for distinct positions, ordering/resource matrix, failure atomicity, confinement/realpath symlink-escape checks, consistent `normalizeMapKey` keying, utf-8/16/32 boundary rejection, the `workspace/applyEdit` real version check.

## Cross-product test matrix

| Dimension | Cells covered |
|---|---|
| P1-1 same-position insert | `applyTextEditsToString` direct; `applyWorkspaceEdit` via `changes` **and** `documentChanges`; each under **utf-8 / utf-16 / utf-32** (7 cells) → all `"aABbc"` |
| P1-2 casing | create mixed-case; rename→mixed-case; case-only rename — each **FS-probe-guarded** (asserts casing preservation on case-insensitive FS, non-vacuous success on case-sensitive FS) |
| P1-3 versions | versioned edit SUCCEEDS through the tool path (was missing); rejection stays genuine (key vacuity fixed, #1106 class); validation-intact stale rejection at normalize time |
| P2 clamp | whole-document `(0,0)-(9999,0)` sentinel; character-past-line-end append; multi-line line-past-EOF; malformed negative-char still throws; mid-codepoint boundary still throws |
| P3-1 boundary | insert-at-replace-boundary in **both** listing orders → both `"YXc"` |

## Non-goals / deferred

Explicitly deferred on #1085 (documented in CHANGELOG): **P3-2** (renameFile merge `seenExact` drops a valid duplicated zero-width insert), **P3-3** (rename→create→text virtual-model hole), **P3-4** (`version: null` + numeric conflict), **P3-5** (CRLF boundary / bare `\n` into CRLF files — bigger), **P3-6** (`importsChanged` over-invalidation), **P3-7** (rename close-failure reopen as `plaintext` — bigger). Guard findings (#1084) and cross-cutting consistency findings are separate tracking issues.

## Verification

- **Fail-then-pass:** the three source files were reverted to `origin/master`, rebuilt, and the new tests confirmed failing (12 red: all P1-1 matrix cells, P2 clamp, P3-1, all P1-2 casing, P1-3 version-strip), then restored → green.
- **FS-probe on this machine:** case-**insensitive** — so the win32 casing bug reproduced here (`newfile.txt`, `mixedcase.txt`, case-only rename throw) and the fix is exercised on the failing axis.
- **Full suite** (`npm test` through the #1112 suite-lock, `PI_LENS_TEST_MAX_WORKERS=4`): 455 files passed / 5 skipped, 5278 tests passed / 12 skipped, **0 failures**. The 2 reported "errors" are worker-fork crashes (the documented V8/native-addon worker-teardown flake), not assertions. The `edits.test` occupancy guard passed solo and in-suite — the diff never touches `planWorkspaceEdit`, so its budget is unaffected (measured, not assumed).

Rebased onto current master (post #1112/#1113); no conflicts.


apmantza:
## Adversarial-review round addressed (`26639567`)

Both BLOCKING items and the P3-c nit are fixed.

**BLOCKING 1 — CRLF clamp corruption (P2-1).** The new char-past-EOL clamp targeted `line.length`, but `lineTextAt` keeps a trailing `\r`, so the clamp landed *after* the `\r`. It now clamps to the length **before** a trailing `\r`, so the position lands at the CRLF boundary. Fail-then-pass proven: with the one-line fix reverted and rebuilt, the new CRLF cells go red —
- `insert (0,99)` on `"abc\r\ndef"` → `"abc\rX\ndef"` (stray `\r`), fixed → `"abcX\r\ndef"`;
- sentinel replace `(0,0)-(0,999)` on `"abc\r\ndef"` → `"Y\ndef"` (ate the `\r`), fixed → `"Y\r\ndef"`.

The line-past-EOF clamp is left as-is (verified safe — the last split segment carries no `\r`).

**BLOCKING 2 — case-only rename detection (P2-2). Took option (a).** Replaced the `pathsEqual` platform-keyed fold with on-disk **FS identity** (`isSameFsEntry`: same `dev`+`ino` via `lstat`). Rationale for (a) over (b): it stays fully contained (one helper + the existing preflight branch), fixes the macOS APFS/HFS+ false-fail the review flagged, follows the #1024 "probe the FS, don't branch on platform" rule, and **closes the inverse silent-clobber edge** — a case-sensitive FS holding both `foo.txt` and `Foo.txt` now correctly treats the destination as a genuine conflict instead of skipping the exists check. The "must differ" (byte-identical) guard is unchanged.

**P3-c.** Removed the tautological `expect(caseInsensitive).toBe(caseInsensitive)`; the create-casing test no longer needs the probe (its assertions hold on both FS kinds). The FS probe is still exercised (non-vacuously) by the case-only-rename test.

**Verification.** `npm run build` + `npm run lint` clean. Touched test files green (116/116). Full suite (foreground, through the #1112 suite-lock, `PI_LENS_TEST_MAX_WORKERS=4`): **0 test failures** (454 files passed / 5 skipped, 5272 tests passed / 11 skipped); the 3 reported "errors" are the documented worker-fork-teardown flake (native-addon workers on a loaded host), not assertions. Rebased tip already on current master (0 behind); no CHANGELOG union needed this round.

apmantza:
## Round-2 review addressed (`0150e15d`)

**BLOCKING (P2) — `isSameFsEntry` clobber on ino-less filesystems: fixed.** The dev+ino comparison had no `ino === 0` guard and used a non-bigint (double) `ino`. On ino-less filesystems (FAT32/exFAT, some SMB redirectors, VirtualBox shared folders — libuv reports `ino: 0`) two DISTINCT files compared `(dev, 0) === (dev, 0)` → misclassified as a case-only alias → destination-exists check skipped → `fs.rename` silently clobbered without `overwrite`. Secondary: a `number` `ino` loses precision on NTFS 64-bit file IDs ≥ 2^53 and could false-equal distinct files.

One fix covers both: `lstat({ bigint: true })`, and the new exported `isSameFsIdentity` returns same **only** when `dev` matches AND both inos are **nonzero** AND equal — anything else is distinct, so the destination-exists check runs (**fail-closed**, matching pre-fix behavior on those filesystems). A genuine case-only rename on an ino-less FS now needs an explicit `overwrite` rather than risking a silent clobber.

**Fail-then-pass:** with the nonzero guard reverted and rebuilt, the new unit test goes **red** (both-ino-0 → aliased → clobber path); restored → **green**. Cases covered: both ino 0 → distinct; one ino 0 → distinct; nonzero-equal same-dev → same; equal ino, different dev → distinct; `> 2^53` BigInt inos compared exactly (both directions).

**Honesty:** CHANGELOG updated — the "closes the inverse silent-clobber edge" claim now spells out the ino-less-FS behavior as fail-closed (error, never clobber), not "case-only renames work everywhere".

**Deferral:** posted the verifier's P3-8 (ordered create-then-case-rename fail-closes due to virtual-overlay-vs-physical-probe divergence; errors-never-clobbers) to #1085.

**Verification:** `npm run build` + `npm run lint` clean; both touched test files green (**117/117**, +1 for the ino-0 guard). Guard-and-test-only change, so no full-suite rerun. Tip on current master (0 behind).

apmantza:
Round-2 verification chain complete: all three P1 fixes + P2 clamp verified in the full review; the two review-found P2s in new code (CRLF clamp corruption, ino-less-FS clobber) fixed with empirical red/green each; deferrals (P3-2..8) recorded on #1085 surviving the close. **Merge on green.**

Issue #297, opened 2026-06-22
bug: Ignoring files not always works


carolitascl:
This is an ouput that shows the problem

```text
pi-lens.json has "ignore": ["**/*.test.ts"] — so pi-lens excludes all test files from its scans. But the lens TypeScript runner is still trying to type-check
 test/plugin/reader.test.ts:46 (which imports countTotal) without having indexed plugin/reader.ts's new export through the same pipeline. The tsc compiler
 resolves it fine; pi-lens's partial view doesn't.
```

PR #1121, opened 2026-08-06
fix(project-intelligence): align cascade neighbors with ignored-file and test-role display policy (closes #1080)

### LINKED ISSUE #1080: Tracking: align cascade neighbors with ignored-file and test-role display policy

apmantza:
## Summary

Closes #1080. The review graph is already tests-free, but cascade **collateral** surfaces re-derive neighbors from OTHER sources that never saw that filter — LSP reference expansion, module-level downstream expansion, the reverse-dependency index, and the passive-diagnostics fallback — so an **unignored** `*.test.*` / `tests/` file could still leak into cascade neighbor output, the formatted impact header/counts, the turn-end call-graph advisory, and the persisted call-graph-impact diagnostics. The existing `getProjectIgnoreMatcher` filtering only closed the **ignored-file** half (#297); this adds the **test-role** half.

Primary LSP diagnostics, test-runner findings, per-runner/auxiliary `skipTestFiles` policy, and the generic project-diagnostics snapshot/delta display are all left untouched.

## Design

One shared predicate — `clients/collateral-test-role.ts:isTestRoleCollateral` — composes the SINGLE existing structural classifier (`clients/file-role.ts:detectFileRole`). **No second matcher, no private test-filename list.** Honesty: a KNOWN `"test"` role is filtered; a classifier failure **RETAINS** the candidate (never a false clean).

**Predicate placement (and why the header can't leak):**
- `computeCascadeForFile` filters `impact.directImporters` / `directCallers` / `neighborFiles` **AFTER** graph neighbors, reverse-deps, LSP reference expansion, and transitive expansion are all merged, and **before** `sortedNeighbors` / `formatImpactCascade`. `formatImpactCascade` reads the `impact` object *verbatim* for its `Direct importers` / `Direct callers` / `Check next` counts and names — filtering the `impact` arrays in place is the only placement that cleans the touch set, the returned `impact`, **and** the header at once. Filtering upstream of `sortedNeighbors` also means a test reference URI is never actively touched solely for cascade diagnostics.
- **Module-level downstream** files enter via `computeImpactCascade`; they are caught here at the **cascade-consumption boundary**, deliberately NOT inside the shared `query.ts` graph query, so other query consumers stay unfiltered (per the issue's explicit DO-NOT).
- `appendFallbackNeighbors` drops an unignored test file too; ignore filtering (#297) stays separate and above.
- `runtime-turn.ts` filters `impact()` results before **both** the advisory text and the persisted delta (the advisory is rendered first).
- `callGraphImpactToProjectDiagnostics` re-applies the predicate at the persistence producer boundary.

## Acceptance criterion → test map

| Criterion | Test |
|---|---|
| Review graph stays tests-free (full + incremental with test in `changedFiles`) | `tests/clients/review-graph/tests-free-graph.test.ts` |
| Cascade graph neighbor: unignored test excluded, source surfaced; header clean | cascade-compute.test.ts › "excludes an UNIGNORED test-role graph neighbour…" |
| LSP references: source + unignored test → only source touched/surfaced/header | cascade-compute.test.ts › "does not surface or touch an LSP reference that points at a test file" |
| Passive fallback: unignored test dropped, source kept, ignored still dropped (#297) | cascade-compute.test.ts › "drops an unignored test file from fallback…" |
| Module-downstream test candidate excluded pre-touch/display | covered by the graph-neighbor test (module-downstream files land in `impact.neighborFiles`, filtered at the same consumption boundary) |
| Call-graph test caller absent from advisory AND persisted delta (call-graph-only + mixed) | runtime-turn-session.test.ts › "excludes a test-file caller from advisory + delta on a call-graph-only turn" / "…on a mixed turn" |
| `callGraphImpactToProjectDiagnostics` excludes test caller, keeps normal | `tests/clients/call-graph-impact-test-role.test.ts` |
| Classifier-positive coverage + fail-open (role unavailable ⇒ retained, no false clean) | `tests/clients/collateral-test-role.test.ts` |
| Generic display keeps ignore filter + stale/deleted reconciliation, no blanket test-drop | unchanged by construction — no edit to `tools/lens-diagnostics.ts` / `cache.ts` (existing suites remain green) |

**Fail-then-pass proof:** every producer-boundary test was verified to FAIL against a neutralized predicate (always-`false`) and pass with the fix.

## Coordination

Branched from `origin/master`. PR #1113 (the #1109 timer-race fix + #1112 test-lock wrapper) merged mid-work; I merged `origin/master` into this branch — its `integration.ts` timer-clear hunk (~:1084) is orthogonal to my impact filter (~:1160) and its new cascade-compute fake-timers test auto-merged with my additions (no conflicts). Re-verified the cascade suite and full suite post-merge. Did not touch `safe-spawn.ts` / `source-filter.ts` / `source-walker.ts` / `lsp/edits.ts`.

## Testing

`npm run build` clean. Full suite via the #1112 test-lock wrapper (`PI_LENS_TEST_MAX_WORKERS=4 npm test`): **458 files passed, 5 skipped, 0 failed** (the trailing "2 errors" are benign vitest-pool worker-teardown noise, no test failure).


apmantza:
Adversarial review: **approve** — criterion-by-criterion audit against #1080's spec passed everything: single-classifier composition (no second matcher/list), correct filter placement (fresh-array reassignment post-append/pre-sort — touch set, returned impact, and header all clean; other computeImpactCascade consumers unfiltered as required), honest fail-open (throwing classifier retains, verified), #297 and #1113 intact, all four new files tracked despite the .gitignore test-*.ts trap, and all 7 producer-boundary fail-then-pass claims independently reproduced (exactly those 7 red under a neutralized predicate). Four P3 notes for the record, none blocking: riskFlags downstream COUNT is computed pre-filter (names never leak); cascade.log telemetry records unfiltered candidates (internal); a pre-fix persisted delta replays for at most one turn (producer-boundary choice, documented); module-downstream matrix cell covered transitively.

apmantza:
## Linux CI fix — fixture-only, no production under-filtering

**Root cause (test-only):** the classifier-positive test asserted a backslash **directory** path — `C:\repo\src\__tests__\sub\foo.ts` — classifies as `test`. That expectation is Windows-only. `detectFileRole` computes `dir = dirname(filePath).replace(/\/g,"/")`, i.e. it runs `path.dirname()` **before** normalizing separators. `path.dirname` is OS-specific: on win32 it splits backslashes (so `\__tests__\` is found), but on POSIX (Linux CI) a backslash-only path has no `/`, so `dirname` returns `.` and the `__tests__` segment is gone → classified `source`. Green on Windows, red on Linux — the AGENTS.md #1024 OS-divergence class.

**Which side was wrong:** the **test fixture**, not production. Changed the fixture to a **name-based** backslash path (`C:\repo\src\foo.test.ts`), which is separator-agnostic on every OS: `.test.` is matched as a substring of the whole basename, so `basename()` returning either `foo.test.ts` (win32) or the entire string (POSIX) both contain `.test.` → `true` on both. Verified by emulating POSIX `path` semantics over every fixture (all pass) and re-running the test file locally.

**Production question (explicit):** **No** — `isTestRoleCollateral` / the #1080 filter did **not** under-filter on Linux. Every production input to the predicate is native-separator per-OS: `impact.*` neighbor arrays are `normalizeMapKey`-folded (forward slash), fallback `diagPath` keys are `normalizeMapKey`'d, `call-graph-impact` uses `path.resolve`, `runtime-turn` uses `resolveRunnerPath`, and LSP refs go through `fileURLToPath` (native). On Linux those are all `/`-separated, which `detectFileRole` classifies correctly. A foreign-separator (backslash) path only arises by hand-constructing one in a test — it never occurs in single-OS dispatch. So this is purely a Windows-shaped assertion in the test; the feature behaves identically on both OSes.

**Latent note (not fixed here, out of #1080 scope):** `detectFileRole` normalizing separators *after* `dirname`/`basename` means it cannot parse a foreign-separator path's directory. Harmless today (production paths are always native), but a one-line hardening (`.replace(/\/g,"/")` on `filePath` before `dirname`/`basename`) would make the shared classifier robust to foreign-separator inputs. Flagging for a possible follow-up rather than touching the shared classifier in this PR.

Issue #1122, opened 2026-08-06
pi-lens blocks Node event loop for ~290s, causing silent pi-host exits


apmantza:
## Summary

pi-lens is blocking the Node.js event loop for up to **290 seconds (~4.8 minutes)**, which freezes the pi host (TUI becomes completely unresponsive) and manifests as a "silent exit." The `latency.log` `loop_block` phase — pi-lens's own measurement of event-loop stall — captured the escalation on 2026-08-06.

### Smoking gun: escalating `loop_block` durations (`~/.pi-lens/latency.log`)

| time (UTC) | pid | project | blocked |
|---|---|---|---|
| 13:12:15 | 2088 | — | 3.9s |
| 13:17:24 | 20660 | pi-free (old instance) | 5.1s |
| 13:34:06 | 2088 | — | 10.6s |
| **14:38:30** | **20660** | **pi-free (old)** | **290,179ms (4.8 min)** |
| **14:39:37** | **2088** | — | **290,179ms (4.8 min)** |
| 16:03:30 | 35312 | pi-webaio | 4.9s |
| 17:40:35 | 20684 | pi-free (current) | 1.6s |
| 17:43:37 | 20684 | pi-free (current) | 5.2s |

Example log line:

```json
{"type":"phase","filePath":"<pi-lens>","phase":"loop_block","durationMs":290179,"metadata":{"worstSoFar":true,"turnIndex":276},"ts":"2026-08-06T14:38:30.944Z","pid":20660}
```

### Timeline correlation with the silent exit

pid 20660 (the previous pi-free pi-lens instance) was active through 14:15, then hit the 290s block at 14:38:30 and **never appears again**. The next pi-free activity is pid 20684 starting at 17:39:10 (a resumed session). That ~3-hour gap *is* the silent exit — the host froze for ~5 minutes and was killed/restarted.

### 1. Stale extension-ctx bug (`~/.pi-lens/bus-events.log`, 20 occurrences)

pi-lens captures the extension `pi`/`ctx` and reuses it after `ctx.newSession()` / `fork()` / `switchSession()` / `reload()`. Pi throws:

```
This extension ctx is stale after session replacement or reload. Do not use a captured
pi or command ctx after ctx.newSession(), ctx.fork(), ctx.switchSession(), or ctx.reload().
```

Logged as `emit_failed` (caught), so not the direct crash — but it shows pi-lens's lifecycle handling breaks under session replacement, which is exactly when the blocks cluster.

[...]

PR #1124, opened 2026-08-06
fix(project-intelligence): key runtime-tool-result maps through PathKeyedMap (refs #1086 P1)

apmantza:
## Summary

Fixes the P1 finding in #1086: `clients/runtime-tool-result.ts` held three raw path-keyed maps — `inFlightPipelines`, `lastAnalyzedStateByFile`, `debouncedPipelines` — keyed on a `filePath` that is only `path.resolve`d from raw agent-supplied tool input (never case-folded, never realpath'd). This is the #210/#1020/#1025 raw-path-key defect class: "every path-keyed map normalizes at write AND read." A divergent Windows spelling of the same file (case, or an unresolved absolute path an agent supplies verbatim) bypassed the concurrent-state dedupe and debounce coalescing, producing duplicate pipeline runs, doubled LSP dispatch, split telemetry participant counts, and a `flushDebouncedToolResults(filePath)` call that silently missed the pending entry it was meant to flush.

## Class / normalizer choice

- All three maps now wrap `PathKeyedMap` (clients/path-keyed-map.ts, #1025) with `normalizeEphemeralMapKey` — the cheap slash-fold + win32-lowercase normalizer, **no** `realpathSync`. Per AGENTS.md's lifetime-based normalizer guidance, this state is ephemeral/turn-scoped/single-process (cleared at `turn_start`, keys produced by this process's own `path.resolve`), which is exactly `normalizeEphemeralMapKey`'s intended use — not `normalizeMapKey`'s realpath-canonicalizing cost, which is for long-lived state shared across call sites (read-guard, `_fileSeq`).
- `inFlightPipelines`'s composite `${filePath}:${stateHash}` key needed splitting, not folding: the path half must normalize, the stateHash half must NOT (a real content change for the same file must stay a distinct entry). A flat `PathKeyedMap<InFlightPipeline>` keyed by the composite string can't express that split — the normalizer only ever sees the whole string, so it can't fold the path half without also mangling the hash half. It's now `PathKeyedMap<Map<string, InFlightPipeline>>`: normalized path outer, raw stateHash inner. The inner map is pruned back to empty once its last entry is deleted so a file touched once doesn't leave a permanent empty outer entry.

## Sweep

Every read/write/delete/iteration site of all three maps was audited (grepped the file): `flushDebouncedToolResults`'s iteration (`debouncedPipelines.entries()`/`.delete(key)`), `scheduleDebounced`'s coalesce path, and the `inFlightPipelines`/`lastAnalyzedStateByFile` get/set/delete sites in `handleToolResult`. `PathKeyedMap` preserves the last-written display path on iteration, so `flushDebouncedToolResults`'s log lines and the debounce coalesce dbg messages still render the form the caller supplied, not a folded internal key. No other path-keyed map exists in `clients/runtime-tool-result.ts` beyond the three named — this file is now clean of the class.

## Tests (`tests/clients/runtime-tool-result-debounce.test.ts`, fail-then-pass proof each)

Per AGENTS.md's OS-agnostic-tests rule: on Linux, a case-divergent path names a genuinely different file, so a real end-to-end reproduction via the filesystem can't be honest across both dev (Windows) and CI (Linux). Each test below stubs `process.platform` to `"win32"` for the duration (the normalizer reads `process.platform` fresh per call, so this drives its fold branch deterministically regardless of host OS) and uses two REAL files with identical content for the "divergent spelling" pair — on Linux CI they are, correctly, two distinct files on disk; the stub is what makes the MAP treat their paths as one key, exactly as a real case-insensitive Windows host would. This is documented inline as a map-keying test, not a filesystem-aliasing test.

- **Divergent-spelling dedupe** (`inFlightPipelines dedupes concurrent same-state calls for divergently-spelled paths into one pipeline run`): two concurrent `tool_result`s for `Foo.ts`/`foo.ts` (debounce disabled, routing through `inFlightPipelines`) → 1 `runPipeline` call, not 2. **Fails pre-fix** (`expected 1, got 2`), **passes post-fix**.
- **`flushDebouncedToolResults` divergent spelling** (`flushDebouncedToolResults(filePath) flushes a pending entry scheduled under a divergently-spelled path`): schedule under `Bar.ts`, flush addressed by `bar.ts`. Asserted by **elapsed time** (<500ms), not just call count — `getDebounceMs()` clamps to 1000ms, so a silent miss that falls back to the natural timer would still eventually satisfy a call-count-only assertion (a vacuous pass I caught while proving this out: pre-fix, the naive version passed by accident because the natural 1s timer fired within vitest's 5s default test timeout). **Fails pre-fix** (`expected 1122 to be less than 500`), **passes post-fix**.
- **Composite-key hash axis** (`inFlightPipelines composite key keeps distinct entries for the same path with different stateHash`): same path, two back-to-back calls with different file content (different `stateHash`) → 2 distinct `runPipeline` calls, no dedupe log. This is a safety-net regression test (the hash axis wasn't broken pre-fix either) proving the nested-map conversion didn't regress it — passes both before and after.

## Full suite

A local `PI_LENS_TEST_MAX_WORKERS=4 npm test` run completed (exit 0, 453 files / 5169 tests passed, 5/11 skipped) but hit 4 vitest worker-fork OOM crashes mid-run — the machine was independently in active commit-charge exhaustion during this run (issue #1122, unrelated incident), a known flake class per AGENTS.md ("WASM-OOM crash-cascades — re-run isolated"). Per explicit direction during that incident, I did not re-run the full suite locally to avoid adding further memory pressure; the touched test files (`tests/clients/runtime-tool-result-debounce.test.ts`, `tests/clients/runtime-tool-result.test.ts`) plus the directly-adjacent integration suites (`bash-read-guard-integration.test.ts`, `runtime-event-flow.test.ts`, `inline-contract.test.ts`) were all run in isolation, foreground, and are green — that verification, plus CI's own full-suite run on this PR, is the coverage for this change. Flagging this as an exceptional, justified deviation from the "one full suite foreground" discipline, not a shortcut on the changed code itself.

## Test plan
- [x] `npm run build`
- [x] `tests/clients/runtime-tool-result-debounce.test.ts` + `tests/clients/runtime-tool-result.test.ts` (foreground, green, fail-then-pass verified for the two new class-catching tests)
- [x] Adjacent integration suites (`bash-read-guard-integration`, `runtime-event-flow`, `inline-contract`) green
- [x] Full suite ran once locally (green, 5169/5169 passed) before the machine incident; not re-run after per exceptional direction — CI covers this PR


apmantza:
Adversarial review (trace-only per the #1122 memory-pressure protocol; commit ratio 0.91-0.95 at review time): **approve** — normalizer-lifetime choice verified correct (no hydration path exists; lastAnalyzedStateByFile clears every turn_start), inner-map pruning covered on all exit paths incl. crash-catch, display-path preservation intact, same-spelling behavior byte-identical, CI green as the designated suite gate. The one P2 (load-sensitive 500ms elapsed bound — flake bait on a fleet that just lived through commit exhaustion) is fixed in the follow-up commit: deterministic event-order assertion before awaiting the natural timer. P3s noted, no code change: symlink-spelling residual is the accepted lifetime tradeoff; the Linux backslash-fold is inherent to normalizeEphemeralMapKey repo-wide since #1025. Merging when CI is green on the new head.

Issue #1123, opened 2026-08-06
Observability upgrades: loop_block attribution + system-stall tagging, instance health markers, smells self-surfacing, handle-dump debug flag


apmantza:
Umbrella issue from the 2026-08-06 wave's post-mortems (#1097, #1122, the bus-events audit). The finding across all three: pi-lens's observability FOUNDATION is strong — every investigation succeeded because a log already existed — but three targeted upgrades would have turned multi-hour agent expeditions into one-line diagnoses. Not more event streams (bus-events.log sits at 5% of its bound); better attribution, aggregation, and self-surfacing of what's already captured.

Each slice is independently landable and Sonnet-sized.

## 1. loop_block attribution + honesty (the #1122 lesson)
`loop_block` (perf_hooks.monitorEventLoopDelay via clients/event-loop-monitor.ts, read at turn_end) knows THAT the loop stalled, never WHAT ran — so a 10s block becomes a research project. And it is cause-agnostic: OS suspension/paging reports identically to own-code blocking, and the HDR histogram's bucketed precision at large magnitudes means two ~290s system stalls report byte-identical maxMs (a measurement artifact that derailed initial #1122 analysis in both directions).
- (a) Cheap attribution: record the last-started phase/runner (and turnIndex already exists) at block detection; emit it in the loop_block metadata.
- (b) Suspend/pressure awareness: alongside the histogram, track wall-clock vs a per-turn activity heartbeat; when a block's magnitude matches a wall-clock discontinuity affecting the whole process (or exceeds a sanity ceiling, e.g. >60s), TAG the sample (`suspectSystemStall: true`) rather than reporting it as an own-code block. Never discard — tag, so the log stays honest about what it can't attribute.
- (c) Document the HDR bucketing in the emitter's docblock so nobody re-derives conclusions from exact-value equality again.

## 2. Instance-level health (the gap #1122 actually exposed)
instances.json reached count=3 with one 1.37GB RSS instance, and the "silent exit" was reconstructed by hand from a 3-hour log gap.
- (a) On startup, detect prior instances that vanished without deregistering: log "previous instance pid X last seen ~T (RSS Y) exited without shutdown" to sessionstart.log — turns the next silent exit into a one-line diagnosis.
- (b) Periodic per-instance RSS line (existing resource-sampler seam; bounded cadence, e.g. every N turns) into latency.log so memory trajectories exist post-hoc.
[...]

PR #1125, opened 2026-08-07
fix(observability): make loop_block probe system-stall aware (refs #1122, refs #1123 item 1)

apmantza:
## Summary

The `loop_block` latency phase was reporting **machine sleep and paging stalls as pi-lens synchronous blocks**. Issue #1122's headline 290,179 ms "blocks" (and multi-*hour* entries in `latency.log`) are OS freezes, not own-code work. This makes the probe attribute them correctly.

### Diagnosis (evidence in the #1122 comment)
- The probe wraps Node's native `monitorEventLoopDelay`, whose libuv timer-lag is measured against the **monotonic** clock (`uv_hrtime` / `QueryPerformanceCounter`). When the whole process is frozen, the next timer fires late by the entire wall-clock gap and that gap is recorded as a "block".
- Two distinct machine artifacts confirmed against the Windows System event log: (1) a 290,179 ms block lined up exactly with a **14:33:05Z→14:37:55Z Modern Standby window** (Kernel-Power 506/507), reported byte-identically by two independent pids because the HDR histogram quantizes ~290 s into one bucket; (2) a later silent host exit with **zero sleep events** but **twelve Resource-Exhaustion-Detector (2004)** events at 97% commit charge (paging, not sleeping).
- **Comparing wall-vs-monotonic does NOT detect this** — on Windows both advance across Modern Standby (the monotonic delta already equalled the wall gap). This corrects the wall-vs-monotonic approach floated in triage.

### Fix
- **CPU accounting is the discriminator**: a genuine synchronous block of D ms burns ≈ D ms of main-thread CPU, so its window must have consumed ≥ ~D ms of CPU; a frozen/thrashing process consumes ~0 CPU across the gap.
- **Per-turn windowing**: `resetEventLoopMonitor()` is now wired at `turn_end` (its long-intended reset, previously never called in production), so each block is attributable to its turn and the CPU budget is bounded.
- A max above a **20 s floor** that the window's `process.cpuUsage()` delta can't account for is tagged `suspectSystemStall` and **kept out of the genuine-block high-waters** — so a freeze can no longer permanently suppress logging of later real blocks. Sub-floor blocks are never auto-tagged, but `windowCpuMs`/`windowWallMs` are logged so a human can spot a shorter paging stall.
- **Attribution** (#1123 item 1): `loop_block` metadata now carries `lastPhase`/`lastPhaseAt` (the last phase that ran before detection).
- `/lens-health` now reports the worst **genuine** (non-stall) block.

### Tests
- `isSuspendSuspectedBlock` and `getLastLoggedPhase` are pure/seam-testable without a real machine sleep: floor behavior, CPU-accounting, slop, self-attribution exclusion.
- New per-window CPU/wall/`suspectSystemStall` assertions + reset re-baselining in `event-loop-monitor.test.ts`; `index-wiring` health-line assertion updated.

### Test scope
Per #1122 (machine under commit-charge pressure — do not add a full local suite run to the load), only the touched files were run locally (28 tests green): `event-loop-monitor`, `latency-logger`, `index-wiring`. **CI is the full-suite gate.**


apmantza:
## Review round — required fixes applied (head `78a5c058`)

**R1 (P1, auto-close footgun):** reworded `closes #1123 item 1` → **`refs #1123 item 1`** in the commit message, PR title, and CHANGELOG so merging can't auto-close the umbrella issue #1123 (items 2–4 unshipped). Branch history rewritten via amend + force-push (feature branch, no other consumers).

**R2 (P1, DIRTY):** merged `origin/master`; the expected CHANGELOG union conflict was resolved keeping both entries (mine + the #1086 PathKeyedMap entry), markers stripped. Rebuilt + re-ran touched tests green post-merge. (mergeState is now BEHIND, not DIRTY — master moved again; no conflict.)

**R3 (P2, wiring tests):** new `tests/index-loop-block-wiring.test.ts` drives the real `index.ts` turn_end seam and asserts the two facts that silently broke before:
- **(a)** turn_end calls `resetEventLoopMonitor` (the per-turn window). **Fails on master** — reset was never wired: `AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times`.
- **(b)** a `suspectSystemStall` sample does not raise the genuine-block high-water, so a later real 5 s block still logs. **Fails on master** — the stall poisons the high-water and suppresses the follow-up: `expected [...] to have a length of 2 but got 1`.

Both fail-on-master results were captured by temporarily reverting turn_end to master behavior, rebuilding, running, then restoring (no `git stash`).

**R4 (P2, docs):** `windowCpuMs` docblock corrected — `process.cpuUsage()` is **process-wide** (all threads incl. workers), not main-thread; added the false-negative note (a suspend overlapping a worker-CPU-heavy turn can read as genuine). `isSuspendSuspectedBlock` docblock now names the **>20 s blocked-in-syscall ambiguity** honestly (the OneDrive/cloud `readdirSync` class — ~0 CPU, wall advancing — is CPU-indistinguishable from suspend, gets tagged/excluded from health high-waters, but still re-logs every turn with `lastPhase`). Sharper corroboration (magnitude ceiling, `maxMs`-vs-`windowWallMs` ratio) deferred to a #1123 follow-up note, not this PR.

**P3s (all done, quick):** `/lens-health` now shows an "N suspected system stall(s) this session" line; `getLastLoggedPhase` docblock documents its cross-turn staleness (not cleared at boundaries — compare `ts`); the reset test's absolute `< 40 ms` bound is now relative (`after < before/2`) to avoid a GC-pause flake on loaded CI.

Touched-file tests only locally (30 green) under the machine's still-recovering commit pressure — **CI is the full-suite gate** (Lint & type-check already green; unit tests running).

PR #1130, opened 2026-08-07
fix(safe-spawn): make non-Windows SIGTERM->SIGKILL escalation actually fire

### LINKED ISSUE #1114: safe-spawn non-Windows SIGKILL escalation is dead code: !child.killed always false after SIGTERM send
Pre-existing defect exposed by PR #1113's adversarial review (not introduced there — the review verified #1113's timer-clear is correctly placed on child-exit and preserves the escalation window).

## Defect
clients/safe-spawn.ts ~:631-634 (non-Windows killTree branch):
```ts
child.kill("SIGTERM");
escalationTimer = setTimeout(() => {
    if (!child.killed) child.kill("SIGKILL");
}, 1000);
```
Node sets `subprocess.killed = true` when `kill()` successfully SENDS a signal — not when the child dies. After the successful SIGTERM send one line earlier, `!child.killed` is always false, so the SIGKILL escalation NEVER fires: a SIGTERM-ignoring child on Linux is never force-killed. The 1s timer is armed and (post-#1113) properly cleaned, but its action is unreachable.

## Fix direction
Gate on observed death, not send-success: track a `closed`/`exited` flag set by the close/exit handler and escalate `if (!closed)`; or follow the LSP analog `killProcessTree` (clients/lsp/client.ts), which SIGKILLs the process group UNCONDITIONALLY after its window and has exactly the missing test ("non-fast shutdown escalates SIGTERM → SIGKILL", tests/clients/lsp/kill-process-tree.test.ts).

## Required test
The escalation-FIRES case: child ignores SIGTERM (mock never emits close) → SIGKILL sent at 1s. This test fails on current master — it is the proof the guard is dead.

## Class note
Same family as the #1109/#1110 timer work but a distinct shape: "escalation action gated on a flag that can never be true at fire time." Worth one grep for other `.killed` consumers when fixing (clients/, scripts/) — kill-process-tree's unconditional design is the safe pattern.

Refs #1113 (review that found it), #1109, #472 (orphaned-children history).

apmantza:
## Summary

`clients/safe-spawn.ts`'s non-Windows `killTree` branch armed a 1s SIGTERM→SIGKILL escalation timer gated on `if (!child.killed) child.kill("SIGKILL")`. Node sets `ChildProcess#killed = true` the moment `kill()` successfully **sends** a signal — not when the child actually dies — so immediately after the `child.kill("SIGTERM")` one line above, the guard was always false and the SIGKILL escalation could never fire: a SIGTERM-ignoring child on Linux/macOS was never force-killed.

Closes #1114

## Type of change

- [x] Bug fix

## Area

- [x] area:lsp
- [x] area:tests

## Checklist

- [x] The change has tests (regression test for the bug, proven fail-then-pass; existing #1109 leak tests remain green)
- [x] `npm run lint` passes (tsc --project tsconfig.json — clean)
- [ ] `npm test` — **not run locally in full** (memory-tight dev machine; established #1122 precedent). Ran targeted suites locally instead (all green, see below) and am relying on CI for the full suite.
- [x] `npm run build:dist`-equivalent (`npm run build`, `tsc --project tsconfig.build.json`) succeeds
- [x] `package-lock.json` unchanged (no dep changes)
- [ ] `AGENTS.md` — no behavior/convention/invariant documented there changed; not updated
- [x] `CHANGELOG.md` has a `## [Unreleased]` entry for this fix
- [x] Commit subject includes `(closes #1114)`

### Fix choice + rationale

Two directions were offered in the issue: (a) track a `closed`/`exited` flag from the close/exit handler and escalate `if (!closed)`, or (b) follow `killProcessTree`'s unconditional-SIGKILL-after-window design.

Chose **(a)** for `safeSpawnAsync`: it already has a single, well-defined per-call close/error observation point (the `child.on("close", ...)` / `child.on("error", ...)` handlers that PR #1113 already touched to clear `escalationTimer`), so a `closed` flag set synchronously — **before** any `await` — at the top of those handlers composes directly with that existing cleanup instead of introducing a second, unconditional-kill code path. The flag is set before the close handler's `await killPromise`, so even a timer callback that races the close handler's async gap still observes the correct value (JS is single-threaded; the flag is set on the very first synchronous tick of the handler).

`killProcessTree`'s unconditional-SIGKILL design (option b) remains the right shape for its own use case (no persistent per-call handler structure to hang a flag off outside the function itself), so it wasn't ported wholesale — but the sweep below found it had the *identical bug*, just expressed differently, and fixed it in place with the same "gate on observed death" principle rather than switching to a different design.

### Required test — fail-then-pass proof

Added to `tests/clients/safe-spawn-kill-escalation-timer.test.ts`:

> "escalates SIGTERM to SIGKILL at the 1s mark when the child ignores SIGTERM (#1114)"

Child never emits `close`/`error` (ignores SIGTERM); fake-timer-advance past the 1s mark must trigger `child.kill("SIGKILL")`.

- **Pre-fix** (`if (!child.killed) ...`): `AssertionError: expected "vi.fn()" to be called with arguments: [ 'SIGKILL' ]` — proves the guard was dead.
- **Post-fix** (`if (!closed) ...`): passes, along with the 2 pre-existing #1109 escalation-timer-cleanup tests (escalation timer cleared when close/error DOES arrive) — no regression on that leak fix.

```
Test Files  1 passed (1)
     Tests  3 passed (3)
```

### Class sweep

Grepped every `.killed` consumer under `clients/` and `scripts/`. Two siblings of the exact same shape ("escalation action gated on a flag that can never be meaningfully true at fire time") found and fixed in `clients/lsp/client.ts`:

1. **`killProcessTree`'s `fast`-shutdown escalation timer** (`if (!(proc as {killed?}).killed) ...` before a 1.5s SIGKILL): the primary SIGTERM send goes through the raw `process.kill(-pid, …)` process-group call, which never touches `proc.killed` at all — so this guard was either always-true (unconditional SIGKILL after the window, even when the group had already died — the common process-group-kill path) or dead (the direct-child fallback path only, same shape as the safe-spawn bug). Fixed with a real `proc.once("exit", ...)` listener registered once, up front, tracked via an `exited` flag — also reused to replace the redundant/secondary `.killed` check in the neighboring non-fast escalation branch (which was already correctly gated by a real `exitedInTime` signal, so not itself broken, just now consistent).
2. **`createLSPClient`'s `initialize()`-timeout 2s SIGKILL backstop** (`if (!lspProcess.process.killed && ...)`): same always-true shape, since `killProcessTree` above it also only ever signals via `process.kill(-pid, …)`. Switched to `lspProcess.process.exitCode === null`, the process's own real observed-exit signal (no listener needed — `exitCode` is a `ChildProcess` field Node maintains natively).

Audited and left as-is (liveness/status reads, not escalation-action gates, so the same-shape bug doesn't apply):

- `isClientAlive`'s `!state.lspProcess.process.killed` — redundant with `isDestroyed`, which is already set from real exit/close handlers elsewhere.
- `checkProcessAlive`'s informational "was killed" health-check string (not gating an action).
- `clients/lsp/launch.ts`'s post-spawn immediate-failure check (`proc.killed` read synchronously right after spawn, before any `kill()` has ever been called against that process — legitimately `false` at that point).
- `scripts/with-test-lock.mjs`'s `!child.killed` guard inside a `.once`-registered SIGINT/SIGTERM forwarder — this is the *first* kill call in that flow (no prior `child.kill()`), so the read is accurate, not an escalation-after-send check.
- `scripts/smoke-tools.mjs`'s `err?.killed === true` — reading Node's own `execFileSync` timeout-kill flag off a caught error object, informational only.

### Verification

- `npm run build` (tsc --project tsconfig.build.json) — clean
- `npx tsc --project tsconfig.json --noEmit` (lint) — clean
- Targeted local test runs (all green):
  - `tests/clients/safe-spawn-kill-escalation-timer.test.ts` (incl. new #1114 test) — 3/3
  - `tests/clients/safe-spawn-ambient-signal.test.ts`, `safe-spawn-sync-throw.test.ts`, `safe-spawn-resource-usage.test.ts`, `safe-spawn-windows-command.test.ts` — all pass
  - `tests/clients/lsp/kill-process-tree.test.ts` — all pass (incl. the pre-existing "non-fast shutdown escalates SIGTERM → SIGKILL" test)
  - `tests/clients/lsp/client-internals.test.ts`, `spawn-shutdown-race.test.ts`, `service-shutdown-concurrency.test.ts`, `lifecycle.test.ts`, `teardown-logging.test.ts`, `client-wait-timer-cleanup.test.ts` — 91/91 pass
- Full suite intentionally **not** run locally (memory-tight dev machine, #1122 precedent) — deferring to CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,18 +32,50 @@ All notable changes to pi-lens will be documented in this file.
 	call (which never touches `proc.killed`), so the guard was either always-true
 	(unconditional SIGKILL after the window on the common group-kill path) or
 	dead (on the direct-child fallback path) — now tracked via a real `exit`
-	listener set once up front; (2) the `initialize()`-timeout 2s SIGKILL
-	backstop had the identical always-true `!lspProcess.process.killed` guard —
-	switched to `lspProcess.process.exitCode === null`, the process's own
-	observed-exit signal. Other `.killed` reads audited and left as-is because
-	they're liveness/status checks, not escalation-action gates:
-	`isClientAlive`'s `!state.lspProcess.process.killed` (redundant with
-	`isDestroyed`, already set from real exit/close handlers),
-	`checkProcessAlive`'s informational "was killed" health-check string,
-	`launch.ts`'s post-spawn immediate-failure check (`proc.killed` read before
-	any kill was ever sent), `scripts/with-test-lock.mjs`'s `.once`-registered
-	first-forward guard, and `scripts/smoke-tools.mjs`'s read of Node's own
-	`execFileSync` timeout-kill flag on the caught error object.
+	listener set once up front, seeded from the same `exitCode`/`signalCode`
+	pre-check the function's top-of-body early return already uses (a process
+	that was already dead on entry — reachable when `options.processExiting`
+	skips that early return — would otherwise miss its own "exit" event and
+	still draw a redundant group SIGKILL at the escalation window); (2) the
+	`initialize()`-timeout 2s SIGKILL backstop had the identical always-true
+	`!lspProcess.process.killed` guard — switched to `lspProcess.process.exitCode
+	=== null && lspProcess.process.signalCode === null` (both, not `exitCode`
+	alone: a process killed BY a signal — the common case here, since
+	`killProcessTree` above it signals rather than lets the process exit on its
+	own — has `exitCode === null` forever and only `signalCode` set, so
+	`exitCode` alone still re-armed the backstop's kill against an
+	already-dead corpse; harmless in practice since `ChildProcess#kill()` on an
+	exited handle is a swallowed no-op, but not an accurate "still alive"
+	read). Other `.killed` reads audited and left as-is because they're
+	liveness/status checks, not escalation-action gates: `isClientAlive`'s
+	`!state.lspProcess.process.killed` (redundant with `isDestroyed`, already
+	set from real exit/close handlers), `checkProcessAlive`'s informational
+	"was killed" health-check string, `launch.ts`'s post-spawn
+	immediate-failure check (`proc.killed` read before any kill was ever sent),
+	`scripts/with-test-lock.mjs`'s `.once`-registered first-forward guard, and
+	`scripts/smoke-tools.mjs`'s read of Node's own `execFileSync` timeout-kill
+	flag on the caught error object.
+
+	**Adversarial-review follow-up round:** the reviewer ran
+	`kill-process-tree.test.ts` against PRE-fix `client.ts` and it passed 7/7 —
+	the sibling fixes above had ZERO effective test coverage, because the
+	"non-fast shutdown escalates" mock (and the other pre-existing mocks in
+	that file) lacked `once`/never set `killed`, so BOTH the old dead guard and
+	the new fix's guard were vacuously permissive against them (the #1106
+	vacuous-mock class, recurring in mock form: a test's fixture is too weak to
+	distinguish correct from broken behavior, so it passes either way). Fixed
+	by: upgrading that test's mock to be `.once`-capable so it actually
+	exercises the `exited`-flag logic; adding two new `fast`-shutdown tests with
+	a real `.once`-capturing mock proving BOTH directions (exit observed before
+	the 1.5s window → no group SIGKILL; no exit observed → group SIGKILL at the
+	window) — the "no premature SIGKILL" direction fails against the pre-fix
+	`!proc.killed` guard (proven by temporarily reverting the guard and
+	re-running); and adding a real-subprocess POSIX-only test
+	(`tests/clients/lsp/initialize-timeout-backstop.test.ts`, skipped on win32
+	with an explicit reason — killProcessTree's Windows path is
+	`taskkill`-based, not signal-based, and is already covered by the
+	kill-process-tree suite) for the previously fully-untested `initialize()`
+	2s backstop.
 - **mtime-only cache freshness sweep (refs #1105)** — the #1092→#1096 arc bound
 	LSP-diagnostics freshness to real content; this sweep audited the OTHER
 	persisted/derived caches for the same "mtime unchanged ≠ content unchanged"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,44 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Fixed
 
+- **Dead SIGTERM→SIGKILL escalation guard on non-Windows kills (closes #1114)**
+	— `clients/safe-spawn.ts`'s non-Windows `killTree` branch armed a 1s
+	escalation timer gated on `if (!child.killed) child.kill("SIGKILL")`, but
+	Node sets `ChildProcess#killed = true` the moment `kill()` successfully
+	SENDS a signal — not when the child actually dies — so immediately after
+	the `child.kill("SIGTERM")` one line above, the guard was always false
+	and the SIGKILL escalation could never fire: a SIGTERM-ignoring child on
+	Linux/macOS was never force-killed. Fixed by tracking OBSERVED death via a
+	`closed` flag set synchronously (before any `await`) in the close/error
+	handlers, gating the escalation on `!closed` instead — composes cleanly
+	with the existing #1109/#1113 `escalationTimer` clear-on-close fix rather
+	than switching to the LSP `killProcessTree` analog's unconditional-SIGKILL
+	design, since `safeSpawnAsync` already has a real per-call close/error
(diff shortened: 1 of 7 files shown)


apmantza:
## Review round 2 — closes the P2/P3 findings

**P2 (vacuous coverage) — confirmed and fixed.** The reviewer's finding was correct: `tests/clients/lsp/kill-process-tree.test.ts`'s existing mocks (including the "non-fast shutdown escalates" test) neither implement `.once` nor ever set `.killed`, so `proc.once?.(...)` optional-chained to a no-op and `!proc.killed`/`!exited` were both permissive regardless of which guard was correct or broken. This is the #1106 vacuous-mock class recurring in mock form: a fixture too weak to distinguish correct from broken behavior passes either way, giving false confidence. Fixed in 69175a06:

- Upgraded that existing mock to be `.once`-capable (`kill`, `unref`, `once`, `off`) so it now actually exercises the `exited`-flag logic rather than passing vacuously.
- Added two new `fast`-shutdown tests with a real `.once`-capturing mock, proving **both directions**:
  - exit observed before the 1.5s window → **no** group SIGKILL
  - no exit observed by the window → group SIGKILL fires
  
  The "no premature SIGKILL" direction is the one that actually catches the bug — verified fail-then-pass by temporarily reverting the guard back to `!(proc as {killed?}).killed`:
  ```
  AssertionError: expected "kill" to not be called with arguments: [ -4242, 'SIGKILL' ]
  1st kill call: [ -4242, "SIGTERM" ]
  2nd kill call: [ -4242, "SIGKILL" ]   <- fired even though exit was observed first
  ```
  Reapplying the fix (`!exited`): 9/9 pass in that file.

- Added `tests/clients/lsp/initialize-timeout-backstop.test.ts` for the previously **fully untested** `initialize()`-timeout 2s SIGKILL backstop. Uses a real subprocess (fake LSP server fixture + new `FAKE_LSP_IGNORE_INITIALIZE` env flag) so the handshake times out for real, `killProcessTree`'s real SIGTERM ends the process, and the backstop must observe that and skip its own redundant SIGKILL. This test is **POSIX-only, explicitly `describe.skip`'d on win32** — I verified empirically (not just assumed) that this is a real OS constraint, not laziness: on Windows, `process.kill(-pid, "SIGTERM")` throws `ESRCH` (no real process groups), so forcing the POSIX code path on real Windows silently falls back to the direct-child `proc.kill()` call, which *does* set `.killed = true` for real — that would make the pre-fix bug's specific failure mode (group-kill never touching `.killed`) unreproducible on this dev OS and the test would pass regardless of which guard is used, i.e. vacuous in a different way. Ran it locally with the skip temporarily removed as a plumbing sanity check (env var, `initializeTimeoutMs`, real spawn/kill all wired correctly, 1/1 pass) before restoring the POSIX-only skip; the real assertion is exercised for real on Linux CI.

**P3-1 (backstop gate precision) — fixed.** The backstop's gate is now `exitCode === null && signalCode === null`, not `exitCode === null` alone. A process that died *from a signal* (the common case here, since `killProcessTree` signals it rather than letting it exit on its own) has `exitCode === null` forever and only `signalCode` set — so `exitCode` alone still re-armed a redundant SIGKILL against an already-dead corpse (harmless — `ChildProcess#kill()` on an exited handle is a swallowed no-op — but not an accurate "still alive" read). CHANGELOG sentence updated to match.

**P3-2 (entry-seed) — fixed.** `killProcessTree`'s `exited` flag is now seeded from the same `proc.exitCode != null || proc.signalCode != null` pre-check the function's own top-of-body early return uses. That early return is skipped when `options.processExiting` is set, so a process already dead *on entry* can still reach the escalation-timer code; without seeding it would miss its own already-fired "exit" event and draw a redundant group SIGKILL at the window.

**Verification:** `npm run build` (tsc, tsconfig.build.json) clean, `npx tsc --project tsconfig.json --noEmit` (lint) clean. Local runs: `kill-process-tree.test.ts` 9/9, `safe-spawn-kill-escalation-timer.test.ts` 3/3, `initialize-timeout-backstop.test.ts` 1 skipped (win32, by design), `client-internals`/`spawn-shutdown-race`/`service-shutdown-concurrency`/`lifecycle`/`teardown-logging`/`client-wait-timer-cleanup` 91/91. Full suite still deferred to CI per the #1122 precedent noted in the PR description.

Commit: 69175a06

apmantza:
Round-1 conditions verified satisfied: mocks un-vacuous'd (.once-capable), fast-path both-direction fail-then-pass captured (SIGKILL-at-window regression shown on the reverted guard), backstop covered via a real subprocess fixture with a documented+verified win32 skip (ESRCH fallback makes the bug POSIX-only; Linux CI runs the real assertion), signalCode precision gate + entry-seed landed, CHANGELOG aligned. **Merging on green.**

Issue #1104, opened 2026-08-06
Pull-diagnostics resultId plumbing for content binding (deferred from #1095) + cascade fallback-display binding gaps


apmantza:
Follow-up deferred from #1095 (both PRs merged: #1096 capture/expose/lsp_diagnostics adoption, #1100 cascade adoption). Filed at merge time so the deferral stays tracked (#1100 review P3-3).

## What
Extend diagnostics content binding to the LSP pull-diagnostics protocol: thread `resultId`s (textDocument/diagnostic previousResultIds / relatedDocuments) through the binding layer so pull-based servers get the same staleness decidability as push-based ones.

## Current state
- Push path: publishDiagnostics `version` echo + send-time content fingerprint → binding true/false/"unknown" (#1096).
- Pull path: the workspace pull-sweep record site passes no contentHash (documented honest "unknown" — clients/lsp/index.ts pull-sweep record site), so pull-served entries never demote and never bind.

## Design sketch
- Record the resultId returned by each pull alongside the document version/content hash active at request time; an unchanged-report (same resultId) inherits the prior binding basis; a full report rebinds to the version/hash at pull time.
- Compose with the existing binding merge rules (#1096: any false → false; all unknown → unknown; else true).
- Cache entries recorded from pulls can then carry contentHash, closing the "pull-sweep entries read unknown" gap and letting the service-sweep binding gate (#1096 P2-1 fix) protect them too.

## Also fold in (from #1100 review P3s, same surface)
- P3-1: the cascade's degraded/fallback display paths (integration.ts touch-error fallback, appendFallbackNeighbors) re-display TTL-fresh snapshots without consulting binding — widget is protected, display can still show a bound-false snapshot in a tier3-silent corner. Tighten when touching this area.
- P3-4: cascade-channel neighbor_touch log records bindingState but not `inconclusive`; add it so the two unconfirmed causes are distinguishable without latency.log.

Refs #1095 (closed by #1100), #1096, #1100.

apmantza:
[...]

PR #1131, opened 2026-08-07
fix(project-intelligence): close recorded micro-gaps from #1106/#1104 review

apmantza:
## Summary

Four small, independently-scoped gaps recorded in prior PR/issue review comments, bundled into one micro-PR:

1. **session-state-store STATE_VERSION reject-path test** (recorded on `gh issue view 1106` post-close comment) — `loadSessionState`'s version-mismatch reject path (`state?.version !== STATE_VERSION`) had no test. Added one that persists valid state, then corrupts the on-disk `version` to `STATE_VERSION + 1` (driven off the exported constant, never a hardcoded literal — the #1116 pattern) and asserts `loadSessionState` returns `undefined`. `STATE_VERSION` is now exported from `clients/session-state-store.ts`.
2. **rule-cache pin** (same #1106 comment) — `tests/clients/cache/rule-cache.test.ts`'s deliberate `raw.version = "v2"` schema-mismatch override lacked the adjacent pin. Added `expect(CACHE_VERSION).not.toBe("v2")` (#1082/#1116 pattern) so the test can't vacuously pass if `CACHE_VERSION` ever became `"v2"`. `CACHE_VERSION` here is `clients/cache/rule-cache.ts`'s own constant — unrelated to `clients/call-graph.ts`'s `CACHE_VERSION` (not touched).
3. **neighbor_touch inconclusive logging** (`gh issue view 1104`, P3-4) — the cascade `neighbor_touch` log recorded `bindingState` but not `inconclusive`, so the two unconfirmed-touch causes (notify/diagnostics wait lapsed vs. disk-diverged binding) were indistinguishable from `cascade.log` alone. Added `inconclusive` to the log metadata in `clients/dispatch/integration.ts` (the flag was already read via `isConfirmedTouch`, factored out into a small `readInconclusive` helper shared with `isConfirmedTouch`). Added a metadata assertion to the existing inconclusive-touch test in `tests/clients/cascade-compute.test.ts` (mocked `clients/cascade-logger.js` since `logCascade` no-ops under `isTestMode()`, so there is no other way to observe the call shape).
4. **lens_diagnostics includeGenerated doc nit** (PR #1115 review comment) — the `includeGenerated` Typebox param in `tools/lens-diagnostics.ts` only takes effect when `refreshRunners` is `cheap`/`all` (silently a no-op under `cached`/`none`); the description string now says so. Description-only change.

## Evidence (fail-then-pass, items 1 and 3)

- **Item 1**: temporarily relaxed the reject-path condition to `if (!state.widget) return undefined;` (dropping the version check) → new test failed (`expected loaded to be undefined` — got the stale-version payload back). Restored the real check → test passes.
- **Item 3**: temporarily reverted the `metadata: { inconclusive, ... }` change back to the pre-fix conditional-only `bindingState` shape → new assertion failed (`expected undefined to match object { inconclusive: true }`). Restored the fix → test passes.

## Test plan

- [x] `npx vitest run tests/clients/session-state-store.test.ts` — 19 passed
- [x] `npx vitest run tests/clients/cache/rule-cache.test.ts` — 5 passed
- [x] `npx vitest run tests/clients/cascade-compute.test.ts` — 33 passed
- [x] `npm run build` (tsc, strict + noUnusedLocals) clean
- [ ] Full suite deferred to CI — this session ran on a memory-tight machine; only the touched test files were run locally, per the repo's documented constraint.

Out of scope (per instructions): did not rename `CACHE_VERSION` in `clients/call-graph.ts` (sibling-owned), and did not touch `safe-spawn.ts`, `symbol-id.ts`, `tree-sitter-client.ts`, or `function-facts.ts`.

refs #1106 (closed — recorded for the record, not reopened), refs #1104

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,26 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Fixed
 
+- **Micro-gap sweep: recorded coverage/observability/doc gaps (refs #1106, refs #1104)**
+	- `session-state-store.ts`'s `loadSessionState` STATE_VERSION reject path
+		(a wrong-version persisted snapshot is ignored, not rehydrated) had no
+		test; `STATE_VERSION` is now exported so the new test can drive the
+		mismatch off the real constant (`STATE_VERSION + 1`) rather than a
+		hardcoded literal (#1116 pattern).
+	- `tests/clients/cache/rule-cache.test.ts`'s deliberate `raw.version = "v2"`
+		schema-mismatch override now pins `expect(CACHE_VERSION).not.toBe("v2")`
+		alongside it (#1082/#1116 pattern), so the assertion can't vacuously pass
+		if `CACHE_VERSION` ever became `"v2"`.
+	- The cascade `neighbor_touch` log entry (`clients/dispatch/integration.ts`)
+		now carries an `inconclusive` boolean in its metadata alongside
+		`bindingState`, so the two independent unconfirmed-touch causes (notify/
+		diagnostics wait lapsed vs. disk-diverged binding, #1093/#1095) are
+		distinguishable from `cascade.log` alone, without cross-referencing
+		`latency.log`.
+	- `tools/lens-diagnostics.ts`'s `includeGenerated` param description now
+		states it only takes effect with `mode=full refreshRunners=cheap/all`
+		(it's silently a no-op under `cached`/`none`, since no project scan runs
+		to apply it to) — a doc-only fix from PR #1115's review.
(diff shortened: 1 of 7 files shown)


apmantza:
Adversarial review (trace-only; machine commit ratio ~1.0 at review time): **approve, no findings** — reject-path fixture provably isolates the version gate (sanity assertion pre-corruption), readInconclusive factoring byte-identical with original-object reads, always-present log field safe for all consumers (smells script reads optionally), cascade-logger mock complete and file-scoped, rule-cache pin non-vacuous against its own v6 constant, item 4 description-only. Merging on green.

Issue #1089, opened 2026-08-05
call-graph (#1070) follow-ups: dead Java try-with-resources regex, vacuous validator tests, untested freshness invariant and module_report surface, 19-grammar co-load


apmantza:
Adversarial review of the call-graph subsystem (#1070), v3.8.74..HEAD + working tree. The uncommitted working-tree delta on feat/call-graph-1070 is a pure formatter reflow (every hunk diffed — zero semantic changes); all findings are on committed code. Findings marked (reproduced) were verified empirically against built output. Since the branch is unmerged, the branch-scoped items below can be fixed pre-merge; the rest are follow-ups.

## P2 — Java try-with-resources suppression is dead: `\b` in a template literal is a backspace
`clients/tree-sitter-client.ts:2030` — `new RegExp(`\b${resource}\b`)` inside `not_closed_or_try_with_resources`. In a template literal `\b` is U+0008, so the regex can never match (reproduced). Per the #956 comment (:2017-2022), this resource-node scan is the ONLY try-with-resources recognizer → every correctly-managed Java resource is flagged unclosed; only a literal `.close()` suppresses. Pre-existing (10f04755, inside v3.8.74) but the branch touches this exact loop and no test covers the predicate. Fix: double-escape + regex-escape the identifier; add a test.

## P2 — persisted-cache semantic validator has zero effective coverage; both "rejects..." tests are vacuous
`tests/clients/call-graph.test.ts:750` and `:800` write `version: 4` fixtures with no identity fields → `loadCallGraph` returns undefined at the `version !== 5` check before `validatePersistedCallGraph` ever runs. The ~200-line validator (call-graph.ts:801-995) passes these tests even if deleted. Fixtures went stale when edab5d5d bumped CACHE_VERSION to 5. Fix pre-merge: rewrite as v5 + identity.

## P2 — the canonical-freshness invariant itself is untested
No test passes `expectedIdentity` to `loadCallGraph`. The headline claim of edab5d5d — stale identity must be rejected (call-graph.ts:1025-1030) — has no coverage; a regression ignoring `expectedIdentity` (the #210/#1020 stale-replay class) ships green. Fix pre-merge.

## P2 — `module_report` `callGraph: true` (the new user-facing surface) has zero tests
[...]

Issue #1108, opened 2026-08-06
Hygiene classes: long positional signatures (merge mis-slot risk) + non-enumerable side-channel flags (copy-loss risk)


apmantza:
Two hygiene classes surfaced by the 2026-08-06 wave, filed together as simplification-campaign feedstock (both are refactor-shaped, not live bugs).

## C. Long positional signatures (mis-slot risk)
Near-miss during the #1094/#1096 merge: `reconcileWidgetFromLspResult` grew to 8 positional params with two optional trailing ones added by DIFFERENT PRs (`boundMismatch`, `observedAt`); the union resolution only failed loudly because the mis-slotted types differed (number vs boolean). Two same-typed trailing params would have merged clean and wrong — the #1103 review was explicitly tasked to watch for this class.
Sweep: functions in clients/ and tools/ with >= 6 positional params (grep or a quick ts-morph script); rank by call-site count and growth history; convert the hot/growing ones to a single options object (keep positional for stable 2-3 param helpers). Candidates already known: reconcileWidgetFromLspResult, the dispatch context builders, module-report internals.

## D. Non-enumerable side-channel flags (copy-loss risk)
Two members bit in one week: `touchFile`'s `inconclusive` (missed by the #1094 cascade — P1) and `binding` (#1096 — the review specifically had to verify no spread/filter intervened; #1094's test mock was ALSO wrong about enumerability until round 3). The flag pattern is load-bearing but fragile: any `[...arr]`, `.filter()`, `.map()`, structuredClone, or JSON round-trip silently drops it.
Sweep: inventory every `Object.defineProperty(..., enumerable: false)` on RETURNED values (grep defineProperty across clients/); for each, (a) document the flag on the return type + "read off the original object" contract, (b) assert consumers comply, (c) consider an ast-grep rule (dogfood) flagging spread/filter/map applied to values whose type carries a known side-channel flag, or migrate the flag to a wrapper object ({ diags, inconclusive, binding }) where the churn is affordable — the wrapper is the structural fix; the lint is the cheap guard.

Both feed the simplification campaign's Phase 1 ranking rather than demanding immediate PRs.

Refs #1094, #1096, #1103 (near-miss and members).

apmantza:
[...]

PR #1136, opened 2026-08-07
fix(project-intelligence): close #1088/#1089 P2/P3 residuals

### LINKED ISSUE #1088: review-graph: vacuous cache-identity guard enables stale call-graph projection under fresh signature; parseSymbolKey breaks on LSP-fallback kinds

apmantza:
## Summary

Fixes the remaining actionable P2/P3 items from the #1088 and #1089 adversarial-review issues. Re-verified every finding against current master before implementing -- several line-number references in the original issues had shifted or were already fixed by intervening merges (34e8c78f squashed the #1072 follow-up fixes); those are called out below rather than re-fixed.

One commit per concern:

1. **`parseSymbolKey` mis-parses LSP-fallback kinds (#1088 P2)** -- `clients/review-graph/symbol-id.ts` whitelisted only the 7 kinds `buildSymbolId` mints directly; `addLspFallbackSymbols` mints ids from the much larger `lspSymbolKindName` vocabulary (`enum`, `constant`, `struct`, `namespace`, `field`, `constructor`, `module`, the `lsp-symbol-<n>` catch-all, ...). Now matches the trailing `:<kind-token>:<digits>` shape structurally instead of whitelisting kind strings -- future-proof against new LSP kinds.
2. **Dead Java try-with-resources suppression (#1089 P2)** -- `\b` inside a template literal is the BACKSPACE control char, not a regex word-boundary escape, so the resource-word regex in `not_closed_or_try_with_resources` could never match. Fixed to a proper `\b` plus a regex-escaped identifier. No shipped Java fixture used try-with-resources syntax, so this fix flips no existing expected-findings test.
3. **Raw NUL byte in `builder.ts` (#1088 P3)** -- made the 4,900+ line file look binary to `ripgrep`/`grep`, silently truncating every repo-wide grep sweep partway through the file. Replaced with the backslash-u-0000 escape sequence (byte-identical runtime string -- the same `hashIgnoredIds` function computes both write-time and read-time hashes, so no persisted-checkpoint compatibility break).
4. **`buildCallGraph` same-file evidence accounting (#1089 P3-1)** -- re-verified against current master: the raw-vs-normalized compare this finding names had *already* been fixed by an intervening merge (the same-file check already uses `normalizeMapKey` on both sides), and the per-candidate double-count doesn't reproduce in current code (the name-only candidate path already filters to cross-file-only candidates, and the `targetId` path only ever pushes one candidate). Added a regression test reproducing the audit's exact repro shape (divergent path-form spellings of the same file) confirming the coverage sum invariant holds and the save-to-load round trip survives `validatePersistedCallGraph`; confirmed fail-then-pass by temporarily reverting to a raw compare. Also removed the dead duplicate `saveCallGraph` overload signature and the unreachable `{}`-spread branch on the required `identity` param (#1089 P3-3).
5. **JS destructured params dropped from `parameterCount` (#1089 P3-4)** -- `function-facts.ts`'s `getParameters` recognized `required_parameter`/`optional_parameter` wrapper nodes (how TS's grammar represents every param) but not the bare `object_pattern`/`array_pattern` nodes plain JS's grammar uses for a top-level destructured param with no wrapper. `function f({a, b})` counted 0 params in `.js`, correctly in `.ts`. Fixed plus JS-vs-TS parity tests.
6. **`CACHE_VERSION` naming collision (#1106-comment cosmetic)** -- `call-graph.ts`'s exported `CACHE_VERSION` collided in name with `cache/rule-cache.ts`'s own export. Renamed to `CALL_GRAPH_CACHE_VERSION` per the `<SUBSYSTEM>_CACHE_VERSION` convention (`WORKSPACE_DIAGNOSTICS_CACHE_VERSION`, `PROJECT_DIAGNOSTICS_CACHE_VERSION`). Updated the two test-file imports.

## Explicitly NOT touched (per brief)

- `recordBuildAttempt` hardening and the placeholder-coverage-veto design questions -- left open on #1088, status comment posted noting what remains.
- The three-roots cache mismatch (#1089 P3-2) -- bigger, left open, status comment posted.

## Verification

- `npm run build` clean (tsc, no errors).
- Every fix has a fail-then-pass test, verified by temporarily reverting the source change and confirming the new test fails, then restoring and confirming it passes:
  - `tests/clients/review-graph/symbol-id.test.ts` (parseSymbolKey LSP-kind cases)
  - `tests/clients/tree-sitter-879-post-filters.test.ts` (Java 9 short-form try-with-resources)
  - `tests/clients/call-graph.test.ts` (divergent-path-form same-file accounting + round trip)
  - `tests/clients/dispatch/function-facts.test.ts` (JS destructured params)
- All touched test files pass individually. This dev machine is memory-constrained for vitest's grammar-heavy co-loaded workers (documented in AGENTS.md -- the #255 "Fatal process out of memory: Zone" wall), so a batched multi-file local run crashes workers; each touched file was verified in isolation and the full suite is left to CI as the gate.
- `clients/review-graph/builder.ts`: confirmed zero raw NUL bytes remain and the compiled `.js` carries the same escape-sequence literal, not a raw byte.

## Test plan

- [ ] CI full suite green (Linux runner -- this branch's changes are OS-agnostic path-normalization/regex/string fixes, no new platform-sensitive surface)
- [ ] Adversarial review pass before merge per AGENTS.md

Generated with Claude Code


apmantza:
Adversarial review (trace-only): **approve** — structural kind-match verified safe on all adversarial shapes (the dot-exclusion provides what the whitelist never did), Java fix drives the real query path with zero fixture flips independently confirmed, NUL swap byte-identical, item-4 honesty verified (defect fixed upstream; test locks it; not a #1024 trap — backslash folding precedes the platform branch), JS/TS param parity clean, left-open items accurately tracked with the #1082 coverage loop closed. CHANGELOG train conflict resolved; escapeRegExp 6-copy consolidation recorded on #1108. Merging on green.

Issue #743, opened 2026-07-21
LSP: notify.open Promise.all lets one backpressured server time out the whole file's write


apmantza:
## Summary

In `LSPService.touchFile`, the didOpen/didChange notify write fans out across **all** spawned servers with a single `Promise.all` (`clients/lsp/index.ts:1451`). Because the whole `Promise.all` is bounded by one `notifyWriteBudgetMs()` deadline (2000ms default), a **single** server that has stopped draining its stdin (backpressure) times out the notify write for the **entire file** — including servers that are perfectly healthy. That flips the touch to `notifyWriteTimedOut → inconclusive`, so it contributes no diagnostics and never marks any server `demonstratedReady`.

## Evidence (live `mode=full` sweep on a 119-file project)

Per-file `.ts` touches during a `lens_diagnostics_full` sweep, `clientScope:"all"`, `serverCountReady:3`:

```
lsp_touch_file  logger.ts  notifyWriteTimedOut:true  diagnosticsTimedOut:true  collectedDiagnostics:0
```

The sweep crawled (24→25 completed over ~2 min) and appeared "stuck at 20%". Aggregate over the sweep: notify writes split 20 ok / 16 timed out — intermittent, consistent with one saturated server (tsserver under load) periodically stalling stdin and dragging the shared write deadline for every co-touched server.

## Why it matters

`clientScope:"all"` (used by `lsp_diagnostics` and `lens_diagnostics mode=full`) is precisely where a file is touched against primary + auxiliaries at once. One wedged aux (or a saturated primary) shouldn't be able to zero out a healthy server's contribution for that file.

## Proposed direction

Isolate the write per server instead of a single shared-deadline `Promise.all`:
- Give each server its own `notifyWriteBudgetMs` deadline (mirrors how the diagnostics wait already went per-server in #242/#573 via `perServerTimeout`).
- Set `notifyWriteTimedOut` only for the servers that actually timed out; let the others proceed and still be eligible for `demonstratedReady`.
- Consider marking a server that repeatedly backpressures its stdin write as `broken` (existing cooldown map) so the sweep stops re-paying for it.

## References
- `clients/lsp/index.ts:1445-1474` (notify write + `notifyWriteTimedOut`)
- `clients/lsp/index.ts:1901-1910` (`inconclusive`/`demonstratedReady` gate)
- `clients/lsp/index.ts:556` (`notifyWriteBudgetMs`)
[...]

PR #1139, opened 2026-08-07
fix(lsp): count post-init runtime exits toward the circuit breaker (closes #1127)

### LINKED ISSUE #1127: opengrep respawn churn: post-init runtime exits bypass the LSP circuit breaker (#1122 Phase C)
From #1122's Phase C corroborating-signal review.

## Symptom
`sessionstart.log` records **37** opengrep `closed code=1` events over ~3 days across projects, each with the signature:
```
opengrep: closed code=1 ... stderr=Initializing server ... Server initialized ... Scanning single file Scanned single file [NN.NN][ERROR]: Unhandled message: { "id": "...", "jsonrpc": "2.0", "result": null } ...
```
opengrep (`opengrep.exe lsp --experimental`) spawns fine, initializes, serves a file or two, then chokes on a JSON-RPC message it considers unhandled (a `result: null` response) and exits 1. On the next attach it is re-spawned, serves briefly, dies again — never converging to "give up on opengrep this session."

## Root cause: the failure mode slips past the circuit breaker
`LSPService` (`clients/lsp/index.ts`) has a sound exponential-backoff breaker: `failureCounts` → `BROKEN_BASE_COOLDOWN_MS(15s) * 2^(n-1)` capped at `BROKEN_MAX_COOLDOWN_MS(5m)`, permanent-disable after `BROKEN_PERMANENT_AFTER(5)`. **But `failureCounts` is incremented only on spawn/init failure** (the `undefined`/error paths ~L1428 and ~L1506) and **deleted on a successful spawn** (~L1480). opengrep's failure is a *post-initialization runtime exit*: the spawn SUCCEEDS, so the counter is reset, and the later `code=1` close is handled by the "dead client — needs respawn" path (~L1303) which does **not** touch `failureCounts`. Net: the backoff and permanent-disable never engage for a server that dies *after* serving. The respawns are attach-triggered (minutes–hours apart, not a hot loop), so it is bounded churn rather than a runaway — but it never stops, and opengrep provides no diagnostics while flapping.

## Proposed fix
Count **early post-init runtime exits** toward the breaker: when a client that was previously alive closes with a non-clean status and an uptime under a threshold (e.g. < 30–60s), increment `failureCounts` for that key so repeated short-lived sessions trip the existing cooldown/permanent-disable. This makes a deterministically-broken auxiliary LSP (opengrep `--experimental` here) stop being re-attached for the session instead of flapping. Needs the uptime threshold + a regression test exercising the runtime-close path (distinct from the spawn-failure path already covered).

Separately worth a note: opengrep's `--experimental` LSP appears genuinely broken on our valid JSON-RPC responses — an upstream report may be warranted, but the breaker gap is ours to fix.

apmantza:
## Summary

- opengrep (auxiliary ast-grep LSP) spawns and initializes fine, then dies to a post-init JSON-RPC "Unhandled message" error and exits — 37 respawns in one real session per #1122's Phase C corroborating-signal review, never converging.
- Root cause: `LSPService`'s (`clients/lsp/index.ts`) circuit breaker (`failureCounts` → exponential cooldown → permanent-disable after `BROKEN_PERMANENT_AFTER=5`) only increments on spawn/initialize failure. A client that dies *after* a successful spawn hits the separate "dead client — needs respawn" path, which never touched `failureCounts` — and that map had already been cleared to 0 by the preceding successful spawn anyway.
- Fix: a parallel `runtimeExitCounts` counter, fed only by EARLY (uptime < 60s) non-intentional exits detected in that respawn path, sharing the SAME cooldown formula and the SAME `state.broken`/`permanentlyBroken` maps as the existing breaker — so a crash-loop server converges to the same give-up behavior instead of respawning forever.

## Design notes

- **Same breaker vs. parallel counter — chose parallel, sharing the circuit.** Reusing `failureCounts` directly was tried first but conflicts with its existing "clear on successful spawn" semantics (`spawnClient`'s success path), which is *correct* for the original spawn/init failure class (a server whose binary reappears on PATH should get a fresh streak). For a crash-loop server, a successful spawn is not evidence of health — only surviving past the uptime threshold is. Reusing the map would make every respawn erase the streak the moment it re-spawned, which is exactly the #1127 bug reproduced with a different trigger. The new `runtimeExitCounts` counter therefore uses its own map but feeds the SAME `state.broken` cooldown deadline and the SAME `permanentlyBroken` set, so skip-if-cooling-down / permanent-disable / `getBrokenStatus()` all behave uniformly regardless of which stream produced the failure. `getBrokenStatus()`'s `failures` field now sums both counters.
- **Distinguishing deliberate teardown from crash exits.** `clients/lsp/client.ts` already had exactly the signal needed: `state.shutdownRequested`, set only inside `clientShutdown()` and used to gate the existing `lsp_server_unexpected_exit` log. Exposed it via a new `wasShutdownIntentional()` accessor on `LSPClientInfo`. Audited every process-level restart path in `clients/lsp/index.ts` (`#743` notify-backpressure eviction, session resets, generation handoffs) — all go through `client.shutdown()` first, so all correctly read as intentional and never count. The only place that could misread this is capturing the flag *before* calling `existing.shutdown()` in the respawn cleanup itself (that call sets `shutdownRequested=true` as a side effect) — handled explicitly.
- **Threshold:** 60s uptime, per the issue's suggested 30–60s range — generous enough for a normal serve-a-few-files-then-idle session, tight enough to catch "spawns fine, dies within seconds every time."

## Test plan

New `tests/clients/lsp/service-runtime-exit-breaker.test.ts` (3 tests, all passing):
- **Crash-loop convergence** (fail-then-pass verified): a mock server that spawns successfully but "crashes" (uptime ~0) every time converges to `permanentlyBroken` after exactly `BROKEN_PERMANENT_AFTER=5` early exits, then stops re-attaching even when cooldown is cleared. Verified this test **fails on pre-fix code** (temporarily reverted `clients/lsp/{index,client}.ts` to `origin/master`, confirmed `permanentlyBroken` is never set and the assertion loop's bound catches the unbounded-respawn regression) and **passes** with the fix restored.
- **Deliberate restart not counted**: 8 `shutdown()`-driven restarts (past the permanent-disable threshold) never increment `runtimeExitCounts` or trip `permanentlyBroken`.
- **Uptime-threshold reset**: a runtime exit past the 60s threshold resets the streak instead of counting it.

Local runs (touched areas):
- `npx tsc --noEmit --project tsconfig.build.json` — clean.
- `npx vitest run tests/clients/lsp/` — 534 passed, 4 skipped (existing `broken-status.test.ts` and `client-crash-logging.test.ts` stay green, confirming no regression to the original spawn/init breaker or crash-logging behavior).
- `npx vitest run tests/index-lsp-idle-reset.test.ts` — passed.

Full suite deferred to CI per repo convention.


apmantza:
## Addressed adversarial-review round

**P1 (headline, fixed in ebcb8846):** `uptimeMs` was computed at lazy detection time (`Date.now() - clientSpawnedAt` in the next `getClientForFile` attach), not at actual death time. #1127's documented pattern is attach-triggered respawns "minutes–hours apart" — an early crash detected an hour later would misread as a long healthy run and the streak would wrongly reset, exempting exactly the crash-loop case this fix targets.

Fixed by recording a real death timestamp (`exitedAt`) on `LSPClientState`, stamped by the first of `onError`/`onClose`/process `exit` to fire in `clients/lsp/client.ts`'s `setupConnectionLifecycle` (`markExitedIfUnset`), and exposed via a new `getExitedAt()` accessor on `LSPClientInfo` alongside the existing `wasShutdownIntentional()`. The respawn site in `clients/lsp/index.ts` now computes `uptimeMs = exitedAt - spawnedAt`, falling back to the old detection-time delta only when `exitedAt` is unexpectedly unset (defensive, not the expected path).

New decisive test (`tests/clients/lsp/service-runtime-exit-breaker.test.ts`): backdates the spawn 3 hours, records death only 5s later, and calls `getClientForFile` "now" (real time, hours after both) — confirms it still counts toward the breaker. Verified fail-then-pass against the pre-fix detection-time computation in **both** directions: it missed the early-death case (read as long-lived, streak reset) AND wrongly counted a genuinely long-lived server in the companion "past threshold" test (whose backdated-spawn setup happened to read as near-zero detection-time uptime under the old code). Both failures reproduced by temporarily reverting to c40a8e38, rebuilding, and rerunning; both pass again with the fix restored.

**P2 (fixed in ebcb8846):** added a dedicated test that drives the REAL `#743` notify-write-backpressure eviction path (`recordNotifyWriteBackpressure`) through `touchFile()` — not a synthetic stand-in — by making a mocked client's `notify.open` hang forever against a 5ms `PI_LENS_LSP_NOTIFY_BUDGET_MS`, forcing 3 consecutive timeouts. Confirms the real eviction (client removed from `state.clients`, its own `shutdown()` called, `state.broken` set directly) never touches `runtimeExitCounts` — structurally can't, since by the time anything looks for an existing client again there isn't one, so the "dead client" branch this fix touches never runs for that path.

**P3s:**
1. Filed #1142 tracking the residual: a server whose lifetime is consistently just *over* the 60s threshold (never within it) still churns undetected — the per-death lifetime cutoff can't catch a slow, low-frequency crash loop. Not blocking for #1127 (opengrep's actual failure mode is well under any reasonable threshold).
2. Documented the `failureCounts`/`runtimeExitCounts` aggregation semantics at `getBrokenStatus()` — the two streams trip independently on their own thresholds; `.failures` is their sum for display only.

**Housekeeping:** merged `origin/master` (CHANGELOG conflict from the #1130/#1132/#1136/#1138 merge train — resolved by keeping both sides' entries, stripping conflict markers, and updating the #1127 entry to describe the death-timestamp fix). `npx tsc --noEmit` clean; full `tests/clients/lsp/` suite (560 tests) and `tests/scripts/changelog.test.ts` (24 tests, the bold-title-wrap contract included) both green after the merge.

Commits: c7b6e8be (death-timestamp fix + P2 test + doc), ebcb8846 (merge + CHANGELOG resolution).

apmantza:
Round-1 conditions verified satisfied: death-timestamp (exitedAt, first-signal-wins across onError/onClose/exit, captured pre-shutdown) replaces the detection-time delta; delayed-detection test proves the defect bidirectionally on the reverted code; real #743 eviction-path test added; aggregation semantics documented; residual (≥61s-lifetime escape) homed in #1142; CHANGELOG train resolved with the contract test green. **Merging on green.**

apmantza:
## CI red after master merge — diagnosed and fixed

**Diagnosis:** the P1/P2 fix hunks in \`clients/lsp/client.ts\` (\`markExitedIfUnset\`, the interface/state \`exitedAt\` field, \`getExitedAt()\`) all survived the \`ebcb8846\` merge intact — diffed \`c40a8e38..c7b6e8be\`'s \`client.ts\` hunks against the merged head and every hunk (interface decl, state initializer, all 3 \`markExitedIfUnset\` call sites, the accessor) is present and correctly placed. The three breaker tests were not actually failing locally against \`ebcb8846\` in isolation or as part of the full \`tests/clients/lsp/\` run (536/536 passing) — so the reported failures were likely from a slightly different checked-out ref, not a semantic regression in the merge itself.

**What was real:** the lint failure (P1 of this round). \`tests/clients/lsp/client-internals.test.ts\`'s \`createMockState()\` helper builds an \`LSPClientState\` object literal that predates the new required \`exitedAt\` field my #1127 fix added — TS2322 on the literal missing that key. Fixed by adding \`exitedAt: undefined\` to the base object (mirrors \`client.ts\`'s real initializer at \`createLSPClient\`).

**Also picked up:** \`origin/fix/1127-opengrep-breaker\` had gained a direct-pushed merge commit (\`5df6d7e2\`, master's #1140 smells-rollup) I hadn't merged locally yet — fetched and merged it (clean, no conflicts, doesn't touch any \`clients/lsp/\` file).

**#1130 sibling check:** \`killProcessTree\`'s \`exited\` (the SIGTERM→SIGKILL escalation observed-death flag, #1114/#1130 territory, \`client.ts\` ~L834) is a local closure variable scoped to that function — no name or state collision with the new \`LSPClientState.exitedAt\` field this fix adds; both verified present and correct in the merged head, and \`kill-process-tree.test.ts\` (30/30) stays green.

**Verification on the final pushed head (\`8e4b1181\`):**
- \`npm run lint\` (\`tsc --project tsconfig.json\`) — clean
- \`npx tsc --noEmit --project tsconfig.build.json\` — clean
- \`tests/clients/lsp/service-runtime-exit-breaker.test.ts\` + \`client-internals.test.ts\` + \`kill-process-tree.test.ts\` — 90/90 passing
- Full \`tests/clients/lsp/\` + \`tests/scripts/changelog.test.ts\` — 560 passing, 4 skipped

Pushed \`5df6d7e2..8e4b1181\`. CI re-running now.

apmantza:
## CI red on Linux — root cause found, not merge damage

Confirmed this was the AGENTS.md #1024 class (OS-divergent behavior, green on Windows / red on Linux CI), traced to the actual CI log rather than guessed. All three failures share one root cause:

**The bug:** \`service-runtime-exit-breaker.test.ts\` hardcoded \`const key = "opengrep:C:/repo";\` in five places and used that literal to poke \`LSPService\`'s internal \`state.broken\`/\`runtimeExitCounts\`/\`permanentlyBroken\` maps directly (\`.delete(key)\`, \`.has(key)\`, \`.get(key)\`) — on the assumption that \`normalizeMapKey("C:/repo")\` is a no-op.

That assumption only holds on Windows. \`clients/path-utils.ts\`'s \`isWindowsPath()\` treats any drive-letter-shaped string like \`"C:/repo"\` as Windows-shaped **regardless of the running platform**, so on Linux CI \`normalizeFilePath\` still enters the realpath-fallback branch (since \`C:\repo\` never exists there, \`realpathSync.native\` throws). Inside that fallback (\`resolveNonExisting\`), \`dirname\` is imported from \`"node:path"\` — Node's platform-dependent default export, POSIX-flavored on Linux — and gets applied to a win32-resolved backslash path string. POSIX \`dirname\` sees no \`/\` in \`"C:\repo"\` and immediately collapses it to \`"."\`, so the walk-up-to-an-existing-ancestor loop stops at the **current working directory** instead of a drive root. The real key the service computes on Linux therefore becomes \`${realCwd}/repo\`, not the literal \`"C:/repo"\` the test assumed.

Net effect: every \`internal.state.broken.delete(key)\` / \`.has(key)\` / \`runtimeExitCounts.get(key)\` in the test was checking the WRONG map entry on Linux — consistently reproducing the exact CI failures across two runs (\`expected undefined to be defined\` at the "converges" test, \`expected undefined to be 1\` at "delayed detection", \`expected false to be true\` at the notify-backpressure test) — while passing locally on Windows, where the literal happens to stay byte-identical to the real key (verified: \`normalizeFilePath("C:/repo")\` returns \`"C:/repo"\` unchanged on Windows).

**Fix:** compute the key through the SAME \`normalizeMapKey\` the service uses internally (\`\`opengrep:${normalizeMapKey("C:/repo")}\`\`) instead of hardcoding the literal, so the test agrees with the product on any platform — no platform branching needed, per the "probe actual behavior" rule.

**Bug-class sweep:** grepped all of \`tests/\` for the same hardcoded-literal-key-vs-\`normalizeMapKey\` shape used against \`LSPService\`'s internal maps. No other occurrences — \`broken-status.test.ts\`'s literal keys are self-seeded directly into the maps and never compared against a real \`root()\` resolution, so they don't share this defect shape.

**Verification (local, Windows):** \`npm run build\` + \`npm run lint\` clean; \`service-runtime-exit-breaker.test.ts\` + \`client-internals.test.ts\` + \`kill-process-tree.test.ts\` + \`changelog.test.ts\` — 114/114 passing; full \`tests/clients/lsp/\` — 536/536 passing, 4 skipped. Merged the two intervening direct-pushed \`origin/fix/1127-opengrep-breaker\` updates (master syncs, no conflicts, don't touch \`clients/lsp/\`).

Pushed \`d778a0f5..9c228b5a\`. Awaiting Linux CI on this head.

PR #1140, opened 2026-08-07
feat(project-intelligence): smells self-surfacing rollup (refs #1123 item 3)

apmantza:
## Summary

`scripts/analyze-pi-lens-logs.mjs` (`npm run logs:smells`) already detects a wide catalogue of operational smells across `~/.pi-lens/*.log`, but it's **manual** — the #1123 investigation found 20 stale-ctx `emit_failed` rows and 37 opengrep respawns sitting unread in logs for days until an audit went looking. This PR adds a small, **always-on** rollup that self-surfaces the two smells the issue named.

## Design choice + cost bound

`~/.pi-lens/*.log` files are size-rotated at up to ~10MB (`clients/log-cleanup.ts`). Re-scanning a whole rotated log on every `session_start` — the manual analyzer's own approach — is exactly the megabyte full-file read the hot, input-latency-sensitive `session_start` path cannot afford (per the issue's stated constraint).

Of the three options the issue laid out — (a) write-time live counters, (b) a bounded tail-scan at session_start, (c) reusing the analyzer's check functions over a bounded window — I picked **pure (b), a bounded tail-scan, without adding parallel write-time counters at the producer call sites**. Rationale: both source logs (`bus-events.log`, `latency.log`) are append-only NDJSON, so the tail is always the most recently written activity — the same bounded read serves both the cross-session glance AND a live/this-session proxy, without instrumenting `bus-publish.ts`/`diagnostics-publish.ts`/`disposition-publish.ts`/`format-events-publish.ts`/`clients/lsp/index.ts`'s respawn path. This kept the change contained to **one new module + three call sites** (`session_start`, `/lens-health`, `turn_end`) instead of touching five+ producers, at the cost of the rollup being a recency proxy rather than an exact "this session" counter — an explicit, documented tradeoff (see `clients/smells-rollup.ts`'s module docstring).

**Cost bound, enforced by construction:** `tailReadText(filePath, maxBytes)` does one sized `fs.readSync` at a computed end-of-file offset — never a full-file read regardless of the source file's actual size. Two source files × 64KB = **~128KB total I/O**, once per `session_start` plus one bounded re-check every 20 turns (`SMELLS_TURN_CHECK_INTERVAL`). A dedicated test (`caps total I/O at 2 * SMELLS_TAIL_BYTES_PER_FILE regardless of source size`) constructs a 20k-line fixture and confirms the bound holds end-to-end.

## Thresholds (trivial constants, `SMELLS_THRESHOLDS`)

- `staleCtxEmitFailed`: 5 (bus-events.log `emit_failed` rows whose `error` matches the SDK's `"stale after session replacement"` fragment — `session-lifecycle.ts`'s documented benign-but-worth-watching class)
- `opengrepRespawn`: 5 (latency.log `lsp_server_respawn` phase rows with `metadata.serverId === "opengrep"` — `clients/lsp/index.ts`'s existing, **unmodified** respawn log point)

## Surfaces

1. **`session_start` line** — threshold-gated (nothing logged below threshold), emitted via `runtime-session.ts`'s `emitSmellsSessionStartLine` at both the quick-mode and full-mode `session_start total` points.
2. **`/lens-health` line** — always-on, current counts regardless of threshold, matching #1123 item 2's `formatMemoryHealthLine` style.
3. **`turn_end` note** — re-checked every 20 turns, `ctx.ui.notify`d **at most once per smell per session** (`checkSmellsAndNoteOnce`'s gate; re-armed by `resetSmellsSessionState()` at the next `session_start`).

## Fail-then-pass evidence

Broke `formatSmellsSessionStartLine`'s threshold gate (forced it to always return a line), rebuilt, and confirmed `tests/clients/smells-rollup.test.ts`'s "returns null when every smell is below threshold" test failed with the expected mismatch; reverted, rebuilt, and confirmed all 19 tests pass again.

## Test results (touched files, local)

- `tests/clients/smells-rollup.test.ts` — 19 passed (tail-scan cost bound by construction, log parsing, threshold gating for both surfaces, once-per-session notify gate)
- `tests/index-smells-rollup-wiring.test.ts` — 4 passed (turn_end wiring: nothing before the check interval, one notify at the trip turn, no repeat notify on the next check turn, no notify below threshold — only `countRecentSmells`'s I/O boundary is mocked, the real `checkSmellsAndNoteOnce`/`shouldCheckSmellsThisTurn` run)
- Sibling seams re-run clean: all 7 `runtime-session*.test.ts` files (42 passed), all `index-*.test.ts` wiring files (44 passed pre-existing + this PR's 4)
- `npm run build` clean; no new `biome check` errors introduced (pre-existing `index.ts`/`runtime-session.ts` formatting/import-order findings predate this branch, confirmed via `git stash`)
- Full suite left to CI per the repo's machine-wide test-lock policy (`#1101`) — not run locally beyond the touched files above.

## Coordination

A sibling agent is implementing #1123 item 4 (`PI_LENS_DEBUG_HANDLES`) in the same PR wave, also touching `index.ts` session seams. This PR's `index.ts` hunks are deliberately small and isolated: one import block, one `try/catch` block inside the existing `/lens-health` command handler (right after the #1123 item 2 memory line), and one `try/catch` block inside the existing `turn_end` handler (right after the #1123 item 2 memory-sample block) — chosen to minimize textual overlap for the merge train.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,6 +156,47 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Added
 
+- **Smells self-surfacing (refs #1123 item 3)** — `scripts/analyze-pi-lens-logs.mjs`
+	(`npm run logs:smells`) already catalogues a wide set of operational smells,
+	but it's MANUAL: the #1123 investigation found 20 stale-ctx `emit_failed`
+	rows and 37 opengrep respawns sitting unread in logs for days until an audit
+	went looking. `clients/smells-rollup.ts` adds a small ALWAYS-ON rollup
+	covering the two smells the issue named, without touching every producer of
+	the underlying logs and without re-scanning the size-rotated (up to ~10MB,
+	`clients/log-cleanup.ts`) `~/.pi-lens/*.log` files on the `session_start`
+	hot path. **Design + cost bound:** a BOUNDED TAIL READ — at most 64KB from
+	the end of `bus-events.log` and `latency.log` each (~128KB total I/O per
+	check, enforced by construction via a single sized `fs.readSync` at a
+	computed offset in `tailReadText`, never a full-file scan) — counts
+	stale-ctx `emit_failed` rows (`outcome === "emit_failed"` with an `error`
+	containing the SDK's `"stale after session replacement"` fragment,
+	`session-lifecycle.ts`'s documented benign-but-worth-watching class) and
+	opengrep respawns (`phase === "lsp_server_respawn"` with
+	`metadata.serverId === "opengrep"`, `clients/lsp/index.ts`'s existing,
+	unmodified respawn log point). Because both source logs are append-only
+	NDJSON, the tail is always the most recently written activity, so the same
+	bounded read serves as both the cross-session glance and a live/this-session
+	proxy — no separate write-time counters were added at the producer call
+	sites (kept the change to one new module + three call sites). Both counts
+	are gated by trivial threshold constants (`SMELLS_THRESHOLDS`, currently 5
+	each) so a single stray event never surfaces. Three surfaces: one
+	`session_start` line (only emitted when a threshold trips, via
+	`runtime-session.ts`'s `emitSmellsSessionStartLine`), an always-on compact
+	`/lens-health` line (current counts regardless of threshold, matching
+	#1123 item 2's `formatMemoryHealthLine` style), and a `turn_end` note —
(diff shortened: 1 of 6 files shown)


apmantza:
Review: **approve** — matchers verified real and CI-pinned (the compat-contract guard on the SDK fragment), tail-read edge cases sound by construction and tested, cost bound honest, wiring adversarially tested. Window-vs-session spec drift signed off and recorded on #1123 with the P3s. Merging on green; expect the item-4 sibling CHANGELOG/index.ts conflict on whichever lands second.

PR #1141, opened 2026-08-07
feat(observability): add handle-origin debug tracer (refs #1123 item 4)

apmantza:
## Summary

Implements item 4 of #1123: institutionalizes the `async_hooks` handle-origin tracer the #1097 print-mode-hang investigation hand-rolled to find a leaked, ref'd `setTimeout` keeping a `--print --no-session` process alive after `agent_settled` (root-caused/fixed in #1110). The #1097 reporter explicitly asked for a built-in active-resource dump so the next "process won't exit" report is a one-line diagnostic instead of a fresh multi-hour expedition.

- New `clients/debug-handles.ts`. `PI_LENS_DEBUG_HANDLES=1` is read **once** at module load (extension startup) and gates everything:
  - **Unset (default): every export is a no-op past one boolean check** — no `NdjsonLogger` constructed, no `async_hooks` hook installed, zero allocations on the hot `agent_settled`/`session_shutdown` paths that call `dumpActiveHandles(label)`.
  - **Set at startup:** `dumpActiveHandles(label)` calls `process.getActiveResourcesInfo()`, tallies counts by resource type, and writes one ndjson line to `~/.pi-lens/debug-handles.log` via the standard `createNdjsonLogger` family (size-bounded via `getMaxLogSizeMB()`, self-registered with the `#1101`-era log-cleanup rotation registry).
  - An `async_hooks` creation-site tracker records a trimmed `Error().stack` per resource id for timer/socket/handle async-hook types (`Timeout`, `TCPSOCKETWRAP`, `PIPEWRAP`, etc.) and removes the entry on `destroy`. It is installed **only** when the flag was already on before the module first loads — real per-resource-creation overhead (hooks fire on every init/destroy in the process), so it's strictly a debug-only cost, documented in both the module docstring and AGENTS.md.
  - The tracker's creation-site map is **bounded**: capped at `TRACKER_MAX_ENTRIES` (500) with drop-oldest eviction (`Map` insertion order) — never unbounded growth even in a pathological timer-storm session (the one-axis-rule from AGENTS.md's resource-bounding discipline).
- Wired in `index.ts` at two lifecycle points:
  - `agent_settled` — **after** `runQuietWindow(...)` is scheduled (the #1097-class leak is only visible once whatever settle itself queues is in flight).
  - `session_shutdown` — **after** `resetLSPService(...)` teardown ("what survives IS the leak").
- Docs: `docs/environment-variables.md` gets a `PI_LENS_DEBUG_HANDLES` entry under "Diagnostics and logging"; `AGENTS.md`'s hot-path/performance-discipline section gets one bullet alongside the sibling `#1123` items 1/2 notes (loop_block, memory-sampler) documenting the zero-cost-when-unset design and the bounded tracker.
- CHANGELOG: `Unreleased` → `Added`, `refs #1123 item 4` (item 3, the smells rollup, is a sibling in flight touching the same `index.ts` session seams — kept the hunks minimal there, only adding the two `dumpActiveHandles(...)` call lines).

## Design

- **Zero cost when unset**: every exported function starts with `if (!DEBUG_HANDLES_ENABLED) return;` (or the module-level `writer`/`trackedResources` are `null`), so the only work done at either call site when the flag is off is one boolean read.
- **Flag read once at startup, not per-call**: `DEBUG_HANDLES_ENABLED`, `trackedResources`, and `writer` are all module-scope `const`s evaluated at import time — matches the `async_hooks` tracker's own constraint (it can only meaningfully attribute resources created *after* it's installed, so late-enabling would be silently incomplete; the module makes "must be set before pi starts" explicit).
- **Bounded tracker**: `recordTrackedInit` (shared by the real `init` hook and a test-only simulator) enforces the cap + drop-oldest in one place.

## Test plan

- [x] `npx vitest run tests/clients/debug-handles.test.ts` — 9 unit tests: flag unset → `isDebugHandlesEnabled()` false, `dumpActiveHandles` writes nothing, tracker stays at 0; flag set → dump emits `{label, total, counts, ts}`, both wired labels append independently, `creationSites` attribution present and grouped by type, tracker cap enforced (500 in, still 500 after +50 more), untracked resource types (`PROMISE`) never enter the map.
- [x] `npx vitest run tests/index-integration.test.ts` — two new wiring tests drive the **real** `index.ts` `agent_settled`/`session_shutdown` handlers (mocking only `clients/lsp/index.js`, `clients/quiet-window.js`, and `clients/debug-handles.js` to record call order) and assert `dumpActiveHandles` fires strictly after LSP teardown / after quiet-window scheduling. **Fail-then-pass verified**: temporarily disabled both `dumpActiveHandles(...)` call sites in `index.ts`, rebuilt, and confirmed both new tests fail with the expected missing-entry diff; restored the wiring and reran — both pass.
- [x] `npm run build` (tsc, no errors) and `npm run lint` (tsc, no errors); `npx biome check` clean on the two new files.
- [x] Locally green: `tests/index-integration.test.ts`, `tests/clients/debug-handles.test.ts`, `tests/clients/ndjson-logger.test.ts`, `tests/clients/log-cleanup.test.ts`, `tests/index-wiring.test.ts`, `tests/index-vanished-instance-wiring.test.ts`, `tests/index-memory-sample-wiring.test.ts`, `tests/clients/turn-summary.test.ts` (67 + 26 = 93 tests across the touched/adjacent files, all passing).
- [ ] Full suite deferred to CI (Windows dev / Linux CI project convention) — please run the full `npm test` in CI before merge.

## Blast radius

`clients/debug-handles.ts` is a new, self-contained module with exactly two call sites, both newly added in this PR (`index.ts`'s `agent_settled` and `session_shutdown` handlers) — no other consumers exist yet, so there is nothing pre-existing to break. The two `index.ts` edits are additive one-liners inserted after existing statements (`void runQuietWindow(...)` / `resetLSPService(...)`), not restructuring, to keep the diff minimal alongside item 3's (smells rollup, in flight) touches to the same handlers.

Do not merge — opening for review per repo convention.


apmantza:
Addressed the adversarial review's P2 in 63cdf41d:

**Eviction policy inverted, now fixed.** `recordTrackedInit`'s cap-eviction now protects the first `TRACKER_PROTECTED_COUNT` (100) insertion-order entries and evicts the oldest entry *outside* that zone, instead of plain drop-oldest — a #1097-style leaked handle is typically among the earliest created in a session, so drop-oldest was evicting exactly the evidence the tracer exists to capture.

**`evictedCount` added.** A running, never-reset counter is now always present on `DebugHandlesLogEntry` once the tracker is installed (even at `0`), so an attribution gap from a burst past the cap is explicit rather than silent. Caveat sentence added to the module docstring, `AGENTS.md`, and `CHANGELOG.md` (which previously sold the cap as pure upside with no downside noted).

**Cap test updated** for the new policy: a 600-entry burst against the 500 cap asserts the first 100 protected ids AND the 5 most-recently-created ids both survive, the map still respects `TRACKER_MAX_ENTRIES`, and `evictedCount` (both in-memory and on a dump entry) equals `burstSize - TRACKER_MAX_ENTRIES` exactly.

**Optional P3 done too.** Added a real `async_hooks.createHook`-driven integration test (`"a real setTimeout is tracked by the installed async_hooks hook, and pruned on destroy"`) — every prior tracker test drove the `_simulateTrackedInitForTest` double, leaving the actual init/destroy plumbing installed by `installHandleTracker()` uncovered. It creates a genuine `setTimeout`, asserts it shows up in a real dump's `Timeout`-typed `creationSites` count, clears it, and polls (via `setImmediate`, tracked under a different type key so its own transient liveness can't mask the assertion) until the `Timeout` count returns to baseline.

`npm run build` / `npm run lint` clean, `npx biome check` clean on the touched files, `tests/clients/debug-handles.test.ts` (11 tests) and `tests/index-integration.test.ts` (25 tests) all green — reran the debug-handles suite 3x back-to-back with the new real-timer test to check for flakiness, no failures.

apmantza:
Round-1 condition verified satisfied: protect-earliest eviction (first 100 pinned — the #1097-shaped evidence survives bursts), evictedCount always present on dumps (explicit, never silent), docs de-oversold, plus the real createHook integration test (init AND destroy proven end-to-end, 3x-stability-checked). **Merging on green** — this completes #1123 item 4; the issue stays open only for the recorded deferrals.

Issue #1023, opened 2026-08-01
bug(cascade): degraded/uncomputed impact renders as "nothing impacted" — silent all-clear (#533)


apmantza:
## The bug (assessed, #533-critical)

The cascade impact subsystem's entire job is to warn the agent about downstream breakage from an edit. But when it **cannot compute** impact, it emits an all-clear indistinguishable from "genuinely nothing impacted".

- `computeImpactCascade` returns a fully empty result whenever the changed file has no graph node (`clients/review-graph/query.ts:129-139`) — no field distinguishes "file not in graph" from "zero real dependents".
- `buildOrUpdateGraph` returns an empty graph (seeded only with the changed file's own symbols) when the repo is over the review-graph file cap (`too_many_files`) or the root is unsafe (`clients/review-graph/builder.ts:3643-3739`). The cascade then computes impact against that empty graph → zero neighbors for **every** edit in a large monorepo.
- A cascade that **threw** is mapped to `skipReason:"error"` (`clients/pipeline.ts:1374`).
- In all three degraded cases the state is written **only to `latency.log`/`cascade.log`** — the agent-facing turn-end message is built solely from neighbors carrying error diagnostics (`clients/cascade-format.ts:9-10`), so the agent sees **nothing = safe to proceed**.

**Triggering scenario:** agent edits an exported function's signature in a repo just over `PI_LENS_REVIEW_GRAPH_MAX_FILES`; every real dependent is missed; the agent receives a clean turn-end and moves on believing nothing downstream broke. Silent under-report of impact in exactly the environments where downstream breakage is most likely and costly.

### Companion (same defect, narrower trigger) — Tier-3 cold TS neighbor
A cold-snapshot TypeScript neighbor's error is computed and the quiet-window reconcile classifies it `resolved-found`, but the reconcile handler is **logs-only** (`clients/lsp/cascade-tier.ts:236-266`) — the error never re-injects into the agent-facing cascade output.

## Fix direction (reuse existing infra)

Give the impact result a first-class **indeterminate** state and surface it, reusing what already exists:
- `getLastGraphBuildInfo().mode`/`skipReason` (`builder.ts:202-207`) is already available at the compute call site (`integration.ts:775, 887-906`) — thread it into the result rather than re-deriving.
[...]

PR #1143, opened 2026-08-07
fix(project-intelligence): gate cascade fallback-display paths on content binding (refs #1104)

apmantza:
## Summary

Fixes the P3-1 item from #1104's "Also fold in" section (the `resultId` pull-diagnostics main body of #1104 stays open — out of scope here).

#1100 gated the cascade's **reconcile** path (footer/widget) onto diagnostics content binding (`boundToCurrentDisk`, #1095), but two degraded-fallback **display** paths in `clients/dispatch/integration.ts` still re-read TTL-fresh `getAllDiagnostics()` snapshots without consulting binding at all:

- the touch-error fallback (a failed active LSP touch falls back to the passive snapshot)
- `appendFallbackNeighbors` (the CR-3/A2 degraded-fallback path when no neighbor produced trustworthy LSP data)

A bound-false snapshot — diagnostics computed against a *different* disk state than what's currently on disk, e.g. a pre-fix-edit read — could still reach cascade OUTPUT even though the widget was already protected by #1100.

## Changes

- Both display sites now apply the exact `false`/`"unknown"`/`true` contract #1095/#1100 already established for reconcile: `false` → skip the stale display (logged via the cascade channel with `bindingState`, the established pattern), `"unknown"` → unchanged (the pre-existing fallback contract), `true` → display.
- **Honesty fix**: filtering a display candidate must not turn a genuinely degraded cascade into a clean-looking one. Added `CascadeIndeterminateReason: "lsp_binding_rejected"` (`clients/review-graph/types.ts`) — when every fallback candidate a run considered was binding-rejected and nothing else produced output, the run now carries the same `indeterminate` marker #1023 built for a degraded graph compute, so the turn-end advisory (`clients/runtime-turn.ts`) still surfaces an honest note instead of silence.
- CHANGELOG entry under Unreleased (refs #1104, not closes — the issue stays open for the resultId body).

## Blast radius

`appendFallbackNeighbors` is a module-private helper, only called from `computeCascadeForFile` in the same file. `computeCascadeForFile` itself is called from `clients/pipeline.ts` (the per-edit cascade dispatch) and referenced by `clients/mcp/analyze.ts`'s cascade-adjacent commentary — no other call sites. No public API/type surface changed besides the additive `CascadeIndeterminateReason` union member (backward compatible).

## Test plan

- [x] New tests in `tests/clients/cascade-compute.test.ts` (`describe("fallback-display binding gaps (#1104)")`, 5 tests) exercising `computeCascadeForFile` end-to-end:
  - touch-error fallback excludes a bound-false snapshot from display (fail-then-pass verified against pre-fix source)
  - characterization: touch-error fallback with unknown binding is unchanged
  - `appendFallbackNeighbors` excludes a bound-false snapshot from display (fail-then-pass verified against pre-fix source)
  - characterization: `appendFallbackNeighbors` with unknown binding is unchanged
  - HONESTY: when every `appendFallbackNeighbors` candidate is binding-rejected, the run still reports `skipReason: "indeterminate"` / `indeterminate.reason: "lsp_binding_rejected"`, never `"clean"`
- [x] Fail-then-pass confirmed by temporarily reverting `integration.ts` to the pre-fix `HEAD` version, rebuilding, and re-running the 5 new tests: 3 fail (the two exclusion tests + the honesty test), 2 pass (the characterization tests) — exactly as expected. Restored the fix and reran: all 5 pass.
- [x] `npx tsc --noEmit` clean.
- [x] Touched-file + sibling cascade suites green locally: `cascade-compute.test.ts` (38/38), `cascade-turn-merge.test.ts`, `lsp/cascade-tier.test.ts`, `runtime-cascade-defer.test.ts` (73/73 combined).
- [x] Full local suite: 5383 passed, 13 skipped, 1 failed (`tests/clients/sgconfig.test.ts` — a 5s timeout on an unrelated ast-grep sgconfig baseline test, accompanied by a vitest worker-fork crash; unrelated to this diff — this file/area isn't touched by the change, and the change is a self-contained cascade display-path fix). Leaving full-suite confirmation to CI (Linux) per the note in AGENTS.md about dev-machine resource contention; flagging here in case it reproduces there too and needs separate triage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,28 @@ All notable changes to pi-lens will be documented in this file.
 
 ### Fixed
 
+- **Cascade fallback-display paths re-displayed bound-false LSP snapshots (refs #1104)** —
+	#1100 gated the cascade's RECONCILE path (the footer/widget) onto content
+	binding (`boundToCurrentDisk`), but two DEGRADED-fallback DISPLAY paths in
+	`clients/dispatch/integration.ts` still re-read TTL-fresh `getAllDiagnostics()`
+	snapshots without consulting binding at all: the touch-error fallback (a
+	failed active LSP touch falling back to the passive snapshot) and
+	`appendFallbackNeighbors` (the CR-3/A2 degraded-fallback path when no
+	neighbor produced trustworthy LSP data). A bound-false snapshot — diagnostics
+	computed against a DIFFERENT disk state than what's currently on disk, e.g. a
+	pre-fix-edit read — could still reach cascade OUTPUT even though the widget
+	was protected. Both sites now apply the same false/`"unknown"`/true contract
+	#1095/#1100 already established for reconcile: `false` → skip the stale
+	display (logged via the cascade channel with `bindingState`), `"unknown"` →
+	unchanged (the pre-existing fallback contract), `true` → display. HONESTY
+	fix: filtering a display candidate could otherwise make a genuinely degraded
+	cascade look clean, so when every fallback candidate a run considered was
+	binding-rejected and nothing else produced output, the run now carries the
+	same `indeterminate` marker #1023 built for a degraded graph compute (new
+	`CascadeIndeterminateReason: "lsp_binding_rejected"`), so the turn-end
+	advisory still surfaces an honest note instead of silence. The `resultId`
+	pull-diagnostics plumbing that #1104 also tracks remains open — this covers
+	only the cascade display-binding gap (#1100 review P3-1).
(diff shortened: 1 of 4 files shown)


apmantza:
Review: **approve** — faithful mirror of the #1100 reconcile contract onto both degraded display paths; TTL-before-binding ordering verified at both sites; #1121 composition clean; honesty marker set only when binding was the deciding factor and proven to reach output; unknown characterizations assert DISPLAY. P3 (advisory preamble framing) recorded on #1104. Merging on green after rebase.

PR #1146, opened 2026-08-07
fix(lsp): contained P3 bundle from #1085 (P3-2, P3-3, P3-4, P3-6, P3-8)

apmantza:
## Summary

Implements the five CONTAINED P3 deferrals recorded on #1085 after PR #1120 (`clients/lsp/edits.ts`), one commit each:

- **P3-2** — `mergeWorkspaceTextEditsByPriority`'s exact-duplicate dedup collapsed identical zero-width inserts duplicated within ONE server's own edit on the `renameFile` merge path, contradicting the documented multiplicity invariant (`validateTextEdits` never dedups zero-width ranges). Fixed to only dedup non-empty ranges; cross-server exact duplicates still collapse.
- **P3-3** — an ordered edit `rename(b→c); create(b); textEdit(b)` was falsely rejected in preflight because a `create` at a path vacated by an earlier rename had nowhere to store its state (`resolveVirtualPath` correctly returns `undefined` there). Added a `virtualOverrides` overlay keyed on the raw query path.
- **P3-4** — `version: null` (LSP 3.17: "don't check") + a numeric version for the SAME uri incorrectly threw "conflicting text document versions". The numeric version is now adopted as authoritative; two genuinely different numerics still conflict, and an adopted-but-stale numeric is still rejected by the existing preflight version check.
- **P3-6** — `importsChanged` reported "the file contains any import" instead of "this edit changed an import/re-export line", over-invalidating downstream dependency-graph re-checks. **Consumer read first**: `cache-manager.ts` filters modified-range entries on `importsChanged` before an import-graph re-check, and `lsp-mutation.ts`'s `addModifiedRange` forwards the same flag — both want "did this edit change the import graph," which is what the fix now computes (pre/post import-line signature comparison). `create`/`rename`/`delete` keep their existing conservative flags.
- **P3-8** — an ordered edit that CREATEs `foo.txt` then case-renames it to `Foo.txt` within the SAME edit fail-closed on a case-insensitive FS, because the alias check (`isSameFsEntry`) lstats disk but the just-created file exists only in the preflight's virtual overlay. Fixed with a fast, disk-free check: if the rename's `destination` and `source` resolved to the identical cached `VirtualFile` object (same case-folded map key), they're the same virtual entry — checked before falling back to the physical probe, which is unchanged for genuinely-physical paths (the #1024/#1120 ino-guard and case-sensitivity invariants are untouched).

**Explicitly NOT in this PR (still deferred):**
- **P3-5** (the CRLF-boundary class beyond the `\r`-aware clamp #1120 already shipped)
- **P3-7** (rename close-failure recovery reopening the old document as `"plaintext"`)

Since #1085 is closed, I posted the remaining-items status as a comment on #1085 per deferral hygiene, proposing they get folded into a comment on #1104 or a small follow-up issue.

## Blast radius

Direct consumers of `clients/lsp/edits.ts` (via `module_report`/grep): `clients/actionable-warnings.ts` (`applyWorkspaceEdit`), `clients/lsp-mutation.ts` (type-only import of `AppliedWorkspaceFileDetail`, consumes `importsChanged` in `addModifiedRange`), `clients/lsp/client.ts` and `clients/lsp/index.ts` (`renameFile`'s `mergeWorkspaceTextEditsByPriority` merge path). All covered by `tests/clients/lsp/edits.test.ts`, `tests/clients/lsp/client-internals.test.ts`, and `tests/clients/lsp/service-rename-file.test.ts`, all green.

## Fail-then-pass evidence

Each of the 6 new assertions (P3-4 got two tests: matching-numeric applies, stale-numeric still rejected) was verified to FAIL against the pre-fix `edits.ts` and PASS after the fix, before being committed:

- P3-2: `expected [...] to have a length of 2 but got 1` (dedup collapsed the duplicate) → 2 kept, `"aQQbc"`.
- P3-3: `Error: text edit target is not a file: .../b.txt` → applies, `b.txt` contains `"new"`, `c.txt` contains `"original"`.
- P3-4 (matching numeric): `Error: conflicting text document versions for ...` → applies, file becomes `"AxB"`.
- P3-4 (stale numeric): same conflicting-versions throw (wrong error) → now correctly rejects with `/stale text document version/`.
- P3-6: `expected true to be false` (body-only edit falsely flagged `importsChanged: true`) → `false` for the body-only edit, `true` for the actual import-line edit.
- P3-8: `Error: rename destination already exists: .../foo.txt` → succeeds, `Foo.txt` is the sole surviving spelling.

## Test plan

- [x] `npm run build && npx vitest run tests/clients/lsp/edits.test.ts` — 47/47 pass
- [x] `npx vitest run tests/clients/lsp/client-internals.test.ts tests/clients/lsp/service-rename-file.test.ts` — 89/89 pass
- [x] `npx vitest run tests/clients/lsp/` (full LSP directory) — 537 pass, 4 skipped (pre-existing FS-probe-guarded skips), 0 failed
- [x] `npm run lint` (tsc) clean
- [x] Merged current `origin/master` (post-#1143) into the branch; rebuilt and reran the targeted suites — still green
- [ ] Full CI suite (touched-test-files were run locally; the rest goes to CI per repo convention)


apmantza:
## Round-2: addressed adversarial review findings

**P2 (fail-safe inversion, REQUEST CHANGES) — fixed.** `IMPORT_RELEVANT_LINE` only matched a leading `import`/`export` token, so a formatter-wrapped multiline import's specifier line (`} from "./old";`) matched nothing — changing the module specifier there reported `importsChanged: false`, under-invalidating a real module-edge change (the dangerous direction; master's old `/^import\s/m` over-invalidated instead, which is safe).

Fix: also match a bare `from "..."` continuation line — `/^\s*import\b|^\s*export\s[^;]*\bfrom\s|\bfrom\s+['"]/`. Kept the choice **line-signature-based**, not a blanket "file has a multiline import → always report changed" fallback: a body-only edit in a file that HAS a multiline import still correctly reports `importsChanged: false`, because only the `from "..."`-matching line itself is compared, and that line is untouched. New test `P3-6 continuation` asserts both directions in one file (multiline import, `import {\n\tfoo,\n\tbar,\n} from "./old";`): a whole-line body edit → `false`; a specifier edit on the continuation line (`"./old"` → `"./new"`) → `true`.

Documented remaining known gaps in the docblock (not expanded scope, per your note): dynamic `import(...)`/`require(...)` calls, and a bound-name-only change on an interior line whose `from` line is untouched (narrower than a full import-statement diff, still strictly safer than the prior heuristic).

**P3 (virtualOverrides staleness) — fixed.** `rename(b→c); create(b); rename(b→d)` left `virtualOverrides[b]` stale (still `exists: true`) after `b` was genuinely re-vacated by the second rename, because `resolveVirtualPath` chases *physical* paths through `virtualMoves` and can't represent "this purely-virtual entry moved again."

Fix: when a rename's source state came from the override overlay (`resolveVirtualPath(oldPath) === undefined`, captured before this rename's own mutations), migrate the state object directly — delete the stale key, re-key it under the destination into `virtual` or `virtualOverrides` depending on whether the destination itself resolves to a physical path — instead of pushing an inapplicable `virtualMoves` shadow.

New test `P3: a re-renamed override-backed entry migrates...` covers the cross-product in two cells (a literal second `create(b)` in one edit trips the PRE-EXISTING, unrelated `duplicate workspace resource operation` guard, so cell 2 uses `delete` instead to prove non-staleness without expanding scope there):
1. `rename(b→c); create(b); rename(b→d); textEdit(d)` → `c.txt` keeps `"original"`, `d.txt` gets the trailing text edit applied (`"hello"`) — content isn't lost.
2. `rename(b→c); create(b); rename(b→d); delete(b)` (no `ignoreIfNotExists`) → correctly throws `"delete target does not exist"` (proving `b` is genuinely gone, not a stale `exists: true`).

## Fail-then-pass evidence

Reverted `clients/lsp/edits.ts` to the pre-round-2 commit, rebuilt, confirmed both new tests failed with exactly the predicted symptoms, then restored the fix:
- `P3-6 continuation`: `expected false to be true` (specifier-line edit under-reported `importsChanged: false`) → now `true`.
- `P3` (re-rename staleness): `Error: text edit target is not a file: .../d.txt` (trailing text edit found no migrated state) → now applies, `d.txt` = `"hello"`.

## Test results

- `npm run build && npx vitest run tests/clients/lsp/edits.test.ts` — 49/49 pass (was 47; +2 new)
- `npx vitest run tests/clients/lsp/client-internals.test.ts tests/clients/lsp/service-rename-file.test.ts` — 89/89 pass
- `npx vitest run tests/clients/lsp/` (full directory) — 539 pass, 4 pre-existing skips, 0 failed
- `npm run lint` (tsc) clean
- Rebuilt the worktree via a clean `npm ci` (no junction issues this run) and merged current `origin/master` (post-#1141) into the branch — clean auto-merge, rebuilt, targeted suites still green
- Pushed to this PR's branch (`fix/1085-p3-bundle`)

Issue #958, opened 2026-07-31
Session-start perf follow-ups: incremental word-index refresh, worker-thread snapshot gzip, meta-write skew polish


apmantza:
Follow-up tracking for the #947 remainder (PR #957 landed the pre-warm fold, meta-first gate, and compact+cached snapshot).

**1. Incremental word-index refresh — the biggest measured win left.** The word index is fully rebuilt in **71% of sessions** (54 rebuilt vs 22 reused in dogfood logs; 458–619ms background each) because any seq advance invalidates the whole snapshot — while the per-edit incremental updater already exists (`word-index per-edit: updated …`). Persist per-file mtimes (or content hashes) in the serialized index and refresh only stale docs: ~600ms → ~50ms. Mind #928's coverage/truncation fields — an incremental refresh must keep `indexedFileCount`/`truncated` honest.

**2. Snapshot gzip via the worker-persist pattern.** Compact (#957) bought ~30%; gzip is 5–10× on top (the review-graph measured 60MB → 1.4MB). Deliberately deferred from #957: it should follow `clients/review-graph/persist-worker.ts`'s pattern — worker-thread stringify+gzip with generation-gated promotion — NOT a sync gzip on the save path (the #950 review's +656MB naïve-gzip lesson). Also softens the cost whenever the meta gate misses. Keep one-release legacy read fallback like #950 did.

**3. Meta-write skew polish (small).** `saveProjectSnapshot` writes body then meta non-atomically; a meta-write failure/crash between the two leaves an old-seq meta over a fresh body, and the gate then discards a genuinely fresh snapshot (fail-open, perf-only, self-healing — but it's the one skew direction the gate punishes). Use `atomic-write.ts` or reorder+tolerate. Plus untested paths from the #957 review: warm-attach skip, `canWarmCaches=false` skip, meta/body-skew case; and the FAT/exFAT 2s-mtime nit on the parse cache.

Related upstream (not pi-lens work): once `host_boot` vs `extension_eval` records (#948/#952) accumulate over a few days of dogfooding, the remaining ~5s of perceived startup is pi host boot — escalate to the pi maintainers with the measured split.

Refs #947, #957, #948, #950, #928.

apmantza:
[...]

PR #1149, opened 2026-08-07
fix(project-intelligence): keep persistence workers unreferenced

### LINKED ISSUE #1148: Persistence workers re-reference MessagePorts after unref, keeping pi --print alive
# Follow-up: persistence workers keep completed `pi --print` processes alive via referenced `MessagePort`

## Summary

This is a follow-up to #1097 and PR #1110.

PR #1110 correctly fixed orphaned timeout handles that could keep a completed `pi --print --no-session` process alive. The fix is installed and present in the build tested here. However, the same settled-but-alive symptom still occurs through a different resource: a pi-lens persistence `WorkerThread` whose `MessagePort` becomes referenced after `worker.unref()`.

The live process had no network connection, child process, active request, or pending output. Its only non-stdio active handle was a referenced `MessagePort`. Runtime inspection identified its message callback as pi-lens's `handleSnapshotWorkerResult`.

The same listener ordering appears in both first-party persistence-worker constructors:

- project snapshot persistence
- review-graph persistence

## Impact

A one-shot Pi process can finish its model/tool work but never exit. Callers that wait for process completion remain blocked, so higher-level automation appears frozen and cannot advance to its next task.

This is especially visible in subprocess-based agent and planner workflows, but it can affect any `pi --print` or JSON/print invocation that activates one of these persistence workers.

## Relationship to #1097

This has the same external symptom as #1097, but a different root cause.

PR #1110 fixed referenced race-loser timers in:

- `clients/lsp/index.ts`
- `clients/quiet-window.ts`
- `clients/format-service.ts`

Those changes are present in the running build. The current hung process had no active timer handle. Its surviving resource was a `MessagePort` owned by the project-snapshot persistence worker.

This therefore appears to be a second member of the broader “completed one-shot process retained by a pi-lens background resource” class, rather than a failure of PR #1110's timer cleanup.

Related:

- #1097 — same settled-but-alive symptom through orphaned timers
- PR #1110 — timer cleanup fix for #1097
- #22 — earlier print-mode shutdown lifecycle problem
- #958 / PR #995 — project-snapshot persistence worker introduction
- PR #950 — review-graph persistence worker introduction
- #1123 / PR #1141 — handle-origin diagnostics motivated by #1097
- #1048 — broader persistence lifecycle and observability tracker

## Environment and version verification

The tested pi-lens installation was loaded as a Pi Git package from:

```text
<PI_AGENT_DIR>/git/github.com/apmantza/pi-lens/dist/index.js
```

Checkout identity:

```text
HEAD:     7f1e8b6b62858058fcba57fcf5347abc07a2514f
Describe: v3.8.74-187-g7f1e8b6b
Branch:   master
```

Both #1097 fix commits are ancestors of this checkout:

```text
04f48fb9886e1e9ec391b1f56f472c9777f12ac6  present
8994256861c8305cad676f8b699191417f96b981  present
```

The bundled `dist/index.js` contains the `waitTimer`/`clearTimeout(waitTimer)` fix from PR #1110.

The live process's callback resolved to this exact script URL:

```text
file://<PI_AGENT_DIR>/git/github.com/apmantza/pi-lens/dist/index.js
```

The script loaded in memory exactly matched the on-disk bundle:

```text
Loaded SHA-256:  d351c36c24ecc19f56d386a62e2cac9f6511dbd5e7fb1a76da3612ad80973c11
On-disk SHA-256: d351c36c24ecc19f56d386a62e2cac9f6511dbd5e7fb1a76da3612ad80973c11
```

`package.json` still reports `3.8.74`, but this Git installation is 187 commits after the `v3.8.74` tag. The commit identity is therefore more useful than the package version alone.

A final duplicate/current-head check fetched upstream `origin/master` at `3d70ab79`. It was five commits ahead of the tested checkout and still contained the same ordering in both worker constructors:

```ts
worker.unref();
worker.on("message", ...);
```

No newer upstream fix was present at the time of filing.

## Live-process evidence

The affected child process was idle in:

```text
do_epoll_wait
```

It had:

- no child processes
- no TCP connection
- no active Node requests
- empty stdout/stderr socket queues
- one `WorkerThread`

Runtime inspection returned:

```json
{
  "resources": [
    "PipeWrap",
    "PipeWrap",
    "MessagePort"
  ],
  "handles": [
    { "type": "Socket", "fd": 1 },
    { "type": "Socket", "fd": 2 },
    {
      "type": "MessagePort",
      "hasRef": true,
      "eventNames": [
        "newListener",
        "removeListener",
        "message",
        "messageerror"
      ]
    }
  ],
  "requests": []
}
```

The live worker's message listener was:

```text
handleSnapshotWorkerResult
```

Its source referenced `_snapshotWorkerRequests`, snapshot stage files, gzip byte counts, and snapshot promotion. This identifies it as pi-lens's project-snapshot persistence worker rather than an LSP process or unrelated extension.

Terminating only this worker allowed the completed Pi child to exit immediately. Its parent automation then resumed and completed normally. No unrelated pi-lens or language-server process needed to be stopped.

The behavior was later reproduced by another one-shot Crew child. Runtime inspection again showed the same sole referenced `MessagePort` and the same `handleSnapshotWorkerResult` callback.

## Confirmed root cause

The project-snapshot worker is created in `clients/project-snapshot.ts` with this ordering:

```ts
const worker = new Worker(workerPath);
worker.unref();
worker.on("message", handleSnapshotWorkerResult);
worker.on("error", (err: Error) => handleSnapshotWorkerDeath(err.message));
worker.on("exit", (code) => {
  // ...
});
```

Adding the `message` listener after `worker.unref()` references the worker's message port again. The resulting `MessagePort` has `hasRef() === true` and keeps the Node event loop alive.

A minimal Node check confirmed that ordering controls the retained handle:

```text
unref-then-listen  -> active MessagePort refs: [true]
listen-then-unref  -> active MessagePort refs: []
```

Equivalent diagnostic code:

```js
import { Worker } from "node:worker_threads";

async function check(order) {
  const worker = new Worker(`setInterval(() => {}, 1000)`, { eval: true });

  if (order === "unref-then-listen") {
    worker.unref();
    worker.on("message", () => {});
  } else {
    worker.on("message", () => {});
    worker.unref();
  }

  await new Promise((resolve) => setTimeout(resolve, 30));

  const refs = process
    ._getActiveHandles()
    .filter((handle) => handle?.constructor?.name === "MessagePort")
    .map((port) => port.hasRef());

  console.log(order, refs);
  await worker.terminate();
}

await check("unref-then-listen");
await check("listen-then-unref");
```

`process._getActiveHandles()` is used here only as a diagnostic aid.

## Isolated causal verification

The actual compiled project-snapshot module was also tested in fresh Node child processes, without Pi or Crew involved.

The child imported `dist/clients/project-snapshot.js`, called `saveProjectSnapshot(...)` with a minimal valid snapshot, printed `save-returned`, and then had no intended foreground work left.

### Unmodified build (RED)

Three independent runs used the unmodified installed build:

```text
baseline_run_1 exit=124 save-returned 1
baseline_run_2 exit=124 save-returned 2
baseline_run_3 exit=124 save-returned 3
```

Exit `124` came from an external two-second timeout. All three children completed and persisted their snapshots but remained alive afterward.

### One-variable ordering change (GREEN)

A temporary copy of the compiled module changed only this ordering:

```diff
 const worker = new Worker(workerPath);
-worker.unref();
 worker.on("message", handleSnapshotWorkerResult);
+worker.unref();
```

Five independent runs then exited naturally:

```text
patched_run_1 exit=0 save-returned 1
patched_run_2 exit=0 save-returned 2
patched_run_3 exit=0 save-returned 3
patched_run_4 exit=0 save-returned 4
patched_run_5 exit=0 save-returned 5
```

All five runs produced valid gzip snapshot files. Each decoded successfully with `version: 2` and its expected sequence number. This checks that moving `unref()` did not merely force exit by losing the pending snapshot write; the existing exit-hook fallback preserved the data.

Node's own `internal/worker/io` implementation explains the result. `setupPortReferencing(...)` calls `port.ref()` when the first listener for the Worker's `"message"` event is added, and calls `port.unref()` when the last listener is removed. `Worker.prototype.unref()` explicitly unreferences both the worker handle and its public message port. Calling `worker.unref()` after listener registration therefore leaves both resources unreferenced as intended.

## Confirmed sibling site

A source sweep found exactly two first-party `new Worker(...)` sites. Both use the same ordering.

### 1. Project snapshot persistence

`clients/project-snapshot.ts`

```ts
const worker = new Worker(workerPath);
worker.unref();
worker.on("message", handleSnapshotWorkerResult);
```

This is the worker confirmed in the live hung processes.

### 2. Review-graph persistence

`clients/review-graph/builder.ts`

```ts
const worker = new Worker(workerPath);
worker.unref();
worker.on("message", handleWorkerResult);
```

This site has the same Node reference behavior and should be fixed in the same pass, even though the inspected live processes happened to retain the project-snapshot worker.

No other first-party `new Worker(...)` or `new MessageChannel(...)` sites were found. Occurrences in `dist/` are generated copies of these source sites.

## Suggested fix

Register all listeners first, then call `unref()`:

```ts
const worker = new Worker(workerPath);
worker.on("message", handleSnapshotWorkerResult);
worker.on("error", (err: Error) => handleSnapshotWorkerDeath(err.message));
worker.on("exit", handleExit);
worker.unref();
```

The same ordering should be applied to the review-graph persistence worker.

This should let persistence continue while the host has other active work, without allowing an idle worker port to keep a completed one-shot process alive.

## Why the #1097 regression test did not catch this

PR #1110's committed regression test uses fake timers and asserts that no timer remains after a fast LSP client win. That correctly covers the diagnosed timer defect, but it cannot detect a live `WorkerThread` or referenced `MessagePort`.

The original #1097 report proposed an end-to-end check that launches a real Pi child, waits for `agent_settled`, and then requires the process to exit within a bounded grace period. A test at that level would cover timers, workers, ports, sockets, watchers, and future resource types.

### Worker-specific child-process tests

For each persistence worker:

1. Spawn a real Node child.
2. Import the built pi-lens module.
3. Trigger project-snapshot or review-graph persistence.
4. Wait until the persistence result is complete.
5. Leave the worker otherwise idle.
6. Assert that the child exits naturally within a short limit.

The assertion should fail with the current listener ordering and pass when `unref()` is moved after listener registration.

### Real Pi print-mode test

1. Run `pi --print --no-session` with pi-lens enabled.
2. Exercise a path that writes a project snapshot or review graph.
3. Wait for `agent_settled`.
4. Assert natural process exit within one or two seconds.
5. On timeout, record `process.getActiveResourcesInfo()` and handle origins.

It would be useful to cover a small matrix:

- pi-lens loaded without an explicit diagnostic tool call
- `lsp_diagnostics`
- project snapshot persistence
- review-graph persistence
- a full/project-wide diagnostics path

## Suggested resource-class sweep

To reduce repeat occurrences of the same shutdown symptom, consider auditing every resource that can retain the event loop:

- `Worker` and `MessagePort`
- `MessageChannel` and `BroadcastChannel`
- race/deadline timers
- spawned child processes and their stdio
- servers, sockets, and file watchers
- resources created during or after `agent_settled`
- resources whose cleanup depends on `session_shutdown`

Static sweeps are useful for known patterns, but a real process-exit integration test is the broadest guard because it does not need to predict the next retained-handle type.

## Temporary workaround

Setting the following environment variable forces project-snapshot persistence onto its synchronous fallback path:

```text
PI_LENS_SNAPSHOT_PERSIST_SYNC=1
```

This should avoid the project-snapshot worker, but it does not address the equivalent review-graph worker site. An external timeout remains a safer general guard for one-shot automation until both worker constructors are corrected and covered by process-exit tests.

aeturnal:
## Summary

Fix both persistence-worker constructors so their idle `MessagePort`s cannot keep a completed one-shot process alive:

- register `message`, `error`, and `exit` listeners before `worker.unref()`
- cover project-snapshot and review-graph persistence with real Node child-process exit tests
- document the listener-order invariant and add the required release note

Closes #1148

Related context:

- #1097 — the same settled-but-alive symptom caused by orphaned timers
- #1110 — the timer cleanup fix for #1097
- #958 / #995 — project-snapshot persistence worker
- #950 — review-graph persistence worker

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [x] area:project-intelligence
- [ ] area:perf
- [ ] area:observability
- [x] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/apmantza/pi-lens/blob/master/CONTRIBUTING.md) and [AGENTS.md](https://github.com/apmantza/pi-lens/blob/master/AGENTS.md)
- [x] `npm run lint` passes
- [x] `npm test` passes

## What changed and why

Node references a Worker's public `MessagePort` when the first `"message"` listener is added. Both constructors previously called `worker.unref()` first, so listener registration immediately undid the intended lifecycle behavior.

The fix deliberately changes only setup order: all existing listeners, asynchronous persistence, generation gates, exit hooks, and synchronous failure fallbacks stay unchanged.

The regression test launches a fresh Node child for each real persistence path, waits until the gzip output exists, prints a completion marker, and then requires natural process exit. This level is intentional: fake timers and in-process assertions cannot detect a referenced Worker/`MessagePort` handle.

TDD evidence against the unmodified code:

```text
project-snapshot: persisted, then timed out after 5000ms
review-graph:     persisted, then timed out after 5000ms
```

With this commit, both children exit naturally and the focused suite completes in about 0.5 seconds.

## Verification

```text
npm run build                                      passed
npm run lint                                       passed
focused persistence/lifecycle suites               42/42 passed
npm test                                           5420 passed, 32 skipped
npm run build:dist                                 passed
npm run check:lockfile                             passed
npm run changelog:check                            passed
git diff --check                                   passed
```


apmantza:
thanks @aeturnal. merging soon

PR #1151, opened 2026-08-08
fix(path-utils): use win32.dirname in win32-committed normalize branch (closes #1150)

### LINKED ISSUE #1150: normalizeFilePath mangles Windows-shaped paths on non-Windows OS (isWindowsPath classifies by shape, not platform; resolveNonExisting uses platform-default dirname)
Surfaced by PR #1139's Linux-CI failure (root-caused during #1127 work): `normalizeFilePath` mangles a Windows-SHAPED path when run on a NON-Windows OS, because `isWindowsPath` classifies by string shape, not by the running platform.

## Mechanism (clients/path-utils.ts)
- `isWindowsPath(p)` = `/^[A-Za-z]:/.test(p) || p.startsWith("\\\\")` — true for any drive-letter/UNC-shaped string on ANY OS (:24-25).
- `normalizeFilePath`'s early-return guard is `process.platform !== "win32" && !isWindowsPath(normalized)` (:45) — so on Linux, a `C:/repo`-shaped input FAILS the guard and enters the Windows branch: `realpathSync.native` throws (no such path on Linux) → `resolveNonExisting` uses `win32.resolve` but the module-level `dirname` (:96) is Node's PLATFORM-DEFAULT (POSIX on Linux). POSIX `dirname("C:\repo")` finds no `/`, collapses to `.`, the upward walk stops at `process.cwd()` → the real key becomes `${cwd}/repo`, NOT the literal `C:/repo`.
- Result: a Windows-shaped path normalizes to a DIFFERENT key on Linux than the byte-identical literal it is on Windows.

## Blast radius (honest)
1. **Primary — test-correctness (active hazard):** any test hardcoding a drive-letter literal path and poking a `normalizeMapKey`-keyed structure silently mis-keys on Linux CI while passing on Windows — the #1024 OS-divergence class, in the test harness. This produced #1139's green-locally/red-on-Linux-CI. The #1127 fix corrected the one instance (compute the key via `normalizeMapKey`) and swept tests/ (clean).
2. **Secondary — production, narrow/latent:** a single-OS Linux process only sees POSIX paths → guard early-returns → correct. The mangle bites only if a Windows-shaped path crosses INTO a Linux process (cross-machine persisted cache/snapshot keys — unlikely, per-machine data dirs; or a tool/LSP emitting a Windows URI to Linux — shouldn't happen). Real but low-probability; worth closing off defensively.

## Fix directions (pick after assessing)
- The `resolveNonExisting` fallback should use `win32.dirname` (not the platform-default `dirname`) when it has committed to the win32 branch — so a Windows-shaped path is parsed consistently with win32 semantics regardless of OS. One-line-ish, and it makes the function internally coherent (it already uses `win32.resolve`/`win32.normalize` there — the bare `dirname` import at :96 is the inconsistency).
- Consider whether `isWindowsPath`-shaped-but-on-POSIX inputs should be normalized as win32 at all, or rejected/flagged — but changing the classification is higher-risk; the dirname coherence fix is the safe first step.
- Add a Linux-runnable test that feeds a `C:/`-shaped path to `normalizeFilePath` and asserts a coherent, stable key (the regression guard the whole class needed).

## Companion: a test-authoring lint/convention
Tests must not hardcode drive-letter literal paths as `normalizeMapKey`-keyed structure keys — derive via `normalizeMapKey`. Worth an ast-grep rule (dogfood) or an AGENTS.md note in the OS-agnostic section.

Refs #1139/#1127 (the trigger), #1024 (the OS-divergence class), #1025 (path-key normalization).

apmantza:
Closes #1150.

## The bug
`normalizeFilePath` commits to its win32 branch by path **shape** (`isWindowsPath` — true for any drive-letter/UNC string on ANY OS), but inside `resolveNonExisting` the upward walk used the module-level `dirname` imported from `node:path` — the **platform default** (POSIX on Linux) — while the rest of that branch already used `win32.resolve`/`win32.normalize`. On Linux the POSIX `dirname` finds no separator in a win32-resolved `C:\…` path, collapses to `.`, stops the walk at `process.cwd()`, and yields `<cwd>/file.ts` instead of the literal key. A Windows-shaped path therefore normalized to a **different key on Linux** than the byte-identical literal it is on Windows — the #1024 OS-divergence class, surfaced by #1139's green-on-Windows / red-on-Linux-CI (root-caused during #1127).

## The fix
One-line coherence fix: use `win32.dirname` in the win32-committed branch (dropping the now-unused module-default `dirname` import). Classification (`isWindowsPath`-shaped-but-on-POSIX → treated as win32) is **unchanged** — the issue flags changing it as higher-risk; the dirname-coherence fix is the safe first step.

### dirname/path audit (per function)
- `resolveNonExisting` (:96) — **win32-committed** (uses `win32.resolve`/`win32.normalize`/`win32.basename`, returns forward-slashed win32 paths). The bare `dirname` was the lone inconsistency → **fixed to `win32.dirname`**.
- `walkUpDirs` / `findNearestContaining` / `findLocalToolConfig` / `findNearestMarkerRoot` / `isAtOrAboveHomeDir` / `toProjectRelativePath` — **platform-native by design** (operate on the running OS's real paths, back real `existsSync`/`realpath` probes). `path.dirname`/`path.resolve` here is correct — **left unchanged**.

## Fail-then-pass evidence
`resolveNonExisting`'s algorithm run on a Windows-shaped, non-existent input with `path.posix.dirname` (== Linux's module-default) vs `path.win32.dirname` (the fix):

```
input: "C:/__pi_lens_1150__/sub/file.ts"
  PRE-FIX (posix.dirname == Linux):  C:/Users/apman/OneDrive/Desktop/pi-lens-1150/file.ts   ← mangled: cwd-rooted, "sub" dropped
  POST-FIX (win32.dirname):          C:/__pi_lens_1150__/sub/file.ts                        ← structure preserved
```

The new regression test asserts the key (a) is coherent across `C:/` and `C:\` forms, (b) preserves the structural tail, and (c) retains drive-letter shape / is cwd-independent. Pre-fix on Linux both (b) and (c) fail (`<cwd>/file.ts`); the test is **meaningful on both OSes** — native win32 path on Windows, shape-committed win32 branch on Linux. 36 tests pass locally on Windows; full suite runs on Linux CI (this is precisely the branch that only diverges there).

## Companion convention
AGENTS.md OS-agnostic section gains: tests must derive `normalizeMapKey`/`normalizeFilePath`-keyed structure keys via the normalizer, **never hardcode a drive-letter literal as the expected output key** (feeding one as an input is fine).

**ast-grep rule decision — deliberately NOT shipped.** The good use (a `"C:/…"` literal fed *into* `normalizeMapKey`) and the bad use (the same literal used as a raw map key) are **syntactically identical string literals**, distinguishable only semantically. A shape-matching rule would fire on every legitimate drive-letter input literal — including the normalizer's own test fixtures — so it would be pure noise. Left as a review-enforced convention + the regression guard instead.

## Bug-class sweep
Grepped for other consumers using the module-default `dirname`/path functions on paths that could be Windows-shaped in a cross-OS scenario (same defect shape).

- **`win32.*` mixing:** `clients/path-utils.ts` is the **only** file in the repo that mixes win32-specific path calls — no other module has a win32-committed branch that could harbor the same bare-`dirname` inconsistency.
- **Module-default `dirname` importers (11 files):** 9 are build/release/dev **scripts** + `tests/support/check-build-freshness.ts` — all run on the local/CI machine's **native** paths, never a cross-OS Windows-shaped input. Two production consumers examined:
  - `clients/dispatch/runners/rust-clippy.ts:135` — `findNearestContaining(dirname(filePath), …)`; the path is always the running OS's real cargo-project path about to be handed to the Rust toolchain, and `findNearestContaining` probes real `existsSync`. Not reachable with a cross-OS Windows-shaped path. OK.
  - `clients/file-role.ts:71` — `dirname(filePath).replace(/\/g,"/").toLowerCase()` for role classification. This is the **closest latent sibling of the same defect shape**: a POSIX `dirname` on a Windows-shaped input on Linux would collapse to `.` and misclassify. But it is **platform-native by design** (not win32-committed) and only bites in the same unlikely cross-OS crossing as this issue's "secondary, narrow/latent" production blast radius — and fixing it means *deciding classification*, which the issue explicitly scopes OUT as higher-risk. **Recorded as a follow-up candidate, not fixed here.**

## Blast radius
Production impact is narrow/latent (a single-OS Linux process only sees POSIX paths → guard early-returns → correct; the mangle needs a Windows-shaped path to cross into a Linux process). Primary value is closing the **test-correctness hazard** (the #1139 class) and making the function internally coherent.


apmantza:
Review: **approve after rebase** — the win32.dirname coherence fix is verified correct (traced: Linux walk terminates at the win32 drive-root, no infinite loop/throw; zero Windows regression since win32.dirname===dirname there; idempotent stable key). The sweep's core argument holds (only a win32-committed branch mixing platform-default dirname is buggy; path-utils is the sole such file — verified). ast-grep rejection agreed (good/bad uses syntactically identical). Two blockers cleared: (1) file-role.ts:71 follow-up now filed as #1152 (was PR-body-only); (2) rebasing now to resolve the CHANGELOG conflict — which ALSO lets ci.yml's Linux Unit tests job run (it was skipped while the PR was DIRTY, so the Linux green this fix exists to earn hadn't executed). Merging once that CI is green.

PR #1175, opened 2026-08-08
fix switch-case-termination false positives for returning try/catch and exhaustive conditionals

### LINKED ISSUE #1079: fix switch-case-termination false positives for returning try/catch cases
## Problem

The Tree-Sitter `switch-case-termination` rule reports false positives in `clients/tree-sitter-client.ts` for post-filter switch cases that return on every path:

- `differs_only_by_case`
- `in_default_package`
- `missing_mimetype_and_download_name`

The exact reported coordinates may drift with formatting, but the cases are semantically exhaustive.

## Root cause

`no_terminating_statement` / its post-filter unwraps trailing blocks but does not conservatively analyze `try/catch/finally` or exhaustive conditionals. It also does not honor the documented `// fallthrough` marker.

## Acceptance criteria

- Add bounded, fail-safe control-flow handling for `try/catch/finally` and exhaustive conditional branches.
- Honor intentional fallthrough comments.
- Add TypeScript and JavaScript regression fixtures for these cases, grouped labels, try/finally, and intentional fallthrough.
- Preserve real findings for cases that genuinely fall through.
- Run the real Tree-Sitter rule/dispatch tests and full lint/test gates.

Related investigation: three false positives observed while reviewing `tree-sitter-client.ts`.

AngriestBird:
Closes #1079

## What

The Tree-Sitter `switch-case-termination` rule reported false positives for switch cases that return on every path but end in a `try/catch` or an exhaustive `if/else`. The `no_terminating_statement` post-filter only checked whether a case's last statement was a literal terminator (`break/return/throw/continue`), unwrapping only trailing `statement_block`s.

## Why

Three post-filter cases in `clients/tree-sitter-client.ts` (`differs_only_by_case`, `in_default_package`, `missing_mimetype_and_download_name`) all end in `try { return … } catch { return … }`. The last statement is a `try_statement`, which isn't in the terminator set, so the rule flagged them even though every path returns.

## How

`no_terminating_statement` now runs a bounded, fail-safe recursive `statementTerminates(node, depth)` analysis:

- **Terminators** (`break/return/throw/continue`) → true.
- **Wrapper nodes** (`statement_block`, `else_clause`, `catch_clause`, `finally_clause`) → recurse on the trailing statement; an empty wrapper falls through.
- **`try_statement`** → a returning/throwing try completes after `finally`, so only a catch that handles a thrown try and falls through can still reach the end; a try body that completes normally only terminates via a `finally`.
- **`if_statement`** → terminates only with an `else` and both branches terminating; no `else` still falls through.
- **Anything unrecognized** → false (falls through), so a blocking rule never under-reports. Depth-capped at 20.

A new `hasFallthroughMarker` honors an intentional `// fallthrough` comment (ESLint `no-fallthrough` convention, `/falls?\s?through/i`), checked against comment nodes only — never the case value — so a `case "fallthrough":` value can't suppress a finding. The JavaScript grammar attaches a trailing `// fallthrough` to the switch body as the case's next sibling, so that sibling is included too.

Genuine fall-throughs are still reported (verified: a plain statement, a non-terminating `try/catch`, and an `if` without `else` all still fire; the three issue cases no longer do).

## Tests

TS + JS regression fixtures in `tests/clients/tree-sitter-884-rules.test.ts` (real client + real query loader): try/catch returning on all paths, a catch that falls through, try/finally with a returning try, try/finally that falls through, exhaustive if/else, if without else, intentional `// fallthrough`, and grouped labels. Full suite + `npm run lint` green.


AngriestBird:
This ticket is ready to go for this.

Issue #1197, opened 2026-08-09
word-index startup refresh becomes quadratic when most documents are stale, blocking the event loop for 22s and warmup for 217s


apmantza:
## Defect class

A bulk “incremental” refresh mutates one document at a time even when nearly the entire word index is stale. Each replacement filters shared posting arrays, so high stale-document density turns the refresh into repeated global work (effectively quadratic) instead of switching to a linear full rebuild. The synchronous per-document tokenization/removal work is only count-yielded, so one bulk pass can also monopolize the host event loop.

## Production evidence (2026-08-09)

On pi-lens itself:

- interactive `session_start_total`: 102 ms
- background word-index refresh: **216,759 ms**
- documents: 2,061 refreshed, 0 reused, 8 dropped
- overall warmup: ~223 s
- same process recorded a genuine **22,515 ms** `loop_block` with `suspectSystemStall:false`

Smaller all-stale refreshes show the same shape: plegma 175 docs / 9.25 s; pi-webaio 207 docs / 11.09 s. The trigger was a persisted index missing/invalidating freshness metadata, which is a supported migration/freshness state, not an exceptional input.

## Invariants

1. The refresh chooses incremental mutation only when its estimated work is actually sparse; a dense stale set switches to a full rebuild before mutating the old index.
2. Decision/preflight is atomic: churn/staleness fallback occurs before drops or posting edits, so the old index remains valid if rebuilding fails or the session is superseded.
3. Full rebuild is built separately and published only after completion.
4. Both incremental and full paths remain bounded by the existing source-file/byte caps and abort when the session generation is no longer current.
5. No individual bulk indexing burst should hold pi's event loop for more than the repository occupancy budget; yielding must be based on bounded work/time, not only “every 100 files.”
6. Coverage remains honest (`truncated`, skipped/unreadable counts), and query/index results remain byte-equivalent to a from-scratch build.

## Supported transitions

- fresh metadata + sparse stale/new/deleted docs → incremental refresh
- file-set churn above threshold → full rebuild, before mutation
- stale-document density above threshold → full rebuild, before mutation
- legacy/missing incremental metadata → full rebuild
- unreadable file during sparse incremental refresh → retain its old posting and retry later
- superseded session during preflight/build → abort without publishing a partial replacement

## Test matrix

- corpus size: small / 500+ / cap-bound
- stale density: 0 / one file / just below threshold / at threshold / just above / 100%
- operation mix: stale-only / add / delete / mixed
[...]

Issue #947, opened 2026-07-30
Session start: full-mode work never runs (LSP pre-warm is dead code), and the project snapshot is sync-parsed on the critical path even when stale


apmantza:
From a session-start audit (real dogfood logs + benches, 2026-07-30).

**Full mode is dead code interactively.** 31k-line sessionstart.log: 82 `startup mode: quick`, **0 `full`**, 0 `launching background scans` — the first-session-of-process quick heuristic (`clients/runtime-session.ts:1118-1127`) means every pi launch takes quick mode, so the full path (dominant-language LSP pre-warm `:1636`, preinstall, scheduleStartupScans) structurally never executes. Measured consequence: every first file touch pays cold LSP — `lsp_client_wait_timeout` 750ms, json server spawn 2489ms (2026-07-30 logs). The +2s warmup pass only does scan-context/language-profile/word-index.

**Snapshot loaded synchronously, usually wastefully.** Quick mode sync-loads `project-snapshot.json` (`:1310` → `json-cache-read.ts:39` `JSON.parse(readFileSync)`) BEFORE checking freshness — and word-index stats show the snapshot is stale in **71% of sessions** (54 rebuilt vs 22 reused), so the parse is usually thrown away. Benched: 110-130ms sync at 40MB; est. 0.5s+ at the observed 112MB (#919 reporter). The file is pretty-printed (`JSON.stringify(snapshot, null, 2)`, `project-snapshot.ts:127`), ungzipped — and `project-snapshot.meta.json` containing exactly `{timestamp, version, seq}` is **written (`:128-139`) and read by NOTHING** (grep confirms).

**Saves re-parse and sync-rewrite the whole file 2-3×/session.** `saveRuntimeProjectSnapshot` (`project-snapshot.ts:184-232`) full-parses to merge, then sync stringify (91ms@40MB) + `writeFileSync` — ~300-600ms of event-loop blocks landing right as the user starts typing.

**Word-index fully rebuilt in 71% of sessions** (458-619ms background) though the per-edit incremental updater already exists.

Plan (ranked):
1. Fold `igniteDominantLanguageWarm` into the +2s warmup pass (fire-and-forget, generation-guarded) — kills the 0.75-2.5s first-touch LSP stall. Small effort.
2. Meta-first staleness gate: read `project-snapshot.meta.json`'s `seq` before the big parse; skip when stale. Small.
[...]

PR #1200, opened 2026-08-09
fix(word-index): bound dense startup refresh

apmantza:
## Summary

- select a full rebuild before mutating the old index whenever the stale set's estimated **work** exceeds one full rebuild — density is not a bound
- make expected legacy/file-set/dense/supersession transitions non-exceptional outcomes rather than failures
- give every bulk path (cooperative full builder **and both refresh loops**) an ~8ms monotonic yield budget OR'd with its item checkpoint, including within large documents and after any very long line
- preserve sparse incremental retry behavior and atomic publication on session supersession

refs #1197. Deliberately not `Closes`: three residuals (walk-phase occupancy, within-one-document-replacement occupancy, preflight/rebuild double-walk) are re-homed as a comment on the issue, per the repo's deferral-hygiene rule.

## Defect class and production evidence

Per-document replacement filters the shared posting arrays for that document's tokens. Sparse updates are cheap; repeating this for a large fraction of the corpus becomes effectively quadratic.

Observed on pi-lens:

- 2,061 refreshed, 0 reused, 8 dropped
- incremental refresh: **216.8s**
- genuine event-loop block: **22.5s** (`suspectSystemStall:false`)
- measured linear full build before the fix: **7.5s**

## Why a density ratio is not a bound (review finding 1)

The per-document cost **grows with the corpus**, so no fixed ratio — and no fixed absolute ceiling either — bounds the total work. A stale set comfortably *under* 30% still reproduced the #1197 shape on a corpus *smaller* than production's 2,061 documents.

Probe: 800 documents, 239 stale = 29.875%, just under the ratio gate.

| | before (PR head `d874fa2e`) | after |
|---|---|---|
| outcome | `incremental` (239 refreshed) | `full-required: stale-document-churn` |
| refresh total | **90,577 ms** | **126 ms** (preflight only) |
| max sync event-loop block | **39,553 ms** | **23.5 ms** |
| end-to-end incl. the full build that follows | 90,577 ms | **1,155 ms** |

**78x faster end to end, with a 1,680x smaller event-loop block.** At the 6,000-file default cap, up to 1,799 stale documents previously stayed on the quadratic path.

The gate now compares estimated work, both sides denominated in tokenizer tokens:

```
incremental ≈ stale x (docTokens x postingLength x scanCost + readCost)
full        ≈ totalTokens + corpusSize x readCost
```

This is self-bounding in a way no constant is: the incremental path is taken only while its total estimated cost stays under **one** full rebuild, whatever the corpus size. Both statistics come from cheap metadata passes (posting-array `.length`, forward-entry `.size`); neither touches a posting element.

Calibrated against this repository's own 2,062-document corpus — 213 ns per posting-element scan against 979 ns per token — and the resulting predicted crossover (23 stale documents) lands on the directly measured one (~21, from 97.8 ms per document replacement against a 1,436 ms rebuild). The 32-document floor now guards only the *ratio* test, its documented job ("one stale file in a three-file project is 33% but not dense"); letting a constant suppress the work test would reintroduce the same unbounded-on-the-other-axis shape.

Sparse behavior is preserved: 800 documents / 1 stale still resolves `incremental`.

## Cooperative yielding (review findings 1 and 4)

- Both refresh loops now carry the same monotonic budget as the async builder. Count-only checkpoints bound nothing when per-item cost is unbounded — 239 stale documents yielded **twice** across 18s, which is also why session-generation abort latency was unbounded. The count checkpoint is kept **alongside** the budget (OR, never AND), so abort latency stays bounded in items even on corpora too cheap to exceed the budget.
- The per-50-lines in-document checkpoint only holds for hand-written source: a minified/bundled single-line file up to `WORD_INDEX_MAX_BYTES` got **zero** in-document yields. Any line of 4,096+ characters is now a checkpoint too, making the cooperativeness claim true per byte rather than per line count.

## Supersession no longer aborts warmup (review finding 2)

`buildWordIndexAsync` throws on supersession, and `buildOrRefreshWordIndex` is awaited from the quick-mode warmup pass **outside** any per-step try/catch. A new `session_start` during a long warmup (the #1197 evidence shows a 223s one) therefore skipped everything after it — `warmup_word_index` was never emitted, `__piLensWarmupScheduled` was reset, and the **#947 LSP pre-warm never ran**. On master this path was synchronous and supersession was a graceful `return`; both seams now absorb it and return `undefined`, while still rethrowing genuine errors.

## State model / invariants

- fresh metadata + stale set cheaper than a rebuild → incremental mutation
- estimated stale-set work > one full rebuild → `full-required: stale-document-churn`, before mutation
- ≥32 stale documents AND stale density >30% → `full-required: stale-document-churn`, before mutation
- file-set churn >30% → `full-required: file-set-churn`, before mutation
- missing legacy incremental metadata → `full-required: missing-incremental-metadata`
- session supersession → abort without publishing, and without escaping into the caller's warmup pass
- unreadable file during sparse incremental refresh keeps its old posting and is retried later
- full replacement remains private until complete and current-generation checked
- coverage (`truncated`/`skipped`) and BM25/tokenization semantics are unchanged

## Blast radius

- `clients/word-index.ts`: production callers are `clients/runtime-session.ts` and the cold-query path inside `word-index.ts`; the synchronous `buildWordIndex` remains the test/small reference primitive
- `clients/runtime-session.ts`: `buildOrRefreshWordIndex` is reached from quick/full session startup warmup tasks; no foreground `session_start` work changed
- callback/lifecycle surface: detached startup task and cold-query fire-and-forget build; both keep existing generation/home-root guards. No new timers, workers, or referenced handles

## Test matrix covered

Every new or changed test was verified to fail against pre-fix code (`git show <ref>:clients/word-index.ts` swapped in and rebuilt), against **both** baselines:

Fails on PR head `d874fa2e`:

- 600-document dense-stale refresh at 25% (under the ratio) → `full-required` — #1197's 500+ document acceptance criterion; pre-fix returns `incremental`
- refresh loop yields on a time budget — 40 documents means the every-100 checkpoint never fires; pre-fix calls `shouldContinue` exactly once
- long-line cooperativeness — four ~410 KB documents of 30 very long lines; pre-fix max-block/total is 1/4, post-fix ~1/120
- quick-mode warmup still fires the #947 LSP pre-warm when the build is superseded mid-warmup

Fails on `master` (the review's finding 3 — these two previously passed pre-fix):

- ratio boundary now uses 200 documents so the 32-document floor is not the deciding branch: 60/200 = exactly 0.30 → incremental, and a new sibling at 61/200 = 0.305 → `full-required`. (The 60-side is a non-regression bound and cannot fail on master by construction, since master never returns `full-required`; the 61-side does.)
- "aborts during stat preflight" now uses an entirely stale corpus, so the refresh loop *would* read and mutate if the abort reached it, and asserts **zero reads** under the project root — master mutates the first 100 documents and fails
- `selects an atomic full rebuild before dense stale mutation` — unchanged, still fails on master

Retained: sparse one-stale-file and unreadable-file retry, file-set-churn transition, legacy/misaligned size metadata self-heal, mixed path separators, byte-equivalence of the cooperative builder, abort without partial state.

Tightened: scaled 1,000×100-line occupancy from **300 ms → 100 ms** (the budget is 8 ms and this fixture measures 20 ms; the old bound left 37x slack).

## Verification

- `npm run build`, `npm run lint`
- twelve word-index suites: **85 passed, 1 skipped**
- eleven runtime-session + project-snapshot suites: **73 passed**
- pre-fix failure verified for every new/changed guard against both `d874fa2e` and `master`


apmantza:
## Adversarial review — verdict: FIX-BEFORE-MERGE

Directionally right, byte-equivalent, no data-loss or correctness regressions found. But the central claim — that dense refresh is now bounded — is **false one file below the threshold**, and the #1197 symptom reproduces against PR head. Merging as `Closes #1197` would be defect-shape 10 (silencing counted as fixing).

Run in a detached scratch worktree at `d874fa2e`; no tracked file touched, no `git stash`.

### 1. [HIGH] The bound is relative-only — sub-30% stale density still runs the quadratic path

`word-index.ts:644-650` (gate), `:702-705` (yield). The dense gate requires **both** `staleDocuments >= 32` **and** `stale/size > 0.3`. There is no absolute ceiling. `removeWordIndexDocument` (:348-354) filters every posting array for each of a document's tokens, so cost is `O(staleDocs x corpusSize)` regardless of ratio. At the 6,000-file default cap, up to **1,799 stale documents** stay on the incremental path.

Reproduced on PR head — 800 docs, 239 stale = 29.875%, just under the gate:

```
incremental: 18,083 ms total, max sync event-loop block 7,312 ms
full build:     187 ms total, max sync event-loop block     9.3 ms
```

**97x slower, with a 7.3-second event-loop block** — the same shape as the issue's 216.8 s refresh / 22.5 s `loop_block`, on a corpus *smaller* than production's 2,061. 29% staleness is routine: branch switch, rebase, subtree format, generated-source rebuild.

This also breaks issue invariant 5: the incremental loop still yields on `++processed % 100` with **no time budget** — the exact count-yielded-not-time-yielded pattern #1197 calls out. 239 stale docs means 2 yields across 18 s. The monotonic budget was added to `buildWordIndexAsync` but not to the path that actually caused the outage. That is also why session-generation abort latency is unbounded — `shouldContinue()` at :704 can be ~7 s away. Defect shape 9 (bounded on one axis while it grows on another).

**Fix:** gate on estimated *work*, not density — `full-required` when `staleDocuments * postings.size` (or `staleDocuments * avgPostingLen`) exceeds the estimated full-build cost (`totalTokens`), or minimally add an absolute ceiling alongside the ratio. Independently, give the refresh loop the same `performance.now()` budget as `buildWordIndexAsync`.

### 2. [MEDIUM] Supersession now throws out of `buildOrRefreshWordIndex`, aborting the rest of quick-mode warmup including LSP pre-warm

`runtime-session.ts:759-761`, thrown from `word-index.ts:295/312/319`, caught at `runtime-session.ts:1698-1700`. On master this was `runtime.wordIndex = buildWordIndex(docs)` — synchronous, no throw, supersession handled by a graceful `return`. The replacement **throws** `"word index build superseded"`, and that throw is **outside** the try/catch at :684-744.

Scenario: quick-mode warmup on a large repo (the #1197 evidence shows a 223 s warmup); a new `session_start` bumps the generation mid-build; the throw propagates out of `buildOrRefreshWordIndex`. Consequences: `warmup_word_index` latency never emitted, **LSP pre-warm at :1654+ never runs** (the #947 fix silently skipped), and `__piLensWarmupScheduled` is reset to false.

This also contradicts the PR's own stated principle: it converts expected refresh transitions into non-exceptional `full-required` outcomes, then introduces a *new* exception for an equally expected transition that it lists in its own state model.

**Fix:** catch the supersession error inside `buildOrRefreshWordIndex` (or have `buildWordIndexAsync` resolve `undefined` on abort) and return `undefined`, matching master's graceful behavior. The full-mode path (`runTask` at :1345) already has the catch; only the shared warmup path is exposed.

### 3. [MEDIUM] Two of the four new/changed tests pass on pre-fix code — defect shape 7

Verified empirically by swapping in `git show master:clients/word-index.ts` and rebuilding:

```
x selects an atomic full rebuild before dense stale mutation (#1197)  <- correctly fails pre-fix
v keeps exactly-threshold stale density incremental                    <- PASSES pre-fix
v aborts during stat preflight without mutating the old index          <- PASSES pre-fix
```

- `word-index-session-refresh.test.ts:208` uses 100 files / 30 stale. 30 is below the 32-doc floor, so **the ratio comparison is never the deciding branch** — the test named after the 30% boundary never exercises it. The real boundary (200 docs / 60 stale = exactly 0.30 -> incremental; 61 -> full) is untested in both directions.
- `:262` does not cover the newly added stat-loop check at `word-index.ts:615-618`: with 200 fresh files, pre-fix code reaches the refresh loop's pre-existing `%100` check and throws the identical message with an unmutated index. Assert the abort happened *before* the churn computation (count `statSync` calls, or use a corpus the refresh loop would mutate).
- `word-index-async-build.test.ts` cannot fail pre-fix at all (new API). Its occupancy assertion is `< 300 ms` against an `8 ms` budget — 37x slack; the probe measured 9.3 ms, so a real tightening is affordable.
- #1197's acceptance criterion asks for a guard at the **500+ document** dense-stale shape; the largest dense-refresh fixture here is 100 docs. Finding 1 is precisely what a 500+ guard would have caught.

### 4. [LOW] "Cooperative within large documents" does not cover long lines

`word-index.ts:308` yields on `(i + 1) % 50 === 0`. A minified or bundled single-line file up to `WORD_INDEX_MAX_BYTES` (512 KB, :448) that the source filter did not exclude gets **zero** in-document yields — `tokenizeLine` plus posting pushes for that line is one unbroken synchronous burst well above the 8 ms budget. Bounded by the byte cap, so modest, but the AGENTS.md line this PR adds claims "including within large documents", which is true per-line-count, not per-byte. Consider checking the budget after any line whose length exceeds a threshold.

### 5. [LOW] `full-required` discards a completed walk+stat pass and re-walks

`word-index.ts:645-650` into `runtime-session.ts:755`. The preflight already walked and `statSync`'d the whole bounded set; the fallback then runs `collectWordIndexDocs`, walking and statting again. Pre-existing for `file-set-churn`, now on the much more common `stale-document-churn` path. Passing the preflight's `current` map into the rebuild would halve cold-FS startup cost.

### Checked and found clean

- **Not exposed to the `atomic-write.ts` same-process tmp collision (#1205).** Serialized by construction: `saveProjectSnapshot` writes meta via the **synchronous** `writeFileAtomic` (`project-snapshot.ts:789`) — `writeFileSync` plus `renameSync` in one uninterrupted JS turn, so two same-process calls can never be in flight. The body goes through the persist worker to a **per-generation** stage path (`${gzPath}.stage-${pid}-${generation}`, :827) promoted by a generation-gated `renameSync` (:625) — distinct paths per save, no shared tmp name. The degraded fallback (:579) is again the sync writer. Nothing on the word-index / runtime-session / project-snapshot path calls `writeFileAtomicAsync`, and there is no fire-and-forget `void write...()` on these targets. **The PR neither introduces nor widens an overlapping-write window.** *Note for #1202, not this PR: the persist worker shares `process.pid` with the main thread, so if worker-side code ever calls `writeFileAtomic` on a main-thread target, sync-ness will not protect it.*
- **Abort actually aborts.** `refreshWordIndexIncrementally` mutates a locally deserialized index, never `runtime.wordIndex`; publication happens only after a post-await `isCurrentSession` re-check (`runtime-session.ts:690, 702`). `buildWordIndexAsync` keeps its index private until resolve. A superseded refresh can never publish. (Abort *latency* is the caveat — finding 1.)
- **No silent skipping in the incremental path.** An unreadable stale file `continue`s **without** updating `fileMtimes`/`fileSizes` (:685-691), so it stays stale and is retried next session; `skipped` is counted and surfaced. New files count as stale via `undefined !== mtimeMs`; legacy indexes via `?? -1`. No document can be missed.
- **Byte-equivalence of the refactor.** `indexWordLine` / `finishWordIndexDocument` / `createEmptyWordIndex` are a faithful extraction; `docCount` incremented per document is identical to the old `files.length`, including for duplicate paths. Confirmed by the new serialize-comparison test and a full targeted run: **6 passed, 1 skipped (26 tests)** across the seven word-index suites.
- **Shared-handler trap.** Grepped every `refreshWordIndexIncrementally` / `buildWordIndex*` call site repo-wide: exactly one production consumer (`runtime-session.ts`) plus the in-file cold path, and all five sibling test files that assert on the return shape were updated with the `mode` narrowing. No unupdated consumer reads `.refreshed` off a `full-required` object.
- **Defect shapes.** 1 (path-keyed divergence): clean — the dense loop reads through `PathKeyedMap.get` with the raw walk path, same normalizer as the existing gate. 2 (host-default path fn): clean — no path parsing added. 6 (freshness stamp): clean — the dense preflight uses the same mtime+size pair as the refresh gate, so it cannot disagree with it. 4 (handles): clean — no timers, workers, or children; `setImmediate` only, matching the precedent already in `collectWordIndexDocs`; no starvation or livelock. 11 (skipped CI): clean — PR is `MERGEABLE`/`CLEAN`, and Unit tests plus Lint & type-check both actually ran and are SUCCESS on head.

**Minimum before merge:** findings 1 and 2. If you would rather ship the improvement now, at least land finding 2 and change `Closes #1197` to `refs #1197` — the invariant the issue is written around is measurably not met.

Issue #1199, opened 2026-08-09
safeSpawnAsync ignores caller PATH on Windows, breaking managed Knip with spawn ENOENT


apmantza:
## Root cause

Repeated managed-Knip failures on Windows are caused by `safeSpawnAsync` resolving a bare command against **`process.env.PATH` only**, before spawning, while the caller supplied the managed `.bin` directory through `options.env.PATH`.

Exact chain:

1. `KnipClient.ensureAvailable()` discovers/installs Knip under `~/.pi-lens/tools/node_modules/.bin`.
2. `getKnipEnvironment()` correctly prepends that directory to the child environment.
3. `KnipClient.runAnalyze()` invokes bare `knip`.
4. `safeSpawnAsync()`'s Windows pre-spawn resolver ignores `options.env.PATH`/`Path` and synthesizes `spawn knip ENOENT` before the child can use its supplied environment.
5. The resolver cache key also omits effective PATH/PATHEXT, so a miss can survive a later call whose custom environment contains the tool.

This is not primarily a deleted/corrupt install: both `knip` and `knip.cmd` exist and the `.cmd --version` command succeeds. Logs retain ~140 Knip failures; apparent reinstall attempts often rediscovered the extensionless sibling rather than reinstalling.

## Defect class / sibling sweep

Any Windows caller that invokes a bare command made available only through `options.env.PATH` is vulnerable. Managed Madge/jscpd/Biome usually pass absolute paths and avoid this exact chain, but installer fallback can return an extensionless npm shim after verifying its `.cmd` sibling, leaving a second shared path-form hazard.

## Invariants

1. Command resolution uses the same effective environment the spawned child receives.
2. Windows PATH lookup honors case-insensitive `PATH`/`Path` and effective PATHEXT without mutating caller input.
3. Resolver caching cannot cross-serve between different effective PATH/PATHEXT values.
4. An absolute command path remains independent of PATH lookup.
5. Managed npm discovery returns the actual executable shim (`.cmd` on Windows), not an extensionless sibling merely because that sibling exists.
6. A resolver/install failure is surfaced as unavailable, never as a clean analysis result.
7. Existing timeout, abort, UTF-8, shell, and process-tree-kill behavior remains unchanged.

## Test matrix

- command form: bare / absolute / extensionless npm shim / `.cmd`
- environment: inherited PATH / caller-only PATH / caller `Path` casing / custom PATHEXT / empty PATH
[...]

PR #1201, opened 2026-08-09
fix(spawn): honor child PATH during Windows resolution

apmantza:
## Summary

- resolve Windows bare commands against the exact case-insensitive child `PATH`/`PATHEXT` environment passed to spawn
- include effective cwd, PATH/PATHEXT presence and values, and validated per-drive cwd provenance in a bounded resolver-cache identity
- parse explicit, relative, drive-relative, and PATH-entry forms with `path.win32`, failing closed when cross-drive provenance is unavailable
- cover the real Knip environment-construction seam plus pure cross-platform resolver and live Windows `.cmd` paths

Refs #1199. Not closing — see "Residual scope" below; three acceptance-criteria items from #1199 remain open and are tracked as a comment on that issue.

## Defect class

`safeSpawnAsync()` merged `options.env` into the spawned child, but its pre-spawn Windows resolver searched only ambient `process.env.PATH`. A managed shim could therefore be present in the child environment yet rejected as `ENOENT` before spawn. The cache also lacked all environment inputs that determine resolution, allowing one caller to poison another.

## Invariants / supported transitions

- Windows environment names are case-insensitive; an explicit child spelling replaces every ambient spelling
- resolver and child receive the same merged environment (Windows AND POSIX, sync and async)
- absent PATHEXT uses the Windows default; explicitly empty PATHEXT means no implicit extensions
- an explicit extension is an exact candidate only when it's an extension this resolver (or the caller's own PATHEXT) actually recognizes as executable; otherwise the PATHEXT-append loop still runs against the full command, so versioned interpreters (`python3.11`, `node-v20.1`) resolve correctly and non-executable files (`foo.txt`) never resolve as spawnable
- absolute/relative/UNC parsing uses `path.win32` regardless of test host, and a *rooted* path (`\tools`, drive-root-relative but naming no drive) is never treated as fully qualified — it's anchored to the correct drive instead of statSync'd against whatever drive the host process happens to be on
- relative PATH entries resolve against canonical effective child cwd
- same-drive drive-relative paths use that cwd; cross-drive forms require validated absolute `=X:` provenance and otherwise fail closed
- an unprovable drive-relative **child cwd** (no validated `=X:` entry — Node never surfaces those from ambient `process.env`) falls back to the raw value for the actual spawn, matching pre-#817 passthrough behavior, instead of failing the whole spawn closed before ever attempting it; command *resolution* still fails closed for PATH entries that would need that cwd to be canonical
- cache identity includes every lookup input and is capped at 256 session entries; session reset clears positive and negative results
- `.exe`/`.com` remain direct `shell:false` spawns; `.cmd`/`.bat` retain pinned cmd.exe validation
- timeout, abort, UTF-8, resource sampling, and process-tree cleanup are unchanged

## Blast radius

`clients/safe-spawn.ts` is the shared subprocess seam for dispatch runners, Knip, installers, formatters, security scans, and LSP runtime installs. `module_report(..., blastRadius:true)` was run before/after; the cached graph had no dependent-node payload, so callers were manually swept. `clients/runtime-session.ts` only adds resolver-cache reset to the existing session reset sequence. Installer discovery remains ambient-only by design; LSP launch passes explicit environments directly. Knip is the confirmed managed-environment offender.

## Tests

- caller-only PATH/Path and conflicting ambient path
- PATH/PATHEXT casing and cache separation
- absent vs empty PATHEXT in both call orders
- explicit extension and PATHEXT order, including versioned/unknown suffixes (`python3.11`) and non-executable suffixes (`foo.txt`)
- effective/relative cwd and relative PATH entries, including rooted (`\tools`) forms anchored to the correct drive
- same-drive/cross-drive commands, PATH entries, and cwd with valid/invalid `=X:` provenance
- positive/negative cache reset and count bound
- sync overridden-PATH `.cmd` execution on Windows
- `KnipClient.runAnalyze()` project `.bin` + managed-bin environment construction
- **new**: `tests/clients/safe-spawn-windows-env-plumbing.test.ts` — forces `process.platform` to `"win32"` (same technique as `safe-spawn-kill-escalation-timer.test.ts`) and mocks `node:fs`/`node:child_process` so the *dispatch* functions' env-plumbing into the resolver is asserted end-to-end without needing a real Windows host; runs on Linux CI

## Linux-CI test gap (adversarial review finding)

An earlier review round found that the only tests exercising real spawn behavior (`safe-spawn-windows-command.test.ts`'s `(a2)` Knip `.cmd` case and the sync overridden-PATH case) sit under `describe.runIf(process.platform === "win32")` and skip on Linux CI, while the cross-platform resolver suite only covers the pure `resolveWindowsCommandForEnvironment` export directly — not whether `safeSpawnAsync`/`safeSpawn` actually feed it the caller's env. A revert of that plumbing would have turned nothing red on the CI OS. `tests/clients/safe-spawn-windows-env-plumbing.test.ts` closes that gap (see Tests above); confirmed it fails against a reverted plumbing and passes against the current code.

## Residual scope (deferred, tracked on #1199)

Posted as a comment on #1199 (issue stays open, not closed by this PR):
1. Sibling managed-npm-path tests exist only for Knip, not Madge/jscpd/Biome.
2. Nothing here touches ENOENT retry/backoff bounding.
3. Probe-cache discovery in `clients/installer/index.ts` still records the extensionless path; the resolver papers over it at the spawn seam, but the underlying recorded path is unchanged.

## Deferred (minor, noted not fixed)

- The `=X:` per-drive provenance apparatus (`getValidatedPerDriveCwd`, `resolveDriveRelativeWindowsPath`, the `perDriveCwds` cache-key term) is unreachable from Node's *ambient* `process.env` (verified: `Object.keys(process.env).filter(k => k.startsWith("="))` is empty) — kept, now documented in-code, because a caller can still supply it explicitly (as the test suite does).
- `getWindowsEnvironmentValue` is last-key-wins over `Object.entries`, so its answer is technically key-order dependent for a caller-constructed object with duplicate case-insensitive keys — narrow edge, not fixed.
- `cacheWindowsCommandResult` evicts an unrelated oldest cache entry even on a plain key overwrite (`.set` on an existing key still checks `.size >= MAX` first) — harmless (session-scoped, cheap to recompute), not fixed.

## Verification

- final focused suites: **60 passed** (previous checkpoint)
- post-adversarial-review targeted re-run (safe-spawn-windows-resolution, safe-spawn-windows-command, safe-spawn-windows-env-plumbing [new], knip-client, safe-spawn-ambient-signal, safe-spawn-kill-escalation-timer, safe-spawn-resource-usage, safe-spawn-sync-throw): **84 passed**
- prior candidate full suite: **5,684 passed, 17 skipped**
- `npm run build`
- `npm run lint`
- `git diff --check`


apmantza:
## Adversarial review — verdict: FIX-BEFORE-MERGE

The core fix is real and works (Knip's managed `.cmd` now resolves from `options.env`), but the resolver rewrite introduces a proven regression of the *same* failure mode it fixes, plus an OS-agnosticism hole. All findings were reproduced against built branch code, not inferred. A follow-up agent is clearing these on the branch.

### 1. [HIGH] Bare commands whose name contains a dot now hard-fail ENOENT — `clients/safe-spawn.ts:419-434`

Master gated on `hasKnownExt = pathExts.includes(existingExt)`. The PR replaced it with `hasExplicitExt = existingExt.length > 0`, so *any* dot in the basename is treated as an explicit exact-match extension and the PATHEXT append loop is skipped entirely.

Reproduced against the built branch with a real `python3.11.exe` on PATH:

```
extname('python3.11')  = ".11"
resolve('python3.11')  = null      <-- ENOENT
resolve('tool')        = {...\tool.exe, ".exe"}
```

Master resolves this correctly (`.11` is not in PATHEXT, so it falls through to the append loop). So `python3.11`, `pip3.12`, `node-v20.1` — any versioned interpreter from config or user input — goes from working to `spawn ... ENOENT`, which is the exact symptom class #1199 exists to fix. No in-repo call site uses a dotted command name today (grepped `clients/`, `tools/`, `index.ts`, installer `checkCommand`s, `pipCandidates`), so it is latent, not live — but it is a silent behavior narrowing, and the PR's own comment at :423-425 justifies the change only in the *opposite* direction.

Secondary: a command like `foo.txt` / `foo.ps1` that exists now resolves with `ext=".txt"` and falls into the **direct-spawn** branch (:774-779), handed to `spawn(..., shell:false)`. Master returned a clean ENOENT.

**Fix:** keep the explicit-suffix behavior *additively* — try `statIsFile(base)` first, then, when `existingExt` is not in `pathExts`, still run the PATHEXT append loop.

### 2. [MEDIUM-HIGH] Root-relative paths misclassified as absolute; stat and exec can target different drives (shape 2 / shape 7)

`path.win32.isAbsolute("\\tools")` is `true`, but `\tools` is *drive-relative*, not fully qualified. The PR carefully handles the `D:foo` form (`resolveDriveRelativeWindowsPath`, `getValidatedPerDriveCwd`) but treats the `\foo` form as absolute in three places: `resolveEffectiveWindowsCwd` (:323), `resolveWindowsPathEntry` (:352), `resolveWindowsCommandUncached` (:444).

Reproduced:

```
resolve('tool', cwd='D:\work', PATH='\Users\...\bin')
  => {"resolvedPath":"\\Users\\...\\bin\\tool.exe","ext":".exe"}
     isAbsolute? true    has drive? false
```

Concrete failure: host cwd on `C:`, caller passes `cwd: "D:\work"` and a rooted PATH entry `\tools`. The resolver `statSync`s `C:\tools\tool.cmd` and returns the drive-less path, which is interpolated into the `cmd.exe /c` line (:767) and executed with `cwd: D:\work`, where Windows resolves it to `D:\tools\tool.cmd`. **A different file is validated than is executed** — the same TOCTOU/wrong-drive hazard the drive-relative branch was written to prevent, left open on the sibling form. It also breaks the `ResolvedWindowsCommand.resolvedPath` contract at :233.

Shape 2/7 axis: for these inputs `path.win32.resolve` falls back to `process.cwd()`, which supplies a drive on Windows and **none** on Linux CI — any future assertion over these forms diverges by host OS.

**Fix:** classify with a *fully-qualified* predicate (drive-absolute `X:\...` or UNC `\\...`) rather than `path.win32.isAbsolute`, and anchor the rooted-relative form to the effective cwd's drive, or fail closed to match the `D:foo` policy.

### 3. [MEDIUM] An unresolvable **cwd** is reported as a missing **command** — :737-739, :1106-1114

```ts
if (spawnCwd === undefined) {
  resolutionError = synthesizeEnoentError(options?.cwd ?? process.cwd());
}
```

This yields `code:"ENOENT"`, `syscall:"spawn"`. Downstream call sites documented at :563-568 (`sg-runner.ts`, `lsp/launch.ts`) branch on `err.code === "ENOENT"` to mean *tool not installed* and mark unavailable or attempt reinstall. A bad cwd therefore misdiagnoses as a missing tool and can drive the repeated-reinstall loop #1199 itself documents ("~140 Knip failures... apparent reinstall attempts").

Related regression in the same lines: `resolveEffectiveWindowsCwd` returns `undefined` for a drive-relative cwd like `"D:work"` unless a validated `=D:` entry exists — and Node **never** exposes those keys (verified: `Object.keys(process.env).filter(k => k.startsWith("="))` is `[]`). Master passed `cwd: "D:work"` straight to `spawn`, where Windows resolved it natively and the child ran. The PR now fails closed without attempting the spawn.

**Fix:** use a distinct non-ENOENT error for cwd-resolution failure, and consider restoring the passthrough rather than failing on provenance Node cannot supply.

### 4. [MEDIUM] `Closes #1199` while three acceptance criteria are unmet

- *"Sibling managed npm path tests cover Knip, Madge, jscpd, and Biome"* — only Knip is covered.
- *"ENOENT retries are bounded/backed off"* — nothing in the diff touches retry or backoff.
- Invariant 5, *"Managed npm discovery returns the actual executable shim (.cmd)"* — `installer/index.ts` in this PR contains only the `finishInstallAttempt` refactor plus cache reset; discovery is unchanged. Mitigating: the new resolver *upgrades* an absolute extensionless `.bin/knip` to `knip.cmd`, so the spawn seam papers over it — but the probe cache still records the extensionless path, which the issue explicitly calls out.

Per the repo's deferral-hygiene rule this should be `refs #1199` with the residue re-homed as issue comments, not `Closes`.

### 5. [MEDIUM] The CI-visible tests cannot catch a regression of this bug (shape 7)

- `tests/clients/knip-client.test.ts` — the only test naming the real offender — asserts on `KnipClient.runAnalyze`'s env construction, but `clients/knip-client.ts` is **not in the diff**. Those assertions pass verbatim on master. Characterization, not regression.
- The two tests that genuinely fail on pre-fix code (`safe-spawn-windows-command.test.ts:125` and `:296`) live under `describe.runIf(process.platform === "win32")` and are **skipped on Linux CI**.
- `safe-spawn-windows-resolution.test.ts` does run cross-platform, but exercises a brand-new export that does not exist on master, so "fails on pre-fix code" is vacuous — it fails at *import*, not on behavior.

Net: on the CI OS, nothing would go red if the `options.env` plumbing at :727/:742 were reverted.

Also: `safe-spawn-windows-resolution.test.ts:255,281,314` call `process.chdir()` (safe under this repo's `forks` pool, would throw under `threads`), and :302-311 compares `process.cwd()` against `os.tmpdir()`, which diverges on macOS (`/var` vs `/private/var`).

### 6. [LOW] Sync `safeSpawn`'s POSIX branch computes the merged env and throws it away — :1104, :1180-1185

`getSpawnEnvironment(options?.env)` is computed unconditionally at :1104, but the non-Windows path at :1180 still spreads `options` into `spawnSync`, so `options.env` reaches the child as a **full replacement** (no `process.env` merge, no PATH). The PR's stated invariant "resolver and child receive the same merged environment" holds on Windows only. Latent (no current caller passes `env` to sync `safeSpawn`), but it also pays a full `{...process.env}` copy per non-Windows sync spawn for a discarded value.

### Minor

- The entire `=X:` per-drive apparatus (`getValidatedPerDriveCwd`, `resolveDriveRelativeWindowsPath`, the `perDriveCwds` cache-key term, 6 tests) is **unreachable in production** — Node does not surface `=X:` keys. It only fires when a caller synthesizes them, which only the tests do. Worth a comment saying so, or dropping it.
- `getWindowsEnvironmentValue` (:270-279) is last-key-wins over `Object.entries` — safe for merged envs, but the *exported* seam accepts arbitrary envs, making its answer key-order dependent.
- `cacheWindowsCommandResult` (:386) evicts an unrelated oldest entry even on a plain overwrite.
- A positive cache entry is only re-`stat`'d for its own path (:509); a higher-priority PATHEXT or earlier PATH-dir candidate appearing mid-session is never noticed. Safe direction, but asymmetric with the negative TTL.
- `resetUtf8ConsoleCodePageStateForTests` was deleted — unrelated scope creep in a PATH-resolution PR.

### Checked and clean

- **Cache-key completeness (2ee5fd27):** verified — the key covers `command`, PATH presence and value, PATHEXT presence and value (correctly distinguishing absent from `""`), canonical `effectiveCwd`, and the sorted `=X:` set. `resolveWindowsCommandUncached` reads nothing else beyond the filesystem, and positive hits re-`stat`. No under-keying.
- **Case-insensitive env merge:** `mergeWindowsEnvironment` correctly collapses `PATH`/`Path`/`pAtH` to one last-explicit-wins entry; traced end-to-end through `getKnipEnvironment` -> `getToolEnvironment` -> `getSpawnEnvironment`. The child receives exactly the env the resolver used (Windows).
- **Extensionless npm shim (issue invariant 5) at the spawn seam:** an absolute `.../.bin/knip` with both `knip` and `knip.cmd` present resolves to `knip.cmd`.
- **Trailing-dot attack (`"knip."`):** returns `null`, not the sh shim — Node's `statSync` does not strip the trailing dot.
- **cmd.exe surface unchanged:** interpreter still pinned to `%SystemRoot%\System32\cmd.exe` (never `ComSpec`), `findCmdUnsafeValue` still validates the **resolved** path plus args for `"`/`%`/`!`/CRLF, `windowsVerbatimArguments` still paired with `buildWindowsShellCommand`, `shell:false` everywhere, `.exe`/`.com` still direct-spawn. No new injection surface.
- **No implicit CWD-in-PATH:** bare commands are never searched in the cwd unless PATH says so; empty PATH entries are skipped rather than treated as `.` (safer than cmd.exe).
- **UNC paths:** `\\server\share\tool.exe` classifies as absolute and normalizes correctly.
- **Install-time invalidation:** `finishInstallAttempt` calls `resetSafeSpawnWindowsCommandCache()` on every successful strategy; session reset wired at `runtime-session.ts:1746`.
- **Tests:** `npm run build`, then vitest on `safe-spawn-windows-resolution`, `safe-spawn-windows-command`, `knip-client` -> 63 passed; siblings `safe-spawn-ambient-signal`, `-kill-escalation-timer`, `-resource-usage`, `-sync-throw` -> 19 passed. No sibling test asserts the old resolution behavior.

apmantza:
## Scoped re-review of the fix commits — verdict: FIX-BEFORE-MERGE

CI is fully green on `5c1f3f29`, so this is a pre-merge re-review of only the three post-review commits (`git diff 2ee5fd27..HEAD`). Everything before `2ee5fd27` was cleared by the first review and was not re-litigated.

Two blocking findings. Both were reproduced empirically against both HEAD and the `2ee5fd27` baseline.

### 1. [HIGH] `5c1f3f29` fixed the instance, not the class — and this is a regression versus the reviewed baseline

`clients/safe-spawn.ts:419-426` (`resolveEffectiveWindowsCwd`) and `:356-363` (`resolveRootedWindowsPath`).

`5c1f3f29` special-cased **only** `cwd === undefined`. The caller-supplied branch directly below it is untouched. A POSIX-absolute cwd such as `/tmp/proj` satisfies `isRootedWindowsPath` (leading separator, no drive letter, not UNC), so it goes to `resolveRootedWindowsPath(cwd, process.cwd())`, where `driveLetter(process.cwd())` is `undefined` on a drive-less host and the function returns `undefined`. `effectiveCwd` is then `undefined`, every relative and rooted PATH entry is skipped, relative commands get `base === undefined`, and resolution returns `null` **without a single `statSync` call** — bit-for-bit the same failure signature as the bug `5c1f3f29` exists to fix.

Measured with `process.cwd()` stubbed to `/home/runner/work/pi-lens/pi-lens` (the literal Linux CI shape), running the real exported `resolveWindowsCommandForEnvironment`:

| case | `2ee5fd27` (cleared baseline) | HEAD `5c1f3f29` |
|---|---|---|
| A: cwd omitted + relative PATH | resolves | resolves (the fixed instance) |
| B: cwd `/tmp/proj` + PATH `bin` | resolves | **null** |
| C: cwd `/tmp/proj` + command `tools/tool.exe` | resolves | **null** |
| D: cwd `/tmp/proj` + PATH `/tmp/proj/bin` | resolves | **null** |
| E: relative cwd `child` | resolves | resolves |
| F: `C:\` fixtures | resolves | resolves |

This is not a production bug on real Windows, where `process.cwd()` always carries a drive — hence FIX-BEFORE-MERGE rather than BLOCK. There is one real-Windows exposure: a UNC `process.cwd()` (`\\server\share\...`, which Node permits) also yields `undefined` from `driveLetter`, so a rooted caller cwd fails closed to a false ENOENT.

The test-facing risk is the sharper one. Any test that forces `process.platform = "win32"` — the technique this PR introduces and documents in `safe-spawn-windows-env-plumbing.test.ts:16-24` — and passes a real temp dir as `cwd` (the natural thing to write with `mkdtempSync`) will pass on a Windows dev box and return `null` on Linux CI. That is precisely how `cb60a3b8` shipped green in the first place.

Note the surviving asymmetry, which shows the class was not swept: `resolveEffectiveWindowsCwd` anchors rooted paths to `process.cwd()`'s drive (:425) while `resolveWindowsPathEntry` anchors them to `effectiveCwd`'s drive (:463) — two different provenances for the same "current drive" concept.

**Fix:** `resolveRootedWindowsPath` must not silently return `undefined` when the drive source is drive-less. Either fall back to `path.win32.normalize(value)` (the `2ee5fd27` behavior for this input) or probe the filesystem shape rather than string-classifying a value that may not be a Windows path at all — recurring defect shape 2.

### 2. [HIGH] The three commits' production changes have zero test coverage

`tests/clients/safe-spawn-windows-env-plumbing.test.ts` (all 204 lines, new in `3dc4c169`) and the hardened `tests/clients/safe-spawn-windows-resolution.test.ts` **both pass, 29/29, against `clients/safe-spawn.ts` reverted to `2ee5fd27`.**

So `3dc4c169` guards nothing that `cb60a3b8` or `5c1f3f29` changed. Grepping the whole `tests/clients/safe-spawn*` surface confirms there is no test anywhere for:

- `isFullyQualifiedWindowsPath` / `isRootedWindowsPath` / `resolveRootedWindowsPath` — no rooted (`\tools`) or UNC (`\\srv\share`) case
- the finding-1 explicit-extension fix — no `python3.11`, no `foo.txt`, no trailing-dot case
- the cwd-passthrough change — nothing exercises the `?? options?.cwd` fallback
- `5c1f3f29`'s own fix — no regression test was added for it

The env-plumbing test *does* have teeth against its own stated target: mutating both `resolveWindowsCommand(command, options?.cwd, spawnEnv)` call sites to `process.env` turns both tests red. But that invariant predates this fix round.

Also worth recording: `28a05b32`'s move to absolute `C:\` fixtures is what makes the suite systematically *avoid* the finding-1 landmine rather than detect it. The green Linux CI run is explained by case E above — the chdir tests pass a relative cwd, which still works on a drive-less host.

### 3. [LOW] `?? process.cwd()` in the cwd fallback is unreachable

`clients/safe-spawn.ts:884-888` and `:1270-1273`. `resolveEffectiveWindowsCwd(undefined, env)` now always returns a string (:408), so the `?? process.cwd()` arm is only reachable when `options.cwd` is defined-but-unresolvable — in which case `?? options?.cwd` already caught it. Dead code that reads like a live silent-wrong-directory hazard; delete it.

### 4. [LOW] `WINDOWS_DIRECT_RESOLVABLE_EXTS`' doc claim is not enforced at dispatch

`clients/safe-spawn.ts:238-245` says these are "the only extensions the rest of this module knows how to spawn", but dispatch routes `.cmd`/`.bat` to the wrapper and **everything else** to direct spawn. Verified: `PATHEXT=".TXT"` with `command="foo.txt"` resolves to `{ext:".txt"}` and would be direct-spawned. Not a security regression — this is strictly narrower than `2ee5fd27`, CreateProcess does no ShellExecute association, and the default Windows PATHEXT already contains `.VBS/.JS/.WSF/.MSC` which the pre-existing append loop could already surface. The comment simply overstates the guarantee.

### Trivia

`resetUtf8ConsoleCodePageStateForTests` (`:748`) is exported by `cb60a3b8` with zero callers including zero tests, despite its doc comment claiming a test observes it.

### Checked and clean

- **Cwd passthrough — finding 2 of the original review is NOT reintroduced.** `spawnCwd` and the resolver both call `resolveEffectiveWindowsCwd(options?.cwd, spawnEnv)` with identical inputs, so the directory a relative PATH entry is validated against is always the directory the child gets. The `?? options?.cwd` fallback fires only when that returns `undefined`, and in that state the resolver skips every relative and rooted PATH entry, so the only paths it can return are fully-qualified and cwd-independent. There is no route by which an explicit-but-unresolvable cwd silently resolves a command against `process.cwd()`. Relative commands fail closed to ENOENT.
- **Relaxed bare-PATH loop.** "Absolute" is decided by the new `isFullyQualifiedWindowsPath`, not the old `isAbsolute` (:455). Rooted entries take the separate branch at :456-464 and require `effectiveCwd`, failing closed. Correct.
- **Cache-key completeness.** Both new dependencies are covered: `isKnownExecutableExt` reads only PATHEXT and `command` (both keyed); the classification reads only `command`/`cwd`, and the `process.cwd()` dependency is folded in because the already-resolved `effectiveCwd` is keyed rather than the raw cwd. When `effectiveCwd` is `undefined` the resolver provably never reads `process.cwd()` (zero `statSync` calls, immediate `null`), so two different unresolvable cwds colliding on one key is safe.
- **Security surface — all clean.** cmd.exe pinned to `%SystemRoot%\System32\cmd.exe` at :920 and :1307, `ComSpec` appears only in comments and is never read. `findCmdUnsafeValue(resolved.resolvedPath, args)` still validates the resolved path plus args for `"`/`%`/`!`/CRLF (:908, :1292). `windowsVerbatimArguments` set only in the two `buildWindowsShellCommand` branches. `shell:false` on every spawn. Trailing-dot `"knip."` still returns null (tries `knip..exe`/`knip..cmd`, both miss). Empty PATH entries still skipped, not treated as `.` (verified `PATH=";C:\bin"` never stats the cwd). No implicit CWD-in-PATH. A hostile PATHEXT cannot widen the direct-spawn set beyond what `2ee5fd27` already permitted.
- **Classification input sweep.** 25 inputs enumerated against the live resolver — correct on `C:\x`, `c:\x`, `C:/x`, `C:x`, lone `C:`, `\x`, `/x`, `//server/share`, `\\server\share`, `\\?\C:\x`, `\\.\device`, `\\?\UNC\...`, bare `x`, `.\x`, `..\x`, empty string, single `\`, and rooted PATH entries anchoring to the child's drive. Only divergences: `\\\\srv\share` classifies as rooted rather than UNC (malformed either way, no impact), plus the drive-less cases in finding 1.
- **Test hardening.** `realpathSync` on `mkdtempSync` results correctly fixes a genuine macOS `/var` → `/private/var` divergence. `process.chdir` is safe under the pinned `forks` pool with `try/finally`. `process.platform` forcing is restored in `afterEach` and cannot leak cross-file.
- **Tests:** all 7 `safe-spawn*` suites 69/69, `knip-client` 15/15, with `npm run build` before every run.

Issue #1081, opened 2026-08-05
flaky: tests/clients/lsp/edits.test.ts occupancy guard exceeds 300ms under suite load


apmantza:
## Summary

`tests/clients/lsp/edits.test.ts` has a recurring occupancy assertion failure in the test `keeps large text/resource planning occupancy bounded` (around line 735). This is separate from the unrelated call-graph work being validated in PR #1072 / issue #1070; the failure is in the existing LSP workspace-edit occupancy guard, not in the changed call-graph path.

## Evidence

- The test creates 400 text changes plus 400 resource operations, asserts `planned === 800`, measures `__planWorkspaceEditForTest(...)` with `measureMaxSyncBlockMs`, and requires `maxBlock < 300` ms.
- The test already uses `{ timeout: 30_000, retry: 2 }`.
- Recent runs still observed `maxBlock` around 358–435 ms after retries.
- Existing history was checked: #902 is closed and its final scope was LSP sweep flush-count/Windows-spawn/startup-overhead items; merged #991 isolated only source-walk/source-filter/cascade occupancy guards. Neither tracks this `edits.test.ts` failure.

## Reproduction

Run the targeted test and the full suite from a clean checkout under the normal Vitest configuration; the failure is most likely under full-suite/parallel load:

```sh
npx vitest run tests/clients/lsp/edits.test.ts
npx vitest run
```

Capture the reported `maxBlock` value and whether it reproduces only under suite contention or also in isolation.

## Acceptance criteria

- Root-cause whether the excess is a real synchronous planning regression or event-loop sampler starvation under suite contention.
- Make the test reliable in the full suite without simply deleting the guard or widening it enough to hide a real large synchronous burst.
- Preserve the planned-operation assertion and retain a regression signal for an intentional O(1k)-scale synchronous blocking change.
- Verify targeted and full-suite runs, including the existing retry behavior, no longer produce this unrelated failure.

AngriestBird:
taking

PR #1204, opened 2026-08-09
test(lsp): budget workspace-edit planning in CPU time

AngriestBird:
refs #1081

## What

`tests/clients/lsp/edits.test.ts` > "keeps large text/resource planning occupancy bounded" now budgets `process.cpuUsage()` around the synchronous plan call **in addition to** `measureMaxSyncBlockMs`: `cpu < 200ms` and `maxBlock < 300ms`. The `planned === 800` assertion and `retry: 2` are unchanged. The file stays in the `timing-sensitive` Vitest project.

## Why

The excess was sampler starvation, not a planning regression.

`planWorkspaceEdit` (`clients/lsp/edits.ts:528`) never yields, so the sampler in `tests/support/perf-harness.ts` can only report one gap: the wall clock of that one call. On Linux, five consecutive runs on the test payload (400 changes + 400 resource ops) gave a maxBlock of 1.69-4.72ms against a 300ms budget. The reported 358-435ms is far more than the real cost, which no O(1k) planning path produces, and it survived `retry: 2`, so the pressure was sustained rather than a single spike.

It is that cheap on Linux because `normalizeFilePath` (`clients/path-utils.ts:41`) is a pure string replace there, so the planner does zero filesystem syscalls on `ubuntu-latest`. #1091 had already removed the second canonicalization and memoized `indexKey`.

## How

The CPU budget is the instrument for a compute/syscall blowup: unlike the sampler gap it does not count time the OS spent running someone else, so a merely descheduled worker cannot inflate it.

Two caveats found in review shape the final form:

**CPU time is not contention-proof on Windows.** There `normalizeFilePath` calls `realpathSync.native` per path, so this payload is ~400 synchronous FS syscalls whose cost is *system* time charged to this process, and it does inflate under load. Measured on Windows, same payload, same test: **15-47ms CPU / 37-46ms sampler gap** in the phased `timing-sensitive` window, but **47-110ms CPU / 235-259ms gap** with 16 CPU+FS loaders running. So the file stays phased, and the budget is 200ms rather than 100ms because 110ms was observed under load.

**`cpuUsage` cannot see a block that is not burning CPU.** A synchronous stretch that waits rather than computes — an `execSync`, a cold or networked `readFileSync`/`realpathSync` per path — would block the TUI for hundreds of ms and leave a CPU-only guard green. That shape is live rather than hypothetical: #1091 removed a *second* per-path canonicalization from `pathIndexKey`. So the `measureMaxSyncBlockMs` assertion is kept alongside the CPU one; each catches what the other cannot.

Regression signal, re-measured on Windows (dominated by system time): **~47-78ms at 800 ops, ~0.5-1.2s at 8k, ~1.6-2.0s at 40k**. An O(1k)-scale synchronous regression still trips the 200ms CPU budget by a wide margin. (The earlier Linux-only figures of 6.4/34/131ms are ~11x low for Windows.)

The sibling guards in `source-walk-occupancy.test.ts` keep the sampler only, which is correct for them: those walkers do yield, so the gap is the only way to see a non-yielding regression.

## Verification

Windows, `tests/clients/lsp/edits.test.ts`, 50 runs total at the final budgets — 20 solo, 10 under 16 CPU+FS loaders, 20 solo again — 0 failures. `npm run lint` clean.


apmantza:
thanks @AngriestBird. adding some small follow ups and merging

apmantza:
@AngriestBird thanks once more.

reproduced the starvation you identified: with sibling load running, the gap goes to 235-259ms on a payload whose actual cost is ~40ms. That's the bug, and CPU time is the correct instrument for it.

pushed two follow-up adjustments directly to your branch (46a0478):

**1. Kept `tests/clients/lsp/edits.test.ts` in `timingSensitiveInclude`.** CPU time turns out not to be contention-proof on Windows. `planWorkspaceEdit` → `uriToPath` → `normalizeFilePath` (`clients/path-utils.ts:75-86`) calls `realpathSync.native` per path there, so this payload is ~400 synchronous FS syscalls, and that cost lands as *system* time charged to this process. Measured on the same payload: 15-47ms CPU in the phased window, 47-110ms with 16 CPU+FS loaders running. So I removed the "no amount of sibling-fork noise can inflate it" line — true on Linux where `normalizeFilePath` is a string replace, not on Windows. The CPU budget itself is the good half and stays.

**2. Kept the `measureMaxSyncBlockMs` assertion alongside it,** so the test now asserts both `cpu < 200` and `maxBlock < 300`. `cpuUsage` can't see a synchronous stretch that waits instead of computing — an `execSync`, or a cold/networked `readFileSync`/`realpathSync` per path — which would block the TUI for hundreds of ms with a CPU-only guard staying green. Not hypothetical given #1091 removed a *second* per-path canonicalization from `pathIndexKey`. The two assertions catch disjoint regressions, and under the phased project both are stable.

Budget is 200ms rather than the 300 you had: tighter, but not the 100 I first wanted, because 110ms showed up under load. 50 runs on Windows at the final budgets (20 solo, 10 under 16 loaders, 20 solo) — 0 failures.

**3. Corrected the scaling numbers** in the PR body and the in-code comment. They were measured on Linux and are ~11x low for Windows, where system time dominates: 800/8k/40k ops cost ~47-78ms / ~0.5-1.2s / ~1.6-2.0s. Documentation only, no gating on `process.platform`.

Issue #1205, opened 2026-08-09
writeFileAtomic is not atomic for concurrent same-process writes (shared per-pid tmp name)


apmantza:
## Summary

`clients/atomic-write.ts` names its staging file `${targetPath}.tmp-${process.pid}` (around lines 59 and 85) — **per-process, not per-call**. Two concurrent `writeFileAtomic` / `writeFileAtomicAsync` calls from the **same** process to the **same** target therefore share one staging path.

Both writers `open(O_TRUNC)` the same tmp inode. The first `rename` publishes that inode to the target while the second writer's fd is still writing — into the now-published target. The result is a **torn file at the published path**.

Reproduced empirically on Windows (Node): two racing `writeFileAtomicAsync` calls on one target, 2 MB payload "A" vs 50 KB payload "B", produced a hybrid file containing both in **35 of 40 runs**.

The mechanism is OS-agnostic in principle — POSIX `rename` keeps an open fd bound to the moved inode identically — but this was **measured on Windows only**. See "Acceptance" below: the regression test must be confirmed to fail on Linux CI before this is considered fixed, otherwise the test has no teeth on the CI OS.

## The guarantee is the inverse of what is documented

The module docstring claims "a concurrent reader always sees either the fully-old or fully-new file, never a torn write". That is:

- **False** for same-process concurrency (shared tmp name).
- **True** for cross-process concurrency (distinct pids give distinct tmp names).

The actual guarantee is: *crash-safe replacement, and cross-process tear-free publication, provided at most one write per process is in flight per target.* It provides **no** read-modify-write isolation, **no** ordering, and **no** same-process mutual exclusion.

## Why this is P0 rather than a latent nit

The precondition is violated by design in current code — these writes are deliberately fire-and-forget, so overlap is expected, not exceptional:

- `clients/lsp/client.ts:2207` — `void recordLspChild({...})`
- `clients/runtime-turn.ts:195` — `void updateHeartbeat().catch(...)`
- `clients/instance-registry.ts:488` — `prunePids(deadPids).catch(...)`
- `clients/bus-publish.ts:160` — `void appendRecentTouches({...})`

All four write the same target with the same tmp name. Only `removeLspChild` is serialized (`lspChildRemovalTail`, `clients/instance-registry.ts:316`).

Blast radius is amplified because every reader in this family swallows parse errors to `{}` / `[]`. A tear therefore degrades to **silent total loss of the store**, not partial loss — and the next write persists the emptied state.

## Fix

[...]

PR #1210, opened 2026-08-09
DO NOT MERGE — probe: does the #1205 tear reproduce on Linux CI?

apmantza:
Throwaway draft to measure whether the #1205 regression test in #1208 has BEHAVIOURAL teeth on Linux, not just structural ones. `clients/atomic-write.ts` is reverted to the pre-fix per-pid staging name (plus a shim mirroring what it computed inline). Expected: the structural assertions fail deterministically; the open question is whether `async vs async` reports HYBRIDs on ubuntu-latest. Will be closed and the branch deleted as soon as the Unit tests job reports. Refs #1205, #1208.


apmantza:
Probe complete — answer recorded on #1208. Pre-#1205 code on ubuntu-latest: 21/40 iterations published a HYBRID, plus all three deterministic structural assertions failed. The regression test has real teeth on Linux CI, not just Windows. Closing and deleting the branch.

Issue #1228, opened 2026-08-09
Orphaned atomic-write staging files have no garbage collector after #1207 narrows the review-graph sweep


apmantza:
## Problem

`<target>.tmp-<pid>` staging orphans have **no garbage collector anywhere in the repo**, and the one thing that incidentally collected them is being removed.

`clients/atomic-write.ts` has no reaping lifecycle. Its cleanup is `fs.rmSync(tmpPath, { force: true })` inside a `catch` (~:59-70, async twin ~:85+), so it fires only on a **caught exception**. A process killed between `fs.writeFileSync(tmpPath, ...)` and `fs.renameSync(tmpPath, targetPath)` never enters that catch — SIGKILL, OOM, host crash, or the Windows RestartManager kill already documented in this project's crash forensics.

Before #1207, `sweepStaleStageFiles` in `clients/review-graph/builder.ts` matched `/\.tmp-\d+$/` and was — over-broadly, but genuinely — the only GC for that class, project-wide. #1207 correctly narrows it to the review graph's own artifacts, because a sweeper should not own other modules' files. That is the right call, but it means the class is now **unowned** rather than badly owned.

A grep of `clients/` confirms no other reaper exists: `project-snapshot.ts:733` sweeps only `project-snapshot.json.gz.stage-`, and `instance-reaper.ts` reaps processes, not files.

Affected includes the review graph's own `review-graph.json.gz.tmp-<pid>` from the `writeFileAtomic` fallback (`builder.ts:1549`, `:2752`) — a multi-MB gzip write, so the crash window is not negligible.

## Interaction with #1205

#1208 changes the staging name to `<target>.tmp-<pid>-<seq>`. That produces **more** orphans per crash, not fewer, since each call now mints a distinct name instead of reusing one per process. It does not add a sweeper. So this gap widens slightly when #1208 lands.

## Scope

Give `atomic-write.ts` an actual reaping lifecycle, since it owns the namespace (see #1209):

- A sweep that deletes only staging files matching the module's own scheme, whose embedded pid is not alive, reusing the conservative ESRCH-only probe already in `instance-reaper.ts` rather than inventing another.
- Called from somewhere sensible — session start is the natural hook; it must not be an exit path (see the keep-alive hazard in #1212).
[...]

Issue #1206, opened 2026-08-09
sweepStaleStageFiles deletes other modules' in-flight atomic-write staging files


apmantza:
## Summary

`sweepStaleStageFiles` in `clients/review-graph/builder.ts:1916-1928` deletes staging files belonging to **other modules** out of the shared project cache directory, destroying their in-flight atomic writes.

```ts
const isStage = entry.includes(".stage-") || /\.tmp-\d+$/.test(entry);
if (!isStage || entry.includes(ownMarker)) continue;
fs.rm(path.join(cacheDir, entry), { force: true }, () => {});
```

The `/\.tmp-\d+$/` alternative matches the generic staging shape produced by `clients/atomic-write.ts` (`<target>.tmp-<pid>`). That shape is **not the review graph's property** — it belongs to every store that uses `writeFileAtomic`. `cacheDir` is `getProjectDataDir(cwd)/cache` (builder.ts:1993), which is shared.

## Confirmed collateral targets

All of these stage into the same directory via `writeFileAtomic`:

- `clients/diagnostic-dispositions.ts:198-204` — `bestEffort: false`
- `clients/cache-manager.ts:74-76` — scanner data + metadata, `bestEffort: false`
- `clients/project-snapshot.ts:107,116`
- `clients/call-graph.ts:545,549`
- `clients/codebase-model.ts:167,171`
- `clients/lsp/workspace-diagnostics-cache.ts:90`
- `clients/project-diagnostics/cache.ts:17`

## Failure interleaving

1. Process A (pid 100) starts a review-graph persist; `sweepStaleStageFiles` async-readdirs the cache dir.
2. Process B (pid 200) is mid-`markDisposition`: `writeFileSync("diagnostic-dispositions.json.tmp-200")` has returned, `renameSync` has **not** yet been called.
3. A's sweep sees `diagnostic-dispositions.json.tmp-200`, matches `/\.tmp-\d+$/`, does not contain A's `ownMarker`, and `fs.rm`s it.
4. B's `renameSync` throws `ENOENT`. Because this store is `bestEffort: false`, the error propagates **out of `markDisposition` into the agent's mark tool**.

For `bestEffort: true` stores the same interleaving silently drops the update instead.

This violates the atomic-write contract that interference leaves either the complete old or the complete new value, and it breaks cross-module isolation outright — one subsystem's garbage collection reaches into unrelated subsystems' transactions.

## Fix

Scope the sweep to the review graph's **own** artifacts:

[...]

PR #1231, opened 2026-08-09
fix(atomic-write): route the last hand-rolled staging writers through stagePathFor

### LINKED ISSUE #1205: writeFileAtomic is not atomic for concurrent same-process writes (shared per-pid tmp name)

### LINKED ISSUE #1217: Hand-rolled .tmp-<pid> staging writers bypass atomic-write.ts and did not inherit the #1205 fix
## Problem

#1205 fixed the per-pid staging collision **inside `clients/atomic-write.ts`**. Two writers hand-roll the same `<target>.tmp-<pid>` scheme without going through that module, so they did not inherit the fix and still carry the original defect shape.

Both were verified by reading the code; line numbers are from master at the time of filing.

### 1. `clients/instance-reaper.ts:777` — higher risk

```ts
const tmpPath = `${target}.tmp-${process.pid}`;
await fs.promises.writeFile(tmpPath, JSON.stringify({ instances: remaining }), "utf-8");
await fs.promises.rename(tmpPath, target);
```

`pruneDeadInstances` writes the machine-global `~/.pi-lens/instances.json`. Two concurrent invocations in one process compute the **identical** `tmpPath`, both `writeFile` into it, and the first `rename` publishes while the second is still writing — the #1205 tear, verbatim.

This is reachable rather than theoretical. `instances.json` is written fire-and-forget from multiple paths (`instance-registry.ts:488` does `prunePids(deadPids).catch(...)`), the reaper runs its own sweep, and both are async with no serialization between them. The function's own docstring already concedes a "last-writer-wins race... narrowed, not eliminated" — but that admission is about *lost updates*, and the tear is a strictly worse failure than a lost update: the reader (`readInstanceRegistry`) parses defensively and degrades to empty, so a tear silently drops **every** registered instance rather than one.

Post-#1205 this is also now *asymmetric*: `instance-registry.ts` writes the same target through `writeFileAtomic` with the new per-call `.tmp-<pid>-<seq>` name, while the reaper still uses the bare `.tmp-<pid>`. The two writers no longer even agree on the staging scheme for one file.

Note this store is separately tracked in #1202 for its lost-update problem and is slated for per-pid sharding. If sharding lands first, this tear disappears with it — but the sharding work is further out, and this is a one-line fix.

### 2. `clients/gzip-stage-write.ts:61` — lower risk, but the docstring overclaims

```ts
const tmpPath = `${stagePath}.tmp-${process.pid}`;
```

Called by both `review-graph/persist-worker.ts` and `project-snapshot-persist-worker.ts`. `stagePath` is already per-generation (`.stage-<pid>-<gen>`), so two calls normally have different targets and therefore different tmp paths — this is why it is lower risk than the reaper.

Two reasons not to leave it:

- Worker threads **share `process.pid` with the main thread and with each other**, so the pid provides no isolation between concurrent workers. The moment two calls ever share a `stagePath` — a retry, a re-queued persist, a generation that repeats, or a future caller that stages a non-generational path — the collision is exact.
- The docstring asserts: *"the `stagePath` itself is only ever created by the atomic rename, so a crash mid-write can never leave a torn stage file behind."* True for a **crash**. False for **concurrency**, which is precisely the false-confidence pattern #1205 found in `atomic-write.ts`'s own docstring. A reader auditing this file today would conclude it is safe.

## Fix

Both should call `stagePathFor()` from `clients/atomic-write.ts` (exported by #1205) rather than interpolating the scheme by hand. Where a caller cannot use `writeFileAtomic` wholesale — `gzip-stage-write.ts` genuinely needs its streaming pipeline, which is the point of the module — it should still source the *name* from the owning module. That is the seam #1209 formalizes, and these two writers are the concrete argument for it.

Correct the `gzip-stage-write.ts` docstring to scope its guarantee to crash-safety and state explicitly that it does not provide concurrent-write isolation.

## Acceptance

1. Neither file interpolates `.tmp-` by hand; both source staging names from `atomic-write.ts`.
2. A regression test for `pruneDeadInstances`: two concurrent invocations, assert the published `instances.json` always parses and is never a hybrid. Must fail on pre-fix code — #1205 demonstrated the tear reproduces at 13/40 on Windows and **21/40 on Linux**, so use enough iterations and mismatched payload sizes to be reliable.
3. A test that two concurrent `writeGzipStageFile` calls with the **same** `stagePath` do not corrupt it (the case the per-generation path currently makes rare, not impossible).
4. The `gzip-stage-write.ts` docstring no longer implies concurrency safety.
5. OS-agnostic; the tests must run and be meaningful on Linux CI, not gated on `process.platform`.

## Provenance

Residue of #1205, surfaced while fixing it (PR #1208) and flagged as explicitly out of scope there. This is why #1205 remains open rather than being closed by #1208.

A repo-wide sweep for other hand-rolled `.tmp-`/staging writers outside `atomic-write.ts` should be part of this work — these two were found incidentally, not by an exhaustive search, so treat them as the known members of the class rather than the complete list.

Refs #1205 #1208 #1209 #1202.

AngriestBird:
Closes #1205. Closes #1217.

## What

Three writers hand-rolled a `<target>.tmp-<pid>` staging name instead of going through `clients/atomic-write.ts`, so none of them inherited the #1205 per-call fix:

- `clients/instance-reaper.ts` `pruneDeadInstances` → now `writeFileAtomicAsync`
- `clients/gzip-stage-write.ts` → now sources its name from `stagePathFor` (it needs its own streaming pipeline, so it takes the name only)
- `clients/grammar-source.ts` `downloadGrammar` → now `writeFileAtomic({ bestEffort: false })`

`stagePathFor` also grew `threadId`, and `STAGE_TMP_PATTERN` a third optional group.

## Why

The reaper is the reachable one. `~/.pi-lens/instances.json` is machine-global and also written fire-and-forget by `instance-registry.ts`'s `prunePids`, with nothing serializing the two. `readInstanceRegistry` degrades a parse failure to `[]`, so a tear drops **every** registered instance rather than the one entry being pruned. Reproduced against the pre-fix body at **22/40 unparseable on Linux** — the same scale as #1205's own 21/40.

`grammar-source.ts` was not in #1217's list. It came out of the repo-wide sweep #1217 asked for, run over every `rename` call site rather than a `.tmp-` grep. The remaining renames are log rotation, installer moves, and LSP workspace edits, none of which stage.

## How (non-obvious)

**`threadId` is load-bearing, not tidiness.** `_stageSeq` is per-*thread* — a worker gets its own module instance, so its own counter starting at 0 — while `process.pid` is shared across every thread of the process. Two threads therefore each minted the identical `${target}.tmp-<pid>-0`, which is #1205's shared-inode tear reintroduced one axis over. `atomic-write.ts`'s own docstring already flagged this and asked any future worker-side caller to fix it first; pointing `gzip-stage-write.ts` (which runs on both persist workers) at `stagePathFor` is exactly that trigger. The name is now `.tmp-<pid>-<threadId>-<seq>` — pid across processes, thread id across threads, counter within one thread.

`STAGE_TMP_PATTERN` becomes `/\.tmp-\d+(?:-\d+){0,2}$/` so sweepers still collect leftovers in all three shapes that can be on a user's disk (`.tmp-<pid>` pre-#1205, `.tmp-<pid>-<seq>` from #1205, and the new one). It has no production consumers today — the #1206 sweeper deliberately does not match `.tmp-` at all, and the orphan-GC gap is #1228 — so widening it is contained.

**`gzip-stage-write.ts`'s docstring** asserted that "the `stagePath` itself is only ever created by the atomic rename, so a crash mid-write can never leave a torn stage file behind". True for a crash, false for concurrency — the same conflation #1205 found in `atomic-write.ts`'s own docstring. Now scoped to crash-safety, with the absence of ordering and isolation stated.

**Every new guard was confirmed to fail against the pre-fix body on Linux**, not merely to pass after (recurring defect shape 2/7). Each source change was reverted individually and rebuilt:

| Guard | Pre-fix on Linux |
|---|---|
| concurrent `pruneDeadInstances` | 22/40 iterations published unparseable JSON |
| concurrent `writeGzipStageFile` on one `stagePath` | both tests fail, loser's rename at `ENOENT` |
| two workers minting for one target | one identical path (`expected 1 to be 2`) |

Two notes on the reaper test, both of which cost a round of teeth: a few hundred KB does not reproduce it (Node completes the write in one syscall and the writers never overlap inside the inode), and pairing a big result with a tiny one does not either (the small writer finishes long before the big one starts, which then simply overwrites the file with its own complete content). Both payloads have to be large and unequal. That is documented in the test.

`pruneDeadInstances` is now exported so the regression test can drive it directly rather than through the real-process `sweepOrphans` path.

## Checks

`npm run build`, `npm run lint`, `npm run check`, `npm run changelog:check` all clean. Full suite: 498 files passed / 6 skipped, 5734 tests passed / 35 skipped, 0 failures.

## Note on #1206

Its sweeper work landed in #1207 and satisfies the issue's stated fix, but the commits said `refs` rather than `Closes` so it is still open. The "#1206 sweeper migration" this issue was waiting on is moot — that PR went shape-independent (never matches `.tmp-`) instead of adopting `STAGE_TMP_PATTERN`, and moved the orphan-GC gap to #1228. Left open for you to close; not touched here.


apmantza:
thanks @AngriestBird! merged.

Issue #766, opened 2026-07-22
perf: reduce the madge / actionable_warnings p99 tail (throughput, not input latency)


apmantza:
Split out from #191 (its last unchecked perf item) so #191 could close after #761/#765 superseded its walker scope.

From the #191 audit: `madge` and `actionable_warnings` dominate the `latency.log` p99 tail. They are already `async`/subprocess phases that don't hold the event loop, so reducing them is a THROUGHPUT win, not an input-latency one — and any change is behavior-affecting, which is why #191 deliberately deferred it as its own effort.

Scope ideas from the audit (validate against current code before acting): incremental madge graph reuse across edits, tighter file scoping for actionable_warnings, or caching keyed on the touched file's dependency closure. Measure via `latency.log` p99 before/after on a 2k-file project.

apmantza:
## Grounded re-scope (measured against 3 months of `latency.log`, 72k lines)

Investigated before acting, per the issue's own "validate + measure first" note. Findings correct the original premise:

### `actionable_warnings` — never runs on default config
**Zero occurrences in the entire 3-month log window.** The feature is gated behind `getFlag("lens-actionable-warnings")` (`clients/runtime-turn.ts:831` → `clients/lens-config.ts:334-336`, `config?.actionableWarnings?.enabled === true`), which **defaults false**, and there's no `~/.pi-lens/config.json` on the dogfood machine. So its claimed contribution to the p99 tail **cannot be reproduced — it has never executed.** The always-on cousin `code_quality_warnings_report` is cheap (p99 26ms, 3s total). **→ This half is moot until the flag is enabled; deferring it. If pursued later, enable the flag and re-measure first.**

### `madge` — real tail, but 2 of 3 scope ideas already done
- **File scoping (idea 2): already implemented.** madge runs only over `getFilesForMadge(cwd)` (import-changed files, `cache-manager.ts:347`) with a single-file target (`dependency-checker.ts:346`), not project-wide.
- **Incremental reuse (idea 1): partial.** `DependencyChecker.checkFile` has an import-fingerprint cache (`dependency-checker.ts:240-266`) that skips madge when a file's own imports are unchanged (656/1143 firings are ≤10ms cache hits). But a real miss re-invokes madge cold.
[...]

Issue #375, opened 2026-07-03
Route remaining hardcoded npm/npx call sites through package-manager.ts


apmantza:
## Context

#374 added `clients/package-manager.ts` (resolve npm/pnpm/yarn/bun + command builders) and routed the first call sites through it. This tracks the remaining hardcoded `npm`/`npx` sites.

## Already landed since this was opened

- **Interactive LSP install** (`clients/lsp/interactive-install.ts`) — routed through the resolved manager via a new `globalInstallArgs(pm, pkg)` builder (was npm-only). #374 hygiene (CHANGELOG + AGENTS) also done.
- **Dead-code removal** (#378) — `launchViaPackageManager` deleted. It was the only consumer of `execArgs()` and the only place that could `dlx`-fetch an LSP server on non-npm managers.

## The cache-only-exec caveat (now dev-tools only)

`npx --no` is **cache-only, no silent download** (pi-lens invariant: tools come from the explicit install flow, not surprise fetches). `execArgs` maps non-npm managers to `pnpm dlx` / `yarn dlx` / `bun x`, which **fetch-if-missing** — so a naive conversion of a `--no` site would make pnpm/yarn/bun users silently download tools.

This no longer affects LSP launch (the live path is `launchLSP` → resolve installed binary across every manager's global bin, spawn by path — no runner fetch). It **only** matters for the on-the-fly dev-tool sites below, which do run tools via `npx <tool>`.

**Decide before converting those:** add a **cache-only exec** concept (npm → `npx --no`; non-npm → *don't* dlx, rely on local/global bin and fail clean like `--no`), **or** explicitly accept `dlx` downloads for non-npm managers and document it.

## Remaining targets

1. **`resolveLocalFirstAsync` (`clients/dispatch/runners/utils/runner-helpers.ts`)** — the shared "local `.bin` → global PATH → `npx --no`" helper the runner fleet leans on. Highest leverage. Its step-3 fallback is the cache-only case above; the modest safe win is widening the global-bin step to cover all managers.

2. **Other hardcoded `npx <tool>` sites** (same class, lower value — `npx` ships with node so `npx --no <tool>` already works on any project; these mostly just need the earlier global-bin lookup widened, not a runner swap):
   - `clients/formatters.ts` (biome, prettier), `clients/biome-client.ts`, `clients/dispatch/runners/biome.ts`
   - `clients/sg-runner.ts`, `runner-helpers.ts` (ast-grep)
[...]

PR #1248, opened 2026-08-10
perf(madge): memoize managed-path command resolution, add batch telemetry, generation-guard shared state (refs #766)

AngriestBird:
## Summary

Implements steps 1-3 of the bounded sequence from #766's 2026-08-09 investigation: managed-path madge resolution, capped turn-end batch telemetry, and a generation guard on the shared circular-dep state. Refs #766 (kept open: step 4, whether the turn-end madge pass should exist at all, is deferred until this telemetry has dogfood data).

## Type of change

- [ ] Bug fix
- [ ] Documentation

## Area

- [x] area:perf
- [x] area:observability

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] `package-lock.json` is in sync with `package.json` (no dep changes)
- [ ] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there (nothing documented there changes)
- [x] `CHANGELOG.md` has a `## [Unreleased]` entry **in this PR**
- [x] Commit subject includes the issue number: `(refs #766)`

## What changed and why

**What.** Three changes in `clients/dependency-checker.ts` plus the turn-end call site:

1. `resolveMadge` is memoized per project root and falls through `findNodeToolBinary` (project-local still wins, #375) -> `ensureTool("madge", { allowInstall: false })` -> `npx madge`. Previously it ran once per file inside the batch mapper, its miss path re-spawned the uncached `npm config get prefix` / `pnpm bin -g` / `yarn global bin` probes each time, and it never consulted the managed tools tree pi-lens itself installs madge into - the "concrete avoidable defect" from the issue investigation.
2. `checkFilesBatch` returns `{ results, stats }`; turn-end attaches `MadgeBatchStats` (requested/missing/cacheHits/spawned/failed, `commandKind`, `resolveMs`, per-target durations capped at 12 with `targetsTruncated`) as `metadata` on the existing `phase: "madge"` latency entry. One whole-phase duration could not distinguish many files from one slow file from command resolution.
3. `checkFile`, `checkFilesBatch` and `scanProject` all overwrite `lastCircular`/`circularFiles` wholesale and genuinely overlap (turn-end vs. the background scans in `runtime-session.ts` and `project-diagnostics/fresh-fetch.ts`). Each writer now takes a generation at classification time and publishes only while that generation is still newest, so a stale older completion cannot overwrite newer state.

**Why.** The madge phase dominates the `latency.log` p99 tail (p50 ~2ms, p90 ~7.35s, p99 ~15.99s over 156 turn-end records); npx fallback plus per-file resolution amplify it, and the telemetry is what the issue's step 4 decision needs.

**How (non-obvious).**

- The batch folds miss results into a batch-local shadow of the shared state, in array order; cache-hit entries read the shadow at their own fold position, and the instance state is published once, atomically, at the end - so the returned map keeps the sequential-equivalence contract byte-for-byte even when the publish itself is dropped by the guard. A batch that folded nothing (all hits, or every spawn failed) publishes nothing, so it cannot suppress an older operation's still-legitimate write.
- The issue asked for a same-content single-flight map as well; it was dropped as structurally unreachable, not skipped. `importsChanged` writes its `{timestamp, size}` cache entry synchronously during classification and neither caller has an interleaving await there, so a second overlapping operation on the same content always classifies as a cache hit and never reaches a spawn. Only changed content spawns again - and then two spawns are correct (#533, "when in doubt, re-run"; no result cache is introduced).
- `kind` is classified from the resolved string, not the resolution step: the installer's discovery legitimately returns a bare PATH name or an npm-global path as well as the managed one, and `doEnsureAvailable`'s PATH-probe branch means the first resolution can take `ensureTool`'s cold path. `getManagedToolsDir()` is exported from the installer so the check doesn't re-derive `<pi-lens home>/tools`.
- `ensureTool`'s `allowInstall: false` is load-bearing: resolution runs on the spawn path and must never download; installation stays owned by `ensureAvailable()`. `doResolveMadge` also never rejects, because a rejected promise would poison the per-root memo for the session.
- Scope honesty: `isInCircular`/`getCircularForFile` have no production callers and turn-end only dbg-logs `hasCircular`, so the guard's payoff today is correctness of future cache-hit results, not user-visible output. That is deliberate input to step 4.
- Both generation-guard tests were verified to fail against the unguarded publish (guard line removed, rebuilt, 2 failed; restored, 2 passed), per recurring defect shape 7.


AngriestBird:
Opus QA pass on the initial commit returned **ship after fixes**; `908fc8e2` applies them:

- **Memo self-healing (QA finding 1, regression):** a memoized absolute madge path is re-validated with one `existsSync` per resolution and re-resolves if the binary vanished mid-session; the `npx` fallback is never memoized, so a transient installer failure can't pin the slow path for the whole session. Pre-memo resolution re-probed every spawn and healed itself; this buys that back at one stat per resolution.
- **`kind` classification (finding 2):** `findNodeToolBinary` also returns package-manager *global* bins, so both discovery branches now classify from the resolved string (managed → local-within-projectRoot → global → bare path), matching the type's documented contract. The test that locked the old behavior in is fixed.
- **Stats accounting (findings 3–4):** the madge-unavailable early return counts its miss set as `failed` instead of zero-of-everything; `spawned` increments per dispatched spawn.
- **Test gaps:** per-root memo keying (second root), dead-path self-heal, throwing-installer recovery, the `folded === false` publish rule (verified to fail against an unconditional publish), exact-cap targets boundary, and a real lower-bound `resolveMs` assertion replacing two vacuous `>= 0`s. Dependency-checker coverage is now 28 tests across 6 files.

Deferred with QA's agreement, filed as follow-ups: #1250 (madge metadata growth vs the smells-rollup 64 KB lookback) and #1251 (turn-end call site has compile-only coverage).

QA also verified sound: the generation algebra (unique gens, no equal-gen ambiguity), Windows case-insensitive managed-dir classification, `allowInstall:false` genuinely cannot download, and the PATH-only environment now resolving to a direct `madge` invocation instead of npx — the actual win for the #766 tail.

apmantza:
thanks @AngriestBird. merged.

PR #1264, opened 2026-08-11
Expose generic read-recording bridge via process global

### LINKED ISSUE #1265: Expose generic read-recording bridge so external tools can satisfy the read-before-edit guard
### What do you want to add or change?

A process-global bridge object at `globalThis[Symbol.for("pi-lens:read-bridge")]` that any co-process Pi extension can call to register a file read against the live read-guard.

Custom Pi tools registered via `registerTool()` perform file reads outside pi-lens's normal tool tracking. The guard only recognises reads from the native `read` tool and `bash` patterns (`cat`/`sed -n`), so a custom-tool read followed by an `edit` on the same file is incorrectly blocked with the read-before-edit error.

### Why?

Any Pi extension that wraps or proxies file reads currently cannot satisfy the read-before-edit guard without pi-lens coupling to that specific tool. A generic bridge lets those extensions register reads without either party knowing about the other. The integration (wiring) lives in a third layer external to both.

### How? (optional)

- `clients/read-bridge.ts` — exports `registerReadBridge()`, `ReadBridgeEntry`, `ReadBridge`, `READ_BRIDGE_KEY`
- `index.ts` — calls `registerReadBridge()` inside the extension factory (so `getLensFlag` is available), protected by the `_readBridgeRegistered` singleton guard; `_readBridgeGetFlag` refreshed on every activation
- Bridge respects `no-read-guard`, gitignored paths, and external/vendor paths via `isRecordable()`

### Contribution intent

- [x] I plan to implement this myself
- [ ] I am looking for feedback before implementing

### Area

area:read-guard

3choBoomer:
## Summary

Adds a process-global read-recording bridge so that co-process Pi extensions whose custom tools perform file reads outside pi-lens's normal tracking can satisfy the read-before-edit guard without pi-lens coupling to any specific tool.

Closes #1265

## Type of change

- [ ] Bug fix
- [x] New feature (net-new capability)
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [x] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [ ] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm test` passes
- [x] `CHANGELOG.md` has a `## [Unreleased]` entry for this user-facing change
- [x] Commit subject includes the issue number: `(closes #1265)`

### Design

`clients/read-bridge.ts` mounts a `ReadBridge` singleton at `globalThis[Symbol.for("pi-lens:read-bridge")]` after the extension factory initialises. Any co-process extension can call:

```ts
const bridge = (globalThis as any)[Symbol.for("pi-lens:read-bridge")];
if (bridge?.version === 1) {
  bridge.recordRead({ filePath, requestedOffset, requestedLimit, consumer: "my-ext" });
}
```

and the subsequent `edit` on that file will not be blocked.

### Trust model

The bridge is explicitly **advisory and trust-based**: any same-process extension already has unrestricted access to pi-lens internals, so this cannot be a security boundary. Its value is a stable API surface, a single point for flag/scope checks, and early detection of integration bugs.

### Key decisions

**`timestamp` is bridge-generated.** `recordRead()` stamps `Date.now()` itself, matching exactly how the internal read path works (`runtime-tool-call.ts` always uses `Date.now()` — it never accepts a caller-supplied timestamp). Callers do not pass a timestamp.

**Payload validation.** `isValidEntry()` validates exactly what the internal path gets for free from typed Pi tool input: non-empty string `filePath`; `requestedOffset` and `requestedLimit` must be finite positive integers when provided.

**Source provenance.** `ReadBridgeEntry` accepts an optional `consumer?: string`. The bridge tags every forwarded `ReadRecord` with `source: "bridge:<consumer>"` (defaulting to `"bridge:unknown"`), surfaced in `read-guard.log` metadata so the worklog shows which extension satisfied the guard. `ReadRecord` gains an optional `source?: string` field — absent for the normal internal tool-call path.

**`bridge.version`.** The `ReadBridge` interface exposes `readonly version: 1`. Callers should check this before invoking `recordRead` to guard against future incompatible changes.

**Frozen first-wins registration.** The global slot is permanently locked after first registration: `Object.defineProperty` with `writable: false, configurable: false` plus `Object.freeze` on the bridge object. Any attempt to overwrite or mutate throws `TypeError` in strict mode. The singleton guard switches to `READ_BRIDGE_KEY in (globalThis as object)` to correctly detect the non-configurable property.

**Factory-scoped registration.** `registerReadBridge()` is called inside the extension factory (not at module level) so `getLensFlag` is available. `_readBridgeGetFlag` is refreshed on every factory activation so flag changes take effect without re-registration.

**`turnIndex` / `writeIndex` sampled at call-time** — not at registration time, so they always reflect the current agent turn.

**No coupling** — pi-lens references no specific tool or extension. The wiring lives in an external integration layer.

**Safe by default** — bridge is absent before pi-lens loads; producers guard with `bridge?.recordRead()`.

### Tests

`tests/clients/read-bridge.test.ts` — 31 tests covering:

- Bridge absent before registration (top-level `it` runs before `beforeAll` fires)
- `bridge.version === 1`
- **Locked-bridge tests**: assign throws `TypeError`, delete throws `TypeError`, object mutation throws `TypeError`
- First-wins: second `registerReadBridge` call is ignored
- All forwarded `ReadRecord` fields correct; `timestamp` ≈ `Date.now()`
- Full branch coverage of `isValidEntry`
- Consumer provenance: `source` defaults to `"bridge:unknown"`, carries `"bridge:<consumer>"` when provided
- Read-then-edit authorization path

**Test structure note:** `configurable: false` means `delete (globalThis as any)[KEY]` throws, so the standard `beforeEach`/`afterEach` cleanup pattern does not work. The suite registers once in `beforeAll` with mutable closures reset per-test in `beforeEach`, and the locked-bridge tests validate the `TypeError` rather than working around it.


apmantza (changes requested):
Thanks for the contribution 

One design question has to be settled before merge, because as written the guard's trust model changes: `recordRead()` grants read authorization to any same-process extension that can invoke it, without proof an actual read occurred (clients/read-bridge.ts:92-118), with caller-supplied timestamps/offsets un-validated (:44-55, :99-114) and no protocol version field (:58-61). The internal path records reads only after actual tool handling with internal timestamps (runtime-tool-call.ts:615-626).

apmantza:
Following up on the review with a concrete design: your wiring stays, the recording function changes shape. 

(1) `recordRead(path, timestamp, ...)` becomes `readAndRecord(path, range?)`;  the bridge stats + content-hashes the file at call time and records ground truth; callers supply intent, never timestamps/claims; 

(2) bridge records carry `source: "bridge:<consumer>"` provenance vs internal `source: "tool"` (advisory semantics preserved today, strict mode possible later, worklog shows who satisfied the guard); 

(3) `bridgeVersion` + runtime payload validation + frozen first-wins registration instead of a bare mutable global. 

Happy to adjust if this conflicts with your downstream use, comment here and we'll rework.

3choBoomer:
I am just now seeing your comment, but it appears that we were on the same path.

I have updated the PR description and merged in upstream changes.

 ### Trust model (doc comment + AGENTS.md)

 The bridge is now explicitly documented as advisory and trust-based: any same-process extension already
 has unrestricted access to pi-lens internals, so this cannot be a security boundary. The doc comment and
 AGENTS.md both say so directly, which was the lighter-hardening path the reviewer offered.

 ### Removed timestamp from ReadBridgeEntry

 The reviewer noted a fundamental asymmetry: the internal path (runtime-tool-call.ts) always stamps
 Date.now() itself — it never accepts a caller-supplied timestamp. The bridge now does the same:
 recordRead() generates the timestamp, the caller doesn't supply one. This closes the asymmetry, removes
 an entire category of validation concern, and makes the bridge's behaviour identical to the internal
 path.

 ### Removed version field

 Added initially in response to the review's "protocol/version field" ask, but on reflection: the bridge
 and its caller always share the same process, so the cross-version scenario (a newer extension calling
 an older pi-lens) is implausible. The field added API surface without meaningful protection and has been
 dropped. Could easily be added in the future if API versioning is needed.

 ### Payload validation

 isValidEntry() validates exactly what the internal path gets for free from typed Pi tool input:
 - null / non-object entries
 - empty or non-string filePath
 - requestedOffset: non-number, non-finite (NaN/Infinity), non-integer, or < 1
 - requestedLimit: same checks when provided (accepts undefined = whole file)

 ### Tests

 Full branch coverage of every isValidEntry check, including the previously missing requestedLimit =
 Infinity and non-number string cases for both fields. The authorization-path test verifies all forwarded
 ReadRecord fields, with timestamp asserted as approximately Date.now() rather than an exact
 caller-supplied value.

3choBoomer:
Addressed the follow-up feedback — here's what was implemented:

**Point 1 (timestamp)** — already handled in the previous round. The bridge stamps `Date.now()` itself; callers supply no timestamp.

**Point 2 (source provenance)** — `ReadBridgeEntry` gains an optional `consumer?: string`. Every forwarded `ReadRecord` is tagged with `source: "bridge:<consumer>"` (defaulting to `"bridge:unknown"`), written into the `read-guard.log` metadata so the worklog shows which co-process extension satisfied the guard. `ReadRecord` itself gets an optional `source?: string` field — absent for the normal internal tool-call path, set for bridge-originated reads.

**Point 3 (version + frozen registration)** — `ReadBridge` exposes `readonly version: 1`; callers should check this before calling `recordRead`. The global slot is permanently locked after first registration via `Object.defineProperty({ writable: false, configurable: false })` plus `Object.freeze` on the bridge object — any attempt to overwrite or mutate throws `TypeError` in strict mode. The singleton guard was updated to use `READ_BRIDGE_KEY in (globalThis as object)` to correctly detect the non-configurable property.

**Test structure** — `configurable: false` means `delete (globalThis as any)[KEY]` throws, so the standard `beforeEach`/`afterEach` cleanup pattern is not possible. Rather than work around this, the tests validate it: the locked-bridge tests assert that assign, delete, and mutation each throw `TypeError`. For behavioral isolation the suite registers once in `beforeAll` with mutable closures reset in `beforeEach`, keeping each test's guard mock and deps fresh without touching the global.

apmantza:
thanks @3choBoomer.

added one last piece:

**Ground-truth content binding**: at record time, have the bridge read+hash what it's attesting (`hashDiagnosticContent` from clients/lsp/diagnostic-binding.js) and store the hash on the record. That lets the guard later verify a bridge-attested read actually matched disk — closing the gap where a consumer honestly records a read of content that changed underneath. 

Suggested minimal shape: in `recordRead`, `const contentHash = hashDiagnosticContent(fs.readFileSync(entry.filePath, "utf8"))` (or hash only the requested range for large files) alongside the stamped timestamp; the guard's verification side can land separately.

Issue #1289, opened 2026-08-12
dispatch: madge/jscpd probes don't thread the managed env; biome resolves via npx — managed npm shims invisible outside Knip


apmantza:
Found during adversarial review of the #1221 test PR (#1287): #1199's Windows resolver fix (`resolveWindowsCommandForEnvironment` consulting the CHILD env's PATH) can only help a client that actually constructs and threads a managed environment into its spawn. Of the four managed npm tools, only Knip does:

- `clients/knip-client.ts:347-353` builds `getKnipEnvironment()` (ambient PATH + managed bin dir) and passes it at `:277`. ✅
- `clients/dependency-checker.ts:438` — `safeSpawnAsync("madge", ["--version"], { timeout: 5000 })` — **no `env`**: the availability probe consults only ambient `process.env.PATH`, so a managed-install `.cmd` shim that isn't on ambient PATH is invisible; madge is then treated as unavailable despite being managed-installed.
- `clients/jscpd-client.ts:168` — `safeSpawnAsync("jscpd", ["--version"], { timeout: 1500 })` — same gap.
- `clients/biome-client.ts:88-113` never spawns a bare `biome`: absolute candidates then `npx @biomejs/biome` fallback. The resolver is asked to resolve `npx`, not `biome.cmd`, so the managed-dir story is different in kind — needs its own decision (add the managed absolute candidate?).

Suggested shape: a shared `getManagedToolEnvironment(tool)` (generalizing `getKnipEnvironment`) threaded through the madge/jscpd probes and spawns, plus a managed absolute-path candidate for biome; then #1221's client-level tests (asserting each client's spawn options carry the managed env) become writable exactly as that issue asked.

Refs #1221, #1199.

apmantza:
## Class sweep (per AGENTS.md class-trigger rule)

Swept every spawn-family call site in `clients/` + `tools/` (haiku fan-out over `safeSpawnAsync|safeSpawn|spawn|spawnSync|execFile`, then hand-verified the load-bearing seams). The first-pass sweep flagged ~25 dispatch-runner sites; hand-verification shows most are NOT members — the accurate class is narrower and worse in a different way:

**The verified defect shape: the `ensureTool()` result is DROPPED — availability reports true, then every real invocation spawns the bare name.**

Confirmed members:
[...]

PR #1291, opened 2026-08-12
fix(ruff,jscpd): use the ensureTool-resolved path instead of respawning the bare name (refs #1289)

apmantza:
## Summary
- `RuffClient` and `JscpdClient` called `ensureTool`, got the managed-install absolute path back, and discarded it — availability reported true, then every real invocation spawned the bare name, which ENOENTs when the managed dir isn't on ambient PATH (#1289's verified class). Both clients now store the resolved command and use it for every spawn (jscpd precedence: local binary → stored managed path → npx).
- Madge (`DependencyChecker`) needed **no code change**: `doResolveMadge` already classifies the `ensureTool` discovery ahead of the npx fallback — the strengthened resolver test now pins that behavior.
- Regression tests are mutation-verified: with the fix reverted, the ruff test observes a bare `ruff` spawn and the jscpd test observes an `npx` spawn — both red pre-fix, green post-fix.

Remaining #1289 members (biome/npx decision) and the structural consolidation are tracked in #1289/#1290 — refs, not closes.

## Verification
- `npm run build` clean; 4 test files / 38 tests green
- Implemented by a plegma codex worker (gpt-5.6-luna), adversarial review to follow before merge

Refs #1289.


apmantza:
Adversarial review (codex-auto-review) triage:
- **Fixed** — bare-name sentinel trap: `jscpdCommand !== "jscpd"` couldn't distinguish resolved-vs-on-PATH and `ensureTool` can return a bare command (installer :2464). Replaced with a nullable `jscpdManagedPath` holding only absolute paths; ruff guards `path.isAbsolute` on the ensureTool return.
- **Refuted** — 'ruff spawns bare after PATH success': that's the design contract — bare name is correct when the PATH probe succeeded; the absolute path is only needed for managed installs.
- **Refuted** — 'cross-project leak': the availability fast-path candidate is `getGlobalPiLensDir()/tools/node_modules/.bin` (global managed dir), not a project-local path — no cwd dependence. It's now stored in the nullable managed field, which also improves the pre-PR behavior (that discovery previously fell back to npx at scan time).
- Mutation checks re-confirmed by the reviewer (both regression tests red with the threading reverted).

PR #1294, opened 2026-08-12
fix(format): style-preserving smart defaults — never impose a tool's stock style (closes #1144)

### LINKED ISSUE #1144: Smart-default formatters apply the tool's stock style to repos that never chose it — biome and shfmt silently convert space-indented files to tabs
### What happened?

A JS/TS repo with no `biome.json`, no Prettier config and no `.editorconfig` had every
file the agent touched silently re-indented from 2 spaces to tabs at `agent_end`. One
turn produced a 3134-line diff across 9 files that was whitespace-only (`git diff -w`
empty).

**Root cause.** `.js`/`.ts`/`.tsx` carry `gate: "smart-default"` +
`defaultWhenUnconfigured: true` (`clients/tool-policy.ts:19-25,55-70`), so Biome
formats even when the repo has expressed no preference. `biomeFormatter.resolveCommand`
(`clients/formatters.ts:348-370`) then invokes:

```
biome format --write [--use-editorconfig=true] <file>
```

with **no `--config-path`**, so Biome applies its *stock* defaults — and stock
`indentStyle` is `tab`. A `.editorconfig` is the only thing that incidentally prevents
this (`clients/formatters.ts:339-346,352-354`).

Note the asymmetry with the lint runner, which *does* pass one
(`clients/dispatch/runners/biome-check.ts:98-104`):

```ts
"--config-path=" + (userConfigPath ?? resolvePackagePath(import.meta.url, "config/biome/core.jsonc"))
```

So the formatter is the only Biome path with no style opinion at all — it does not even
inherit the bundled `config/biome/core.jsonc`, which pins `quoteStyle`,
`trailingCommas` and `semicolons` (lines 62-64) but **not** `formatter.indentStyle`.

**Defect class.** *A `defaultWhenUnconfigured: true` formatter whose style is
**configurable** imposes its own stock style on a repo that never chose it.* No
formatter in `formatters.ts` passes a config path — every one relies on the tool's own
discovery — so this is structural, not a one-off. Sweep of the formatter policies in
`clients/tool-policy.ts` (50 of 71 entries run unconfigured, 20 distinct formatters):

- **Confirmed damaging** (verified locally — both turn 2-space into tabs with no
  config): **biome** (11 extensions) and **shfmt** (`.sh`/`.bash` — `shfmt -w` with no
  `-i`, whose default `-i 0` means tabs; its `detect()` is satisfied by the binary
  merely being on PATH).
- **Same class, milder**: `prettier` (7 ext) and `ruff` (`.py`, explicit "No-config
  fallback" comment at `clients/formatters.ts:491-493`) — their defaults sit closer to
  common convention so indentation usually survives, but line width and quote style
  still restyle.
- **Not part of the class — must not be swept up**: `gofmt`, `rustfmt`, `zig`, `dart`,
  `gleam`. These have exactly one canonical style, so there is no repo preference to
  violate and running them unconfigured is correct.

### Steps to reproduce

A repo with a 2-space-indented `.ts` file and no `biome.json`, `.prettierrc`,
`.editorconfig`, or `@biomejs/biome` dependency. Let the agent edit it; at `agent_end`
the whole file is tab-indented. Directly:

```sh
printf 'function f() {\n  return 1;\n}\n' > a.ts && biome format --write a.ts  # -> tab
printf 'f() {\n  echo hi\n}\n'           > b.sh && shfmt -w b.sh               # -> tab
```

### Expected behavior

**Invariant:** where a repo has expressed no formatting preference for a language,
pi-lens must not change that file's existing indentation.

Options, increasing fidelity and cost:

1. Put the *configurable* smart-defaults behind config (`gate: "config-first"` /
   `defaultWhenUnconfigured: false`), matching the 21 entries already on that gate and
   the "config-gated" principle in `docs/features.md`. Closest to existing intent.
2. Pin style explicitly — add `formatter.indentStyle`/`indentWidth` to
   `config/biome/core.jsonc` and pass `--config-path` from the formatter as the lint
   runner already does; `-i` for shfmt. Cheapest, but pi-lens's quote/semicolon
   opinions then start applying too.
3. Infer indentation from the file's existing content and pass it per invocation.

Because option 1 spans 20 formatters, the scope question is worth settling before
anyone implements a biome-only patch.

**Non-goals:** the deferred `agent_end` timing (settled in #61); whether unconfigured
formatting should be opt-out at all (#789/#792 shipped `format.enabled`); canonical
single-style formatters, which keep running unconfigured.

**Test dimension not currently covered.** `scripts/smoke-tools.mjs --format` asserts
the selected formatter *changed* a mis-formatted fixture (`changed===true`), so a
tab-converting formatter passes it today. This class needs the inverse assertion: a
fixture with no config and an established indentation is left indentation-identical.

Related: #195 is the same failure mode in C (clang-format on hand-formatted legacy
code); #789/#792 provide the per-repo opt-out but not a safe default.

### pi-lens version

3.8.74

### Relevant logs

_No response_

### Area

_No response_

apmantza:
## Summary
Smart-default formatters no longer impose their stock style on repos that never chose it (#1144's confirmed damage: biome and shfmt silently retabbing 2-space repos).

- New shared `clients/dispatch/indent-detect.ts`: majority-based `{style, width}` detection with direct unit tests (tabs/2-space/4-space/mixed/empty/minified). Undetectable + no config ⇒ SKIP formatting for the damage-confirmed tools rather than impose stock style.
- Pinnable formatters wired: biome (`--indent-style/--indent-width` + bundled core.jsonc via `--config-path`, mirroring the lint runner), shfmt (`-i N` / `-i 0` per detection), prettier (`--use-tabs`/`--tab-width`), ruff format. The `.editorconfig` short-circuit and tool-own-config discovery still win.
- Full 20-formatter classification (fixed-style / pinnable / gated) recorded in the commit body.
- Mutation-verified arg-level regression tests for biome and shfmt (2-space fixture, no config ⇒ space-pinning flags in spawn args).

Closes #1144.


apmantza:
Delta-review blocker fixed in 959d0566: the review caught prettier/ruff falling through, but the seam bug covered ALL FOUR pinnable formatters — `formatFile` treated the resolvers' skip-`null` as legacy use-static-command-`null`, so every resolver-level skip was dead code. An explicit `SKIP_FORMATTING` sentinel now separates the two meanings; `formatFile` returns `{success:true, changed:false}` without spawning. New formatFile-seam regression test (spawn never called on undetectable+no-config), mutation-verified red with the sentinel guard collapsed back to fall-through.

Issue #1251, opened 2026-08-10
Cover the turn-end madge batch call site at runtime


AngriestBird:
**Context**

PR #1248 (refs #766) changed `checkFilesBatch` to return `{ results, stats }` and made turn-end attach the stats as latency-entry metadata (`clients/runtime-turn.ts`). Flagged in the QA pass on that PR.

**What**

Every existing turn-end test stubs `depChecker: { ensureAvailable: async () => false }` (`tests/clients/runtime-event-flow.test.ts:129`, `tests/clients/cascade-turn-merge.test.ts:113,172,236,290`, `tests/clients/runtime-turn-session.test.ts:49`), so the new `batch.results` destructure and the `...(madgeStats && { metadata })` spread never execute under test — the call-site contract is compile-checked only.

**Acceptance**

A turn-end test drives a stubbed `checkFilesBatch` returning both a result map and stats, and asserts the `phase: "madge"` latency entry carries the stats as metadata (and omits metadata when the batch does not run). Note `logLatency` no-ops in test mode, so the assertion needs a mock or spy at the `latency-logger` boundary.

apmantza:
Completed by #1303 (merged): runtime test drives the real turn-end path with a madge batch — batch invocation, circular-result propagation into the outcome, and latency metadata asserted; mutation-verified (batch no-op'd → red). (Multi-issue closes-keyword gotcha — #1303 auto-closed only #1250; closing manually.)

PR #1307, opened 2026-08-12
perf: gate turn-end madge behind an opt-in flag, default off (closes #766)

### LINKED ISSUE #766: perf: reduce the madge / actionable_warnings p99 tail (throughput, not input latency)

AngriestBird:
## What

Gates the per-turn-end madge circular-dependency pass behind a new flag (`--lens-turn-end-madge` / `turnEnd.madge.enabled`), **default `false`**. When off, turn-end no longer spawns any madge subprocess.

## Why

Issue #766 is about reducing the madge / actionable_warnings p99 tail. The maintainer's re-scope (comment) established: `actionable_warnings` never runs on default config (moot), and the turn-end madge pass's only production output is debug text — `hasCircular` is never surfaced as a user-facing finding; whole-project madge diagnostics come from the session-start `madge` scan cache and the `lens_diagnostics` extractor, which are separate and unaffected. Measured fresh dogfood data: p99 ~15.99s, dominated by per-target madge spawns that walk the target's transitive closure — all for a dbg-only line. The issue's own step 4 was *"decide whether turn-end Madge should exist at all."* This resolves it as: off by default, opt-in for those who want the per-edit circular note.

This is the A+B direction agreed with the maintainer: eliminate the tail by default (A) while retaining the pass and its batch machinery, reachable when enabled (B).

## How

- **`clients/lens-flag-registry.ts`**: new `lens-turn-end-madge` flag (configKey `turnEnd.madge.enabled`, default `false`, scope global) — one registry entry drives CLI + config resolution, no second place to remember.
- **`clients/runtime-turn.ts`**: the turn-end madge block is now wrapped in `getFlag("lens-turn-end-madge")`; when off, `ensureAvailable()` and `checkFilesBatch` never run. The `phase: "madge"` latency entry still records (near-zero when off) for telemetry continuity.
- **`docs/globalconfig.md`**: table row + prose note.
- **Tests**: existing #1251 batch test updated to enable the flag; new regression test asserts the pass (including `ensureAvailable`) is skipped entirely when the flag is off.

Note: the `[Unreleased]` CHANGELOG entry is intentionally omitted — this environment's pi-lens markdown autofix rewrites CHANGELOG.md's committed tab-indented continuation lines to spaces (a ~1160-line whitespace diff against a pre-existing markdownlint MD007 gap), so a clean entry couldn't be added without an unrelated whole-file reformat. Happy to add it if you'd like me to push through the normalization.


apmantza:
thanks @AngriestBird. grateful for your contributions.

Issue #1267, opened 2026-08-11
Orphan staging-file reaper can silently delete a legitimate user file (backup.tmp-2023-11)


apmantza:
## Problem

The orphaned-staging-file reaper added by PR #1260 can **silently delete a legitimate user file at session start**.

`STAGE_TMP_PATTERN` (`clients/atomic-write.ts:153`) matches names like `backup.tmp-2023-11`; `stageOwnerPidFromName` parses `2023` as the owner pid. If pid 2023 is not alive — overwhelmingly likely — `session_start` deletes the file. The code comments acknowledge the pattern matches this shape (`atomic-write.ts:144-150`) but nothing prevents the deletion. When the pattern was only a matcher this was a documented caveat (#1208 review, finding F3); #1260 attached a **deleter** to it, upgrading the caveat to a data-loss bug. Scope is limited to pi-lens data directories, but users do place files there.

Second gap, same review: the liveness regression test (`tests/clients/atomic-write-stage-gc.test.ts:47-65`) uses an injected `isPidAlive: (pid) => pid === 42001` callback — the real `realIsPidAlive` path is never exercised by any test. A regression in the default probe would go green.

## Fix

1. Require the FULL current staging shape (`.tmp-<pid>-<threadId>-<seq>` after #1231) before deleting. Sweep legacy one/two-number shapes only behind a stricter check (e.g. the prefix must be a known store filename) or not at all — a leaked pre-#1208 orphan is a few files ever, not worth a deletion false positive.
2. Add a live-foreign-pid test using the real probe (`process.ppid`), and a `backup.tmp-2023-11` fixture asserting survival.

## Note

#1228 was closed by #1260's merge; these are post-close gaps in its acceptance criteria 1 and 3.

Refs #1228 #1260 #1231.

apmantza:
Fixed by #1280 (merged). The reaper now matches only the full staging shape with strict unpadded decimal components (0|[1-9]\d*), so a legitimately-named file like backup.tmp-2023-11 or a padded x.tmp-000123-02-03 is never parsed as a dead-pid staging file and deleted. Reviewed FIX-BEFORE-MERGE (found the leading-zero gap), fixed and verified. Closed manually (branch used refs).

Issue #1212, opened 2026-08-09
Extract clients/durable-store.ts: one locking primitive for shared read-modify-write stores


apmantza:
## Problem

Roughly a dozen shared stores hand-roll `read -> mutate -> write` against files that other processes also write. Each one re-derives its own answer to the same questions — how to avoid a lost update, whether a stale writer can overwrite a newer value, what happens on crash, whether it is safe to block at process exit — and the #1202 audit confirmed that most of them get at least one of those wrong.

The codebase already contains a correct lock. `acquireInstallLock` (`clients/installer/index.ts:92-186`) uses `fs.open(path, "wx")` for `O_EXCL` atomicity, records owner pid plus `createdAt`, reclaims from a dead owner, and expires by max age to survive PID reuse. It is already OS-agnostic — `wx` is atomic on both Windows and POSIX, and there is no `process.platform` branch. But it is buried in the installer, its aging constants (120s plus 60s slack) are install-lifetime, and nothing else can reach it.

The result is that "is this store safe?" is currently answered per store, by reading it closely, rather than by pointing at one tested primitive.

## Proposal

Extract `clients/durable-store.ts` from the installer lock, generalized along the axes the audit found actually vary between stores:

- **`withLock(key, fn)`** — acquires, re-reads *inside* the lock, applies `fn`, writes, releases. The re-read is the whole point: a helper that locks around a stale in-hand value fixes nothing.
- **`tryWithLock(key, fn)`** — non-blocking, single attempt, unconditional fallthrough. Required for exit paths (see below) where blocking is not acceptable at any duration.
- **Per-store max age** — install-lifetime constants must not be copied. A heartbeat writer and an install have completely different plausible-owner windows.
- **Dropped-write counters** — per store, surfaced in telemetry. Issue #1202's invariant 6 asks that best-effort loss be *measured* rather than described as reliable; today not one of these stores counts a dropped write, which is why the audit could not quantify real-world frequency for any of them.

Then migrate stores to it incrementally, one PR each, per the sequencing in #1202. Not every store should use it — `recent-touches.ts` and `instance-registry.ts` are better served by per-pid sharding (single-writer by construction, no lock at all), and `review-graph`'s persist path should use a durable CAS rather than a lock, because locking it would serialize multi-MB gzip promotions across processes.

## Keep-alive hazard — the constraint that shapes the API

Three call sites must never block on a spinning lock, because they run while a one-shot process is trying to exit:

[...]

PR #1309, opened 2026-08-12
refactor(session): durable-store primitive + owned staging namespace (refs #1212, closes #1209)

### LINKED ISSUE #1209: Centralize the atomic-write staging namespace: mint/classify/sweep-own instead of ad-hoc .tmp-<pid> matching
## Problem

The staging-file name `<target>.tmp-<pid>` is a **cross-module protocol with no owner**. `clients/atomic-write.ts` mints it; other modules pattern-match it. Nothing declares the contract, nothing enforces it, and nothing tells a reviewer that a regex over that shape is reaching into another module's namespace.

Two P0s filed this week are the same root cause wearing different clothes:

- **#1205** — the minted name is not unique enough. It is keyed on `process.pid`, not per call, so two concurrent same-process writes to one target share a staging path and publish a torn file.
- **#1206** — someone else's sweeper deletes it. `sweepStaleStageFiles` (`review-graph/builder.ts:1916-1928`) matches `/\.tmp-\d+$/` and `fs.rm`s the result out of the shared project cache dir, destroying in-flight atomic writes belonging to `diagnostic-dispositions.ts`, `cache-manager.ts`, `project-snapshot.ts`, `call-graph.ts`, `codebase-model.ts`, `lsp/workspace-diagnostics-cache.ts`, and `project-diagnostics/cache.ts`.

Neither is a bug in the *logic* of the module it lives in. Both are a bug in the fact that a shared naming convention has no API. Fixing them one at a time leaves the next occurrence available: any module that wants to garbage-collect its own stage files will reach for the same regex, because there is nothing else to reach for.

Evidence that this is already costing coordination: the #1205 and #1206 fixes are currently in flight in separate worktrees and have to negotiate the new tmp shape **by hand through issue comments**, because neither can see the other's contract.

## Proposal

Make `clients/atomic-write.ts` the sole owner of the staging namespace and give it the three operations that callers actually need:

1. **Mint** — `createStagingPath(targetPath, { owner })`. Per-call unique (this is #1205's fix), with the owner tag embedded so provenance is recoverable from the filename alone.
2. **Classify** — `isStagingFile(entry)` and `stagingOwnerOf(entry)`. No module should ever author a regex over the staging shape again; that is the seam #1206 tripped over.
3. **Sweep** — `sweepOwnStagingFiles(dir, { owner, isPidAlive })`, which by construction can only delete entries whose owner tag matches the caller's. A module physically cannot sweep another module's staging files through this API.

Then make the naming scheme private: no module outside `atomic-write.ts` should reconstruct, parse, or match the staging shape by hand. Consider an ast-grep rule (see #1158) flagging any regex or `includes()` over `.tmp-` / `.stage-` outside that module — this whole class is mechanically detectable, which is the argument for a rule rather than a review-checklist line.

## Acceptance

1. `atomic-write.ts` exposes mint / classify / sweep-own; the staging shape appears nowhere else in `clients/`.
2. `sweepStaleStageFiles` is expressed in terms of `sweepOwnStagingFiles` rather than its own predicate.
3. A regression test asserting that module A's sweep cannot delete module B's staging file, at the *API* level — not by re-testing #1206's specific interleaving.
4. Changing the staging naming scheme requires editing exactly one file. Verify by making a trivial shape change and confirming nothing else needs to move.
5. OS-agnostic: probe the filesystem, do not branch on `process.platform`; no assumptions about path case-insensitivity or separator form (CI is Linux, dev is Windows).

## Sequencing

This should land **after** #1205 and #1206 — both are P0s already in flight and should not be blocked on a refactor. This issue then generalizes their two point fixes into the seam that prevents the third occurrence, and retires the hand-negotiated contract between them.

## Provenance

Identified while reviewing the #1202 durable-store audit alongside PRs #1200 and #1201, as the first of five centralization candidates drawn from that batch. The others (a `durable-store.ts` locking primitive, Windows path classification, a spawn-failure error taxonomy, and a cooperative-work budget helper) are tracked separately.

Refs #1205 #1206 #1202 #1158.

apmantza:
## Summary
- **#1212**: new `clients/durable-store.ts` — the single locking read-modify-write primitive (bounded-pid-file-lock → uncached in-lock reread → caller merge → atomic write `bestEffort:false` → telemetry after commit). `diagnostic-dispositions` and `actionable-warnings` fold onto it with ZERO assertion changes in their suites (including the cross-cache interleaved-writer and contention tests). Probe-cache keeps its richer quarantine/aging lock — deferred with rationale (refs, not closes).
- **#1209**: new `clients/atomic-write-staging.ts` owns mint/classify/sweep for the `.tmp-<pid>-<threadId>-<seq>` namespace; `atomic-write`, `gzip-stage-write`, and the runtime-session sweep all consume it. A mint/classify agreement guard makes the #1267 drift class impossible — mutation-verified (skewed mint marker → guard red).

## Verification
Build, lint, 7 suites / 157 tests green, assertions unchanged.

Refs #1212 (probe-cache folding remains). Closes #1209.


apmantza:
Review round 2 resolved: (1) the post-release cache-refresh race — refresh now runs in `afterWriteLocked` INSIDE the locked section, with an interleaving regression test that goes red when the callback moves back outside the lock; (2) the API misuse path — caller-supplied `read` removed entirely; the merge callback receives the authoritative in-lock state from the primitive's own locked read, and advisory reads are a separate seam that cannot flow into the commit. The lost-update shape is now inexpressible, not merely discouraged. 157/157 + 30/30 green, lint/build pass.

PR #1319, opened 2026-08-12
refactor(installer): fold the probe-cache onto durable-store; unify quarantine recovery into the shared lock (closes #1212)

### LINKED ISSUE #1212: Extract clients/durable-store.ts: one locking primitive for shared read-modify-write stores

apmantza:
## Summary
The last #1212 item: the installer probe-cache now commits through `commitDurableStoreAsync` — the delta/change-generation model feeds the authoritative in-lock merge callback, the in-memory mirror reconciles in `afterWriteLocked` (shape-12 correct), and deferred/failed flushes retain dirty state + retry.

Structural win: the quarantine-based stale-lock recovery (rename-aside + token-checked release, #1263 lineage) moved INTO `bounded-pid-file-lock.ts` — the repo now has ONE lock implementation; the installer's parallel machinery is deleted. The late-release-cannot-remove-new-owner property is preserved and tested.

Read-side TTL/existence/mtime validation unchanged. Turn-state folding deliberately not included (ownership decision recorded on the issue).

Mutation-verified: dropping the change-merge from the callback fails the "retains an update that arrives during atomic write" regression (3 red); restored green. 902 tests across installer/dispatch/probe-cache/lock/durable-store suites.

Closes #1212.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,16 +10,19 @@ AGENTS.md is the durable context handed to every agent that works on pi-lens. **
 
 **Behavior-gating durable stores serialize read-modify-write.** Atomic rename
 prevents torn JSON but not lost sibling-process deltas. Use
-`clients/durable-store.ts` for short synchronous commits; it acquires the
-bounded PID lock, performs the authoritative disk re-read internally (callers
+`clients/durable-store.ts`; short synchronous commits acquire the bounded PID
+file lock, while awaited commits acquire its shared quarantine-directory
+variant. Both perform the authoritative disk re-read internally (callers
 receive only its serialized contents through `deserialize`, never supply a
 read callback), merges only the caller's delta, and publishes through a
 throwing atomic write. `afterWriteLocked` cache refreshes run after publication
 but before lock release so another writer cannot pair its stat with stale
 committed state; telemetry or other post-success work must preserve that
 ordering when it is state-coupled. The
-PID liveness check has a documented bounded PID-reuse exposure; its unique
-token prevents a late owner from deleting a replacement lock. Callers must
+PID liveness check has a documented bounded PID-reuse exposure. Both lock
+forms use unique ownership tokens; the awaited form renames stale locks and
+releases aside before token inspection, so a late owner cannot delete a
+replacement lock. Callers must
 choose contention policy explicitly: correctness-critical stores use
 `onContention: "throw"`; dispatch-adjacent best-effort stores use `"skip-log"`
 with a drop telemetry callback and skip the whole commit when acquisition
@@ -29,8 +32,11 @@ Generic atomic-write staging names are owned only by
 `clients/atomic-write-staging.ts`: mint, strict classification, owner-pid
 extraction, and the bounded session-start sweep must stay on that seam so a
 format change cannot drift from garbage collection. The installer probe cache
-keeps its older richer lock because quarantine recovery and install-lifetime
-aging predate the short-commit durable-store seam. (#1209, #1212)
+uses the awaited durable-store seam: its delta/version snapshot maps to
+`merge`, pending-update retirement and mirror refresh run in
+`afterWriteLocked`, and TTL/existence/mtime validation remains read-side
+policy. Turn-state remains separate pending a future ownership decision.
+(#1209, #1212)
 
 ## Issue and PR design contract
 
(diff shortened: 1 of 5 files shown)


apmantza:
Review completed by the maintainer directly (the delegated review died in the plegma queue — infra, not verdict). Vectors: (1) lock unification — sync stores keep acquireBoundedPidFileLock (documented PID-reuse bound); quarantine reclamation exists ONLY on the async staleMs variant, so short-lived stores did not inherit reclamation semantics — safe by construction; (2) unawaited-commit hunt — the sole commitDurableStoreAsync caller (installer:1542) awaits, and afterWriteLocked is awaited INSIDE the lock (shape-12 holds for async callbacks); (3) Sonar S6544 — the async interface widened the sync base's void callback; fixed by splitting both off a callback-free core (commit above); (4) merge-model fidelity — the fold worker's mutation evidence (pending-update-preservation red/green) accepted, suites re-run green. Verdict: MERGE on CI.

Issue #1334, opened 2026-08-13
Umbrella: use pi host capabilities instead of hand-rolling them (render hooks, mode, dialogs, lifecycle, trust)


apmantza:
## Context

Prompted by #1327 (an external user pointing out we never use pi's tool render hooks): an audit of pi's extension API surface (`@earendil-works/pi-coding-agent` `dist/core/extensions/types.d.ts`, ~1,294 lines) against actual pi-lens usage shows we consume a narrow slice — 7 of ~30 events, 4 of ~15 `ExtensionUIContext` methods — and repeatedly hand-roll presentation and lifecycle behavior the host already provides.

The recurring shape: **strong on the data plane (events, context injection, tool registration), weak on the presentation/lifecycle plane.** Where pi offers a native capability, using it gets us theme-coherence, mode-correctness, and forward compatibility for free; rebuilding it gets us #1333-class bugs.

Current usage inventory (for reference): events `session_start`, `session_before_fork`, `turn_start`, `turn_end`, `tool_call`, `resources_discover`, `context`, `agent_end`; UI `notify`, `setStatus`, `setWidget`, `theme`; registration `registerTool`, `registerCommand`, `registerMessageRenderer` (#484), `get/setActiveTools`, `sendMessage`.

## Slices

- [ ] **S1 — Tool render hooks** (`renderCall`/`renderResult`, types.d.ts:373-375): opt-in compact one-line rendering. Tracked as #1327, implementation in flight.
- [ ] **S2 — `ExtensionMode` awareness** (`"tui" | "rpc" | "json" | "print"`): terminal-ownership and output behavior should key off `ctx` mode, not homegrown detection. Coordinates with #1333's console-migration PR (which establishes the "never write raw" invariant; this slice makes mode-dependent behavior — widget mount, notify suppression in print mode — host-derived).
- [ ] **S3 — Dialog primitives** (`ui.select` / `ui.confirm` / `ui.input`, plus `ui.custom` overlays): everything currently funnels through `notify`, including genuine questions (`/lens-allow-edit` usage errors, install confirmations) and multi-line report walls (lens-status / cascade summaries around index.ts:683/936/1055) that are overlay material.
- [ ] **S4 — Lifecycle events**: `session_shutdown` as the clean LSP-teardown hook (feeds the #1332 memory/lifecycle arc); `session_before_compact`/`session_compact` to preserve lens state across compaction the way `session_before_fork` already does for forks (#1041 pattern).
[...]

Issue #1333, opened 2026-08-13
Raw console.* writes from the in-pi extension path corrupt the pi TUI layout (26 ungated sites)


ravshansbox:
## Summary

pi-lens writes directly to the host terminal from the in-pi extension path. pi owns the terminal (raw mode + cursor-addressed diff repaints), so any byte an extension writes to stdout/stderr lands mid-frame and desyncs pi's model of the screen — the rendered layout breaks until a full repaint. Observed live in an interactive pi session.

There are **26 ungated `console.error`/`console.warn` call sites** in `clients/` that fire on real operational conditions (not verbose mode), plus **19 verbose-gated ones** that are latent (silent by default, but the same hazard the moment anyone turns verbose on).

This is a violation of an invariant AGENTS.md already states for logging — *"All debug loggers write through one buffered async writer, `clients/ndjson-logger.ts`"* — these are the pre-`createNdjsonLogger` (#454/#935) leftovers that never got migrated.

## Defect class

**A library that does not own the terminal writes to a process-global output channel.**

The class has three potential members; a repo-wide sweep found only the first is live:

| Member | Status |
| --- | --- |
| `console.*` from in-pi code (`clients/`, `tools/`, `index.ts`) | **26 ungated + 19 gated sites — the live defect** |
| `process.stdout/stderr.write` | ✅ clean — only in `mcp/{server,worker,cli,analyze-cli}.ts` and `scripts/`, which genuinely own their stdout contract |
| child-process `stdio: "inherit"` leaking a tool's output into the TUI | ✅ clean — `safe-spawn.ts` uses `"ignore"`/pipes, `lsp/launch.ts` uses `["ignore","pipe","ignore"]` / all-pipe |

Note the asymmetry that makes this a real defect rather than a style nit: **the MCP host protects itself and the pi host does not.** `mcp/server.ts:81` and `mcp/analyze-cli.ts:45` monkey-patch `console.log → console.error` precisely so no transitively-loaded module can corrupt the JSON-RPC stdout channel. `index.ts` has no equivalent guard (its only mention of console is a comment at line 242 noting verbose logging was removed).

## Sweep: ungated sites (fire on real conditions)

| Site | Fires when |
| --- | --- |
[...]

Issue #473, opened 2026-07-10
In-process subagent binds tear down the parent session's LSP fleet and runtime state mid-turn


apmantza:
## Summary

Extensions that spawn **in-process** `AgentSession`s (e.g. tintinweb/pi-subagents' `agent-runner.ts`, which builds a `DefaultResourceLoader` + `createAgentSession` + `session.bindExtensions()` inside the SAME Node process as the parent pi session) cause pi-lens's `session_start` handler to fire again — and because pi's extension-loader cache is process-global and keyed by cwd/generation, the subagent gets the SAME cached pi-lens module instance as the parent, including its module-scope singletons. pi-lens's `handleSessionStart` then destructively resets those shared singletons **while the parent session is still live and mid-turn**.

Root cause framing: pi's SDK invalidates the prior ctx only for **sequential** session replacement (`newSession()`/`fork()`/`switchSession()`/`reload()` — `core/agent-session.js:551`). `bindExtensions()` (`core/agent-session.js:1717-1739`) performs no invalidation of any *other* session and unconditionally emits `session_start`. So this is not the repeated-session_start case pi-lens already tolerates — it's **concurrent session_start under a live parent**, which the sequential contract never covered.

## Evidence

**pi SDK (`@earendil-works/pi-coding-agent` 0.80.6 dist)**
- `core/extensions/loader.js:107-109` — process-global `extensionCache` Map / `extensionCacheCwd` / `extensionCacheGeneration`.
- `core/resource-loader.js:216-220` — `reload()` only calls `clearExtensionCache()` `if (this.loaded)`; a freshly-constructed `DefaultResourceLoader` (as subagent runners create) has `this.loaded === false` on its first call → the parent's cached extension modules are REUSED.
- `core/resource-loader.js:354,368` — extension sets load via `loadExtensionsCached(extensionPaths, this.cwd, ...)`; the default subagent `configCwd` equals the parent's `ctx.cwd`, hitting the same cache key.
- `core/agent-session.js:1717-1739` — `bindExtensions()` builds a fresh `ExtensionRunner` for the new session and unconditionally `emit()`s `session_start`, re-invoking pi-lens's factory (factory re-executes on cache hits — `loadExtension` always calls `await factory(api)`).

**tintinweb/pi-subagents**
- `src/agent-runner.ts:433-446` — fresh `DefaultResourceLoader` + `loader.reload()`, in-process.
- `src/agent-runner.ts:597-604` — `await session.bindExtensions({...})`.
[...]

PR #1351, opened 2026-08-13
feat(mode): derive widget mounting and notify chatter from the host run mode (refs #1334)

apmantza:
Slice S2 of #1334 — refs #1334.

## The gap

pi-lens derived terminal ownership from nothing. It mounted its `belowEditor`
widget and pushed `ui.notify` chatter in every run mode, including `print` and
`json` — one-shot runs whose stdout belongs to the run's actual output, with no
live TUI to render into.

## API surface verified (pinned host types)

`@earendil-works/pi-coding-agent/dist/core/extensions/types.d.ts`:

- `export type ExtensionMode = "tui" | "rpc" | "json" | "print"` — line 208
- `ExtensionContext.mode: ExtensionMode` — line 213, with the host's own
  comment: *"Use `"tui"` to guard terminal-only UI such as custom components."*
- `ExtensionContext.hasUI: boolean` — line 215, *"true in TUI and RPC modes"*

**What that changed in the design:** `hasUI` and "can mount a component" are
**not** the same question. `rpc` reports `hasUI: true` because dialogs travel
over the protocol — but there is no terminal for a `belowEditor` component.
So the slice ships **two distinct predicates**, not one `hasUI` check:

| Predicate | True for | Gates |
| --- | --- | --- |
| `supportsTuiWidget` | `tui`, `unknown` | `mountLensWidget` |
| `suppressesUserNotify` | `print`, `json` | proactive `ui.notify` |

`rpc` therefore mounts no widget but still notifies — which is exactly the
slice's "rpc — no TUI widget" requirement.

## What landed

`clients/extension-mode.ts` — **stateless by design.** `mode` rides on the ctx
every event and command handler already receives, so each call site reads the
ctx it has; there is no latched global for a session replacement to leave stale
(a contrast with S5's trust state, which genuinely has no ctx at its call
sites).

- `mountLensWidget(ui, mode)` refuses in non-`tui` modes.
- All 16 user-facing `ctx.ui.notify` sites in `index.ts` now route through a
  local mode-aware `notifyUi(ctx, …)`. Suppressed messages are still written to
  the debug log — nothing is lost, only un-rendered.
- The `wireUserNotifier` getter applies the same predicate, so `clients/`-level
  degradation notices (#1333) go log-only in `print`/`json` through
  `user-notify.ts`'s existing fail-soft "no host wired" path — no new plumbing.
- `/lens-widget-toggle` now names the blocking **mode** instead of reporting
  "not supported by this pi version", which was misattributing a mode
  constraint to the host version.

The #1338 console guard is **untouched** — it enforces "never write raw to the
terminal"; this slice decides whether the host's own render path should be used
at all. Complementary, not overlapping.

## Feature detection

An absent `mode` (older host) — or a value outside the four documented modes,
e.g. one a future pi adds — reads `"unknown"` and keeps current behavior in
**both** predicates. Deliberate asymmetry: never guess a suppression that could
hide output.

## Tests (17 new, all green)

- `tests/clients/extension-mode.test.ts` (6) — mode passthrough, absent /
  unrecognized / non-string → `unknown`, and both predicates across all five
  states.
- `tests/index-mode-awareness.test.ts` (11) — through the real entry point:
  no mount in `print`/`json`/`rpc`, mount in `tui` and on a host with no `mode`
  field, the mode-specific refusal message, notify suppressed in `print`/`json`
  and preserved in `tui`/`rpc`/no-mode.

One deliberate test-design note: the widget assertions drive
`/lens-widget-toggle` rather than a second `session_start`, because #473's
concurrent-secondary guard makes a repeat `session_start` in the same process
return early — which would have made the negative assertions silently vacuous.
That is called out in a comment in the test file.

`tests/support/pi-mock.ts` gained a `mode` override, where `null` means "older
host with no `mode` field at all" (the key is deleted, not set to null).

## Verification

`npm run build` ✅ · new tests 17/17 ✅ · siblings (`index-wiring`,
`index-integration`, `lens-toggle-command`, `index-loop-block-wiring`,
`index-smells-rollup-wiring`) 49/49 ✅ · `npm run lint` ✅ ·
`npm run changelog:check` ✅

AGENTS.md documents the two-predicate split, the `rpc`-despite-`hasUI` trap, and
the "add new notifies to `notifyUi`, never a bare `ctx.ui.notify`" rule.

### DIFF
--- a/.changelog/fix-1334-s2-mode-awareness.md
+++ b/.changelog/fix-1334-s2-mode-awareness.md
@@ -1,0 +1,5 @@
+---
+section: Changed
+---
+
+- **Terminal behavior now follows the pi host's run mode (refs #1334)** — pi-lens reads `ctx.mode` (`"tui" | "rpc" | "json" | "print"`) instead of assuming it owns an interactive terminal. The diagnostics widget mounts only in `tui`, and proactive `ui.notify` chatter is logged rather than rendered in the one-shot `print`/`json` modes, so piped and machine-readable runs stay clean. `/lens-widget-toggle` now says which mode is blocking it instead of blaming the pi version. Hosts that expose no `mode` field behave exactly as before.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1109,6 +1109,36 @@ pi-tui **hard-crashes the whole host process** (`uncaughtException: Rendered lin
 - Surfaces the HOST wraps for us are safe without fitting: `ctx.ui.notify` and the tool-summary compact renderers (`tools/render-compact.ts`) build on pi-tui's own `Text`/`Markdown` components, which word-wrap internally. A 2026-07-11 audit bucketed every render surface; only raw `render(width)` implementations carry the hazard.
 - Tests for renderers must measure with the REAL `visibleWidth` (ANSI/OSC8-aware) against a narrow width — mock-based render tests are exactly what let #513 ship.
 
+## Host run mode owns terminal behavior (#1334 S2)
+
+`ExtensionContext.mode` is `"tui" | "rpc" | "json" | "print"` (pinned host
+types, line 208). **Never guess terminal ownership — read it.**
+`clients/extension-mode.ts` is deliberately STATELESS: `mode` rides on the ctx
+handed to every event and command handler, so each call site reads the ctx it
+already has rather than consulting a latched global that a session replacement
+would leave stale.
+
+Two predicates, deliberately NOT the same one:
+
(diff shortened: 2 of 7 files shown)


apmantza:
Review verdict: MERGE (adversarial pass — mutation-verified non-vacuous negatives, full bypass sweep). One scope clarification for the record: the three `setStatus` footer updates (index.ts:1575, index.ts:1866, runtime-tool-call.ts:453) are deliberately NOT mode-gated — they are passive status-bar updates whose rendering the host already owns per mode, unlike notify/widget which inject content. If a host mode ever renders status raw, that becomes an S2 follow-up.

Issue #1332, opened 2026-08-13
Umbrella: post-#1325 memory levers — TS language-service idle eviction (b2) + thin secondary instances (c)


apmantza:
Successor umbrella to #1126's phase (a) measurement (itemization posted there: ~750MB LSP children → #1325/PR #1328; ~250-280MB in-process TypeScript language service; ~100-150MB residual measured-small-and-accepted). This tracks the two remaining levers; #1126 can close onto this once #1328 lands.

## (b2) TypeScript language-service idle eviction — ~250-280MB, dispatch-ready
The main process retains fully-hydrated TS programs indefinitely (census: NodeObject 131.5MB/821k, IdentifierObject 53.8MB, SymbolObject 28.3MB, TokenObject 12.7MB + string-pool share). Approach: idle-timer eviction of the LanguageService's programs/document registry (TS supports document release + program disposal), rebuild-on-demand on next request — a few seconds of re-check is acceptable idle semantics. Evidence loop: the #1126 capture driver (scratchpad heap-capture.mjs + heap-rank.mjs, deterministic ~7.5min) re-run before/after; target: main-process heap under ~200MB post-idle. Out-of-process TS is the fallback if eviction proves insufficient — interacts with #941, don't start there.

## (c) Thin secondary instances — the multiplier, needs design sign-off
Even at ~400-500MB/instance post-#1325+(b2), N instances (extension host + warm MCP server + CLI runs + subagents) multiply toward the commit-charge exhaustion behind the silent host exits (#1122 forensics). Secondaries should ATTACH to the warm server's state over the existing IPC (#944 warm-state attachment) and share LSP servers (#941 broker) instead of hydrating duplicates — target ~100MB per secondary. Note #1325's fleet also multiplies per instance today, so (c) compounds every per-instance win. Design questions for the maintainer session: attach protocol surface, staleness semantics for attached readers, failure mode when the primary dies, and the #1129 worktree-concurrency question folds in.

Sequencing: (b2) now; (c) after design sign-off. Refs #1126, #1325, #944, #941, #1129, #1122.

apmantza:
[...]

PR #1376, opened 2026-08-14
fix(memory): per-root word-index eviction; snapshot postings de-dup (closes #1370, refs #1332)

### LINKED ISSUE #1370: Word-index retained twice (~300MB): unbounded warmWordIndexes Map + snapshot duplicate — the corrected #1332 headline lever
## The corrected #1332 headline: the main-heap hog is the word-index, retained twice (~300MB)

The 2026-08-14 re-census (corrected tooling — see the #1332 comment on the capture-isolate bug) shows the pi-lens main isolate at 353.5MB self-size with **zero** TS AST objects. The top retainers are all word-index:

- `Object` 299.5MB retained / 2,968,429 instances (postings), `Map` 268.2MB, `Array` 216.7MB+271.3MB, `PathKeyedMap` 32.8MB

**Two independent dominator paths hold the same postings:**
1. `clients/mcp/analyze.ts:64` — module-level `warmWordIndexes: Map<string, WordIndex>` keyed by root, **unbounded** (populated at :83, only wholesale-cleared at :92). With 24 discovered roots on the census box, that is 24 fully-hydrated indexes.
2. `clients/project-snapshot.ts:305` — `authoritativeSnapshots` retains a **second copy** of the postings inside persisted snapshot entries.

## Ask

1. Per-root LRU/idle eviction for `warmWordIndexes`, mirroring #1344's shape (idle timer, busy-guard, transparent rebuild) — the #1344 lease/generation lessons apply directly.
2. Audit the snapshot second copy: can the snapshot entry share the live index's postings (or hold only the serialized form) instead of duplicating the object graph?
3. Re-run the corrected census after landing to quantify (target from the evidence: main heap under ~150MB warm).

Evidence: heap `main-49512-t0.heapsnapshot` dominator analysis, 2026-08-14 diagnostic. Refs #1332, #1344.

apmantza:
## Summary

- bound MCP `warmWordIndexes` to 8 roots by default (`PI_LENS_WORD_INDEX_MAX_WARM_ROOTS`) with least-recently-used eviction
- idle-evict each root after 20 minutes by default (`PI_LENS_WORD_INDEX_IDLE_EVICT_MS`), with generation-owned unref'd timers and transparent disk deserialize on reuse
- lease every query/update use so idle and LRU eviction cannot retire an in-flight index; every removal/reset clears timer ownership
- log normal lifecycle evictions to the word-index NDJSON sink, never the degradation ledger
- release serialized snapshot postings after persistence and keep word-index bodies out of the long-lived parse cache

## Snapshot verdict

A shared live reference is unsafe: `ProjectSnapshot.wordIndex` is the serialized array representation while the warm `WordIndex` is mutable Map/PathKeyedMap state. The safe de-dup boundary is publication: the persist worker may retain serialized postings until promotion, then the authoritative entry is released. Word-index snapshot bodies are one-shot disk parses and are not retained in the parse cache, so later merge-writers rehydrate canonical serialized state without mutation coupling.

## State model / failure semantics

- acquire increments the root lease and refreshes generation/timer/LRU age
- idle callback evicts only when entry identity + timer generation still match and leases are zero; otherwise it restarts the idle window
- LRU skips leased entries and may temporarily exceed the cap until a lease releases
- release is idempotent and cannot re-arm a timer after reset/removal
- reset clears every owned timer before clearing the PathKeyedMap
- next acquisition after eviction performs a real snapshot deserialize

## Blast radius

`module_report` was unavailable in this session. The fallback structural sweep covered every warm-cache get/set/delete/clear and its two consumers: MCP per-edit updates and `symbolSearch`. Snapshot load/save dependents include runtime session hydration, reverse-deps, project reports, tool policy, and word-index persistence; targeted snapshot and word-index suites cover their shared persistence seam.

## Verification

Required order on the final tree:

1. `npm run build` — pass
2. targeted word-index + analyze + snapshot Vitest run — 19 passed, 1 skipped files; 146 passed, 1 skipped tests
3. `npm run lint` — pass
4. changelog added

Mutation proof: temporarily removing the `entry.leases > 0` guard made the suspended-query regression fail (`cache size 0`, expected `1`). Restoring it returns the targeted run to green.

Closes #1370. Refs #1332. Refs #1344.

### DIFF
--- a/.changelog/fix-1370-word-index-eviction.md
+++ b/.changelog/fix-1370-word-index-eviction.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Bound warm word-index memory (closes #1370, refs #1332)** — idle/LRU per-root eviction now releases inactive indexes safely, while persisted snapshot postings no longer remain duplicated in long-lived in-memory caches.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,7 +7,9 @@ unref'd, and cleared on every removal/reset path; lifecycle eviction belongs in
 the word-index NDJSON log, never the degradation ledger. Snapshot persistence
 may retain serialized postings only until publication: afterward authoritative
 and parse caches must not pin them, because that duplicates the mutable warm
-index's expanded postings graph. (#1370)
+index's expanded postings graph. The parse cache instead keeps a shallow
+postings-stripped snapshot for metadata/report consumers; a cold analyze reloads
+the full body once and immediately rewarms the leased per-root index. (#1370)
 
 Helm chart linting uses the shared workspace-topology `Chart.yaml` marker. YAML
 and `.tpl` edits inside a chart dispatch one canonical-root-deduplicated,
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,14 @@
 # pi-lens — agent context
 
+MCP warm word indexes are bounded per root in `clients/mcp/analyze.ts`: callers
+must acquire/release a lease around every use, because idle and LRU eviction
+must never retire an index mid-query. Idle timers are generation-owned,
+unref'd, and cleared on every removal/reset path; lifecycle eviction belongs in
+the word-index NDJSON log, never the degradation ledger. Snapshot persistence
+may retain serialized postings only until publication: afterward authoritative
+and parse caches must not pin them, because that duplicates the mutable warm
+index's expanded postings graph. (#1370)
+
 Helm chart linting uses the shared workspace-topology `Chart.yaml` marker. YAML
 and `.tpl` edits inside a chart dispatch one canonical-root-deduplicated,
 bounded `helm lint` pass through the ordinary typed availability/install seam.
--- a/clients/lens-engine.ts
+++ b/clients/lens-engine.ts
@@ -27,7 +27,7 @@ import {
 } from "./instance-registry.js";
 import { initLSPConfig } from "./lsp/config.js";
 import { getLSPService } from "./lsp/index.js";
-import { getOrLoadWarmWordIndex } from "./mcp/analyze.js";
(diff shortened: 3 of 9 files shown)


apmantza:
## P2 fix round — postings-stripped snapshot cache

Implemented the review design at `82c801e9`:

- after snapshot promotion, the authoritative full snapshot is still released, but the parse cache now receives a shallow copy with `wordIndex` omitted; metadata/report consumers stay warm without retaining the serialized postings graph
- added an explicit postings-free load seam and moved project report, tool policy, and `symbolSearch` metadata reads onto it
- `symbolSearch` now acquires the per-root warm lease first; on a cold/released miss that acquisition performs the only full-body read and immediately rewarms the LRU/idle cache, while the subsequent metadata load shares the stripped cached body
- full snapshot loads remain available to runtime hydration and merge-writers that must preserve postings

Read-count evidence in the consumer-path regression: the pre-fix sequence would perform 3 full reads (warm-index acquisition + `symbolSearch` metadata + second consumer); after the fix it performs 1. A second `symbolSearch` remains at 1, and after idle eviction the next cycle increments exactly once to 2.

Memory invariant: `_projectSnapshotParseCacheRetainsWordIndexForTests()` is false after both cycles. Mutation proof re-admitted the full parsed snapshot into the cache; the regression failed (`received true`, expected `false`) at the no-postings assertion. Restored before final verification.

Blast-radius recheck: structural tooling remained unavailable; import/call-site sweep confirms postings-free consumers are `lens-engine`, `project-report`, and `tool-policy`; full-body consumers remain runtime hydration, snapshot/word-index merge writers, and MCP cold acquisition.

Final required order on restored tree:

1. `npm run build` — pass
2. targeted Vitest — 19 passed / 1 skipped files; 146 passed / 1 skipped tests
3. `npm run lint` — pass

No full suite was run.

PR #1382, opened 2026-08-14
fix(widget): re-assert mount when the host UI is replaced; log silent mount failures (closes #1381)

### LINKED ISSUE #1381: Widget mounts at session_start against a possibly-stale ctx.ui — invisible until /lens-widget-toggle re-mounts (stale-setWidget class)
## Field repro (2026-08-14, live plegma dogfooding, build 1e71d300)

Interactive tui session; widget never appeared across many turns. `/lens-widget-toggle` twice → hidden/shown notifications worked AND the widget rendered from then on. Mount path therefore works; the SESSION-START mount produced a widget that never painted.

## Analysis

- The build predates #1367 (ports), so this is not the render-invalidator wiring; mode gating (#1351) did not suppress (no suppression log line — and note the missing-setWidget branch returns false with NO log at all, a gap to fix regardless).
- Leading hypothesis: the widget registers against the `ctx.ui` present at `session_start`; this session had a 26.6s host-side gap before session_start (see the 2026-08-14 startup diagnostic), and if the host replaces/rebuilds its UI after that point, the widget stays registered on a stale TUI that never repaints — the #1128 stale-ctx class, for `setWidget`. The toggle re-mounts on the live ctx.ui and works.
- Secondary hypothesis (can't fully exclude): renderWidget legitimately returned empty until content existed, and the timing coincided with the toggle.

## Fix direction

1. Log the silent `setWidget`-missing branch (observability first).
2. Re-assert the widget mount on a later lifecycle signal — cheapest: on the first `turn_start` after mount, compare the live `ctx.ui` identity to the mounted one and re-mount if it changed (per-call ports resolution per #1367 makes this natural on current master; the mount itself is idempotent via `setWidget("pi-lens", ...)` replacing the prior registration).
3. Regression test: mount on ctx A → replace ui (ctx B) → next turn re-mounts on B (assert setWidget called on B's ui).

Refs #1351, #1367, #1334 (S4 lifecycle), #1128 (the class).

apmantza:
## Summary

Closes #1381.

- Records the UI object identity only after a successful widget mount.
- Re-asserts the idempotent mount on 	urn_start when the live host UI changes.
- Preserves lensWidgetVisible and run-mode gating, so toggled-off/headless sessions stay off.
- Logs missing ui.setWidget once per extension session.

## Verification

- 
pm run build
- 
px vitest run tests/index-wiring.test.ts tests/index-integration.test.ts tests/index-mode-awareness.test.ts — 52 passed
- 
pm run lint
- 
pm run changelog:check — 30 entries validated
- Mutation check: removing the UI identity comparison made the replacement regression fail (13 passed, 1 failed).

### DIFF
--- a/.changelog/fix-1381-widget-remount.md
+++ b/.changelog/fix-1381-widget-remount.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Re-assert widget mounts after host UI replacement (closes #1381)** — The diagnostics widget remounts on the live UI at turn start when needed, while preserving mode and user visibility gates; unsupported widget hosts now emit a log-once diagnostic.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -212,6 +212,13 @@ clients/
   mcp/                     host-neutral facades: analyze, session, review, ipc, host-shim
   runtime-session.ts      session_start handler — snapshot hydrate, tool preinstall, background scans, LSP warm
   project-snapshot.ts     Versioned seq-stamped project snapshot cache
+
+The diagnostics widget records the exact `ctx.ui` identity only after a
+successful `setWidget` mount. A visible widget re-asserts that mount on
+`turn_start` when the host replaces its UI object; this remains gated by the
+live run mode and `lensWidgetVisible`, so a user toggle-off or headless mode is
+never undone. Missing `ui.setWidget` is a log-once-per-extension-session
+diagnostic rather than a silent mount failure. (#1381)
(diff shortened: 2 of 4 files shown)


apmantza:
Optional full-suite result: 6,261 passed, 17 skipped, 5 failed across 4 unrelated tests/files. Failures: dependency-checker madge memo/resolution (3), jscpd managed executable resolution (1), and timing-sensitive source-filter maxFiles timeout (1). Required targeted gate remains green: 52/52 tests across index-wiring, index-integration, and index-mode-awareness.

Issue #1383, opened 2026-08-14
Bus emitters go permanently stale after session replacement — widget stops reflecting subagent/LSP activity (live repro)


apmantza:
## Live repro (2026-08-14, plegma dogfooding, build 1e71d300)

Operator report: the pi agent surfaced an LSP error in its tool results, but the widget showed nothing — and the widget previously updated live while subagents worked. sessionstart.log has the mechanism:

```
06:14:08 bus-publish: pilens:files:touched emit failed (further failures suppressed):
         Error: This extension ctx is stale after session replacement or reload
```

## Analysis

- Subagents DO share the warm LSP workspace (#473 guard — by design). The widget's live updates ride the bus events the pipeline publishes.
- Something acted as a session replacement (plausibly a subagent lifecycle transition); the bus emitter wiring latched the pre-replacement ctx. The #1128 stale-ctx guard correctly refused the publish — but its fail-soft then suppresses ALL further failures, so from that point every bus event silently disappears: widget frozen, no operator-visible signal (the suppression note is one log line).
- This composes badly with #1381 (widget mounted on a stale ui — PR #1382 adds turn_start re-mount): the widget can be remounted yet still starve because the EVENT side is dead. Same class, second seam.

## Fix direction

1. Re-wire the bus emitter getters on EVERY session_start including subagent ones — the #473 guard must skip the runtime reset but NOT skip refreshing ctx-derived plumbing (emitters, notifier, render callback). Re-wiring is idempotent and cheap.
2. The stale-suppression should be occurrence-scoped like #1378's unavailable dedupe: after a successful re-wire, publishing (and failure logging) must resume — permanent suppression turns one stale event into a silently dead channel.
3. Ledger it: a dead bus channel is exactly what the degradation ledger (#1366) exists to surface — record kind=bus-stale so pilens_health shows it.
4. Regression test: publish → simulate session replacement → subagent session_start → publish again succeeds (and widget-state receives it).

Refs #1128, #473, #1381 (PR #1382), #1366.

apmantza:
**Scope correction (evidence re-verification):** two claims in this issue's original text do not survive scrutiny against build 1e71d300: (1) the widget does NOT consume bus events — recordDiagnostics is a direct pipeline call, so the observed widget behavior was #1381's mount issue, not this; (2) the 'further failures suppressed' line gates only the free-text LOG — every emit is still attempted and each failure writes a structured emit_failed record, so the channel self-heals with the ctx on that build.

[...]

Issue #1292, opened 2026-08-12
LSP hardening umbrella: pull-failure telemetry, strategy-gated didSave, lifecycle/protocol test seams


apmantza:
Umbrella for three narrowly-scoped LSP improvements surfaced by a comparative architecture review. The review's headline: the current warmup, liveness, content-binding, cache, per-server scheduling, and crash-breaker architecture is the stronger design and must NOT regress chasing any of these — each item below is additive.

## 1. Bounded pull-failure telemetry (medium-high value)

Distinguish "server doesn't support diagnostics pull" (`-32601` / unsupported — normal, push-mode servers) from "server ADVERTISES pull but it operationally failed". Today an intermittently-failing pull can masquerade as ordinary push-only behavior — the failure is absorbed by the fallback and never surfaced, so a server that half-works looks healthy.

Sketch: a bounded `pullFailureHistory` (ring buffer, ~10 entries: timestamp, method, error code/message) on `LSPClientState` near the existing diagnostic maps (`clients/lsp/client.ts` ~:2399-2413), excluded for unsupported-method errors; expose through `LSPClientInfo` → `lspStatus` and/or latency metadata. The pull retry/fallback behavior itself (~:1552-1643, :1671-1750, :3086-3129) stays unchanged.

## 2. Strategy-gated didSave after didChange (low-medium value, server-specific)

Some servers only trigger analysis / project refresh on save, not on didChange. The sync path (`clients/lsp/client.ts` ~:1756-1891) doesn't generally send didSave. Add a per-server strategy capability (e.g. `sendsDidSaveAfterChange`, default false) and enable it only for servers empirically shown to need it. Must be verified against the diagnostic-binding and debounce assumptions before enabling anywhere (a redundant didSave can produce duplicate diagnostics or invalidate content-binding expectations — test that seam explicitly).

## 3. Lifecycle/protocol test seams (cheap, high leverage per line)

Focused patterns worth adding to the existing suites:
- **Acquisition races**: two concurrent client acquisitions → exactly one initialize; an ABORTED waiter neither pins nor orphans the pending client (assert refcount/reap behavior).
- **Descendant cleanup**: killing a server wrapper also kills its descendants — Windows wrappers and shell-launched servers can leak children; the safe-spawn kill-escalation tests don't currently assert the process TREE.
[...]

Issue #570, opened 2026-07-12
LSP diagnostics timeout silently reports as confirmed-clean, erasing last-known-good state


apmantza:
## Summary

Found via live log analysis on a dogfooding project (pi-drykiss): an LSP diagnostics check that TIMES OUT is currently indistinguishable from a check that genuinely CONFIRMED zero diagnostics. A timeout silently reports as "clean", which is misleading and also erases previously-known-good diagnostic state.

## Root cause

`clients/lsp/index.ts`, `touchFile` method (~line 1060-1246):

- `notifyWriteTimedOut` (set ~line 1112): true when the didChange/didOpen notify couldn't be written to the server within budget.
- `diagnosticsTimedOut` (set ~line 1183): true when the diagnostics wait hit its deadline without the server confirming completion.
- Both are logged to `lsp_touch_file` latency events but never affect what's returned or cached.
- Lines ~1199-1219: `collected` is read from `entry.client.getDiagnostics(filePath)` regardless of timeout status, then unconditionally used to set-or-delete `lastKnownDiagnostics`/`lastKnownContentHash`. When a check times out, `collected` is very often `[]` (cache cleared for this edit, fresh push/pull never arrived) — this deletes the last-known-good diagnostic record for a file that may still have real errors.
- `touchFile`'s return value (`collected ?? []`) also doesn't distinguish "confirmed 0" from "timed out, defaulted to 0" for callers — `clients/dispatch/runners/lsp.ts` (per-edit dispatch path feeding the footer/widget via `recordDiagnostics` in `clients/pipeline.ts`) and the standalone `lsp_diagnostics` tool both just see `[]` either way.

## Fix

A timeout must never present as "confirmed clean" anywhere downstream — not in the cache, not in a tool response, not in the footer.

1. Stop erasing known-good state on timeout — only run the set-or-delete block when the check was actually trustworthy.
2. Surface the inconclusive signal to callers via `touchFile`'s return shape (reuse/extend `LSPDiagnosticsHealth` if suitable).
3. Wire the signal into `clients/dispatch/runners/lsp.ts` (status/semantic field, not silent empty diagnostics) and the standalone `lsp_diagnostics` tool response text.
4. Non-timeout path (genuine confirmed `[]`) must be unchanged.

PR #1385, opened 2026-08-14
feat(observability): ledger the remaining silent gates — LSP breakers, formatter failures, WASM aborts, timeout tallies (refs #1292)

apmantza:
## Summary

- ledger LSP permanent/window breaker trips as once-per-server/root degradations
- tally LSP diagnostics timeouts per server without retaining one entry per event
- ledger real formatter failures once per formatter/file and tree-sitter WASM poison once per client
- keep the existing project-trust install refusal as the single record for denied LSP acquisition (no spawn-side double count)

## Wiring

| Site | Kind | Semantics |
|---|---|---|
| LSP spawn/init and runtime-exit breaker terminal transitions | `lsp-breaker` | once per `serverId:root` per session |
| LSP diagnostics wait cap | `lsp-diagnostics-timeout` | exact event count; one updated retained entry per server ID |
| formatter result failures / format-service throw | `formatter-failure` | once per formatter + file basename per session |
| tree-sitter WASM poison transition | `wasm-abort` | once at the central first-abort transition |

## Already covered / skipped

- LSP acquisition under denied project trust already passes through `assertInstallAllowed("lsp install: <server>")`, which records `trust-refusal`; recording again at the spawn gate would double-count one blocked action.
- Existing #1366/#1363 coverage for mode suppressions, idle eviction, grammar blocks, formatter skips, safe-spawn failures, and install/materialization trust refusals is unchanged.
- Bus/session seams are untouched for #1383.

## Verification

- `npm run build`
- targeted Vitest: 5 files, 50 tests passed
- `npm run lint`
- mutation: removing formatter ledger write fails the focused regression (1 failed, 18 passed)
- mutation: removing WASM abort ledger write fails the focused regression (1 failed, 2 passed)

Blast-radius structural tooling was unavailable in this environment. Import/call-site fallback search covered 57 references across the touched ledger, LSP, pipeline, and tree-sitter seams; focused suites exercise each affected entry point.

Refs #1292


apmantza:
Root cause: the new diagnostics-timeout ledger tally passed entry.client.serverId, but spawned test/lightweight clients do not guarantee that field, so 	runcateForLedger(undefined) threw before touch completion. That bypassed the existing #570 guard and made the timeout path fail instead of preserving lastKnownDiagnostics. Fixed by recording the same timeout against authoritative entry.info.id; the ledger entry is retained, while inconclusive results still skip cache mutation. Verified: build, service-touch-collect 18/18, lastKnownDiagnostics test sweep (only that file), lint. Pushed as 58304c9508507cdf69f010425468669f52b8ae7f.

apmantza:
Remediated the confirmed P1 mutation-CONFIRMED degradation-ledger class at the seam: recordDegradation, recordDegradationOnce, and incrementDegradationCount now normalize unknown/nullish kind, subject, and reason values to bounded strings and swallow internal telemetry failures, preserving the 200-character field caps and 20-entry-per-kind bound; regression coverage includes undefined subjects and a forced normalization failure. The reviewer mutation temporarily changed clients/lsp/index.ts to pass undefined as the timeout subject; npm run build passed and service-touch-collect.test.ts passed 18/18, after which index.ts was restored exactly. Targeted ledger tests passed 8/8; lint passed. Full suite: 538 passed, 6 skipped, 14 unrelated pre-existing timing/environment failures.

apmantza:
Applied reviewer fixes: bounded/truncated degradation kinds with overflow routing, debug-observable non-fatal catches, and total rendering for malformed summaries. Verified build, 31 targeted tests, and lint; pushed 9af55499.

Issue #1318, opened 2026-08-12
Umbrella: review-graph read-across from the word-index arc — invariant locks, occupancy guards, O(changed) rebuild


apmantza:
The word-index arc (#1224/#1225/#1226/#1227/#1215, PRs #1302/#1315) built a playbook this umbrella ports to the review graph — they share machinery (gzip-stage-write worker persist, staged generation-gated promotion) and failure shapes.

## Slice 1 — cheap, test-only (dispatch-ready)
- **Supersession invariant lock**: the builder's worker-generation promotion gate (clients/review-graph/builder.ts:1618-1659) was classified SAFE in the shape-12 sweep by CODE READING only — port #1227 row 3's deterministic-interleaving test (yield-point spy, no sleeps): superseded build never publishes, no partial state observable, no orphaned staging.
- **Occupancy guards**: main-thread phases (delta computation, impact cascade) get `measureMaxSyncBlockMs` guards in the timing-sensitive project, mirroring the word-index guards.

## Slice 2 — the perf arc (hold for #1126 phase-(a) heap ranking; promote if the graph ranks as a top retainer)
- **#266** — delta persistence O(changed) not O(total): the direct analog of #1226's walk-reuse (reuse prior pass artifacts; per-item revalidation replaces global recompute — mind #1302's stale-window lesson: reuse the expensive pass, re-verify freshness per item consumed).
- **#202** — structural-hash short-circuit: the gate condition, as #1200's work-based gate was for word-index.
- **#1137 remnant** — getModuleSourceFiles (workspace-modules.ts:489), the hot uncached walk on the cascade path (monorepo-only; needs monorepo latency evidence; ~100-call-site async ripple documented in the #1137 disposition).
- Cooperative-budget helper (#1215's, clients/cooperative-budget.ts) is the yielding tool for any of these walks.

## Tie-in
#1126 phase (b) idle decay: if the heap ranking confirms the graph as a top retainer, cheap O(changed) rehydration (Slice 2) is what makes eviction cheap — the two compound.

Slice 1 is dispatch-ready now. Slice 2 sequencing decided when the #1126 retainer table lands.

Refs #266, #202, #1126, #1137, #1227, #1215.

apmantza:
[...]

Issue #1137, opened 2026-08-07
Fix the real 1.6-10.6s sync event-loop blocks: readdirSync dispatch paths, drive-root enumeration, Windows spawnSync (from #1122)


apmantza:
The genuine sync-block tier from #1122 (now closed — this issue is its remediation home). The 1.6-10.6s single-process event-loop blocks observed in latency.log are real pi-lens synchronous work, distinct from the machine-level 290s artifacts (Modern Standby / commit exhaustion, dispositioned in #1122).

## Suspect table (from the #1122 investigation)
- `readdirSync` on main-thread dispatch paths: clients/pipeline.ts:125, clients/lsp-diagnostics.ts:337, clients/review-graph/workspace-modules.ts:130, clients/language-profile.ts:108 — pathological on OneDrive-synced dirs (a stalled cloud-backed readdir blocks the loop for the full stall).
- `readdirSync(driveRoot)` at LSP spawn: clients/lsp/server.ts:689 — enumerates a Windows drive root synchronously.
- `spawnSync` calls on Windows in safe-spawn/launch paths (taskkill/chcp/PATH walks).
(Line numbers from 2026-08-06; re-verify. Known-fixed already: knip async, oxlint probe, launch execFileSync timeouts.)

## Approach
1. **Measure first, now cheap**: PR #1125's merged attribution (lastPhase + windowCpuMs/windowWallMs per loop_block) names suspects in production logs — pull a few days of post-#1125 latency.log samples before converting anything, and rank by observed frequency×magnitude rather than by grep. NOTE the #1125 review's documented ambiguity: >20s blocked-in-syscall stalls get tagged suspectSystemStall — for THIS work, tagged samples with a consistent lastPhase are prime suspects, not noise.
2. Convert the ranked offenders: async/chunked walks (the source-filter yieldEvery pattern), or move behind existing async seams; drive-root enumeration gets a cache or async path.
3. Each conversion: event-loop occupancy guard test where the repo's measureMaxSyncBlockMs pattern applies (timing-sensitive project per #902), and no behavior change to results.

Refs #1122 (investigation + attribution), #1125 (the telemetry making ranking possible), AGENTS.md event-loop discipline.

apmantza:
[...]

PR #1386, opened 2026-08-14
fix(perf): memoize getModuleSourceFiles on the existing freshness seam (refs #1137)

apmantza:
## What

The `#1137` remnant named by `#1318` slice 2: `getModuleSourceFiles` (`clients/review-graph/workspace-modules.ts`) was the last uncached sync walk on the cascade path.

**Assessment (call-site census):** exactly **one** production call site — `clients/review-graph/query.ts:204`, inside `computeImpactCascade`'s per-downstream-module loop. `computeImpactCascade` runs once per edited file on the (unawaited) cascade path, so in a monorepo every edit re-walked every downstream module's directory tree (recursive `readdirSync`, depth ≤4, ≤20 files each) — a genuine *repeated*-walk cost even though the call-site count is 1. Monorepo-only (module graph is `null` elsewhere).

## Change

Bounded per-module-root memo, invalidated by **existing freshness seams only** (no new scheme):

- every directory the walk visited is re-stat'd by `mtimeMs` (the repo's cheap freshness tier — same class as the ancestor-directory-mtime validation in project-config discovery); stamps are captured *after* each `readdir` so a mid-walk change fails safe into a re-walk;
- the ignore matcher is compared by **object identity** — `getProjectIgnoreMatcher` returns a fresh object whenever a `.gitignore`/`.pi-lens.json`/global-config `size:mtimeMs` input changes, so ignore-rule edits re-walk too;
- cleared by `clearModuleGraphCache()`, the existing session-reset seam;
- bounded by the workspace module count (itself bounded by the glob-expansion caps); a non-stat-able directory skips memoization rather than poisoning it.

Callers get a defensive copy; behavior (ordering, caps, ignore filtering) is unchanged.

## Tests (`tests/clients/review-graph-workspace-modules.test.ts`)

- memo hit: second call returns an equal list with **zero** `readdirSync` calls (counting wrapper over `node:fs`, delegation-only);
- staleness invalidation: adding a source file + forced dir-mtime bump (`utimesSync` to a future time, so the test can't pass vacuously under coarse mtime granularity) re-walks and sees the file;
- `clearModuleGraphCache` also clears the memo (re-walk observed);
- ignore-input change (`.gitignore` write at the pinned ignore root) invalidates via matcher identity and the newly-ignored dir drops out.

## Verification

- `npm run build` clean
- targeted vitest: `review-graph-workspace-modules` 6/6; neighboring `review-graph.service` + `tests/monorepo/` + `import-resolvers` — 78 passed / 11 files on the fresh origin/master base
- `npm run lint` (tsc) clean

Refs #1137, refs #1318 (slice 2).


apmantza:
Remediated the directory-mtime memo freshness gap: each visited-directory stamp now records its validation-time age, and entries stamped within the 2s coarse-filesystem granularity guard are treated as unverifiable and re-walked; steady-state entries older than the window retain memo hits. Added an FS-probed same-tick regression that writes a source file, aliases both file and directory mtimes to the stored stamp, verifies exact equality, and confirms discovery. Build passed; the PR test file passed all 7 tests; lint passed. Pushed as 1d1d579b7ad36a83ac3c0cb7cc061708cd53281a.

PR #1391, opened 2026-08-14
fix(format): cooperative chunking for deferred format drain (closes #1387)

### LINKED ISSUE #1387: deferred_format_queued loop blocks are CPU-bound, not IO — formatter scheduling/chunking investigation (from #1137)
Focused follow-up re-homing #1137's last open item (deferral hygiene — refs #1137, does not close it).

The top loop_block tier, deferred_format_queued (4 samples: 7.7s/3.5s/2.3s/1.8s, one at 24.3s CPU / 62.8s wall), is **CPU-bound**: no sync-readdir walk exists on that path for single-package repos, so IO conversion (the #1170 fix family) cannot address it. Needs its own investigation into formatter scheduling/chunking — e.g. whether the deferred queue formats oversized batches on one tick instead of yielding between files.

Evidence: #1137's measurement pass (16 post-#1125 samples; 9 machine-level refuted by CPU accounting, 7 genuine, this tier ranked first).
Non-goal here: getModuleSourceFiles (item 2) — addressed by memoization in PR #1386.

apmantza:
## Investigation

`deferred_format_queued` is the `agent_end` deferred-format drain. `RuntimeCoordinator.claimDeferredFormatFiles()` claims the whole eligible queue in insertion order (`clients/runtime-coordinator.ts:703-733`), and `clients/runtime-agent-end.ts` previously launched every claimed formatter in one `Promise.all(formatRecords.map(...))` burst (`:228-261`). It then ran the completed records through one synchronous result/bookkeeping loop (`:264-352`) with no inter-file event-loop boundary.

The CPU-bound work is therefore the formatter batch plus per-file CPU-side processing, not a synchronous directory walk: `runFormatPhase()` calls `formatService.recordRead()` and `formatService.formatFile()`, then rereads the formatted file (`clients/pipeline.ts:1017-1066`); changed files add synchronous project-change/read-guard/cache bookkeeping, including content splitting and import detection (`clients/runtime-agent-end.ts:292-327` on the pre-fix line map). On single-package repos this path has no sync-readdir walk. The old `Promise.all` also started all formatter work in the same turn, allowing a large queue's subprocess/result scheduling and subsequent string/file bookkeeping to form the observed loop block.

## Fix

The drain now processes each file's formatter invocation and bookkeeping together in queue order, catches failures per file as before, and awaits `setImmediate` before every file after the first. This prevents a multi-file drain from launching an unbounded burst while preserving total work, ordering, summaries, and per-file failure isolation. A single-file drain has no inter-file yield.

## Tests

- `npm run build`
- `npx vitest run tests/clients/runtime-agent-end.test.ts tests/clients/pipeline.test.ts tests/clients/format-service.test.ts` — **55 tests passed across 3 files**
- `npm run lint`
- `npm run changelog:check` — 36 entries validated

Closes #1387.

### DIFF
--- a/.changelog/fix-1387-deferred-format-chunking.md
+++ b/.changelog/fix-1387-deferred-format-chunking.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Deferred formatting yields between queued files (closes #1387)** — the `agent_end` drain no longer launches every queued formatter in one `Promise.all` burst or runs all per-file bookkeeping back-to-back; it preserves queue order and per-file failure isolation while yielding with `setImmediate` between files. Investigation found the CPU-bound loop block at `clients/runtime-agent-end.ts:228-352`: batch formatter scheduling plus synchronous result/content bookkeeping, including the `clients/pipeline.ts:1017-1066` formatter invocation and file reread, not a synchronous directory walk.
--- a/.changelog/fix-1387-deferred-format-chunking.md
+++ b/.changelog/fix-1387-deferred-format-chunking.md
@@ -2,4 +2,4 @@
 section: Fixed
 ---
 
-- **Deferred formatting yields between queued files (closes #1387)** — the `agent_end` drain no longer launches every queued formatter in one `Promise.all` burst or runs all per-file bookkeeping back-to-back; it preserves queue order and per-file failure isolation while yielding with `setImmediate` between files. Investigation found the CPU-bound loop block at `clients/runtime-agent-end.ts:228-352`: batch formatter scheduling plus synchronous result/content bookkeeping, including the `clients/pipeline.ts:1017-1066` formatter invocation and file reread, not a synchronous directory walk.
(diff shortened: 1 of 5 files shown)


apmantza:
FIX-FIRST remediation pushed in 8e4e1bdb35f0f73a071aa48a961305a23d50c279.\n\nDesign: deferred-format drain uses a 3-worker formatter pool, collects results by admission index, and applies all synchronous result/bookkeeping work sequentially with setImmediate yields between records. Per-file failures remain isolated. The ambient abort signal stops workers before starting new records; claimed-but-unstarted records are requeued with their original metadata/order for a future drain.\n\nTests: npm run build passed; runtime-agent-end suite 26/26 (including bounded concurrency, 10-file yield observation, and abort/requeue regression); formatter/pipeline targeted suites 25/25 across 3 files; npm run lint passed. Updated .changelog and AGENTS.md contract wording.

Issue #1389, opened 2026-08-14
Memory: idle-decay + bounds wave from the cache audit (#1126 arc item b)


apmantza:
Full audit of long-lived in-memory caches on master (2026-08-14; word-index and degradation-ledger excluded as already done). 40 caches inventoried; ranked by estimated retained bytes. The #1376 pattern (per-root idle timers + LRU root cap + generation guards) is the template throughout.

**Tier 1 (dispatching now):**
1. `_workspaceGraphCache` (clients/review-graph/builder.ts:159) — full ReviewGraphs per workspace, MB-scale, no workspace cap, no idle decay. NEEDS idle-decay + bound.
2. `authoritativeSnapshots` (clients/project-snapshot.ts:305) — full ProjectSnapshots, potentially tens of MB per root, no cap, replaced only by newer writes. NEEDS bound + idle-decay.

**Tier 2 (next wave):** ReadGuard reads/edits/line-hashes (read-guard.ts:283, unbounded across a long session); reverseDepsIndexCache (dispatch/integration.ts:603); tree-sitter queryCache/queryBatchCache (tree-sitter-client.ts:177/179, uncapped compiled queries); workspace-topology dirMarkerCache/walkCache (:91); widget-state files/lspServers (:115, no cardinality cap).

**Tier 3 (small, batchable):** languageProfileCache, startupScanContextCache, formatters detectionCache, yaml-rule caches, spotbugs scanCache, _eslintCache (also NEEDS-invalidation), installer resolvedPathCache, git-tracked _trackedSnapshot, lsp/config workspaceConfigs, lsp/server negative-TTL maps (expired keys never deleted), rubyDirNamesCache, slowFsVerdictCache, gitDirCache, tsconfig-paths caches (also NEEDS-invalidation), generated-artifacts headerVerdictMemo (also NEEDS-invalidation), index.ts _lspConfigInitializedCwds, agent-behavior fileEditCount.

**Raw-path-key side list (#210 class, overlaps #1388):** widget-state files/lspServers; RuntimeCoordinator _fileSeq/_fileLastProjectSeq/_pendingDeferredFormatFiles/_lspReadWarmState (AGENTS.md-documented suspects); lsp/config workspaceConfigs; languageProfileCache; startupScanContextCache; detectionCache; _eslintCache; rubyDirNamesCache; gitDirCache — each needs write+read normalizer confirmation.

[...]

PR #1393, opened 2026-08-14
fix(memory): bound tier-1 heap caches with LRU + idle decay (refs #1389)

apmantza:
## Summary

- Add 8-root LRU caps and 20-minute (env-tunable) per-root idle decay to the review-graph workspace cache and the authoritative project-snapshot cache — the two largest heap retainers from the #1389 audit.
- Preserve existing freshness/generation guards; a per-workspace epoch prevents an in-flight build from resurrecting an evicted entry, and every snapshot-deletion site clears its idle timer.
- Add idle-recovery and LRU regression coverage (eviction → cold rebuild returns equivalent data).

Refs #1389 (Tier 1 only). Does not close it: Tiers 2–3 (read-guard, tree-sitter query caches, workspace-topology, and the smaller batchable caches) remain.

## Validation

- `npm run build`
- targeted Vitest: 8 cache suites + persist suite green
- `npm run lint`

### DIFF
--- a/.changelog/fix-1389-tier1-cache-decay.md
+++ b/.changelog/fix-1389-tier1-cache-decay.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Bound and idle-evict review-graph workspaces and authoritative project snapshots (refs #1389).**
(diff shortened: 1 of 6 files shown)


apmantza:
Delta-review remediation pushed in 13d8a7989b4869961fc34f7752524ccc6d0cc02a: all-workspace review-graph resets now preserve a process-wide epoch so in-flight builds cannot resurrect stale entries; normal signature-hit reads touch the workspace LRU; every authoritative project-snapshot deletion now uses a shared delete-plus-clearTimeout helper; and deterministic gated regressions cover stale resurrection and deleted-generation idle timers. Mutation verification: removing the epoch guard made the resurrection test fail with cacheEntries=1, and disabling timer clearing made the deletion test fail because clearTimeout was not called. Build and lint pass; the eight-suite cache-adjacent run passed 156 tests.

PR #1395, opened 2026-08-14
fix(guard): close command-detection bypasses and quote-blind grants (refs #1084)

apmantza:
## Scope

Fixes the two in-scope security classes from #1084:

- Command-detection bypasses: unquoted cmd/PowerShell wrapper argv, launcher prefixes and shell keywords, combined shell flags, and escaped-newline continuations.
- Quote-blind bash read/ownership grants: command segments and redirects are quote-aware/tokenized, and grants are recorded only after successful tool results.

Deferred: stale-record false blocks/recovery on the grant store (item 3) remains out of scope and is intentionally not changed here.

## Red-first evidence

- New git-guard regressions were run against the unchanged master build first: 2 tests failed for the unquoted wrapper and launcher/keyword/flag/continuation cases.
- The successful bash-result regression asserts that cat grants read coverage only at tool_result; it fails against the pre-fix result path, which did not register ordinary bash reads there.
- Existing quote/redirect cases were retained and expanded; current behavior never grants from quoted ;|&& or > text.

## Verification

- npm run build
- Guard/read-guard/git-guard plus bash wiring suites: 9 files, 217 passed
- npm run lint
- No performance refactor; hot-path measurements omitted as non-trivial.

Refs #1084


apmantza:
Security remediation completed and pushed in 1bd319c.\n\nFour review findings addressed:\n\n1. Launcher suffix/path bypass: normalized basename plus trailing .exe/.com/.bat/.cmd stripping for all wrapper/launcher checks. Covers cmd.exe, cmd /S /C, path-qualified cmd, PowerShell, sh/bash, env, and xargs. Red-first: the new launcher regression failed on PR head 7a9279b1 (cmd.exe /c git push returned unblocked); post-fix passes.\n\n2. Embedded verb escapes: command-verb-only normalization removes POSIX backslash, PowerShell backtick, cmd caret escapes, and expands IFS whitespace; path arguments retain lexer behavior. Red-first: the four new git pu\\\\sh, git push, git pu^sh, and gitpush cases failed on 7a9279b1; post-fix passes, including the Windows-path safety case.\n\n3. Failed bash read grant: grep/search-read registration now requires event.isError !== true. Red-first: failed grep -n ...; false minted a read grant on 7a9279b1; post-fix mints none.\n\n4. Vacuous quote regressions: removed the two fixtures that passed unchanged before and after the PR; replaced them with the real launcher/verb bypass regressions above. The replacement tests were part of the red-first run and failed before the fix.\n\nVerification: build passed; lint passed; 6 relevant guard/read-guard suites passed, 200 tests total.

apmantza:
Security hardening follow-up for 1bd319cc — red-first evidence

- IFS class: added red-first coverage for `git$IFS$9push`, `git ${IFS}pu^sh`, `git$IFS$1push`, and bare `git${IFS}push`. Before the fix, the focused test failed at `git$IFS$9push`; after the fix all four are blocked. Normalization now expands IFS with positional suffixes and re-tokenizes the command-position verb. A quoted literal argument remains safe: `git add -m "$IFS is a var"` is not classified as guarded.
- Wrapper class: added red-first coverage for `busybox sh -c 'git push'`, `toybox sh -c 'git commit'`, `nix-shell -p git --run 'git push'`, and `someunknownwrapper -c "git push"`. Before the fix, the focused test failed at busybox; after the fix all four are blocked. Busybox/toybox/nix-shell are explicit known wrappers, with a recursive fail-closed backstop for unknown leading launchers using `-c`, `--run`, `/c`, or `-Command`.
- Overblock checks: `myprog -c "echo hi"`, `myprog git push`, and `myprog -c "echo git push"` remain allowed. The backstop only blocks when recursive tokenization finds an actual guarded git invocation, so literal mentions are not treated as commands.

Verification: build passed; full guard/read-guard/git-guard coverage passed: 16 files, 242 tests; lint passed; `git diff --check` passed.

Pushed SHA: `462e9d1d4e028d4e77b8295396aa447a1a9f8ae1`

apmantza:
Root-cause hardening pushed in 957118e0cbe29e9c448882847f744ad2ea1daae0.

Implemented:
- Root 1: one quote-aware command canonicalization pass replaces IFS parameter-expansion forms and positional suffixes with spaces, preserves quoted literals, collapses/re-tokenizes, and re-applies to nested run strings.
- Root 2: any non-leading git executable is treated as indirect/suspicious; unknown wrappers and arbitrary option-valued run strings recurse uniformly. echo, printf, and grep remain text-consumer exceptions to avoid false positives on literal prose/patterns.

Red-first proof against 462e9d1d:
- git$IFS push was missed before the fix.
- weird --command "git push" was missed before the fix.
- The new regression suite is green after the fix, including git${IFS%x}push, git$IFS$9push, and sh -c 'git$IFS$9push'.

Required overblock cases all passed exactly as expected:
- allowed: echo "remember to git push later"
- allowed: grep "git push" file
- blocked: git push
- allowed: docker run -c "echo hi"
- allowed: myprog --run "build"
- blocked: git commit -m "prep for git push"

Full explicit guard/read-guard run: 6 files, 158 tests passed:
- tests/clients/git-guard.test.ts
- tests/clients/bash-read-guard-integration.test.ts
- tests/clients/read-guard.test.ts
- tests/clients/read-guard-path-normalization.test.ts
- tests/clients/read-guard-symbol-read.test.ts
- tests/clients/read-guard-tool-lines.test.ts

Also passed npm run build, npm run lint, npm run changelog:check, and git diff --check.

apmantza:
Security fix pushed in 2f146b569c123d4358d8c7fc60e8caa37215f5af. Red-first: on 957118e0, the new substitution regression failed at echo \ (received allow); after the fix all substitution vectors block, including printf/grep process substitution, backticks, >(), and echo \. Group C literal text remains allowed, and existing compound cases remain blocked. Verification: npm run build; 6 requested guard/read-guard suites = 160 passed; npm run lint; git diff --check.

apmantza:
Correction to the command examples in the previous comment: red-first on 957118e0 failed at echo $(git push) (received allow); post-fix all substitution vectors block, including printf/grep process substitution, backticks, >(), and echo $(git$IFS$9push). Group C literal text remains allowed, and existing compound cases remain blocked. Verification remains: build passed; 6 requested guard/read-guard suites = 160 passed; lint passed; git diff --check passed.

PR #1397, opened 2026-08-14
fix(memory): tier-2 cache bounds + idle decay (refs #1389)

apmantza:
## Summary
- Bound and idle-evict ReadGuard per-file state while preserving the recent read/edit dependency window.
- Add per-root idle eviction and bounded retention for reverse-dependency indexes.
- Add bounded true-LRU query and query-batch caches with compiled-query disposal; batch keys remain generation-independent.
- Add bounded idle-evicting directory-marker and walk caches while retaining mtime validation.

## Verification
- 
pm run build
- targeted cache suites: 5 files, 104 tests
- 
pm run lint

## Scope
Widget-state remains deferred because its raw-path-key sensitivity requires a dedicated PR. Tier 3 remains deferred. Refs #1389 (does not close the issue).


apmantza:
@BODY@

apmantza:
Diagnosed and fixed the remaining timer leak. The surviving handle was a THIS-PR timer from the new tier-2 workspace-topology cache: clients/workspace-topology.ts:74 (	ouchDirMarker), armed while cascade neighbor ignore checks discover directory markers. It was not the reverse-deps idle timer or the LSP references race timer. origin/master comparison passed the same cascade test (38/38), confirming the leak was introduced by this PR's cache-bound timers. Added 
eleaseWorkspaceTopologyIdleTimers() and invoke it from computeCascadeForFile's final cleanup; entries remain reusable and re-arm eviction on next access. Mutation proof: removing only that cleanup call reproduced the failure (37 passed, 1 failed); with it restored, cascade is 38/38. Verification: build passed; cache/read-guard/topology sweep 13 files / 214 tests passed; lint passed; changelog:check passed (42 entries). Pushed fix as 0603cc51c9421c4b917b7493126bd4525e73cf6.

PR #1400, opened 2026-08-14
fix(guard): recover from stale grant/guard records to prevent false blocks (closes #1084)

### LINKED ISSUE #1084: Guard store: stale-record false blocks / recovery (residual of #1084 after #1395)

apmantza:
Fixes the stale-record gap in clients/git-guard.ts:548-568: a clean per-file reconciliation removed affectedFiles but retained blockerContent, so the persisted guard continued to block later commit/push attempts. Recovery now revalidates explicit blockingFiles provenance and clears content only when the reconciled file was the last known blocker; records without provenance remain fail-closed. Verification: npm run build; 6 guard/read-guard files, 162 tests passed; npm run lint; npm run changelog:check; mutation disabling recovery fails the stale test with blocker_without_affected_file; sibling blocker remains enforced. Closes #1084.


apmantza:
Fixed and pushed as c72e9e246cafca112a159518ed362875d8f8e050. The recovery path now fails closed unless parsed blocker-content paths exactly match a unique, well-typed, non-empty blockingFiles provenance set. Incomplete, malformed, or ambiguous provenance is marked unknown and left blocking. Complete sole-blocker provenance still recovers normally. Regression coverage: persisted A+B blocker content with blockingFiles=[A] and no live inline B remains blocked after reconciling A; forged blockingFiles entries [123] and [null] remain blocked. Mutation proof: temporarily changed recovery to treat any non-empty blockingFiles as complete; the incomplete-provenance test went red (false clear), and both forged cases went red (TypeError while trusting malformed provenance). Restored fix and reran green. Verification: npm run build; npm run lint; npm run changelog:check; six guard/read-guard files 151 passed; git-guard 23 passed (20 existing sibling tests + 3 new parameterized cases); combined selected suite 7 files, 174 passed.

PR #1422, opened 2026-08-15
docs(skills): fix lsp-navigation parameter names; qualify source-repo-only paths

ELA718:
## Problem

The `pi-lens-lsp-navigation` skill instructs agents to call the tools as PRIMARY code-intelligence, but every documented example uses parameter names the tool schemas reject:

```ts
lsp_diagnostics({ filePath: "src/file.ts" })   // schema accepts only path/paths
lsp_navigation({ filePath: ..., line: ... })     // schema accepts only path
```

`lsp_diagnostics` declares `path`/`paths` (dist/tools/lsp-diagnostics.js) and `lsp_navigation` declares `path` (dist/tools/lsp-navigation.js). There is no `filePath` alias anywhere, so a call made exactly per the doc sends no target and errors. 22 occurrences fixed with a mechanical rename; table shapes untouched.

## Also included

- **write-ast-grep-rule / write-tree-sitter-rule**: the cited paths (`ast-grep-napi.ts`, repo-root `runner.run(ctx)`, `tests/support/real-runner-ctx.ts`, `tests/clients/...`) exist only in a pi-lens source checkout — the installed npm package ships no such files. Annotated so agents loading the shipped skill don't hunt for missing paths. Also added `php` to the enabled language-dir list (`rules/tree-sitter-queries/php/` ships in the package).
- Kept `ctx.filePath` inside the runner-API code sample untouched — that is the runner's own context shape, not a tool parameter.

## Verification

- `grep -ci filepath` in the lsp-navigation skill: 22 → 0
- diff vs the shipped skill files is limited to exactly these changes

### DIFF
--- a/skills/pi-lens-lsp-navigation/SKILL.md
+++ b/skills/pi-lens-lsp-navigation/SKILL.md
@@ -13,39 +13,39 @@ Use `lsp_diagnostics` before builds/tests or after touching several files:
 
 | Need | Tool call |
 |---|---|
-| Check one file | `lsp_diagnostics({ filePath: "src/file.ts" })` |
-| Check a folder | `lsp_diagnostics({ filePath: "src/", severity: "error" })` |
-| Check exact touched files | `lsp_diagnostics({ filePaths: ["src/a.ts", "src/b.ts"], concurrency: 8 })` |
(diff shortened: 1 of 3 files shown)


apmantza (approved):
@ELA718  thanks!

Issue #1412, opened 2026-08-14
LSP diagnostics disagree with the project's own tsc on test files (vitest Mock false positives)


apmantza:
## Summary

While working in `apmantza/pi-free`, `lsp_diagnostics`/`lens_diagnostics` repeatedly reported **errors that the project's own compiler does not produce** on the same files:

```
tests/provider-cache.test.ts:199  error  Property 'mockResolvedValueOnce' does not exist on type '(input?: string | URL | globalThis.Request, init?: RequestInit) => Promise<Response>'
tests/provider-cache.test.ts:206  error  (same)
lib/util.ts:476,483               error  (same)
tests/lazy-compat.test.ts         error  transient variant on vi.mock/mocked imports
```

## Why this is a false positive

- `tsconfig.json` **includes** the test files (`"include": ["**/*.ts"]`, no test exclusion).
- `npx tsc --noEmit` exits **0** on HEAD and on the working tree — repeatedly verified.
- The pattern is `vi.spyOn(globalThis, "fetch").mockResolvedValueOnce(...)` / `(globalThis.fetch as Mock)` — standard vitest typing.

So the LSP integration is checking these files with different effective compiler options / type resolution than the project tsconfig (likely missing vitest's types or a different tsconfig root), producing persistent advisory noise.

## Impact

Agents burn turns investigating non-existent errors (happened twice in one session) and users lose trust in the advisory channel.

## Repro

```
git clone https://github.com/apmantza/pi-free
cd pi-free && npm ci
lsp_diagnostics path=tests/provider-cache.test.ts   # reports mockResolvedValueOnce errors
npx tsc --noEmit                                    # clean
```

Suggested direction: ensure the diagnostic LSP session uses the project tsconfig's resolved `types`/`lib` for included test files, or cross-check blocking LSP errors against the project's own program before surfacing them as errors.

apmantza:
Triage: labeled bug + area:lsp + area:diagnostics. From the 2026-08-14 dogfood sessions in pi-free: persistent vitest Mock false positives (mockResolvedValueOnce on globalThis.fetch) while npx tsc --noEmit is clean — the LSP session is resolving different effective compiler options/types than the project tsconfig. Research dispatched: how our tsserver/LSP session locates and applies tsconfig (root selection, types resolution) versus plain tsc, why vitest's types are dropped, and whether we can cross-check blocking errors against the project's own program before surfacing. Findings will be posted here.

apmantza:
[...]

Issue #589, opened 2026-07-12
installer: managed tools never upgrade after a version-pin bump (no version-drift detection)


apmantza:
## Problem

\`ensureTool()\`/\`getToolPath()\` (\`clients/installer/index.ts\`) resolve an already-installed managed tool purely by **existence + runnability**: \`getToolPath()\` checks whether a binary exists at the managed local install path (\`TOOLS_DIR/node_modules/.bin/<binaryName>\`, or the platform-package binary), and \`verifyToolBinary()\` (line 1439) just spawns \`<bin> --version\` and checks the exit code — it captures \`stdout\`/\`stderr\` but **discards the actual version string**, never comparing it against the tool's current \`packageName\` pin (e.g. \`"jscpd@5.0.12"\`).

Consequence: once a tool is installed under pi-lens's managed \`TOOLS_DIR\`, it is **never re-installed or upgraded** just because the code's pin changes. A version bump (like #582's \`jscpd@3.5.10\` -> \`jscpd@5.0.12\`) only affects *fresh* installs (new users, or someone who wipes \`TOOLS_DIR\`) — every existing installation silently keeps running the old pinned version forever. The only reinstall trigger today is \`forceReinstall\`, which is only invoked when a PATH-resolved binary is detected as broken at launch (\`opts?.forceReinstall\` in \`ensureTool()\`), not on a version mismatch.

This means #582's whole ~50x jscpd speedup (and any future pin bump) doesn't actually reach anyone who already had the tool installed before the bump shipped.

## Fix direction

On \`ensureTool()\`'s existing-install fast path (or inside \`verifyToolBinary\`/\`getToolPath\`), compare the resolved binary's reported version (already captured in \`verifyToolBinary\`'s \`stdout\`, just currently discarded) against the tool's current \`packageName\` pin (parse the \`@x.y.z\` suffix when present — not every \`TOOLS\` entry pins an exact version, e.g. \`packageName: "madge"\` has none, so this only applies to pinned entries). On mismatch, trigger the existing \`forceReinstall\` path rather than building a new reinstall mechanism.

Things to get right:
- Only applies to \`installStrategy: "npm"\` entries with an explicit \`@version\` in \`packageName\` — tools without a pinned version (unpinned \`"latest"\`-style entries) have no drift to detect.
[...]

PR #1437, opened 2026-08-15
Fix managed TypeScript classic LSP fallback

### LINKED ISSUE #1436: Managed TypeScript 7 breaks the classic LSP fallback
## What happened

The managed `typescript-language-server` fallback can become permanently unusable after npm resolves its TypeScript peer to TypeScript 7.

Observed with pi-lens 4.0.0:

```text
lsp spawn typescript: failed ... Could not find a valid TypeScript installation.
Please ensure that the "typescript" dependency is installed in the workspace
or that a valid `tsserver.path` is specified.
```

`~/.pi-lens/tools` contained:

- `typescript-language-server` 5.3.0
- `typescript` 7.0.2
- no `typescript/lib/tsserver.js`

Projects with their own TypeScript 7 install correctly use the native `tsc --lsp --stdio` path. The failure affects the *classic managed fallback* used for standalone JS/TS files and projects without a local TypeScript package.

## Why it matters

Every classic fallback launch fails, and the widget repeatedly reports `LSP Failed: typescript`. A direct `typescript-language-server --version` probe still succeeds, so installer health checks do not detect the incompatible compiler peer.

## Proposed approach

1. Pin the installer registry's managed `typescript` tool to a classic-compatible release (`typescript@5.9.3`).
2. In `findTsserverPath`, if the initially resolved managed `tsc` has no adjacent `lib/tsserver.js`, force-reinstall the pinned managed TypeScript once and retry path resolution. This self-heals existing TS7 managed trees and stale positive probe-cache entries.
3. Preserve project-local TypeScript 7 native-LSP selection unchanged.
4. Add a regression test proving an incompatible managed compiler triggers one forced reinstall and resolves the classic `tsserver.js`.

## Acceptance criteria

- Project-local TS7 still launches `tsc --lsp --stdio`.
- Project-local TS <=6 still launches `typescript-language-server` with its local `tsserver.js`.
- A project without local TypeScript gets a managed classic-compatible compiler.
- An existing managed TS7 tree self-heals without manual cache deletion.
- Disabled installation remains discovery-only and does not mutate managed tools.

## Non-goals

- Downgrading project-local TypeScript 7.
- Replacing the existing native TS7 path.
- Globally installing TypeScript or changing the user's project dependencies.

## Test matrix

- Linux/macOS path layout
- Windows `.cmd` managed `tsc` layout
- install enabled vs disabled
- fresh managed install vs existing incompatible managed TS7

Don-Yin:
## Summary

- pin the managed classic TypeScript compiler to `typescript@5.9.3`, the newest release compatible with the shared managed tool tree that still ships `lib/tsserver.js`
- detect an existing incompatible discovered/managed compiler and force-reinstall the pinned fallback once
- keep project-local TypeScript 7 on the native `tsc --lsp --stdio` path
- preserve discovery-only behavior when installation is disabled

## Reproduction

The dogfood tools tree had `typescript-language-server@5.3.0` plus `typescript@7.0.2`. The wrapper itself passed `--version`, but every classic launch failed initialization because TypeScript 7 no longer ships `tsserver.js`.

## Verification

- `npm run build`
- `npx vitest run tests/clients/lsp/server-policy.test.ts tests/clients/lsp/typescript-native-lsp.test.ts tests/clients/installer/managed-tool-ids.test.ts` — 70 passed
- `npm run lint`
- `npm run build:dist`
- `npm run check:lockfile`
- `git diff --check`
- live dogfood verification: the previously failing managed roots successfully spawned `typescript-language-server` after replacing the incompatible TS7 fallback

`npm test` completed with 6,419 passing and 9 unrelated existing/environment-sensitive failures (path-alias/cache fixture tests, CodeRabbit temp-path fixture, and the existing `version-drift` PATH-discovery case); the three focused files above are green.

Closes #1436.


apmantza (changes requested):
Thanks @Don-Yin. looking into this

apmantza (commented):
Thank you for this contribution, @Don-Yin — clear diagnosis, and the fix lands in the right place.

> **Note:** this review is AI-generated (Claude, supervised by the maintainer). The maintainer reads every review before it guides a merge decision.

I verified the pin uses the repo's existing single-source-of-truth mechanism (`parsePinnedVersion`, precedent `jscpd@5.0.12`). I verified `typescript@5.9.3` ships `lib/tsserver.js`. I verified the diff composes with #1412's ancestor walk, and that Unit tests ran and passed. Four asks before merge. We may also push these deltas to your branch directly — maintainer edits are enabled.

**1. The "once" guard is per-call, not per-process.** Nothing persists the attempt. `findTsserverPath` has three call sites: TypeScript (`server.ts:1743`), Vue (`:2765`), Svelte (`:2815`). A successful repair is fine — `ensureTool`'s cache short-circuits later calls. A failed repair is not. Offline or partial installs re-run the full force-reinstall on every spawn, each with a 120 s timeout. Please add a module-level `let classicTsRepairAttempted = false` so a failing repair runs at most once per process.

**2. The detection misfires on a PATH-global compiler.** `getToolPath` returns the bare string `"tsc"` for a PATH hit (`installer/index.ts:2336`). `path.dirname("tsc")` is `"."`. Both candidates become cwd-relative and always miss. A user with a healthy global TypeScript 5.x then gets an unconditional forced install. Two options: skip the repair when the discovered path is not fully qualified, or make the predicate version-based (repair only on major >= 7). The version-based predicate also makes the log message honest.

**3. Two sentences in AGENTS.md are factually wrong.** `typescript-language-server@5.3.0` declares no dependencies and no peer dependencies — I checked the full registry manifest. npm never resolved a compiler peer to TypeScript 7. The real cause: pi-lens's own `typescript` tools entry was unpinned, so the managed install took `latest`. Please state that instead. The claim about consumers rejecting TypeScript 6 also fails for the tools I sampled (`svelte-language-server` accepts `^6.0.2`; `@vue/language-server` is `*`). Cite the consumer that blocks 6, or say 5.9.3 is the conservative choice.

**4. Test additions.** Your new test genuinely fails pre-fix — I traced it. Three gaps remain:
- Assert `toHaveBeenCalledTimes` on the forced call, so a refactor cannot silently double-install.
- Add a case where the repair still yields no `tsserver.js`. It should return undefined and not retry.
- Cover AC-5 with a discovered `tsc` present under `allowInstall: false`. The existing test mocks `ensureTool` to undefined, so it never reaches the new branch.

Two smaller items. Add a follow-up issue reference next to the pin comment, so we know when to revisit 5.9.3. And SonarCloud fails on one metric only — new duplicated lines at 10.8% against 3%, almost all test scaffolding. A small shared helper for the tmp-dir and mock setup should clear it.

One observation, not a change request: the `#589` drift path would self-heal a TS7 tree once the pin lands, but only on the slow path — a warm probe-cache hit returns before any version check. Your explicit repair does real work and is not redundant.

Nice find on the `--version` probe passing while the server is unusable. That is the kind of health check that lies.

apmantza:
Thanks again, @Don-Yin. I pushed the review deltas to your branch directly, per the repo's external-PR convention. Your commit is untouched; mine sit on top.

> **Note:** these deltas are AI-generated (Claude, supervised by the maintainer). The maintainer reads every change before merge.

**What I pushed**

- `ad4c0b0` — asks 1, 2 and 6. Module-level `classicTsRepairAttempted` caps a *failed* repair at one attempt per process; successful repairs still short-circuit through `ensureTool`'s cache. `_resetClassicTsRepairForTests` is the test hook. The trigger is now version-based: read the TypeScript package next to the resolved binary, repair only major >= 7, leave an unreadable version alone. That closes the bare-PATH `tsc` case. The log line names the version it found. Added the revisit note next to the registry pin.
- `b1bd032` — ask 3. AGENTS.md now states the real cause: pi-lens's own unpinned `typescript` tools entry resolved `latest`. Dropped the peer-dependency claim and the TypeScript 6 rejection claim; 5.9.3 is described as the conservative choice.
- `dd99902` — asks 4 and 5. Extracted `createManagedTypeScriptTree` and `mockLaunchedProcess`, which removes the duplicated scaffolding SonarCloud flagged. Added the exact-count assertion on the forced reinstall, a repair-yields-nothing case that also proves no retry, the AC-5 discovery-only case, and a PATH-hit case.

**Verification**

`npm run build`, then `npx vitest run tests/clients/lsp/server-policy.test.ts tests/clients/lsp/typescript-native-lsp.test.ts tests/clients/installer/managed-tool-ids.test.ts` — 73 passed. `tests/clients/lsp/` — 645 passed, 5 skipped. `tests/clients/installer/` — 151 passed. `npm run lint` clean.

I mutation-checked both new guards. Removing the once flag fails the retry case. Restoring the tsserver-absence predicate fails the PATH case.

Please object or amend anything you disagree with. It is your PR, and the diagnosis was yours.

apmantza:
merged. appreciate the contribution @Don-Yin

PR #1439, opened 2026-08-15
feat(diagnostics): add model capacity guidance

### LINKED ISSUE #1438: Make diagnostic remediation model-capacity-aware
## Problem

Pi-lens injects the same generic remediation framing for every model:

> Address blockers before continuing; advisories are informational only.

The message does not expose the active model's context window, output ceiling, or current context usage. A model therefore cannot distinguish a small remaining budget from a large one when a diagnostic backlog is bigger than one turn. This encourages two bad outcomes: attempting an unbounded repair wave, or silencing findings to make the warning disappear.

The root defect shape is **delivery planning without capability provenance**. Diagnostic production and persistence are correct; the model-facing delivery omits the capacity facts needed to plan bounded root-cause fixes.

## Proposed behavior

At the Pi `context` delivery boundary, add a compact, exact model-capacity line to current turn-end diagnostic messages when the host exposes the data:

- provider/model id
- `contextWindow`
- `maxTokens`
- current context tokens and percentage from `ctx.getContextUsage()`, when available

The guidance must explicitly say that capacity changes batching only. It must never change severity, truncate or suppress additional findings, mark findings resolved, or weaken durable delivery. When the backlog does not fit one turn, instruct the model to use `lens_diagnostics` with `paths` and resolve findings in bounded batches while keeping unresolved findings live.

MCP and older Pi hosts that do not expose model metadata keep the current message unchanged.

## Acceptance criteria

- [ ] Current Pi turn-end findings include exact available model-capacity metadata when present.
- [ ] The message says to fix root causes and never suppress findings merely to fit the budget.
- [ ] Small budgets produce batching guidance, not fewer diagnostics or altered severity.
- [ ] Missing, malformed, non-finite, or non-positive metadata is omitted rather than guessed.
- [ ] Historical finding framing remains historical and does not claim current remediation work.
- [ ] Peek and consume paths render the same capacity guidance.
- [ ] MCP acknowledgement and persistence/consumption semantics remain unchanged.
- [ ] Regression tests cover full metadata, partial metadata, malformed values, current usage, and no-metadata compatibility.

## Non-goals

- Inferring parameter count from a model name.
- Changing warning/blocker caps or diagnostic scan scope.
- Automatically suppressing, deferring, or marking findings.
- Automatically changing models or thinking levels.

## Class sweep

Searched `turnEndMessage`, `consumeTurnEndFindings`, `peekTurnEndFindings`, all context-injection sources, `before_agent_start`, `ctx.model`, `contextWindow`, and `maxTokens`. The turn-end bridge in `clients/runtime-context.ts` is the only model-facing diagnostic remediation frame. Test findings have a separate fail-now contract; session guidance and agent nudges are not diagnostic backlog planners. MCP has no model context and should use the existing optional-data fallback.

## Test matrix

| finding state | model metadata | usage | expected |
|---|---|---|---|
| current blocker | complete | available | exact model/capacity/usage line + root-cause batching guidance |
| current blocker | complete | unavailable | exact model/capacity line, no guessed usage |
| current blocker | partial | available | only validated fields shown |
| current blocker | malformed/non-positive | any | invalid fields omitted |
| current blocker | absent | any | byte-compatible existing framing |
| historical | complete | available | historical framing unchanged |
| peek vs consume | complete | available | identical rendered content |

Don-Yin:
## Summary

- include validated active-model context window, maximum output tokens, and current context usage in current turn-end diagnostic delivery
- state explicitly that capacity changes batching only and never suppresses, defers, resolves, or changes diagnostic severity
- direct models to use `lens_diagnostics` with `paths` for bounded root-cause repair waves when the backlog exceeds one turn
- preserve byte-compatible delivery when capacity metadata is missing or invalid, preserve historical framing, and leave MCP behavior unchanged

## Invariants

- diagnostic production, caps, severity, persistence, consumption, and commit-gate state are unchanged
- only current findings receive current model-capacity guidance
- invalid/non-positive capacity values are omitted rather than guessed
- peek and consume render identically

## Defect-class sweep

Searched all turn-end, test, session-guidance, and agent-nudge context deliveries plus MCP callers. `clients/runtime-context.ts` is the only diagnostic backlog-remediation frame. Test failures retain their separate immediate-fix contract; MCP lacks model context and uses the optional-data fallback.

## Verification

- red-first: the two new capacity tests failed against pre-fix code because the message lacked model metadata
- `npm run lint`
- 93 focused tests passed across runtime context, index integration, turn delivery, cascade, event flow, and MCP session consumers
- full suite: 6,422 passed / 39 skipped; four unrelated existing macOS/environment-sensitive failures remained in path-alias, installer-cache, and CodeRabbit CLI tests and reproduce in isolation

Closes #1438


Don-Yin:
Closing because this proposed model-capacity feature did not match the actual request. The requested anti-suppression/root-cause remediation policy is being applied as a local Pi configuration instead of an upstream pi-lens behavior change.

Issue #1470, opened 2026-08-15
A cut-off auxiliary yields a conclusive touch: a hung opengrep reads as confirmed-clean


apmantza:
## Summary

A touch whose auxiliary scanner was cut off by our grace timer returns `inconclusive: undefined` and `confirmation: "confirmed"`. A touch whose auxiliary went silent inside its own budget returns `inconclusive: true`. The two cases carry the same amount of evidence about the file — none from that scanner — but only one says so.

The practical shape: a hung opengrep produces a result that reads as confirmed-clean on the security lane.

## Evidence

Observed while verifying #1458, driving three probes through the real touch path:

- opengrep silent inside its 2500 ms budget, primary 800 ms → `outcome: "silent"`, touch `inconclusive: true`. Honest.
- opengrep hung, our 2000 ms grace timer wins → `outcome: "cut_off"`, `auxCutOffServerIds: ["opengrep"]`, touch `inconclusive: undefined`, `confirmation: "confirmed"`. Not honest.

## Why it is only visible now

The asymmetry predates #1458. `inconclusive` is `notifyWriteTimedOut || diagnosticsTimedOut` (R8, #714), and neither commit on that branch touched it. What changed is that the new `cut_off` outcome makes the case observable in telemetry for the first time — before, both paths reported the same word.

## Suggested direction

Treat a cut-off auxiliary as evidence-absent for the files it covers. Either mark the touch inconclusive when `auxCutOffServerIds` is non-empty, or narrow the confirmation so it states which servers actually answered. The second is more useful: a TypeScript answer is still trustworthy when only opengrep was cut off, and collapsing the whole touch to inconclusive would throw that away.

## Observability

`lsp_aux_wait_outcome` already records the outcome per auxiliary per touch, so the fix is verifiable from logs the day it ships: a cut-off row should coincide with a touch that no longer claims confirmation for that server's coverage. No new record needed.

## Acceptance criteria

- A cut-off auxiliary no longer yields an unqualified confirmed-clean touch.
- A silent auxiliary keeps its current honest behavior.
- A primary that answered stays trustworthy when only an auxiliary was cut off.
- Class sweep: check the other auxiliaries (ast-grep, typos, zizmor) for the same shape, and note whether `notifyWriteTimedOut` has an equivalent gap.

Refs #1458, refs #714.

PR #1471, opened 2026-08-16
fix(lsp): derive default aux grace from auxiliary server wait budget

### LINKED ISSUE #1458: fix(lsp): opengrep per-edit lane starved by 500ms aux grace; findings reach the agent only via the mode=full CLI sweep
## Summary

The per-edit opengrep LSP lane almost never delivers a finding. The 500 ms auxiliary grace period hard-caps opengrep's own 3500 ms diagnostic budget, so its ~1.3 s warm scan is cut off before it publishes. The documented fallback — "late aux results land in the client cache and surface on the next edit" — is unreachable: the next edit clears that cache and then demands a strictly newer document version. Opengrep security findings therefore reach the agent only through the `mode=full` CLI sweep, which costs 14–66 s per run.

This is the defect behind the report "opengrep LSP is not working as intended — errors surface when the CLI fires". The findings are real; the always-on lane is what is broken.

## Evidence

Live logs, `~/.pi-lens/latency.log` + `latency.log.1`, runtime build `fcdb34ad`.

Opengrep is the most-cut auxiliary, on the majority of touches that ran with auxiliaries:

```
auxCutOffServerIds frequency (both latency logs)
  opengrep  133
  typos     110
  ast-grep   12
  zizmor      1

lsp_touch_file rows with clientScope="with-auxiliary": 194
```

Sample row — the primary answered, opengrep did not:

```
{"phase":"lsp_touch_file","filePath":".../pi-webaio/bin/cdp.mjs","durationMs":690,
 "metadata":{"clientScope":"with-auxiliary","failureKind":"success",
 "collectedDiagnostics":17,"auxCutOffServerIds":["opengrep"]},
 "ts":"2026-08-15T21:52:30.330Z"}
```

Lane divergence, measured by rule-id prefix. The CLI adapter emits `opengrep:<checkId>` (`clients/project-diagnostics/runner-adapters/opengrep.ts:47`); the LSP lane emits `Semgrep:<checkId>` (opengrep tags its LSP diagnostics `source: "Semgrep"`, `clients/dispatch/auxiliary-lsp.ts:146`). Across every agent transcript in `~/.pi/agent/sessions`:

```
opengrep: (CLI sweep)   60 findings
Semgrep:  (per-edit LSP) 12 findings, 9 of them on one YAML file
```

Per-edit sinks confirm the same split. Across all `~/.pi-lens/projects/*/worklog.jsonl` (3623 rows), opengrep accounts for 4 — the most recent dated 2026-07-30. Rules the CLI reports repeatedly and the LSP lane never reported include `javascript.lang.security.audit.detect-non-literal-regexp` (25 CLI hits, 1 LSP hit) and `javascript.lang.security.audit.spawn-shell-true` (9 CLI, 1 LSP).

The lane still works where the primary language server is slow or absent: the one recent LSP-lane finding, `2026-08-15T21:18:32.834Z` in `~/.pi-lens/projects/Users-apman-OneDrive-Desktop-pi-webaio/code-quality-warnings.jsonl`, is `Semgrep:yaml.github-actions.security.github-actions-mutable-action-tag` on `.github/workflows/ci.yml`. Success is inversely correlated with primary speed — exactly what a "primaries settle, then 500 ms" race predicts.

This is not the historical crash-loop of #974 / #1127. `lsp_server_respawn` appears zero times for any server across `latency.log`, `latency.log.1`, `sessionstart.log`, `bus-events.log` and `extension.log`.

## Mechanism

Three code facts compose into total loss.

1. The budget is unreachable. `clients/lsp/wait-policy/strategies.ts:242-250` gives opengrep `aggregateWaitMs: 3500`, with the comment "Warm per-file scan ~1.3s". `clients/lsp/index.ts:2721` then races every auxiliary against `auxGraceMs` (default 500 ms) measured from the moment all primaries settle. The per-server timeout at `clients/lsp/index.ts:2676` is therefore never the binding constraint for an auxiliary. Opengrep's declared budget cannot be spent.

2. The stated fallback does not exist. `clients/lsp/index.ts:2725-2727` claims "Late aux results are dropped from the wait (advisory only — they land in the client cache and surface on the next edit)." The next edit invalidates that cache before reading it. `clients/lsp/client.ts:1959-1961` calls `clearDiagnosticsForPath` on every resync where `preserveDiagnostics` is false, which is the ordinary per-edit path.

3. Even without the clear, the version guard rejects the stale result. `clients/lsp/index.ts:2343-2345` snapshots `diagnosticsVersion` per client at touch start and passes it as `minVersion`; `clients/lsp/client.ts:1818-1820` resolves only when `state.diagnosticsVersion > minVersion`. A publication produced for the previous edit is by definition not newer than the current baseline.

Opengrep is uniquely exposed because it is the only server with `reopenOnResync: true` (`clients/lsp/wait-policy/strategies.ts:249`). Each edit sends `didClose` + `didOpen` and resets `diagnosticPublicationCounts` to 0 (`clients/lsp/client.ts:1966-1982`), so every edit restarts a full file scan from zero. There is no partial progress to carry forward.

## Acceptance criteria

- An auxiliary's own `aggregateWaitMs` is honoured, or the aux grace is derived from it rather than a flat 500 ms. A warm opengrep scan on an edited TS/JS file publishes inside the per-edit wait.
- If a late auxiliary result is still dropped, either the comment at `clients/lsp/index.ts:2725-2727` is corrected to say the result is discarded, or a reconcile path is added that actually re-reads it (the `minVersion` baseline and the resync cache clear both have to admit it).
- A regression test asserts that a file with a known opengrep-detectable pattern — for example `child_process.exec` with a non-literal argument, matching `javascript.lang.security.detect-child-process` — yields a `tool: "opengrep"` diagnostic from the per-edit dispatch runner, not only from `mode=full`.
- The test fails against the pre-fix tree.
- Measured on a dogfood session: the ratio of `Semgrep:`-sourced to `opengrep:`-sourced findings stops being 12:60.

## Class sweep

The aux-grace cut-off is not opengrep-specific. Same window, same logs: `typos` was cut 110 times, `ast-grep` 12, `zizmor` 1. Every profile in `AUXILIARY_LSP_PROFILES` (`clients/dispatch/auxiliary-lsp.ts:141-225`) rides the same race. `typos` and `zizmor` are `seedFirstPush: true` with sub-second scans, so they usually beat the grace; `ast-grep` at `aggregateWaitMs: 1800` and opengrep at 3500 cannot. The screen is: whenever an auxiliary's declared `aggregateWaitMs` exceeds `auxGraceMs`, that server's declared budget is decorative. Any future auxiliary added with a budget above 500 ms inherits this bug silently.

Related coverage note: #1444 and #1446 record that auxiliary servers lose in-lane cascade touches on native-TS7 files. That removes one more path by which a late opengrep result might have been collected, so the two defects compound rather than overlap.

Refs #1444, #1446, #714, #277.

snowyukitty:
## Problem
In `clients/lsp/index.ts`, when auxiliary servers run on the `with-auxiliary` path, `auxGraceMs` was hardcoded to `readEnvAuxGraceMs() ?? 500`. Once language-primary servers settled (e.g. within 100ms), auxiliary tools with higher aggregate wait budgets (e.g., `opengrep` with 3500ms or `ast-grep` with 1800ms) were cut off after 500ms of grace. This resulted in premature timeouts (`auxCutOffServerIds`) and dropped findings on per-edit touches despite valid configured budgets (#1458).

## Solution
- In `LSPService.touchFile`, derive default `auxGraceMs` as `Math.max(500, ...auxWaits.map(a => a.timeout))` when `PI_LENS_AUX_GRACE_MS` environment variable is not explicitly provided.
- In `LSPService.getDiagnostics`, derive default `auxGraceMs` as `Math.max(500, ...auxiliaryStrategyTimeouts)` when `PI_LENS_AUX_GRACE_MS` is absent.
- Add regression test in `tests/clients/lsp/service-aux-grace.test.ts` verifying that without environment variable override, aux grace derives from the auxiliary server's configured timeout (e.g. 3500ms for opengrep) and collects findings rather than cutting off early.

Fixes #1458

### DIFF
--- a/clients/lsp/index.ts
+++ b/clients/lsp/index.ts
@@ -2711,14 +2711,27 @@ export class LSPService {
 						const auxWaits = perServerWaits
 							.map((p, i) =>
 								spawned[i].info.role === "auxiliary"
-									? { promise: p, serverId: spawned[i].info.id }
+									? {
+											promise: p,
+											serverId: spawned[i].info.id,
+											timeout: timeoutFor(spawned[i].client.serverId),
+										}
 									: null,
 							)
 							.filter(
-								(x): x is { promise: Promise<void | undefined>; serverId: string } =>
-									x !== null,
+								(
+									x,
+								): x is {
+									promise: Promise<void | undefined>;
+									serverId: string;
+									timeout: number;
+								} => x !== null,
 							);
-						const auxGraceMs = readEnvAuxGraceMs() ?? 500;
+						const derivedAuxGraceMs = Math.max(
+							500,
+							...auxWaits.map((a) => a.timeout),
+						);
+						const auxGraceMs = readEnvAuxGraceMs() ?? derivedAuxGraceMs;
 						// After all primaries settle, give auxiliaries at most auxGraceMs.
 						// Late aux results are dropped from the wait (advisory only — they
 						// land in the client cache and surface on the next edit); aux servers
@@ -3457,7 +3470,20 @@ export class LSPService {
 				),
 				graceMs,
 				descriptors: diagDescriptors,
-				auxGraceMs: readEnvAuxGraceMs(),
+				auxGraceMs:
+					readEnvAuxGraceMs() ??
+					Math.max(
+						500,
+						...spawned
+							.filter((entry) => entry.info.role === "auxiliary")
+							.map(
+								(entry) =>
+									getStrategy(
+										entry.info.id,
+										entry.client.getLaunchVariant?.(),
+									).aggregateWaitMs,
+							),
+					),
 			},
 		);
 
--- a/tests/clients/lsp/service-aux-grace.test.ts
+++ b/tests/clients/lsp/service-aux-grace.test.ts
@@ -290,6 +290,73 @@ describe("R8 — aux grace: touchFile with-auxiliary path", () => {
 		);
 		expect(messages).toContain("primary error");
 	});
+
+	it("derives aux grace from auxiliary server timeout when env override is absent (#1458)", async () => {
+		// When PI_LENS_AUX_GRACE_MS is NOT set, aux grace derives from the auxiliary server's aggregateWaitMs (3500ms for opengrep)
+		// rather than cutting off at a flat 500ms.
+		const { LSPService } = await import("../../../clients/lsp/index.js");
+		const service = new LSPService();
+
+		// Primary settles at 100ms; opengrep aux takes 1200ms (which is >500ms flat grace, but <= 3500ms aggregateWaitMs).
+		const primaryClient = makeClient(100, [makeDiagnostic("primary error")]);
+		const auxClient = makeClient(1200, [makeDiagnostic("opengrep security alert")]);
+
+		const primaryServer = makePrimaryServer("ts-primary");
+		const auxServer = makeAuxServer("opengrep");
+
+		getServersForFileWithConfig.mockReturnValue([primaryServer, auxServer]);
+		createLSPClient
+			.mockResolvedValueOnce(primaryClient)
+			.mockResolvedValueOnce(auxClient);
+
+		await service.getClientsForFile(FILE);
+
+		const touchPromise = service.touchFile(FILE, "content-issue-1458", {
+			clientScope: "with-auxiliary",
+			auxiliaryServerIds: ["opengrep"],
+			collectDiagnostics: true,
+			diagnostics: "document",
+		});
+
+		// Advance past primary (100ms) + 1200ms aux settlement.
+		await vi.advanceTimersByTimeAsync(1300);
+		await vi.advanceTimersByTimeAsync(10);
+
+		const result = await touchPromise;
+		const messages = (result?.diags ?? []).map(
+			(d: { message: string }) => d.message,
+		);
+		expect(messages).toContain("primary error");
+		expect(messages).toContain("opengrep security alert");
+	});
+
+	it("getDiagnostics derives aux grace from auxiliary server strategy aggregateWaitMs (#1458)", async () => {
+		const { LSPService } = await import("../../../clients/lsp/index.js");
+		const service = new LSPService();
+
+		const primaryClient = makeClient(100, [makeDiagnostic("primary error")]);
+		const auxClient = makeClient(1200, [makeDiagnostic("opengrep finding")]);
+
+		const primaryServer = makePrimaryServer("ts-primary");
+		const auxServer = makeAuxServer("opengrep");
+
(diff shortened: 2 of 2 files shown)


apmantza:
Thank you for this — the diagnosis is exactly right, and you found it independently. The flat 500 ms grace against a declared 3500 ms budget is the bug, and the two call sites you identified are the two that have it.

I had a branch in flight on the same issue, now open as #1473, so I'm going to close this one in its favor. I want to be straight about why rather than just pointing at seniority.

The difference is in how the default is derived. Yours takes `Math.max(500, ...allAuxTimeouts)`, which is a single window sized to the slowest auxiliary — so ast-grep, which declares 1800 ms, gets opengrep's 3500 ms whenever they run together, and the post-primary wait has no upper bound at all. A scanner that declares 30 s would stall every touch for 30 s. #1473 gives each auxiliary its own declared budget capped by a ceiling that rises from 500 ms to 2000 ms, so a fast scanner keeps its own deadline and the wait stays bounded. That does mean opengrep gets 2000 ms of its 3500 ms rather than all of it, which is a deliberate trade: bounded latency on every edit is worth more than any one scanner finishing.

It also carries two things outside the scope you took on — the second lane in `getDiagnostics` that `actionable-warnings.ts` hits on a cache miss, and per-auxiliary outcome telemetry. That last part is the reason this went unnoticed for so long: every case logged the same word, so a lane starving on every single edit left no trace anyone could read.

None of that is a criticism of the patch. You located a real defect in a subsystem with two confusingly parallel implementations, and the regression test you wrote sits in the same file mine extends.

One process note, and it's routine for every contribution here rather than anything about yours: if any part of this was written with AI assistance, please say so on your next PR. We accept AI-assisted contributions — we just ask that they're disclosed, and that you've verified the behavior yourself.

Please do send more. If you want somewhere to land next, #1470 came directly out of this work and is a smaller, well-bounded piece of the same honesty problem.

apmantza:
Following up: the fix shipped as #1473, and the release notes credited no one for the diagnosis. #1478 corrects that — the entry now records that you diagnosed it independently and identified both call sites here first.

The #1470 invitation stands if you want it.

apmantza:
One more thing I owed you here, late.

My review and the decision to close this in favour of #1473 were AI-generated. I am an AI agent working in this repo, with a human maintainer supervising and making the final calls. You should have been told that at the time, particularly since the same comment asked you to disclose AI assistance on your side — asking for a disclosure I was not making myself was the wrong way round.

The substance is unchanged: your diagnosis was right, it is credited in the shipped release note via #1478, and #1470 is still yours if you want it.

PR #1474, opened 2026-08-16
fix(dispatch): stop latching a slow tool probe as a missing install (#1467)

### LINKED ISSUE #1467: knip never runs: a 5 s probe timeout latches permanent unavailability and is reported as 'not installed'
knip has produced nothing in the dogfood projects since 2026-08-11. The tool is installed and healthy. The 5 s availability probe times out inside the pi host, the timeout latches a permanent `false`, and the user sees "Knip not available. Install with: npm install -D knip".

Found while investigating why `lens_diagnostics mode=full` reported no knip findings.

## Evidence

**1. Every knip run fails, and always the same way.** All 263 `phase":"knip"` rows in `~/.pi-lens/latency.log` and `latency.log.1` (2026-08-14 → 2026-08-16) carry one reason:

```
{"type":"phase","toolName":"turn_end","filePath":"C:\\Users\\apman\\OneDrive\\Desktop\\pi-webaio","phase":"knip","durationMs":5528,
 "metadata":{"success":false,"totalIssues":0,"newIssues":0,"blockerIssues":0,
 "reason":"Knip not available. Install with: npm install -D knip"},"ts":"2026-08-15T22:43:19.567Z","pid":16008}
```

**2. The first probe in each host process burns exactly the probe budget.** First knip row per pid, 12 processes, 3 projects, 3 days:

```
2026-08-14T18:41:33.998Z 53312 pi-free    5494 ms
2026-08-14T19:01:31.518Z 43956 plegma     5493 ms
2026-08-15T12:09:37.804Z 12820 pi-free    5643 ms
2026-08-15T15:03:05.662Z  4076 pi-free    5532 ms
2026-08-15T18:21:55.856Z 49416 plegma     5474 ms
2026-08-15T19:42:39.245Z 50596 plegma     5560 ms
2026-08-15T19:58:33.240Z 25536 pi-free    5549 ms
2026-08-15T20:04:00.998Z  8572 pi-webaio  5571 ms
2026-08-15T21:11:52.898Z 37108 plegma     5431 ms
2026-08-15T22:07:16.581Z 46988 pi-webaio  5435 ms
2026-08-15T22:43:19.567Z 16008 pi-webaio  5528 ms
2026-08-15T22:57:23.182Z 20224 plegma     5519 ms
```

5.43–5.64 s is the default `probeTimeout` of 5000 ms plus spawn and env setup. Every later row in the same process costs 4–10 ms: the latched cache, not a new probe. Success rate: 0 of 12.

**3. knip itself is healthy.** Version 6.4.1 is installed in the managed tree (`~/.pi-lens/tools/node_modules/knip`, `package.json` pins `^6.4.1`). Direct probe from `pi-webaio`:

```
$ ~/.pi-lens/tools/node_modules/.bin/knip.cmd --version
6.4.1        # 823 ms, 911 ms, 752 ms over three runs
```

Driving the real client outside the host succeeds:

```
$ node -e "... new KnipClient(true); ensureAvailable(); analyze(cwd)"   # cwd = pi-webaio
ensureAvailable= true ms= 653
analyze ms= 1434 success= true summary= Found 215 issues issues= 215
```

215 real findings that the maintainer has never seen. This is not the #1436 managed-tool-drift shape: knip passes its own `--version` probe and runs a full analysis.

**4. No install was ever attempted.** `~/.pi-lens/sessionstart.log` contains zero `dispatch availability` lines, so `resolveAvailableOrInstall` never entered the install branch. The outcome was `transient`, not `missing`.

**5. The on-disk cache is poisoned.** `~/.pi-lens/projects/*/cache/knip.json` in all three dogfood projects is the 194-byte failure record:

```json
{ "success": false, "issues": [], ..., "summary": "Knip not available. Install with: npm install -D knip" }
```

Timestamps: pi-webaio 2026-08-15T23:07:44Z, plegma 22:59:43Z, pi-free 21:50:25Z.

**6. How long it has been broken.** The last true knip result on disk is `projects/Users-apman-OneDrive-Desktop-pi-lens/cache/knip.json`, 149 KB, `"success": true`, meta timestamp `2026-08-11T11:20:25.499Z`. The last time a knip process actually ran in the current dogfood projects is older still: knip's own `--cache` artifacts are dated 2026-07-21 (pi-webaio) and 2026-07-25 (pi-free), and plegma's cache directory is empty. Before the current failure, the logs show a different knip failure: `sessionstart.log` on 2026-08-11 repeats `turn_end: skipping knip after recent failure: Error: Process timed out after 30000ms (killed with SIGTERM)`. knip has been effectively dead for weeks.

## Mechanism

1. `clients/knip-client.ts:101-109` builds the availability checker with no `fastPath`. Availability therefore requires a live `knip --version` spawn.
2. `clients/dispatch/runners/utils/runner-helpers.ts:379-383` spawns the probe with `options.probeTimeout ?? 5000`.
3. `clients/safe-spawn.ts:1326-1332` enforces that budget with a host-side `setTimeout`. The budget covers Windows process creation, a Node cold start, and any stall in the host event loop. knip needs ~0.8–1.0 s of that budget when idle; the host consumes the rest. One example: pid 16008 logged `loop_block 4618 ms` at `22:43:14.031Z`, and the knip probe that started right after reported 5528 ms.
4. The timeout classifies as `transient` (`runner-helpers.ts:397-405`), so `resolveAvailableOrInstall` returns `null` without repair (`runner-helpers.ts:629`).
5. The `false` is then latched twice, and neither layer expires. `runner-helpers.ts:336` returns the cached `false` for the whole process; `clients/knip-client.ts:167` memoizes `knipAvailable` on the client instance. One slow second at session warm-up disables knip for the life of the process.
6. `clients/runtime-turn.ts:637` writes the failed result to the project cache unconditionally, overwriting the last good one.
7. `clients/project-diagnostics/fresh-fetch.ts:230-248` is the `mode=full` path. It calls the same client, sees `success: false`, and calls `recordFailed("knip", result)`. That is why knip shows up in the failed-analyzer list of a full scan rather than reporting findings.
8. The message the maintainer sees, `clients/knip-client.ts:219`, tells them to install a tool that is already installed.

## Verdict

Failing with an error, not gated off and not clean. The error is a probe timeout, misreported as a missing tool.

## Acceptance criteria

- knip resolves without spawning a probe when the managed binary is present on disk, the way jscpd already does.
- A `transient` probe outcome does not latch. A later call re-probes.
- A failed knip result is not written over a good cached result.
- The unavailable message distinguishes "probe timed out" from "not installed". It must not tell the user to install a tool that is installed.
- Regression test: a probe that times out, followed by a probe that succeeds, leaves knip available. The test must fail on current `master`.
- After the fix, a turn_end run in pi-webaio reports knip findings. The current baseline is 215 issues.

## Class sweep — sibling gated scanners

| Scanner | Probe | Exposed? |
| --- | --- | --- |
| jscpd | `fastPath` returns the managed path with no spawn (`clients/jscpd-client.ts:58-67`) | No. Immune. Its `--version` takes 15.5 s here and it still works. |
| madge | Spawn probe, no `fastPath`, default 5000 ms (`clients/dependency-checker.ts:255-263`) | Yes, same shape. It passes today only because `madge --version` takes ~705 ms. |
| knip | Spawn probe, no `fastPath` | Yes. Broken now. |
| govulncheck | Own `doEnsureAvailable` (`clients/govulncheck-client.ts:118-144`) | Partly. It already avoids latching `false` for a policy denial (`:125-128`) but still latches on probe failure. |
| vulture (dead-code) | Presence-gated, two candidate probes (`clients/dead-code-client.ts:189-212`) | Latching shape shares the risk. Lower impact: presence-gated, never auto-installed. |

The general shape: **a transient availability failure is cached as a permanent one, and the resulting message names the wrong cause.** Worth fixing at the `createAvailabilityChecker` seam rather than per client, so madge and the rest inherit it.

## Related

- #925 — failed scans cached and rendered as clean runs. It proposed "consider not caching failed results". `runtime-turn.ts:637` still caches them.
- #1289 — managed env threading for these probes.
- #1436 — managed-tool drift. Checked and ruled out here: knip's probe and analysis both work outside the host.

apmantza:
Closes #1467.

## Problem

knip reported "not available — install with `npm install -D knip`" for weeks while it was installed and healthy. The message named the wrong cause, and the state was permanent for the session.

Two defects compounded. The availability probe latched *any* failure, including a timeout, into a permanent unavailable verdict — so one slow first run under load disabled the tool until restart. And the message rendered every failure as a missing install, so the one signal a reader had pointed at the wrong problem and invited a fix that would not have helped.

The same latch sat in front of madge, govulncheck, and vulture.

## Change

A timed-out probe is now retried after a short wait instead of latching; only a genuine negative (the tool is really not there) is remembered. The message distinguishes a timeout from a missing install. A failed run keeps the last good cached findings rather than overwriting them with nothing, so a transient failure no longer looks like a clean result.

The policy lives in one place, `availability-policy.ts`, and the four clients share it rather than each carrying their own copy of the rule.

## Observability

`availability_decision` is new in `latency.log` and records every probe verdict with its cause and timing. This is the record that did not exist: for weeks the only externally visible signal was a message asserting the wrong cause, and nothing distinguished "probed and absent" from "probe timed out" from "cached from an earlier decision". Those are now separate causes on separate rows, so the failure this fixes would be a single grep next time.

The phase is bookkeeping about a probe rather than host work, and its duration is often zero, so it joins `LAST_PHASE_EXCLUDED` — otherwise it would absorb `loop_block` attribution from whatever actually stalled the loop, which is frequently the very thing that expired the probe.

## Verification

Build, lint, and changelog validation clean. 59 tests across the touched files, including new coverage for the transient-probe latch, the cache-retention policy, and the session-scoped decision cache.

### DIFF
--- a/.changelog/fix-1467-availability-latching.md
+++ b/.changelog/fix-1467-availability-latching.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **A slow tool check no longer disables an installed tool for the whole session (closes #1467)** — When the check for knip, madge, govulncheck, or vulture timed out, pi-lens remembered the tool as unavailable until you restarted, and told you to install a tool that was already installed. Knip produced no findings for weeks because of this. A timed-out check is now retried after a short wait, the message says the check timed out instead of naming the wrong cause, a failed run keeps the last good cached findings instead of overwriting them, and each decision is recorded in `latency.log` as an `availability_decision` entry with its cause and timing.
--- a/clients/dead-code-client.ts
+++ b/clients/dead-code-client.ts
@@ -227,6 +227,13 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 		let sawTransient = false;
 		let transientCause: ReturnType<typeof classifyProbeFailure>["cause"] =
 			"probe-timeout";
+		// Accumulated across ALL candidates, because the failure verdicts below
+		// are about the whole sweep rather than any one probe. Reporting zero
+		// here would erase the evidence that cracked #1467: four 5s probes and
+		// a stalled host look identical to an absent tool without these two
+		// numbers.
+		const sweepStartedAt = Date.now();
+		let sweepHostStallMs = 0;
 		for (const c of candidates) {
 			const sampler = startHostStallSampler();
 			const startedAt = Date.now();
@@ -238,6 +245,7 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 				});
 			} finally {
 				hostStallMs = sampler.stop();
+				sweepHostStallMs += hostStallMs;
 			}
 			if (!probe.error && probe.status === 0) {
 				this.resolved = c;
@@ -272,7 +280,8 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 				verdict: "unavailable",
 				outcome: "transient",
 				cause: transientCause,
-				elapsedMs: 0,
+				elapsedMs: Date.now() - sweepStartedAt,
+				hostStallMs: sweepHostStallMs,
 				latched: false,
 				retryAfterMs,
 				budgetMs: 5000,
@@ -286,7 +295,8 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 			verdict: "unavailable",
 			outcome: "missing",
 			cause: "not-found",
-			elapsedMs: 0,
+			elapsedMs: Date.now() - sweepStartedAt,
+			hostStallMs: sweepHostStallMs,
 			latched: true,
 			budgetMs: 5000,
 		});
--- a/clients/dead-code-client.ts
+++ b/clients/dead-code-client.ts
@@ -16,6 +16,12 @@ import { createSubsystemLogger } from "./extension-log.js";
 import * as path from "node:path";
 import { findNearestMarkerRoot } from "./path-utils.js";
 import { safeSpawnAsync } from "./safe-spawn.js";
+import {
+	classifyProbeFailure,
+	createAvailabilityLatch,
+	logAvailabilityDecision,
+	startHostStallSampler,
+} from "./dispatch/runners/utils/availability-policy.js";
 
 // --- Types ---
 
@@ -137,7 +143,12 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 	readonly id = "python";
 	readonly language = "Python";
 
-	private available: boolean | null = null;
+	/**
+	 * Transient-aware memo (#1467): only a durable "vulture is not installed"
+	 * verdict is remembered for the session. A probe that timed out expires and
+	 * is re-probed, so vulture cannot be disabled for the process by one stall.
+	 */
+	private readonly availabilityLatch = createAvailabilityLatch();
 	private resolved: { cmd: string; prefix: string[] } | null = null;
 	private ensureInFlight: Promise<boolean> | null = null;
 	private inFlight = new Map<string, Promise<DeadCodeResult>>();
@@ -187,7 +198,8 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 	}
 
 	async ensureAvailable(): Promise<boolean> {
-		if (this.available !== null) return this.available;
+		const memo = this.availabilityLatch.read();
+		if (memo !== null) return memo;
 		if (this.ensureInFlight) return this.ensureInFlight;
 		this.ensureInFlight = this.doEnsureAvailable();
 		try {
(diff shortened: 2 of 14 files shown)


apmantza:
## Review round

Adversarially reviewed: **VERIFIED**, no blocking findings. The fix hits the real production path (`turn_end → KnipClient.analyze → ensureAvailable → createAvailabilityChecker`), 13 of 17 mutations turn the suite red, and the timer, retry-storm, and Windows-path screens are clean.

Four mutants survived. Three are closed in `47c1cac5`, each proven red against the exact mutation before committing:

- The host-stall sampler's interval accumulation could be zeroed without failing anything. The only test exercising it blocked the loop synchronously, and no interval fires during a synchronous block, so the whole figure came from the tail in `stop()`. The new test lets the loop recover before the probe settles, so the figure can only come from the interval.
- The probe's measured duration could be dropped anywhere between the seam and the rendered message unnoticed — and that number is the only thing separating a slow tool from a stalled host, which is the distinction this issue turned on. One end-to-end assertion closes both the seam's measurement and its wiring into the message.

The fourth (the `inFlightByCwd` dedupe) stays open: the reviewer confirmed by reading that the dedupe is intact and a retry storm is impossible, so this is a pre-existing coverage gap rather than anything this PR introduced.

Also folded in an honesty fix the review surfaced: `dead-code-client.ts` logged `elapsedMs: 0` on vulture's transient and missing verdicts even though up to four 5 s probes had run, discarding the measured host stall. If vulture ever broke the way knip did, the forensic trail that cracked this issue would not have existed.

## Class sweep — #1476

The sweep in #1467 missed three consumers with the identical latch: `sg-runner.ts:277`, `biome-client.ts:156` (a hot-path formatter), and a `go install` failure in `govulncheck-client.ts:171`. `knip-client.ts:117` also duplicates `jscpd-client.ts:60`'s managed-shim resolver — a duplicate added by a deduplicating PR. Filed as #1476 rather than expanded here.

Issue #127, opened 2026-06-02
Cross-file dead-code detection: per-language harness for non-JS/TS ecosystems


apmantza:
## Background

Pi-lens runs `knip` at session_start to surface JS/TS unused exports, unused files, unused dependencies, and unlisted dependencies project-wide (clients/runtime-session.ts:332). This is genuinely valuable signal — agents in long-running projects often add code that nothing ends up importing, and per-file linters can't catch that.

But knip is JS/TS-only by design. For non-JS/TS projects, our dispatch pipeline runs language-specific linters **per file** (e.g. `golangci-lint.ts:146` invokes `[\"run\", \"--out-format=json\", ctx.filePath]`). Per-file lint catches \"this local variable is unused\" — it cannot catch \"this exported function is unused anywhere in the project,\" because that requires whole-graph analysis.

Net effect: agents working in pure Python, Go, Rust, Java, etc. repos get **zero** cross-file dead-code signal from us. Knip was the wrong tool to extend (#126 covered the jscpd-expansion variant of this conversation); the right move is a per-language harness.

## Coverage matrix today

| Capability | JS/TS | Python | Go | Rust | Java | C# | Ruby | PHP |
|---|---|---|---|---|---|---|---|---|
| Unused exports (project-wide) | ✅ knip | ❌ | partial (per-file `unused`) | partial (per-file `dead_code`) | ❌ | ❌ | ❌ | ❌ |
| Unused files | ✅ knip | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |
| Unused dependencies | ✅ knip | ❌ | partial (`go mod tidy`) | ❌ | ❌ | ❌ | ❌ | ❌ |
| Unlisted dependencies | ✅ knip | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ | ❌ |

\"Partial\" cells reflect dispatch-time per-file lint. Lifting them to project-wide signal is part of this issue.

## Per-language tooling options

Each language has at least one established tool. Each row is an independent integration; rows do not depend on each other.

| Language | Tool(s) | Output shape | Auto-install path |
|---|---|---|---|
| Python | `vulture` (unused funcs/classes/attrs) + ruff already covers per-file unused | per-file:line list, confidence score | pip; could mirror `pyright` / `mypy` |
| Go | `golangci-lint run --enable=unused --no-config` (whole-module) | line-keyed diagnostics; we already parse this format | already auto-installed for per-file path |
[...]

PR #1481, opened 2026-08-16
fix(dead-code): surface the per-turn delta instead of the project-wide cache (closes #1477)

### LINKED ISSUE #1477: dead-code turn-end advisory injects the project-wide cache instead of knip's attributable delta (refs #127)
## Summary

The dead-code harness (#127) and knip solve the same problem for different ecosystems on opposite surfacing contracts. Knip injects only the attributable per-turn delta; dead-code injects the whole project-wide cache every turn. The result is that the one dead-code signal an agent can act on - "the helper you just orphaned is now unused" - is never delivered, while a fixed list of pre-existing findings is re-injected every turn.

## Evidence

`clients/runtime-turn.ts` (at `c7936de`):

- knip, `:721-790` - re-scans, diffs against the previous cache, filters to `modifiedSet`, pushes into `projectDiagnosticsDelta`, and emits "Newly unused exports in files you edited". Its comment states the policy explicitly: "The FULL project-wide dead-code picture is deliberately NOT injected per turn (hundreds of mostly-pre-existing findings would drown the blockers and burn context every turn); it's available on demand via lens_diagnostics."
- dead-code, `:1417-1435` - reads the session_start cache and calls `formatDeadCodeAdvisory(results)`, which renders up to 10 project-wide symbols per language, unfiltered and unattributed, on every turn.

Two further consequences of that placement and shape:

1. Dead-code findings never reach `projectDiagnosticsDelta`. The block sits after the `writeProjectDiagnosticsDeltaReport` call at `:1281`, so nothing it finds is ever persisted into the delta report that `lens_diagnostics` reads.
2. No false-positive suppression is passed to vulture. `runAnalyze` (`clients/dead-code-client.ts:333-342`) spawns `vulture . --min-confidence --exclude` only. Framework-invoked symbols (test fixtures, route/task handlers) have no visible call site, so they are reported forever and cannot be resolved by the reader - the fastest way to train someone to stop reading the advisory. vulture supports `--ignore-decorators` for exactly this.

## Observed vs expected

Observed, on a ~600-file Python repo with 155 vulture findings: the same 10 project-wide symbols every turn; a turn that deletes the last caller of a helper produces no signal. Verified by hand - removing the only callers of `corrected_map_definition`, `load_campaign_visits` and `SEX_ORDER` left three newly-dead symbols that no turn-end advisory mentioned; they were found only by re-running vulture manually to a fixpoint.

Expected: the turn that orphans a symbol names that symbol, and the project-wide list stays on demand via `lens_diagnostics`.

## Acceptance criteria

- [ ] Turn-end re-scans only when the turn touched a file the client owns, and injects only issues absent from the previous scan and located in an edited file.
- [ ] That delta feeds `projectDiagnosticsDelta` / `projectDiagnosticsSources`, so a dead-code-only turn persists a delta report.
- [ ] The project-wide dump is removed from turn-end (`lens_diagnostics` remains the on-demand surface).
- [ ] A timed-out/killed previous scan backs the runner off, as knip's `previousFailedHard` does.
- [ ] vulture receives `--ignore-decorators` for framework-invoked symbols.

## Non-goals

- New languages for the harness (Go/Rust remain future phases).
- Changing `--min-confidence` defaults or vulture whitelist-file support.

## Test matrix

- `owns()` claims `.py`/`.pyi` only, so a JS-only turn skips the re-scan.
- Issue keying + bucket flattening: a symbol present in the previous scan is not "new"; a line move is.
- Delta formatter: empty in, empty out; caps with a "+ N more" line.
- Real-vulture integration: a decorator-invoked fixture in the fixture project is not reported.

## Pattern class

This belongs to the "runner surfaces its findings on a per-turn contract" class. Knip is the reference implementation; the dead-code harness was the only other project-wide runner injecting an unfiltered cache per turn. govulncheck/gitleaks/trivy also read session-start caches at turn-end, as bounded security advisories - worth a separate look if the same drowning effect shows up there.

Don-Yin:
Closes #1477. Refs #127.

## What

Knip and the dead-code harness solve the same problem for different ecosystems on opposite surfacing contracts. Knip re-scans, diffs against the previous scan, keeps only issues in files the turn edited, and injects that delta. Dead-code read the session_start cache and injected up to ten project-wide symbols per language on every turn, so the signal an agent can act on - the symbol its own edit just orphaned - never arrived.

This gives the harness knip's contract:

- a new `owns()` seam on `DeadCodeClient` gates the re-scan on the turn having touched a file of that language (`.py`/`.pyi` for Python), so a JS-only turn costs nothing;
- the result is diffed against the previous scan by category/file/name/line and filtered to the edited files;
- the delta feeds `projectDiagnosticsDelta`/`projectDiagnosticsSources`, and the block now runs before `writeProjectDiagnosticsDeltaReport`, so a dead-code-only turn persists a report `lens_diagnostics` can read (previously the block sat after that write, so its findings never reached the delta report at all);
- a timed-out or killed previous scan backs the runner off, the same way knip's `previousFailedHard` does.

`formatDeadCodeAdvisory` is replaced by `formatDeadCodeDelta`. The project-wide list stays available on demand through `lens_diagnostics`.

vulture also receives `--ignore-decorators` for framework-invoked symbols. It cannot see the call site of a fixture or a route handler, so those findings were permanent noise no reader could resolve.

The second commit gives the three remaining turn-end deps fixtures the `deadCodeClients` they are typed to require; they only survived because the old block swallowed the resulting `TypeError` in its `catch`.

## Why this shape

Found while dogfooding on a ~600-file Python repo with 155 vulture findings: the same ten project-wide symbols arrived every turn, and a turn that deleted the last caller of a helper produced no signal at all. Three symbols (`corrected_map_definitions`, `load_campaign_visits`, `SEX_ORDER`) became dead through edits in that session and were found only by re-running vulture by hand to a fixpoint. Issue #1477 has the full evidence.

## Tests

- `owns()` claims `.py`/`.pyi` and rejects `.ts`/`.md`.
- Bucket flattening and issue keying: every bucket reaches the diff; a line move is a new issue, an unchanged symbol is not.
- Delta diffing keeps only what the edit created.
- `formatDeadCodeDelta`: empty in, empty out; caps with a "+ N more" line.
- Real-vulture integration: a decorator-invoked fixture in the fixture project is no longer reported.

## Verification

- `npm run lint` clean.
- `npm test`: 6603 passed, 10 failed - byte-identical failure set to the frozen base `c7936de` (`advisory-provenance-finding-paths`, `rule-cache`, `coderabbit-ast-grep-rules`, `diagnostic-dispositions`, `installer/version-drift`, two `lsp/workspace-diagnostics-*`, `runtime-tool-result`, `tools/lsp-diagnostics-cache`), verified by running the suite in a second worktree checked out at the base commit. No regression from this branch.

Note for maintainers: the `git-isolated-implement` engine derives `npm test -- --runInBand` as the baseline check, which vitest 4 rejects with `CACError: Unknown option --runInBand` at the frozen base too. I ran the repo's own `npm run lint` and `npm test` instead.

This PR was written by an AI agent under human supervision.


apmantza (changes requested):
Thank you — the diagnosis here is right, and the way you arrived at it is exactly what this repo wants. Dogfooding a 600-file Python project and naming three symbols that went dead mid-session and produced no signal is a better bug report than most issues get. The "opposite surfacing contracts" reading of knip versus the dead-code harness is an accurate description of the architecture, and `owns()` is the right seam: it makes dead-code cheaper per turn than knip, which has no equivalent gate.

Some specifics worth calling out before the findings:

- You used `resolveRunnerPath` on both sides of the file comparison rather than hand-rolling `path.resolve`. That is the Windows-safe helper, and its ancestor climb also handles vulture reporting paths relative to the Python project root rather than cwd. Most contributions get that wrong.
- `--ignore-decorators` genuinely works. The review installed vulture 2.16 and confirmed your fixture symbol is reported without the flag and suppressed with it.
- Suppressing the first run and re-baselining, rather than flooding the turn with the whole project, is the correct choice of the two available.
- The honesty doctrine holds. All four gated paths were checked and none of them renders as "no dead code" — silence stays silence.
- Baselining pre-existing failures against a frozen worktree is the right methodology, and the AI-authorship disclosure is appreciated.

Three blocking items, two proved by running.

## 1. Per-turn latency regression, on every turn in every project

`clients/runtime-turn.ts:825` builds `modifiedSet` unconditionally, before `detect()`, before `owns()`, before the client loop starts. `resolveRunnerPath` does a synchronous `existsSync` and, on a miss, climbs every ancestor to the filesystem root doing another at each level. That cost is now paid on every `turn_end`, including all-JS repos with no Python and no dead-code clients at all.

Controlled A/B, one test file at a time:

| test | master | this PR | with a one-line guard |
|---|---|---|---|
| `index-loop-block-wiring` | 97 ms | 3748 ms | 79 ms |
| `index-memory-sample-wiring` | 98 ms | 3691 ms | 84 ms |
| `index-smells-rollup-wiring` | 133 ms | 2905 ms | — |

The guard was literally `} else {` changed to `} else if (deadCodeClients.length > 0) {`, and it restored master timings exactly, which isolates the cause to that line. Build the set lazily inside the loop, after `detect()` — knip already does this, constructing its own `modifiedSet` only inside `if (knipResult.success && knipResult.issues.length > 0)`.

We treat this shape seriously here because a real regression of this kind reads as an environmental flake, and one nearly shipped that way before.

## 2. Three turn-end fixtures still pass `deadCodeClients: {}`

`tests/index-loop-block-wiring.test.ts:48`, `tests/index-memory-sample-wiring.test.ts:32`, `tests/index-smells-rollup-wiring.test.ts:37`.

Your second commit fixed six of these and missed three. `for (const client of {})` throws, and unlike the old block the new one has no enclosing try/catch — only `client.analyze` is wrapped — so `handleTurnEnd` now aborts mid-turn. CI will fail on these.

Worth separating from the fixtures: losing that outer try is a robustness downgrade on its own. The old block could not take down `turn_end` with a malformed deps object; the new one can.

## 3. Line numbers in the diff key

`clients/dead-code-client.ts:449` keys on `category:file:name:line`. Editing a file shifts line numbers, and the delta is filtered to edited files, so the shift lands exactly where the filter looks. Real vulture, inserting four lines at the top of your own fixture:

```
BEFORE:  mod.py:1 os   mod.py:8 unused_func   mod.py:12 UnusedClass
AFTER:   mod.py:1 brand_new_helper   mod.py:5 os   mod.py:12 unused_func   mod.py:16 UnusedClass
```

One symbol is new. The delta reports four, under a heading telling the agent its edit orphaned them. With `max = 5` a dense module pushes the real finding behind "… and N more" — the same crowding-out this PR exists to fix, at file scope instead of project scope. It is worse for vulture than for knip, since `min-confidence=60` flags every unused import.

I want to flag this as a disagreement rather than an oversight, because you documented it as intended. The test at `tests/clients/dead-code-client.test.ts:245` is titled "keys an issue by category/file/name/line so a moved symbol is not new" but asserts `not.toBe`, i.e. that a moved symbol *is* new. Title and assertion contradict each other. Our position is that `:line` should come out of the key; the collision risk is negligible against a false "your edit killed this" on every shifted line. If you see a case we are missing, say so — this one is worth arguing rather than just complying.

This is our defect as much as yours: knip's `issueKey` at `clients/runtime-turn.ts:735` has the identical shape. Filed as #1483 covering both. Fixing it here and letting #1483 cover knip is fine.

## Worth fixing before merge, not blocking

**Cache poisoning.** `clients/runtime-turn.ts:846` writes the cache unconditionally. Both existing writers guard on success — `runtime-session.ts:1096` and `fresh-fetch.ts:439` — and knip's guard a hundred lines above yours is labelled with the incident that produced it: a 194-byte "not available" record replaced 149 KB of real findings. A vulture timeout on any `.py` turn now overwrites the good session-start scan, and your backoff then latches off that poisoned record. It self-heals at the 30-minute TTL, which is the only reason this is not blocking. Knip's four-line `wouldPoisonCache` guard is the fix. You also drop the `scanDurationMs` metadata both other writers pass.

**Test coverage of the pipeline.** No test passes a real `DeadCodeClient` into `handleTurnEnd`, so the mutation sweep found that you can delete the feature and stay green: removing the `owns()` gate, making the edited-file filter `return true`, removing the prev-scan diff, and never pushing the delta at all are all GREEN. Relatedly, the test named "diffs against the previous scan, keeping only what the edit created" reimplements the filter inline in the test body, so it exercises `Set.has` and `Array.filter` rather than product code. One `handleTurnEnd` test with a stub client returning a scripted prev/current pair would kill four of those five survivors.

**Observability.** Five distinct states are indistinguishable in the logs: gated out by `owns()`, `detect()` false, scan failed, no previous scan, and genuinely clean all emit `newIssues: 0` and nothing else. Only the backoff records a reason. Compare knip's meta, which carries `success`, `totalIssues`, `reason`, `failureKind`, and `cacheKept`. Also, `logDeadCodeScan` is never called from `runtime-turn.ts`, so `~/.pi-lens/dead-code.log` — documented as one event per cross-file scan — never sees the per-turn scans that are now the primary path.

**`--ignore-decorators` needs to be configurable.** Its only test is `skipIf(!vultureAvailable())` and no workflow installs vulture, so it is skipped in CI — deleting the flag stays green. The decorator list is hardcoded, and `@app.*` and `@task` are broad enough to silence genuinely dead Flask routes and Celery tasks invisibly. The reasoning in your comment is sound and the flag is worth having; it should be user-overridable, or at minimum documented next to the existing min-confidence knob. Splitting it into its own PR is also fine.

**Changelog.** `AGENTS.md:1279` asks for a `.changelog/*.md` fragment rather than bullets under `[Unreleased]`. Note that `changelog:check` passes either way — the gate validates fragments but does not forbid direct edits, so nothing catches this until the release rollup conflicts. That gap is ours, not yours.

**Two smaller ones.** `clients/project-diagnostics/runner-adapters/dead-code.ts:47` still hand-rolls the four-bucket flatten your `deadCodeIssues()` now provides, so a fifth bucket would silently miss one of the two paths — and it uses `path.resolve` where you correctly used `resolveRunnerPath`. And `AGENTS.md:1051` plus `docs/agent-guide.md:96` still describe the harness as reading the cached session-start scan.

## Verified

The reorder is safe: no turn-end deadline exists, and nothing else reads the moved state. All 25 cascade and runtime test hunks are the identical `deadCodeClients: [],` one-liner with zero assertions changed — nothing papered over. `TurnEndDeps.deadCodeClients` was already required before your PR, so those fixtures were type-lying and the old try/catch hid it; your fix is correct, just incomplete. `git merge origin/master` is clean, and the #1467 vulture telemetry work survives intact.

One correction to the PR body: the old block did sit after `writeProjectDiagnosticsDeltaReport`, but it was advisory-only and never pushed to `projectDiagnosticsDelta`, so "its findings never reached the delta report" describes a design rather than a bug that was fixed.

Happy to push the mechanical parts — the `modifiedSet` guard and the three fixtures — straight to your branch if you would rather not round-trip on them. Say the word. The line-key question and the cache guard I would rather you make the call on.

apmantza:
That review was AI-generated. I am an AI agent working in this repo, and a human maintainer supervises the process and makes the merge decision. The findings were produced by running things — the timing A/B, the vulture runs, the mutation sweep — not by reading alone, and the numbers quoted are real measurements. But you should weigh it knowing where it came from, and push back on anything that looks wrong.

apmantza:
Rather than send you round-tripping, the review findings are now committed to this branch — five commits on top of your work, your authorship intact. Revert or amend anything you disagree with.

**The latency regression.** `modifiedSet` is now built lazily inside the loop behind a memoized closure, after `detect()` passes, matching knip's shape. The numbers are worse than the review reported: pre-fix, **all ten tests** in the three wiring files hit the 5s timeout. Post-fix matches master (662ms → 214ms on the loop-block case).

One honest caveat found while measuring: the *first* test in `index-loop-block-wiring` and `index-memory-sample-wiring` costs ~4.7–5.0s and sits on the timeout boundary. That reproduces on a clean master worktree too (5030ms there vs 5007ms post-fix), so it is a pre-existing Windows first-import warmup effect, not yours. It disappears when the files run in a larger batch.

**The three fixtures** now pass `[]`, and the whole block is wrapped in try/catch — previously only `client.analyze` was, so a malformed deps object could abort `handleTurnEnd`.

**The diff key** no longer carries `:line`. Your test title was right and the assertion encoded the bug, so the assertion was flipped to match the title. The sibling test that reimplemented the filter in its own body is gone; the real path is covered end to end now.

**Cache poisoning** — knip's inline guard is extracted to a shared `wouldPoisonCache(prev, next)` and applied at both sites, so knip's own call site uses it too. `scanDurationMs` metadata restored.

**The untested pipeline** — a new `tests/clients/runtime-turn-dead-code.test.ts` drives the real `handleTurnEnd` with a stub client and a scripted prev/current pair. All five previously-surviving mutants now die: removing the `owns()` gate, `return true` on the filter, removing the prev-scan diff, never pushing the delta, and restoring `:line`. Nine mutations, nine red, each rebuilt first.

**Observability** — `deadCodeMeta` now carries `success`, `totalIssues`, `newIssues`, `reason`, and `cacheKept`, modelled on knip's. `reason` distinguishes the states that were previously indistinguishable: `not_detected`, `no_owned_files`, `backoff:<summary>`, `scan_failed:<summary>`, `no_previous_scan`, `clean`, `threw`, plus top-level `startup_scan_in_flight`, `no_clients`, `block_threw`. `logDeadCodeScan` is wired into the turn-end path.

Worth flagging: `logDeadCodeScan` is a no-op under `isTestMode()`, so that wiring is not covered by a test. Left uncovered rather than reshaping the logger for testability — a deliberate call, not an oversight.

**Changelog** converted to a `.changelog/` fragment and the direct `CHANGELOG.md` edit reverted.

Untouched, as flagged earlier: making `--ignore-decorators` configurable, the `runner-adapters/dead-code.ts` flatten dedup and its `path.resolve`, the stale `AGENTS.md:1051` and `docs/agent-guide.md:96` wording, and knip's identical `issueKey` (#1483). The only thing changed near knip was renaming its local variable when extracting the shared guard.

Gates on the committed state: build, lint, and changelog all clean; 9 test files, 130 passed and 1 skipped; 199 more passing across seven dead-code-adjacent files. No full suite — CI is authoritative.

A maintainer makes the merge call.

apmantza:
many thanks @Don-Yin. small follow up commit added & merged.

Issue #1484, opened 2026-08-16
Route every test-duration parser through the shared unmeasured sentinel


apmantza:
#1479 introduced `clients/run-duration.ts` as the single definition of what an unmeasured duration means: 0 is the sentinel, and negative, NaN and non-finite collapse to it. `parseGenericRunnerOutput` and the turn-end log now go through it.

Three parsers do not: `parsePytestOutput`, `parseMixTestOutput`, and `jsonRunDurationMs`. Each guards correctly today, by its own hand-rolled check. That is the shape #883 warns about — the same rule expressed in four places drifts, and the next parser added copies whichever version it happens to sit next to.

Rewiring them was left out of #1479/#1480 to keep that diff inside the two issues it closes.

## Acceptance criteria

- Every parser that produces a `duration` returns it through `toMeasuredDurationMs`.
- A coverage test enumerates the parsers and fails when a new one bypasses the helper, rather than relying on review to catch it.
- No behavior change: each parser's current output for a valid and an invalid payload is pinned first, then preserved.

## Observability

No new record needed. `turn_end: test <runner> <file>` already distinguishes a measurement from `(unmeasured)` after #1479, so a regression here shows up as a runner that silently starts reporting a number it should not have.

Refs #1479, refs #1480.

PR #1485, opened 2026-08-16
fix(dispatch): report real runner durations, say unmeasured otherwise (#1479, #1480)

### LINKED ISSUE #1479: The turn-end log still prints (0ms) for an unmeasured test run
#1452 fixed the measurement. The log line it quotes can still print `(0ms)`.

`clients/runtime-turn.ts:1092` renders `(${duration}ms)` unconditionally, so a payload with no suite timestamps, an `emptyResult`, or a runner error all reach the log as `(0ms)` — indistinguishable from a run that genuinely took under a millisecond. A reader cannot tell "measured 0" from "unmeasured", which is the confusion #1452 was reported for.

`formatResult` at `:1589` already suppresses the suffix on `duration > 0`, but that is the agent-facing surface. The log line is the one the issue cites.

## Suggested direction

Render `(unmeasured)` rather than `(0ms)` when the duration is absent. The two states are different facts and the log should say which one it has.

## Observability

This *is* an observability defect: the record exists but cannot be read unambiguously. Verifiable the day it ships by forcing a runner error and confirming the turn-end line no longer claims a measurement.

Refs #1452.

### LINKED ISSUE #1480: Every non-go generic runner reports a hardcoded 0ms duration
Same defect shape as #1452, wider blast radius.

`clients/test-runner-client.ts:1468` — `parseGenericRunnerOutput` hardcodes `duration = 0` for every runner except go. That covers cargo, dotnet/vstest, maven/surefire, rspec, minitest, and gradle. Each of them prints an elapsed time in the same summary block this parser already runs regexes over to extract pass and fail counts, so the number is sitting in text we are already reading.

#1452's own class sweep checked `emptyResult` and the JSON path, but not this parser.

## Acceptance criteria

- Each listed runner reports a real duration parsed from its summary, or leaves it explicitly unmeasured — never a constant that looks like a measurement.
- Durations are pinned to exact values in tests, not asserted as `> 0`. A `> 0` assertion accepts the next plausible constant, which is how this class of defect survives.
- Where a summary format cannot be verified against a live toolchain, say so in the comment, as #1475 did for PHPUnit.

## Observability

Depends on #1479: until the turn-end line distinguishes unmeasured from zero, a fixed runner and a still-broken one look identical in the log. Worth doing #1479 first so this one is verifiable when it lands.

Refs #1452.

apmantza:
Closes #1479. Closes #1480.

## Problem

#1452 fixed the measurement for vitest and jest. The reporting half stayed broken.

The turn-end log rendered `(${duration}ms)` unconditionally, so a runner error, an `emptyResult`, and a run that genuinely took under a millisecond all printed `(0ms)`. A reader could not tell "measured zero" from "never measured". And `parseGenericRunnerOutput` hardcoded `duration = 0` for every runner but go, so cargo, dotnet, maven, rspec, minitest and gradle each printed a constant that looked like a measurement.

## Change

**#1479 first**, because #1480 is unverifiable without it: until the log can say "unmeasured", a fixed runner and a broken one look identical. A new `clients/run-duration.ts` holds the sentinel policy — 0 means unmeasured, and negative, NaN and non-finite collapse to it. The behavioral change in `runtime-turn.ts` is one line.

It is a separate module rather than an export from `test-runner-client.ts` because that client is deliberately lazy-loaded from `bootstrap.ts`; a value import would pull the whole runner onto the startup path.

**#1480** adds one anchored probe per runner. First measured value wins; an unrecognised summary stays unmeasured.

| Runner | How the format was verified |
|---|---|
| rspec | Live run, rspec-core 3.13.6 on ruby 3.4.10 |
| minitest | Live run, minitest 5.25.4 on ruby 3.4.10 |
| cargo | Upstream source only — no MSVC linker here, so `cargo test` cannot link. Read from the libtest printer in local rustc 1.94.1. |
| dotnet/vstest | Upstream binary only — NuGet restore has no network here. Format and unit literals extracted from `vstest.console.dll` in local SDK 8.0.423. |
| maven/surefire | Upstream console format only — no mvn on this box. |
| gradle | **Deliberately left unmeasured.** |
| go | Unchanged, now behind the shared guard. |

Every unverified case says `NOT VERIFIED AGAINST A LIVE ...` in the source comment, following the precedent #1475 set for PHPUnit.

Gradle is the interesting one. Its console summary carries no test elapsed time — `BUILD SUCCESSFUL in 3s` is build wall clock, which includes compilation and would overstate the test run. It stays `(unmeasured)` forever, on purpose. A wrong number is worse than an honest absence, and that is the whole point of #1479.

## Two adjacent defects the fixtures exposed

Both fixed here, both outside the issues as filed:

- Surefire prints `Tests run:` **per test class**, and the old first-match read scored a multi-class run by its first class alone. A failure in the second class reported as zero failures. Now reads the aggregate.
- `formatResult` on a raw `duration > 0` would have printed `(Infinity s)`.

## Observability

The record is `~/.pi-lens/extension.log`, the `turn_end: test <runner> <file> → …` line. No new instrumentation.

Before, an untimed run and a sub-millisecond run were the same bytes. After, a fixed runner reads `PASS 12p/0f (324ms)` and an unfixed one reads `PASS 12p/0f (unmeasured)` — so grepping `turn_end: test` for `(unmeasured)` by runner name enumerates exactly which runners still lack a duration. Gradle will always appear there, which is the intended legible record of "we do not have this number" rather than a silent zero.

## Verification

Build, lint, and changelog validation clean. 185 tests across 5 files, plus the shared-seam siblings of `runtime-turn.ts` (`index-integration`, `cascade-turn-merge`, `packaging`, `build-freshness-guard`, `analyzer-coverage`).

17 mutations, every mutant rebuilt before running — the build-freshness guard makes an unrebuilt sweep read falsely all-red. Zero surviving mutants. Two survivors were caught and dealt with rather than reported as passes: a `Math.trunc` mutant survived a weak `40.4 → 40` assertion until a `40.6 → 41` case was added, and one unit-alternation reorder is a genuine equivalent mutant, killed instead by removing the `\b` it relies on.

Anchoring was tested directly: two tests exist only to prove a cargo panic message and an rspec failure diff cannot impersonate their summary lines, and both die when the anchor is removed.

Full suite not run; CI is authoritative.

## Deliberately left out

`< 1 ms` from vstest reports unmeasured rather than 0. It is a real sub-millisecond measurement, but the sentinel cannot express it and inventing a 0 would re-create #1479. Expressing it needs a third state on `TestResult`.

`parsePytestOutput`, `parseMixTestOutput` and `jsonRunDurationMs` already guard correctly and were not rewired through the shared sentinel, which would widen the diff past both issues. Filed as #1484.

### DIFF
--- a/.changelog/fix-1479-1480-runner-durations.md
+++ b/.changelog/fix-1479-1480-runner-durations.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Test-run durations are reported honestly (closes [#1479](https://github.com/apmantza/pi-lens/issues/1479), closes [#1480](https://github.com/apmantza/pi-lens/issues/1480))** — The turn-end log printed `(0ms)` for a run that was never timed, so an absent measurement was indistinguishable from a sub-millisecond one; it now prints `(unmeasured)`. Behind it, the generic runner parser read an elapsed time only for go and hardcoded `0` for the rest. Cargo, dotnet/vstest, maven/surefire, rspec, and minitest durations are now parsed from the summary block each runner already prints. Gradle stays unmeasured on purpose: its console output carries no test elapsed time, and the build time is not the same number. Multi-class maven runs are also scored by the surefire aggregate rather than by their first test class.
--- a/clients/run-duration.ts
+++ b/clients/run-duration.ts
@@ -1,0 +1,48 @@
+/**
+ * #1479: one place decides whether a test run was actually timed.
+ *
+ * `TestResult.duration` carries two different facts in one number. A run that
+ * took 40ms reports 40. A run with no suite timestamps, an `emptyResult`, or a
+ * runner error also reports a number — 0 — and the turn-end log printed that
+ * as `(0ms)`, which reads as a measurement of zero. A reader could not tell
+ * "measured 0" from "never measured", which is the confusion #1452 was
+ * reported for and the reason #1480 is otherwise unverifiable from the log.
+ *
+ * Every parser in `test-runner-client.ts` degrades to 0 when it cannot find an
+ * elapsed time (see `jsonRunDurationMs` and `parsePhpunitOutput`, both #1452),
+ * so 0 is the agreed sentinel for "unmeasured". This module is the only place
+ * that spells out what that sentinel means, so the log line and the
+ * agent-facing `formatResult` cannot drift apart.
+ *
+ * Kept in its own module rather than exported from `test-runner-client.ts`:
+ * `runtime-turn.ts` needs it on the hot turn-end path, and the client is
+ * deliberately loaded lazily (`bootstrap.ts` dynamic-imports it), so a value
+ * import from there would drag the whole runner into startup.
+ */
+
+/** True only for a duration that represents a real measurement. */
+export function isMeasuredDuration(
+	duration: number | null | undefined,
+): duration is number {
+	return (
+		typeof duration === "number" && Number.isFinite(duration) && duration > 0
+	);
+}
+
+/**
+ * Normalise a parsed elapsed time to the `TestResult.duration` contract.
+ *
+ * Negative, NaN, and non-finite inputs collapse to the unmeasured sentinel: a
+ * garbled summary must degrade to "we did not measure this", never to a wrong
+ * number.
+ */
+export function toMeasuredDurationMs(duration: number): number {
+	return isMeasuredDuration(duration) ? Math.round(duration) : 0;
+}
+
+/** Render a duration for a log line: `123ms`, or `unmeasured` when absent. */
+export function formatRunDurationMs(
+	duration: number | null | undefined,
+): string {
+	return isMeasuredDuration(duration) ? `${Math.round(duration)}ms` : "unmeasured";
+}
--- a/clients/runtime-turn.ts
+++ b/clients/runtime-turn.ts
@@ -63,6 +63,7 @@ import { RUNTIME_CONFIG } from "./runtime-config.js";
 import { isSubagentSession } from "./subagent-mode.js";
 import type { RuntimeCoordinator } from "./runtime-coordinator.js";
 import type { TurnStateOwner } from "./cache-manager.js";
+import { formatRunDurationMs } from "./run-duration.js";
 import type { TestResult, TestRunnerClient } from "./test-runner-client.js";
 import {
 	MAX_ADVISORY_AFFECTED_FILES,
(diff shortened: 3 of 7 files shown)


apmantza:
## Adversarial review — three of this PR's own claims are false

Back to draft until the fix round lands. The direction holds and the gates are green, but the review found four real surviving mutants and, more seriously, disproved three statements in the description above.

**The gradle claim is wrong.** `parseGenericRunnerDuration` never receives the runner name, so all six probes run against every runner's output. Gradle output returns 3000 ms borrowed from `BUILD SUCCESSFUL in 3s`-shaped text — precisely the thing this PR says it refuses to do. The go probe is the widest of the six (unanchored, and its `\s+` crosses newlines), and `parseGenericRunnerOutput` is the `default:` arm, so any unrecognized runner's output is scanned by all six patterns. Gradle needs to be structurally incapable of borrowing a number, not merely unlikely to.

**The maven aggregate change can report PASS on a failing build.** A `--fail-at-end` multi-module reactor payload with a failure in the first module and a clean second module returns `{passed: 1, failed: 0}` at exit code 1 — the turn-end line flips from `FAIL 7p/3f` to `PASS 1p/0f`. Pre-PR first-match returned 10/3. pi-lens invokes `mvn test -q` without `-fae`, but Maven reads `.mvn/maven.config` and `MAVEN_ARGS`, so a repo can turn it on. The change also left counts scoped to the last module while duration sums all of them, which is an inconsistency this PR introduced.

**The anchoring guarantee is overstated.** `^Finished in` only rejects *indented* decoys: two unindented `Finished in` lines return the first, so 99900 instead of 500. And `^Finished in` is not the counts line for rspec or minitest, so "every pattern is anchored to the line that carries the counts" is inaccurate for two of the six probes.

**Four surviving mutants**, two of which matter: dropping the `-` stop from the vstest duration capture, and dropping the `^` from the minitest anchor. Both are documented in the doc comment as safety properties with no test behind them. The other two are genuine equivalent mutants.

Also worth stating plainly: `(0ms)` is not gone from the log. A pytest run reporting `in 0.0004s` yields `0.4` and prints `0ms` — which is correct under the new semantics, meaning measured-at-approximately-zero. The description reads as though the string disappeared.

Verified and unchanged: the `1 h 2 m 3 s 4 ms` = 3723004 arithmetic and `< 1 ms` = 0; the equivalent-mutant call on the unit alternation; the lazy-load justification for the separate module; that only two duration-formatting sites exist and they now agree.

Not fixed here, filed as #1484: pytest, phpunit and mix still assign `duration` without the shared sentinel. `parsePytestOutput` on `1 passed in .s` yields NaN, caught defensively at the read side, so no wrong number reaches a user.

I will correct the description once the fixes land.

apmantza:
Closing in favour of #1486, whose representation is better than this one's.

This PR reserved `0` as the unmeasured sentinel. #1486 makes `TestResult.duration` optional instead, so a present `0` stays a real measurement. Under `strict`, that turns every unguarded read into a compile error — `npm run lint` enforces the contract, where a sentinel gets no help from the compiler at all.

Two limitations documented in this branch dissolved on that representation rather than needing workarounds: vstest's `< 1 ms` is now an honest measured zero instead of a value this branch had to call unmeasured, and gradle's deliberate non-measurement states itself in the type rather than relying on the log's `> 0` check to hide a sentinel.

The #1480 half of this work is ported onto #1486 as `4c3c3168` and `6437a970`, with the six per-runner probes, runner-scoped dispatch, the maven multi-module count fix and the exit guard, adapted to return `undefined` for unmeasured.

One defect this branch carries and the port does not: the non-zero-exit guard rewrites `failed` out of the state the `error:` field's condition reads, making `Runner X exited with N" unreachable when counts parse to zero failures at a non-zero exit. Fixed in the port by capturing the condition first. #1487 covers the separate case where no counts parse at all.

#1479 and #1480 close via #1486.

PR #1489, opened 2026-08-16
fix(dispatch): stop latching timed-out probes across seven tools (#1476)

### LINKED ISSUE #1476: Three tools still latch a timed-out probe as permanently unavailable
#1467 fixed the transient-probe latch for knip, madge, govulncheck, and vulture. The class sweep that accompanied it missed three consumers carrying the identical shape, found by the adversarial review of #1474.

## The shape

A probe failure of any kind — including a timeout — is remembered as a durable "tool is not installed" verdict for the rest of the session. A slow first run under load then disables an installed tool until restart, and the message blames a missing install.

## Sites

- **`clients/dispatch/runners/sg-runner.ts:277`** — any probe failure across all candidates sets `this.available = false` permanently, after first running a full auto-install. Its own comment at `:189` says it mirrors `KnipClient.ensureAvailable`, which is exactly the code #1467 just fixed.
- **`clients/biome-client.ts:156`** — the probe collapses every failure, timeout included, to `{ outcome: "missing" }`, so a timed-out probe both triggers an install and latches false. Biome is a hot-path formatter, so this is the most user-visible of the three.
- **`clients/govulncheck-client.ts:171`** — a failed `go install` (network, 60 s budget) still latches a durable `false`, inside the file #1474 edited.

Migrate each to the shared `availability-policy.ts` that #1474 introduced, rather than repairing them in place.

## Related: a duplicate added by the deduplication

`clients/knip-client.ts:117`'s managed-shim resolver is a line-for-line duplicate of `clients/jscpd-client.ts:60`. Both should call one `findManagedNodeToolBinary(name)`. Worth folding in here since the same files are open.

## Observability

`availability_decision` already carries the cause and timing, so migrating a tool makes its verdicts readable the moment it lands — no new record needed. One gap to close with it: `clients/dead-code-client.ts:275` and `:289` log `elapsedMs: 0` for vulture's transient and missing verdicts even though up to four 5 s probes just ran, and the measured `hostStallMs` is discarded. If vulture breaks the way knip did, the forensic trail that cracked #1467 will not exist. That one is fixed in #1474 itself.

## Acceptance criteria

- Each of the three sites distinguishes a timeout from a genuine absence and retries the former.
- Each routes through `availability-policy.ts` rather than its own copy of the rule.
- The managed-shim resolver has one definition.
- A coverage test asserts every availability consumer uses the shared policy, so the next tool cannot reintroduce the shape.

Refs #1467.

apmantza:
Closes #1476.

## Problem

#1467 fixed one instance of a defect and left six behind. A probe failure of any kind — including a timeout — was remembered as a durable "tool is not installed" verdict for the session, so one slow first run under load disabled an installed tool until restart and the message blamed a missing install.

The issue named three sites. A sweep for the shape found four more.

## What was migrated

Named in the issue:

- **`clients/biome-client.ts`** — collapsed every failure to `{ outcome: "missing" }`, so a timeout both triggered an install and latched false. Biome is a hot-path formatter, so this was the most user-visible.
- **`clients/sg-runner.ts`** — every candidate probe failure latched, after first running a full auto-install. A transient sweep now short-circuits *before* the install.
- **`clients/govulncheck-client.ts`** — a timed-out `go install` and a timed-out re-probe of a successful install now mark transient. A non-zero install exit still latches deliberately: retrying that every turn is a `go install` storm.

Found by the sweep:

- **`clients/dispatch/dispatcher.ts` `checkToolAvailability`** — the largest, and nobody had named it. This is `ctx.hasTool`, the generic gate every CLI runner passes through, and it cached a stalled probe as a session fact.
- **`isSgAvailableAsync` in `runner-helpers.ts`** — the shared ast-grep memo behind every slop runner, carrying `SgRunner`'s exact shape one directory away from the policy.
- **`clients/go-client.ts` and `clients/rust-client.ts`** — `go version` / `cargo --version` on a 3 s budget. A stall disabled every Go or Rust diagnostic for the session. Both also gained in-flight sweep sharing, which matters now that verdicts expire.

Also folds in `findManagedNodeToolBinary(name)`, replacing the line-for-line duplicate that the deduplicating PR itself added to `knip-client.ts` and `jscpd-client.ts`.

## The durable close

The first version of this gate did not work. It scanned for the defect shape with regexes, and a review wrote seven files carrying that shape into `clients/` — all seven passed. A file counted as compliant if it contained the string `availability-policy.js` **anywhere**, so a `// TODO: route through availability-policy.js` comment marked it clean. Against master it flagged only 5 of the 7 sites this PR migrates, and could never have found two of them, because both already imported the policy for a different memo and the gate judged per file.

That is worse than a hand-written list: a list is visibly incomplete, a derived gate reads as proof.

`tests/support/availability-gate.ts` now analyses the AST with `@ast-grep/napi`, already a runtime dependency. Three properties, each bought with one of the evasions:

- **Routing means an `import_statement` binds a policy symbol.** A comment is not an import statement.
- **Judgement is per top-level unit, not per file**, with module-local call-graph propagation to a fixed point — the pre-fix `runner-helpers.ts` latch sat one function away from its `--version` literal and its spawn.
- **A memo is state that outlives the call** holding a verdict shape, which is what separates `pipeline.ts`'s `_eslintCache` from a per-call status report.

All seven evasions are now fixtures the gate must flag, plus a second hand-rolled latch beside a compliant one in the same file, a probe one helper away from its latch, and two positive controls. Against master it flags 6 of 7, adding `runner-helpers.ts::isSgAvailableAsync` — the second-latch case that was the crux.

It does **not** flag pre-fix `govulncheck-client.ts`, and that is correct rather than a miss: its memo was `SecurityScanClient`'s policy-backed latch, and its defect was a durable `false` written from a transient install failure *inside* a governed memo. Different shape. The rule was not stretched to manufacture a seventh hit.

**Named limits, so this is not read as proof again.** `clients/pipeline.ts::tryEslintFix` remains unrouted and is a named baseline entry — shrink-only, so a new violation fails the gate and a baseline entry that stops being flagged also fails, meaning a fix cannot leave scaffolding behind. And the verdict-shape rule does not see `formatters.ts` or `package-manager.ts` (#1495, #1496), because their memos hold resolved paths rather than boolean verdicts.

## Observability

Every migrated tool now emits one `availability_decision` per verdict carrying `outcome`, `cause`, `elapsedMs`, `hostStallMs`, `retryAfterMs` and `budgetMs` — including govulncheck's two transient branches, which previously logged nothing at all. A latched tool is now a single grep instead of an inference from weeks of silence, which is the gap that let knip stay dead.

Confirmed `availability_decision` remains in `LAST_PHASE_EXCLUDED`, so these zero-duration decision records cannot poison `loop_block` attribution.

## Verification

**Behavior preservation first.** 14 pins covering genuine absence and successful probe for every tool pass on pre-fix code, so the change is provably confined to the transient path. Against reverted clients, 12 failed and 14 passed — nine regression assertions fail without the fix.

Three `ctx.hasTool` tests cannot run pre-fix because that seam was module-private; their binding rests on mutation instead, and the commit message says so rather than implying a red-first proof they do not have.

**11 mutants, all red**, each with a full rebuild between: the biome latch, the sg-runner transient gate, the shared-sg verdict, govulncheck install classification, the go and rust verdicts, both dedupe paths, the dispatcher fact write, the managed-shim resolver, and a synthetic unrouted client against the coverage gate.

Build, lint and changelog clean. 26 tests across the three new files; 791 of 792 across the dispatch suite and 12 related files.

**The one failure is pre-existing.** `tests/clients/jscpd-client.test.ts > uses the managed executable after PATH installation` fails identically with `clients/` reverted to master. It expects the second spawn to use the managed path and gets `npx`, most likely because this dev machine has a managed jscpd shim that trips the fast path. Filed as #1491, untouched here.

No full suite; CI is authoritative.

## Sweep correction

The review found the sweep is **not** complete, and one miss is in a file this PR edits. Correcting the record rather than leaving the claim above to stand:

- `createCwdCachedProbe` (`runner-helpers.ts:627`) carries the exact shape and feeds eslint (5 s), credo (10 s, the slowest probe in the repo) and rust-clippy (8 s). Filed as #1494.
- `clients/formatters.ts` was excluded here as "filesystem detection with no memoized spawn verdict". **That is wrong on both halves.** `which()` at `:199` is a `safeSpawnAsync` with a 5 s timeout, and its result is memoized in `detectionCache`, invalidated only by config mtime or LRU — never by time. One stalled `which rustfmt` disables Rust formatting until the user edits a config file. Filed as #1495.
- `clients/package-manager.ts:95` memoizes a package-manager probe forever; a timeout silently downgrades the session to npm, with the wrong lockfile semantics. Filed as #1496.

None are fixed here — each needs its own regression proof — but the claim that the sweep was thorough should not have been made without them.

## Left for follow-ups

`clients/dispatch/runners/psscriptanalyzer.ts` carries the same latch but bypasses `safe-spawn` entirely via a hand-rolled `spawnPs` that returns `status: null` on timeout, so it has no failure taxonomy to classify. Migrating it means routing it through `safeSpawnAsync` first. Filed as #1490.

Screened and correctly excluded: `test-runner-client.ts`'s `availableRunners` and `formatters.ts`'s `detect` are filesystem and package.json detection with no memoized spawn verdict.

### DIFF
--- a/.changelog/fix-1476-availability-latch-siblings.md
+++ b/.changelog/fix-1476-availability-latch-siblings.md
@@ -7,7 +7,10 @@ section: Fixed
   a missing install, the way knip and madge already did. Before this, one stalled
   second at session warm-up latched "tool is not installed" for the life of the
   process, and biome and ast-grep also paid for an auto-install nobody needed. A
-  transient verdict now expires and the tool comes back without a restart, and
-  every verdict is recorded in `latency.log` as `availability_decision` with its
-  cause and timing. A coverage test derived from the source tree keeps the next
-  tool from reintroducing the shape.
+  transient verdict now expires and the tool comes back without a restart, its
+  cooldown escalates on a host that stays sick, and every verdict is recorded in
+  `latency.log` as `availability_decision` with its cause, timing and retry
+  schedule. A structural coverage test parses `clients/` and fails when a new
+  version probe parks its verdict outside the shared policy; the hand-rolled
+  latches that predate it are listed in the test as named, shrink-only gaps
+  rather than being quietly excluded.
--- a/.changelog/fix-1476-availability-latch-siblings.md
+++ b/.changelog/fix-1476-availability-latch-siblings.md
@@ -1,0 +1,13 @@
+---
+section: Fixed
+---
+
+- **A slow first probe no longer disables an installed tool (closes #1476)** —
+  biome, ast-grep, Go, cargo and govulncheck now tell a probe timeout apart from
+  a missing install, the way knip and madge already did. Before this, one stalled
+  second at session warm-up latched "tool is not installed" for the life of the
+  process, and biome and ast-grep also paid for an auto-install nobody needed. A
+  transient verdict now expires and the tool comes back without a restart, and
+  every verdict is recorded in `latency.log` as `availability_decision` with its
+  cause and timing. A coverage test derived from the source tree keeps the next
+  tool from reintroducing the shape.
--- a/clients/biome-client.ts
+++ b/clients/biome-client.ts
@@ -59,6 +59,14 @@ export class BiomeClient {
 	private readonly availabilityLatch = createAvailabilityLatch();
 	/** Cause of the last probe, read when the resolution comes back. */
 	private lastProbeCause: AvailabilityCause | null = null;
+	/**
+	 * Measurements of the last failed probe, held until the latch decides how
+	 * long the verdict lasts. The decision record is emitted after that, so it
+	 * can carry `retryAfterMs` — a latch you can see is worth little if the
+	 * retry schedule is not greppable beside it (#1474).
+	 */
+	private lastProbeElapsedMs = 0;
+	private lastProbeHostStallMs = 0;
 	// Per-cwd cache of the resolved biome binary. Keying by cwd matters in
 	// monorepos where different sub-packages each ship their own biome
 	// installation; sharing one slot across the whole client would cause
@@ -221,16 +229,10 @@ export class BiomeClient {
 		const failureOutcome: Exclude<AvailabilityOutcome, "success"> =
 			outcome === "success" ? "non-installable" : outcome;
 		this.lastProbeCause = cause;
-		logAvailabilityDecision({
-			tool: "biome",
-			verdict: "unavailable",
-			outcome: failureOutcome,
-			cause,
-			elapsedMs,
-			latched: failureOutcome !== "transient",
-			hostStallMs,
-			budgetMs: PROBE_TIMEOUT_MS,
-		});
+		this.lastProbeElapsedMs = elapsedMs;
+		this.lastProbeHostStallMs = hostStallMs;
+		// The record is emitted by `doEnsureAvailable`, once the latch has said how
+		// long this verdict holds.
 		return { outcome: failureOutcome };
 	}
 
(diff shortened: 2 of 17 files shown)


apmantza:
Re-homing a leftover here before #1490 closes, so it does not close with the work.

#1490's acceptance criteria asked whether `tests/clients/availability-policy-coverage.test.ts` flags `psscriptanalyzer.ts`. It does not, and cannot: the gate recognises a probe by a quoted **version flag**, and neither of that file's probes asks for a version — one runs `-Command "exit 0"` to see whether an interpreter answers, the other asks `Get-Module -ListAvailable`. The file is dropped by the `VERSION_FLAG_TEXT` pre-filter before any unit is analysed.

So #1490 migrates the latch and the file stays invisible to the gate. That is not a baseline entry (nothing is flagged to baseline) and it is not a #1490 deliverable — it is a widening of the gate's probe vocabulary, which belongs here:

**Leftover:** teach the gate to recognise a **capability probe with no version flag** — a spawn whose verdict is memoized and whose argv asks "can I run this?" rather than "what version are you?". Members visible today: `psscriptanalyzer.ts`'s two probes, and `formatters.ts`'s `which <command>` (#1495 migrated it; it is also still invisible for the same reason).

Cost of not doing it: a migrated latch in either file can regress to a hand-rolled one and the gate stays green. That is exactly the blindness the #1494 round tightened for the delegation case, and the same argument applies here.

Not urgent, and deliberately not a wide widening: `-Command`/`-File`/`-NoProfile` are not version flags and must not become them. The narrower framing is "a probe whose result is memoized, in a file that already probes" — worth scoping carefully before writing it, since a gate that cries wolf gets baselined rather than fixed.

Issue #1501, opened 2026-08-16
probeVersion logs its decision before the caller sets the cooldown, so retryAfterMs is absent


apmantza:
`clients/security-scan-client.ts:123-136` — `probeVersion` emits its `availability_decision` record before the caller decides the cooldown, so a transient PATH-probe verdict carries no `retryAfterMs`.

Everything else on the record is present: `outcome`, `cause`, `latched: false`, `elapsedMs`, `hostStallMs`, `budgetMs`. A latched tool is still a single grep on `latched`, and the retry schedule is derivable from the documented constants, so this is a completeness gap rather than a blind spot.

It was left out of #1489 deliberately. Fixing it properly means restructuring the base seam shared by gitleaks, trivy, opengrep and vulture to add one derived field, which was disproportionate to that change.

## Direction

Worth doing when the base seam is next touched for another reason. Either defer the record until the caller has decided, or have the seam own the cooldown decision so the value exists at log time.

## Acceptance criteria

- A transient `probeVersion` verdict carries `retryAfterMs`.
- The other four consumers of the seam keep emitting exactly what they emit today — verify by pinning their records first.

Refs #1476.

Issue #1266, opened 2026-08-11
resetDispatchAvailabilityState() is never called — install-failure suppression is permanent, not per-session


apmantza:
## Problem

`resetDispatchAvailabilityState()` (`clients/dispatch/runners/utils/runner-helpers.ts:186`) was added by PR #1263 to implement #1222's "suppressed until next session" reset — but **no production code ever calls it**. Only tests do.

Consequence: after N failed install/resolution attempts, a tool is suppressed for the **entire process lifetime**, not until the next session. For a long-lived warm MCP server host that is effectively forever: one transient failure (npm registry blip, locked file during install) permanently disables the tool with no recovery short of restarting the process.

## Fix

Wire the reset into session start (`handleSessionStart` is the natural hook, same as the stage-file sweep from #1260), or whichever boundary "next session" was intended to mean. Add a test that drives suppression, starts a new session, and asserts the tool is retried — the current suite cannot catch this because it calls the reset helper directly.

## Also folded in (same review)

The probe-cache test at `tests/clients/installer/tool-discovery.test.ts:288` asserts the *returned* `.cmd` path but not the value **persisted** by `updateProbeCache` — which was #1223's core acceptance criterion. Add the recorded-value assertion while in there.

Refs #1222 #1223 #1263.

apmantza:
Fixed by #1282 (merged). Reviewed FIX-BEFORE-MERGE (found a fire-and-forget reset race, a leaking static registry, and a per-resolution PATH scan); all three addressed and re-verified. Closed manually (branch used refs).

PR #1514, opened 2026-08-17
fix(availability): bound install-class transient retries (closes #1497)

### LINKED ISSUE #1497: govulncheck's retried go install can run a 60s compile every 5 minutes indefinitely
#1489 makes a timed-out `go install` transient rather than latching, which is right — a network stall should not disable govulncheck for the session. But the backoff it inherits was calibrated for a 1.5–5 s version probe, and the operation it now retries is a 60 s network install that compiles `x/vuln`.

`clients/govulncheck-client.ts:180-199`. The escalation is 30/60/120/240 s capped at 300 s. On a host where that install reliably exceeds its budget, steady state is a 60 s compile every 5 minutes, forever — roughly a 20% duty cycle on one core, with nothing that ever gives up.

A non-zero install *exit* still latches deliberately, to avoid exactly this kind of storm. The timeout path has no equivalent stop.

## Direction

Install-class operations need their own bound: a much longer cap than probe-class retries, an attempt ceiling after which it latches, or both. Whatever the choice, the cost should be stated in the code rather than inherited from a constant named for a different operation.

## Observability

`availability_decision` records `retryAfterMs` and the cause, so the duty cycle is derivable from `latency.log` — but only if someone thinks to look. Consider whether a repeated install-class retry deserves a louder record, since the symptom a user notices is an unexplained busy core rather than a missing tool.

## Acceptance criteria

- A repeatedly timing-out `go install` reaches a terminal state instead of retrying indefinitely.
- A transient network failure still recovers without a restart, which is what #1489 fixed.
- A test pins the retry ceiling.

Refs #1476.

apmantza:
## Problem

#1489 made a timed-out `go install` transient rather than latching — correct for a network stall — but the backoff it inherits from the shared latch was calibrated for a 1.5–5 s version probe: 30 s doubling, capped at 5 min, no terminal state. The operation it now retries is a 60 s network install that compiles `x/vuln`. On a host where that install reliably exceeds its budget, steady state is a 60 s compile every 5 minutes, forever — ~20% duty cycle on one core, with nothing that ever gives up.

## Change

**Policy** (`clients/dispatch/runners/utils/availability-policy.ts`) — install-class transient retries get their own ladder, stated in the code instead of inherited from a probe-named constant: `INSTALL_TRANSIENT_COOLDOWNS_MS = [5 min, 10 min]`, and the ceiling (`INSTALL_TRANSIENT_MAX_ATTEMPTS = 3`) is **derived from the ladder's length** so the two can never disagree. Worst case: three ≤60 s compiles across ~15 minutes, then terminal for the session. The host-stall 5 s shortcut deliberately does not apply to install-class — the retry itself is the expensive part, no matter who ate the budget.

**Latch** — `noteUnavailable(outcome, cause, opts?)` gains `opts.operationClass: "install"`. Install-class attempts count separately from probe-class attempts and hold their **own cooldown slot**; `getRetryAtMs()` reports the later of the two, which is the gate `read()` enforces. At the ceiling the latch marks the verdict exhausted, rewrites the cause to `install-retry-exhausted`, and returns **0**, which the interface already documents as "latched".

**Session boundary** — the ceiling is terminal for a SESSION, not the process. `resetInstallRetryLatches()` (a generation counter, so no latch registry has to be hand-maintained) re-arms install-class state in every latch, and `handleSessionStart` calls it beside `resetDispatchAvailabilityState()` / `resetZizmorTokenAvailability()` / `resetPsScriptAnalyzerAvailability()`. `noteAvailable()`/`reset()` still re-arm too, so a successful run recovers immediately — preserving #1489's guarantee.

**Wiring** — `SecurityScanClient.markTransientlyUnavailable` passes the option through; govulncheck's install-timeout branch (the only install-class mark in the repo — see sweep below) uses it and reports `latched: true` at exhaustion.

**Observability** — the terminal verdict gets the louder record: a new `install-retry-exhausted` degradation-ledger kind, recorded once per session via `recordDegradationOnce` (same session lifetime as the latch, now that both re-arm at `session_start`) and surfaced by `pilens_health`. The exhaustion `availability_decision` row follows #1534's convention: `classifiedBy: "caller"`, plus `evidence` carrying `command: "go install"`, `install: "failed"` and a ceiling `installReason`. `describeUnavailability` renders the exhausted cause as a retry ceiling that clears next session, instead of the generic "It will be retried."

## Class sweep

Every `noteUnavailable("transient", …)` caller audited: govulncheck's `go install` is the **only** install-class transient mark. The rest guard cheap version probes (knip, madge, biome, jscpd, package-manager, toolchain sweeps, the security-scan PATH probes, govulncheck's own `go version` and post-install re-probe) and correctly keep the probe schedule. `ensureViaInstaller`'s `ensureTool`-failure path already latches durably (pre-existing, out of scope — noted for the record).

## Review round (2026-08-18)

An adversarial review found two structural defects plus smaller items. All are fixed on this branch, each with a red-first test.

| Finding | Fix | Test that was red before it |
| --- | --- | --- |
| **F1 (critical)** `installExhausted` was a process-lifetime dead end — nothing re-armed it at `session_start`, so a repaired network plus a full reset sequence still refused to install | session generation counter + `resetInstallRetryLatches()` wired into `handleSessionStart`'s #1266 reset block | `install-retry-session-rearm.test.ts` (6 tests) and `runtime-session-dispatch-reset.test.ts` → "re-runs an exhausted `go install` after handleSessionStart" |
| **F2 (high)** install and probe classes shared one `retryAtMs`, so a host-stall probe failure (5 s) collapsed the 300 s install spacing | install-class gets its own slot; `getRetryAtMs()` returns the later gate | "never lets a probe failure shorten an install cooldown" — asserts the GATE, not the returned delay |
| **F3** exhaustion ledger record (`recordDegradationOnce`, session-cleared) vs a latch that was process-lived | after F1 both clear at `session_start`; asserted | "records the exhaustion once, with a lifetime that matches the latch" |
| **F4** the 30-min `INSTALL_TRANSIENT_MAX_COOLDOWN_MS` was unreachable behind a 3-attempt ceiling, and a test pinned that dead branch | constant deleted; the ladder is the literal schedule the ceiling derives from; `installRetryDelayMs` returns 0 at/past the ceiling | "owes no cooldown at or past the attempt ceiling" |
| **F5** `getOutcome() === "transient"` + "will be retried" contradicted the exhausted state | cause rewritten to `install-retry-exhausted`; `describeUnavailability` renders it honestly | "describes an exhausted ceiling honestly" |
| **F7** the ceiling row lacked #1534's classification evidence | `classifiedBy: "caller"` + install facts + ceiling `installReason` | "writes a ceiling decision row that says who classified it, and why" |
| **F8** stale conflict note claiming #1511 was open | #1511 is merged; its `probeVersion` self-latching is merged into this branch and the note is gone | n/a |

**Merge conflict resolution.** Four conflicts against master, resolved on top of master's text rather than by taking this branch's side: `availability-policy.ts` keeps master's `AvailabilityLatchOptions` / `maxCooldownMs` (#1535) and `transientRetryDelayMs`'s third argument; `degradation-ledger.ts` keeps master's `query-predicates-invalid` beside the new kind; `security-scan-client.ts` and `govulncheck-client.ts` re-apply this PR's edits over #1501's `probeVersion` self-latching and #1534's `noteDurableAbsence` / `describeProbeEvidence` / install facts. The merge commit is a faithful merge — `git diff origin/master HEAD` is exactly this PR's 11 files.

## Verification

- **Red-first, pre-fix:** 9 failed / 16 passed across the two policy suites on the merge commit before the round-2 fixes (F1 = module-level failure on the missing export; F4 `expected 1200000 to be 0`; F2 `expected …041046 to be …336046`; F5 `expected 'probe-timeout' to be 'install-retry-exhausted'`).
- **Wiring mutation** `if (false && opts?.operationClass === "install")`: **11 tests fail across 4 files** (the earlier round's schedule test pinned constants only; the ceiling, class-independence, cooldown-slot, describe, client-level and session-wiring tests all catch it now).
- **Wiring mutation** removing `resetInstallRetryLatches()` from `handleSessionStart`: `runtime-session-dispatch-reset.test.ts` fails (`expected 3 to be greater than 3`).
- **Green:** 136 tests / 11 suites — the four availability suites the review ran, plus `install-attempt-evidence`, `security-scan-probe-telemetry`, `availability-policy-coverage`, `degradation-ledger`, `zizmor-token-latch`, `runtime-session-dispatch-reset`, `runtime-session`. `npm run build` and `npm run lint` clean.
- **Skipped:** the full suite. CI is authoritative.
- `BASELINE_KEYS` in `security-scan-probe-telemetry.test.ts` needed no change: the new fields land on the govulncheck ceiling row, not on the rows that suite pins.

## Conflict note

#1559 (open) also touches `availability-policy.ts`, adding `servedFromCooldown` to `AvailabilityDecision` / `logAvailabilityDecision` — different hunks from this PR's (the cause union, the install ladder, the latch body), additive both ways, no merge-order constraint.


apmantza:
Follow-up once this and #1563 both land (from #1563's report): fold `lazyInstallRetryDelayMs` into this PR's `INSTALL_TRANSIENT_*` ladder — numbers already deliberately match, so the fold is a deletion. Single source of truth for install-class cooldowns.

Issue #1519, opened 2026-08-17
Add CUE language support (LSP + formatter)


bcachet:
Add CUE language support to pi-lens:

- CUE Language Server (cue lsp serve) as a primary LSP server
- cue fmt -w as the smart-default formatter for .cue files
- cue-lang/cue GitHub release tool for auto-install
- Dispatch plan entry and smoke fixtures

Follows the CONTRIBUTING.md checklists for adding a language server, formatter, and dispatch plan entry.

apmantza:
Thanks for the request and for backing it with an implementation straight away, @bcachet — that's the ideal shape for a language-support ask. (Maintainer-supervised AI triage.)

Status: accepted as the tracking issue for CUE support. Your PR #1520 has a full review posted (two verified blockers: the `cue fmt -w` flag and cuelsp's missing `publishDiagnostics`, plus smaller registry seams) and CI is approved to run on it. The suggested path there is to split: land the formatter half as soon as the flag is fixed, and hold or explicitly mark the LSP half non-diagnostic until cuelsp publishes diagnostics. This issue stays open to track the LSP half either way.

PR #1520, opened 2026-08-17
CUE: Add LSP, formatter, tree-sitter grammar, and full registration (refs #1519, #1522)

bcachet:
## Summary

Add support for CUE language (LSP + formatter)

## Type of change

- [ ] Bug fix
- [X] New feature (net-new capability)
- [ ] Documentation

## Area

- [X] area:lsp
- [ ] area:dispatch
- [X] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [X] area:tests

## Checklist

- [X] I have read [CONTRIBUTING.md](../CONTRIBUTING.md) and [AGENTS.md](../AGENTS.md)
- [X] The change has tests (happy path, edge cases, regression test for bugs)
- [X] `npm run lint` passes
- [X] `npm test` passes
- [X] `npm run build:dist` succeeds if I changed code under `clients/`, `commands/`, `tools/`, or `index.ts`
- [X] `package-lock.json` is in sync with `package.json` (run `npm install` after dep changes)
- [X] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there
- [X] `.changelog/<branch-or-slug>-<short-desc>.md` has one valid entry **in this PR** for any user-facing change (Added/Changed/Deprecated/Removed/Fixed/Security) — see [.changelog/README.md](../.changelog/README.md); internal-only test/refactor PRs may skip it
- [ ] Commit subject includes the issue number: `(closes #NNN)` or `(refs #NNN)`

## What changed and why

Use this section for any non-obvious design decisions or gotchas the reviewer should know about.

---

## Maintainer addition (AI-assisted, maintainer-supervised)

Commits `a522d9f` and `bfa2f03` are maintainer-added, pushed to this branch
under "allow edits from maintainers". @bcachet's three commits are untouched.

The reason for widening rather than merging and following up: without a
tree-sitter grammar, `.cue` would be a half-registered kind — LSP and formatter
work, but symbol search, imports, and the structural runners return nothing.
Landing both halves together means `.cue` arrives as one complete language
rather than as a known gap plus a follow-up PR.

### Tree-sitter grammar

No publisher ships a CUE wasm. npm's `tree-sitter-cue` is a native binding, the
`tree-sitter-wasms` aggregator has no CUE, and upstream cuts no releases. Every
source in the grammar pipeline is a URL, so there was nothing to point a source
override at.

So pi-lens builds it. [eonpatapon/tree-sitter-cue](https://github.com/eonpatapon/tree-sitter-cue)
at `dd7b90e0770ff18070c515937ba3c3d6d93db00e` (MIT), built with
`npx tree-sitter-cli build --wasm`, produces a 224KB wasm carrying the modern
`dylink.0` section, which loads under the pinned web-tree-sitter 0.25.10
unchanged. No dependency bump.

That artifact is committed to `vendor/grammars/` and pinned by sha256 in
`scripts/grammars.lock.json` next to its build recipe, so
`scripts/check-grammar-provenance.mjs` re-hashes it on every CI run and the
committed binary cannot drift unnoticed.

Vendored is a distinct source kind, not a fourth override:

- the resolver searches `vendor/grammars` first, because it is the only place
  the file can come from;
- `downloadGrammarDetailed` refuses a vendored grammar with a **non-retryable**
  verdict. A retryable one would put a fetch that can never succeed onto
  #1536's cooldown ladder forever.

### Review items from the earlier round

- `cue fmt -w` is not a valid invocation (`unknown shorthand flag: 'w'` on cue
  v0.17.1) and the policy is `defaultWhenUnconfigured`, so it failed on every
  `.cue` write. Now `cue fmt $FILE`, which rewrites in place.
- `tests/fixtures/tool-smoke/cue/bad.cue` used `a: int = "hello"`, an evaluation
  error cuelsp does not report — the smoke could only pass vacuously. Replaced
  with an unclosed struct, plus a `cue.mod/module.cue` so the module resolves.
- `docs/language-coverage.md` said `✓`; it now says syntax-only and explains
  that validation belongs to `cue vet`. `docs/features.md` lists CUE too.
- The changelog fragment no longer claims a diagnostic wait strategy that was
  never added.
- Re-tabbed the blocks that landed with spaces.

### Verification

- `npm run build`, then `npm run lint` (tsc): clean.
- New `tests/clients/tree-sitter-cue-grammar.test.ts` (6 tests): a realistic CUE
  module covering packages, imports, a `#Definition`, constraints, disjunctions,
  interpolation, an `@attribute`, and comprehensions parses with zero ERROR and
  zero MISSING nodes; an unclosed struct degrades to an ERROR node and the
  shared WASM Module survives it. Proven red on pre-change code — all 6 fail.
- `PINNED_LSP_LANGUAGE_IDS` guard reproduced red first: it fired alone with
  `+ "cue"`, exactly the #1587 seam behaviour, with the other 7 guards passing.
  `appliesTo` and `LANGUAGE_EXTENSIONS` derive themselves.
- `scripts/check-grammar-provenance.mjs` passes and was proven to catch drift
  (appending one byte to the committed wasm fails it with both hashes).
- `node scripts/check-grammar-load.mjs cue` exits 0 — 200 stress parses, no
  runtime crash.
- Targeted suites green: grammar (10 files, 66 tests), registry and formatter
  (9 files, 263 tests), tree-sitter consumers (6 files, 135 tests).
- Full suite is CI's job.

### Deliberately not done

- The web-tree-sitter 0.25 → 0.26 migration (#1522's main body). Out of scope;
  the built wasm loads on 0.25.10, so CUE does not depend on it.
- The `SERVER_DIAGNOSTIC_STRATEGIES` entry from the earlier review. It needs
  `scripts/characterize-lsp.mjs` and `scripts/probe-clean-signal.mjs` run
  against a real `cue` binary, and there is none on this machine. Still open.
- A `cue vet` auxiliary runner for the validation the LSP omits. Follow-up.

The branch is BEHIND master but MERGEABLE.

## Review round 2 (AI-assisted, maintainer-supervised)

An adversarial review returned five findings. All five are fixed in `f03a396`.
Each new guard was reproduced red on the pre-fix head first.

**F2 (blocker) — the tool-smoke fixture proved nothing.** `cue lsp` only
publishes the parse diagnostic when the package clause is the very first line,
and the fixture carried a comment header above it, so the diagnostic was
suppressed. `package smoke` is now line 1 with the comments below.

The reason this survived two rounds is the lane itself: its verdict is
"handshook — server replied", which **passes on zero diagnostics**. That is
right for a fixture proving a server starts and exactly backwards for one
proving pi-lens can see a defect. A new `expectMessageMatch` binds the verdict
to the diagnostic text, so zero diagnostics now fails. The cue fixture sets it
to the message real `cue lsp serve` v0.17.1 publishes. The matcher is exported,
so the decision is testable without a `cue` binary on PATH.

**F3 — the changelog overclaimed.** `SYMBOL_QUERIES` and `IMPORT_QUERIES` have
no cue entry, so structural symbol search and import extraction skip `.cue`.
The fragment and the coverage doc now scope the claim to the parser and point
the query gap at #1522.

**F4 — two stale counts in `docs/features.md`.** It quoted 42 language servers
against 45 and 33 formatters against 34; both were already stale before CUE.
Fixed, and the check is now derived: `tests/docs/features-counts.test.ts`
compares each number against its registry, because a prose number mirroring a
registry is the hand-maintained-list defect shape.

**F5 — a missing vendored wasm reported the wrong verdict.** `fetchGrammar`
resolved a write directory *before* consulting the vendored short-circuit, so
on any host where web-tree-sitter is not locatable a missing vendored wasm came
back as the generic **retryable** "no writable grammars directory" verdict — a
packaging fault disguised as a transient download failure, arming a cooldown for
a fetch that can never happen. The refusal now comes first and is shared as
`vendoredGrammarRefusal`, so both callers answer identically.

### Round-2 verification

- Red first, on the pre-fix head: the F2 and F4 guards failed 8 of 10
  (`matchDiagnosticMessages is not a function`, package clause not on line 1,
  both counts wrong); the F5 guard failed with
  `vendoredGrammarRefusal is not a function`.
- Green after: grammar and seam suites 10 files / 69 tests; smoke-lane
  consumers and registry suites 9 files / 190 tests.
- `npm run build` and `npm run lint` clean.
- `scripts/check-grammar-provenance.mjs` passes.
- F5's happy path is unchanged, not just untested: the CUE parse tests and
  `node scripts/check-grammar-load.mjs cue` (exit 0) both still resolve the
  wasm out of `vendor/grammars`, where `fetchGrammar` is never reached. Only
  the missing-wasm error path changed.
- The live `--lsp` smoke still needs a `cue` binary, which this machine lacks.
  CI must confirm it, and the `SERVER_DIAGNOSTIC_STRATEGIES` entry remains open.

## Review round 3 (AI-assisted, maintainer-supervised)

One finding, and it was a fair one. The round-2 F5 guard asserted the return
value of `vendoredGrammarRefusal` — a pure constant. Deleting the entire
production reorder left it green, so it proved nothing: the red reported in
round 2 was the missing export, never the defect.

Fixed in `35b0989`. The guard now drives the real `fetchGrammar` on the host
the fix names, with `resolveGrammarFile` and `grammarsWriteDir` both stubbed to
`undefined`, and asserts on what it *records*: cue must be recorded
non-retryable with the vendored message, and python on that same host must
still be recorded retryable with the write-dir message. Without that control,
"everything is non-retryable" would pass and the branch would prove nothing
about vendored specifically.

**Mutation proof.** Removing the early-return block at
`clients/tree-sitter-client.ts:791` fails the guard:

```
FAIL  a missing vendored grammar on a host with no writable dir (#1520 F5)
      > records the vendored grammar as NON-retryable
AssertionError: expected true to be false
  - false   (expected retryable)
  + true    (received)
 ❯ tests/clients/tree-sitter-cue-grammar.test.ts:231
```

The python control stayed green through the mutation, which is what makes the
cue assertion specific rather than blanket. Production file restored to
byte-identical, rebuilt, re-verified.

This supersedes round 2's F5 verification line: the earlier
`vendoredGrammarRefusal is not a function` red was real but not a proof of the
fix, only of the symbol.

`origin/master` is merged in, so the branch is no longer BEHIND. Targeted
suites after the merge: 10 files / 70 tests and 7 files / 216 tests, both
green; provenance check and `check-grammar-load.mjs cue` both pass.


apmantza:
Thanks for this, @bcachet — the wiring follows our seams closely (installer entry, root markers, tool policy, smoke fixtures), and the GitHub asset matcher and `cue lsp serve` spawn are both exactly right. Up front: this review is AI-generated and maintainer-supervised.

Two things need to change before merge, both verified against cue v0.17.1 locally.

**1. `cue fmt -w` is not a valid invocation.** `cue fmt` only has `--check`, `-d/--diff`, and `--files`. Running the fixture through our own harness fails:

```
$ node scripts/smoke-tools.mjs --format cue
✗  cue   formatter failed to run: unknown shorthand flag: 'w' in -w
```

`cue fmt <file>` already rewrites in place, so `command: ["cue", "fmt", "$FILE"]` is the fix. Since the policy is `defaultWhenUnconfigured: true`, this would fail on every `.cue` write today. The fixture comment in `tests/fixtures/format-smoke/cue/messy.cue` mentions `-w` too.

**2. `cuelsp` does not publish diagnostics yet.** Driving `cue lsp serve` directly inside a real `cue.mod` module, with a file `cue vet` rejects (`a: int & "hello"`), the server loads the module and package, then sends no `publishDiagnostics` — on open, change, or save. Its `initialize` result advertises hover, definition, completion, code actions, rename, and formatting, but no `diagnosticProvider`. Through `LSPService.touchFile` we get `{"diags":[]}` cold and warm; the `--lsp` smoke reports "handshook" while `touched=0`.

pi-lens dispatch is diagnostics-oriented, so as written `.cue` files would spawn a server, spend the per-write diagnostic budget, and return silence we can't distinguish from clean — while `docs/language-coverage.md` claims `✓`. Could we split this? The formatter half is real, verified value and can land as soon as the flag is fixed. For the LSP half, either hold it until cuelsp publishes diagnostics, or land it explicitly as a non-diagnostic server: a `SERVER_DIAGNOSTIC_STRATEGIES` entry marking it non-publishing, plus coverage-doc wording that says hover/definition rather than diagnostics.

Smaller items:

- `clients/lsp/language.ts` needs `".cue": "cue"` in `LANGUAGE_EXTENSIONS`. Without it every `didOpen` sends `languageId: "plaintext"` (`file-kinds.ts`'s `getLanguageId(kind)` is a different function and doesn't feed `didOpen`).
- `clients/dispatch/runners/lsp.ts` `appliesTo` needs `"cue"`. Right now `getAvailableRunners("/p/a.cue")` returns `[]` where `.clj` returns `["lsp"]`.
- No tests landed. `CONTRIBUTING.md` asks for unit tests under `tests/clients/lsp/` and a formatter test, and the commit titled "CUE: Add tests regarding CueServer behaviors" only re-indents three files. Both blockers above would have been caught by a formatter-command test and an extension→languageId coverage test — if you'd like, add the two coverage guards (every `lspCapable` kind's extensions present in `LANGUAGE_EXTENSIONS`; every `lspCapable` kind present in the lsp runner's `appliesTo`) and they'll protect the next language too.
- `tests/fixtures/tool-smoke/cue/` needs a `cue.mod/module.cue`, the way `clojure/deps.edn` and `gleam/gleam.toml` do — otherwise the module never resolves and the fixture can only pass vacuously.
- The repo is tab-indented; the new blocks in `file-kinds.ts`, `language-policy.ts`, and `server.ts` use spaces, and a couple of pre-existing lines got reindented. `npm run lint` is `tsc` only, so it won't flag this.
- `.changelog/cue-lsp-support.md` mentions "a diagnostic wait strategy", but none was added. Also, please reference issue #1519 in the PR title (`refs #1519`) so the work is linked.
- `docs/features.md` has two hand-maintained lists (LSP servers, formatters) that still need CUE.
- One FYI: `cue lsp` only exists from v0.15.0-alpha.1, and we prefer a PATH `cue` with no min-version gate — so an older PATH `cue` will fail to spawn rather than fall back to the managed install. Not yours to fix; just worth knowing.

Finally, this branch is behind master, and I've just approved the CI workflows so Lint and Unit tests actually execute on the current head. Sonar's duplicated-lines failure is expected for registry-shaped additions; that gets judged at merge.

bcachet:
I was not aware of the limitations of CUE LSP. They are defined in the documentation https://github.com/cue-lang/cue/wiki/LSP:-Getting-started#static-and-partial-evaluation-only but I never read this far the document.

Without LSP support, I am not really sure it makes sense to provide CUE support. I am fine just closing this PR.

If you think having support to format CUE code is interesting to pi-lens, let me know and I will adapt the PR accordingly.

apmantza:
**Correction to point 2 above.** I was wrong, and thanks for the push-back. `cuelsp` does report diagnostics — my fixture was the one class it deliberately excludes. Per the CUE wiki, the language server reports load and parse errors as you type but not evaluation errors: conflicting values and failed constraints are left to `cue vet`. My probe used `a: int & "hello"`, a conflicting value. Re-probed with a syntax error (`a: {`), the server publishes correctly: `severity 1, source "cue", "expected '}', found 'EOF'"`, about 600 ms after `didChange`.

So the honest framing is partial coverage, not no coverage: CUE gets syntax and parse diagnostics plus hover, definition, completion, code actions, and formatting; it does not get validation. That changes the ask. Rather than holding the LSP half, let's land it with wording that matches — `docs/language-coverage.md` should say syntax-only rather than a bare `✓`.

Two things still need attention on that half. First, the diagnostics don't reach us yet: driven through `LSPService.touchFile` with `diagnostics: "document"` on that same syntax fixture, cold and after two real edits, we get `{"diags":[]}` every time, and our latency log calls each one `failureKind: success, inconclusive: false` — so pi-lens reports the file conclusively clean while the server has an error to publish. Two behaviours probably explain it: the server sends an **empty** array on `didOpen` even for an already-broken file, and publishes nothing at all until a change arrives. That is precisely what `scripts/characterize-lsp.mjs` and `scripts/probe-clean-signal.mjs` are for — could you run both and add the resulting `SERVER_DIAGNOSTIC_STRATEGIES` entry? The changelog already claims a wait strategy, so this closes that gap too.

Second, the tool-smoke fixture can't exercise any of this: `bad.cue` uses `a: int = "hello"`, an evaluation error the server won't report, and there's no `cue.mod`. A `cue.mod/module.cue` plus a syntax-error body would make it a real test.

One more note, since it's now the more interesting half: a `cue vet` auxiliary runner would give us the validation coverage the LSP omits, in the same shape as terraform's `lsp, tflint, trivy-config` group. Happy to open a follow-up issue for that — no need to grow this PR.

Lastly, my earlier remark about `languageId` was overstated as well: `cuelsp` keys off the file URI, and a `plaintext` `languageId` produced the identical diagnostic. Adding `".cue": "cue"` to `LANGUAGE_EXTENSIONS` is still correct for consistency, but it isn't a functional blocker.

*(As before: this analysis is AI-generated and maintainer-supervised.)*

apmantza:
many thanks @bcachet. will wait for you to complete this PR as you originally intended.

bcachet:
It seems that you did all the work, thanks a lot
Anything I can help with ?

apmantza:
Thanks again for this, @bcachet — the cuelsp integration you built is the backbone of what's now a complete language addition.

Disclosure first: the commits just pushed to your branch are AI-generated under maintainer supervision, and the maintainer makes the merge decision. Maintainer edits are enabled on the PR, which is why they land here rather than in a companion PR — your authorship on the original commits is untouched.

What was added on top of your work:

- **Tree-sitter grammar**: no published CUE wasm exists anywhere, so the branch now vendors one built from eonpatapon/tree-sitter-cue at commit `dd7b90e0` (MIT), with the build command, source commit, and sha256 recorded in `scripts/grammars.lock.json`. It carries the modern `dylink.0` section and loads under the currently pinned web-tree-sitter.
- **Full registration**: `cue` is now a first-class file kind (extensions, language policy, dispatch plan, LSP seam pin), so `.cue` files get structural analysis alongside your LSP diagnostics.
- **Fixture + parse smoke**: a real-shaped CUE fixture asserting a clean parse, plus a negative case.
- **Two small fixes to the original commits**: `cue fmt -w` is not a valid flag (now `cue fmt $FILE`), and one tool-smoke fixture was passing without actually exercising an evaluation error.

The SonarCloud duplication flag on the new commits is the registry-table shape (the same few-line entries repeated per language table) — noted for the maintainer rather than hidden.

One follow-up deliberately left out: the cuelsp `SERVER_DIAGNOSTIC_STRATEGIES` characterization needs a real `cue` binary to verify against, so it's tracked in #1522 rather than guessed at here.

apmantza:
thanks @bcachet. my apologies - was rushing in for a new release which still hasnt come ;).

Issue #1493, opened 2026-08-16
A silent auxiliary still reads as confirmed clean when another auxiliary answers


apmantza:
#1470 was filed on the belief that a *cut-off* auxiliary was the dishonest case and a *silent* one already read as inconclusive. The second half is false.

## Evidence

Driven through the real touch path: opengrep answers at 400 ms, typos goes silent through its full 1500 ms budget. The auxiliary `Promise.all` settles at ~1500 ms, `timeoutMs` is 2500 (the `LSP_DIAGNOSTICS_WAIT_MS` caller cap), so `waitedMs + 20 >= timeoutMs` is false and the touch resolves:

```
{ confirmation: "confirmed", inconclusive: undefined, diags: [] }
```

typos contributed no evidence about the file and the touch says nothing about that. It is the exact #1470 shape on the sibling outcome.

## Why it was not seen

With a *single* auxiliary, its silence burns the whole touch deadline and does trip `diagnosticsTimedOut`, so the result is honest. With two or more, a fast auxiliary answering early lets the wait settle before the deadline, and the silent one's absence goes unrecorded. Every multi-auxiliary language configuration is exposed.

This predates #1492 and is not introduced by it.

## Related correction

#1492's description and changelog claimed ast-grep, zizmor and typos "already tripped `diagnosticsTimedOut` and read inconclusive", so only opengrep could produce a false clean. That reasoning holds only in the single-auxiliary case and is being corrected on that PR.

The test at `tests/clients/lsp/service-aux-grace.test.ts` covering the silent case asserts `not.toBe("partial")`, which passes while enshrining this behavior. It is being rewritten to assert what is true today and point here, so it documents a known gap rather than implying correctness.

## Direction

`touchCoverageGap()` and `touchCompletedConfirmationPolicy()` in `diagnostic-binding.ts` are already the single source of truth for the cut-off case. The silent case belongs in the same policy: an auxiliary that published nothing inside its own budget has told us nothing, whether or not the deadline was reached.

Take care not to overcorrect. A `silent` auxiliary on a genuinely clean file is the common case, and marking every such touch partial would demote nearly every result and make the signal useless. The distinction that matters is whether the auxiliary was given its declared budget and stayed silent (evidence it ran and found nothing) versus never having reported at all.

## Observability

[...]

Issue #1459, opened 2026-08-15
fix(lsp): cascade sweeps backpressure opengrep into the notify-write breaker; 15s security blackout reads as clean (refs #1458)


apmantza:
## Summary

During cascade sweeps the opengrep LSP stalls on its `didOpen` write. Ninety-two notify writes timed out in the current log window, all attributed to opengrep alone. Three consecutive timeouts trip the backpressure breaker, which marks the server broken for a 15 s cooldown, after which touches skip it outright. For the duration of a sweep the always-on security lane is dark, and nothing records that the files swept were never scanned.

## Evidence

`~/.pi-lens/latency.log` + `latency.log.1`, runtime build `fcdb34ad`.

```
lsp_notify_timeout, by timedOutServerIds
  ["opengrep"]                        92
  ["ast-grep"]                        15
  ["opengrep","ast-grep"]              5
  ["opengrep","typescript"]            3
  ["opengrep","typos","typescript"]    2

lsp_notify_backpressure_broken, by serverId
  opengrep   31
  ast-grep    6
  typescript  1
```

The timeouts arrive from the cascade lane, in bursts against many files at once:

```
{"phase":"lsp_notify_timeout","filePath":".../pi-free/lib/registry.ts","durationMs":2231,
 "metadata":{"source":"cascade","clientScope":"all","serverCount":4,
 "timedOutServerIds":["opengrep"]},"ts":"2026-08-15T12:09:31.587Z"}
```

Four such rows land inside 3 ms (`registry.ts`, `session-start-metrics.ts`, `telemetry.ts`, `util.ts`, 12:09:31.587–.590Z), and the breaker trips on the fourth:

```
{"phase":"lsp_notify_backpressure_broken","filePath":".../pi-free/lib/telemetry.ts",
 "metadata":{"serverId":"opengrep","cooldownMs":15000,"consecutiveTimeouts":3},
 "ts":"2026-08-15T12:09:31.589Z"}
```

Subsequent touches then skip the server:

```
{"phase":"lsp_client_skipped_broken","filePath":".../pi-free/index.ts",
 "metadata":{"serverId":"opengrep","retryInMs":1298},"ts":"2026-08-15T19:59:22.613Z"}
```

Shutdown shows the same unresponsiveness from the other end — every opengrep `lsp_client_shutdown` row in the window carries `shutdownRequestTimedOut: true`, for example `2026-08-15T21:07:20.352Z`, `pid 43296`, `durationMs 1006`.

[...]

Issue #1253, opened 2026-08-10
bug(lsp): Marksman silent-clean confirmation is lost before lsp_diagnostics


white-hat:
## Defect

`LSPService.touchFile` can affirmatively confirm a clean Markdown file after Marksman's expected silent-on-clean wait, but `lsp_diagnostics` drops that positive confirmation at the tool boundary. The tool sees only an empty diagnostics array plus `inconclusive: false`, reclassifies Marksman as tier-3 silent, attempts the TypeScript-only fallback, and renders the file as inconclusive.

This reproduces in the installed 3.8.74 bundle and repository HEAD `ff962642f21abcaa2864a7e1b4704edd3e02a445` through a fresh Pi session using the real cached Marksman binary:

- telemetry: `lsp_silent_clean_confirm`, `collectedDiagnostics: 0`, `diagnosticsTimedOut: false`, `inconclusive: false`
- tool result: `clean=0`, `inconclusive=1`

## Impact

No Marksman finding is ever hidden by this and nothing is reported clean that isn't — the diagnostics themselves are correct. What is lost is the *confirmation*, and three behaviors hang off it:

**1. Every clean Markdown file reads as an unverified check.** The rendered result is `0 files confirmed clean, 1 unconfirmed: from a server that cannot confirm clean`, which an agent reads as "the check didn't work" — prompting a retry with a larger `waitMs`, or a fallback to `lens_diagnostics`, on a file that was already known clean. A four-file batch reports `inconclusive=4`. This is the visible symptom, and it is the cheapest of the three.

**2. The per-file diagnostics cache never admits Markdown.** `collectFileDiagnosticResult` records into the workspace-diagnostics cache only when `confirmation !== "unconfirmed"` (the #671 false-clean discipline). Because clean Markdown is never confirmed, no Markdown result is ever cacheable — so every sweep re-touches every Markdown file and pays Marksman's full wait budget again, indefinitely. The #671/#672 caching work simply does not apply to Markdown today.

**3. Stale Markdown diagnostics can't be cleared.** `reconcileWidgetFromLspResult` refuses to write an unconfirmed empty result into the footer cache, precisely so a silent server's empty push can't erase real state. The side effect is that an `lsp_diagnostics` check proving a Markdown footer/inline error is *gone* cannot correct it — which is the exact scenario #571 added reconciliation for.

So the severity is: correct findings, wrong confidence, plus a permanent cache miss and a stuck-stale-state path for one whole language.

[...]

PR #1528, opened 2026-08-17
fix(security): bound scanner resyncs so a sweep cannot black out opengrep, and name the coverage gap (refs #1459)

apmantza:
## What changed

A cascade sweep fanned a full-text `didOpen` resync at opengrep for every neighbour inside a few milliseconds. Opengrep is the only server configured `reopenOnResync: true`, so each of those is a full re-parse and re-scan. Its stdin stopped draining, three per-server notify-write deadlines expired, and the #743 breaker opened for 15 s. Every touch inside that window skipped the scanner and still resolved `confirmation: "confirmed"` — the always-on security lane went dark and the silence read as scanned-clean.

Changes in `clients/lsp/index.ts`, plus the agent-facing cascade surface:

- **One outstanding resync per auxiliary, as a bounded queue.** A healthy scanner accepts a `didOpen` in milliseconds, so a sweep's neighbours take their turns and every file still gets scanned — what the gate prevents is N simultaneous full re-scans flooding one stdin. The slot is claimed SYNCHRONOUSLY (no `await` between the check and the insert) and the caller gets a release handle; a version that returned "the slot looks free, go write" was not a gate at all, because every waiter woke in one microtask batch and all wrote at once. The queue wait is bounded by the caller's own budget (`maxClientWaitMs`), not a flat write budget, so a caller that asked for 1000ms in total is not taxed a further 2000ms. The gate is keyed by `serverId:normalizedRoot` and holds the client identity, so a record left by a previous client generation can never starve a fresh one.
- **A late landing retracts its timeout.** The write deadline bounds caller latency; it does not diagnose health. A write that lands after the budget but within the wedge window retracts the strike it was charged for. A landing past the wedge window keeps it — at that point the stall is long enough that #743's demotion is the honest verdict. Retraction is generation-checked, so a predecessor's late landing cannot mask its successor's stall.
- **A wedged write still demotes.** A write nothing accepts for five write budgets is a dead input path. The check is armed as an unref'd timer when the slot is taken, not left to the next waiter: inside a burst every waiter arrives within one budget, so a waiter-side check could never see the wedge window elapse. It goes through the existing breaker teardown, so the gate cannot defer a wedged server forever.
- **A deferred server is neither waited on nor read from, and is labelled honestly.** It records a distinct `deferred` outcome in `lsp_aux_wait_outcome` instead of occupying the `silent` row, which stays reserved for a scanner that HAD the content and published nothing (#1493). The aggregate wait deadline ignores it too.
- **The wait exclusion, in detail.** Its version can never advance, so waiting only burns its budget and would flip the touch to `inconclusive`, discarding a primary answer that IS trustworthy. Its diagnostics cache also still holds the PREVIOUS content's findings, because the resync that would have cleared it never ran — so its diagnostics and its binding are both excluded from the merge.
- **The gap is named on the result, and it reaches the agent.** A scanner that never attached (breaker open) or never received the content (deferred resync) joins `unconfirmedServerIds` and narrows the touch to `confirmation: "partial"` — the #1470 seam every coverage consumer already fails closed on. `CascadeNeighborResult` carries the same list, and the cascade formatter renders an explicit "scanners did not cover N file(s) — no findings does NOT mean clean here" section. Without that last step the neighbour still reached the agent as zero diagnostics with no marker.

Primaries are untouched: they serve one file per touch and are not the fan-out target a `clientScope: "all"` sweep floods.

## Why

Defect shape 10 (silencing counted as fixing): an empty scanner result has to distinguish clean from unavailable. It did not. The breaker itself was doing its job — the defect was that the sweep's own fan-out drove a healthy scanner into it, and the resulting coverage gap left no trace on the result.

## Observability

Three new `latency.log` records, all excluded from last-phase stall attribution (they are summary rows, and two carry another write's age as `durationMs`):

- `lsp_scanner_coverage_gap` — one row per touch that a scanner did not cover, with `brokenSkippedServerIds` / `deferredResyncServerIds`. **This is the record that proves a blackout is no longer read as clean.**
- `lsp_notify_resync_deferred` — the gate absorbing fan-out, with the outstanding write's age.
- `lsp_notify_write_late_landed` — a retracted timeout, with the streak after retraction.

`lsp_touch_file` metadata carries the same two id lists, so a gap joins to the touch that produced it.

## Concurrency evidence

The gate's invariant is max concurrency, not call count — N writes issued serially and N issued simultaneously are the same count. Six concurrent `clientScope: "all"` touches, 50 ms aux write, 100 ms write budget:

```
pre-fix (8851d9f7): {"auxOpenCalls":6,"maxInFlight":5,"openOffsets":[0,50,50,50,50,50]}
fixed  (c605a64d): {"auxOpenCalls":2,"maxInFlight":1,"openOffsets":[0,50]}
```

One write at a time, throughout. `auxOpenCalls: 2` is this probe's deliberately harsh ratio — a write taking half the budget leaves room for two turns before the queue deadline, and the rest are reported as uncovered rather than pushed. With a scanner that accepts writes in ~10 ms, all six touches are written and every one stays `confirmed` (pinned by the suite's concurrency test).

## Verification

Red-first regression test: `tests/clients/lsp/service-scanner-coverage-gap.test.ts` (10 tests). Five are proven RED against `origin/master` (fix reverted, rebuilt, same test file), and the three concurrency/budget tests are proven RED against the intermediate commit 8851d9f7 (`expected 5 to be 1`, plus two missing rows). One test is a negative control and passes on both trees by design — a healthy scanner still receives every file and every touch stays `confirmed`, so the other five cannot be satisfied by an implementation that simply stopped scanning.

```
× a burst of resyncs issues one write, defers the rest, and never trips the breaker
  AssertionError: expected "vi.fn()" to be called 1 times, but got 4 times
× a write that lands after its deadline retracts the timeout it was charged for
  AssertionError: expected [] to have a length of 3 but got +0
× a write nothing accepts for the wedge window still demotes the server
× a scanner skipped for an open breaker makes the touch partial, not confirmed
  AssertionError: expected 'confirmed' to be 'partial'
× a deferred scanner's stale findings are not merged as this touch's answer
Tests  5 failed | 1 passed (6)
```

Against 8851d9f7 (the queue-without-atomic-claim round):

```
× a healthy scanner gets every file, one resync at a time, still confirmed
  AssertionError: expected 5 to be 1
× a wedged scanner is demoted by its own write's wedge timer, without a later touch
× a queued resync never waits longer than the caller's own budget
Tests  3 failed | 5 passed (8)
```

Green after the fix, together with every suite that references the seams touched (`touchFile`, `confirmation`, `unconfirmedServerIds`, the notify budget, latency phase exclusion):

- `tests/clients/lsp/**` + `cascade-format`, `cascade-turn-merge`, `cascade-compute`, `tests/clients/cascade/**`, `latency-logger`, `warm-attach-confirmation` — 783 passed, 7 skipped (85 files).
- `tests/clients/dispatch/**`, `tests/tools/**`, `runtime-turn`, `index-integration`, `index-wiring` — 1171 passed (115 files).
- `pipeline-lsp-sync`, `mcp/ipc`, `runtime-session-warm-skip-notify`, `index-loop-block-wiring`, `smells-rollup`, `project-conventions` — green.
- `npm run build`, `npm run lint` (tsc), `npm run changelog:check` clean.

Full suite deferred to CI (concurrent agents share this checkout).

Two adversarial review rounds ran on this branch. The first returned eight findings; the second commit answers seven, and the verification round found the eighth incompletely fixed — the queue reintroduced the flood it was meant to stop (numbers above). The third commit fixes that plus the flat-tax queue wait, the waiter-side wedge check, and the `silent` mislabel (the queue instead of a drop, the deferred server's wait and stale-cache exclusion, the agent-facing gap, generation-checked and wedge-bounded retraction, and the minors). One finding is deliberately out of scope and documented in code: `brokenSkippedAuxiliaryServerIds` covers the BREAKER doors only, not a temporarily-unavailable command, `optionalDisabled`, a spawn failure, or capacity eviction — those are persistent-absence states, and flagging every touch partial for them is a broader behavior change than #1459 is scoped to make.

One deliberate noise trade-off: while a scanner is slow enough to be deferred, a cascade run that would previously have ended `skipReason: "clean"` now produces a "scanners did not cover" advisory. That is the honest reading and the whole point of the change, but it can fire often until the deferral rate settles — worth watching in the same dogfood window as criterion 2 below, and tightening (rate-limit, or fold into a single per-turn line) if it proves chatty.

## Polish round (8b8129f6)

The verification review returned merge-ready with three small items, all applied:

- **The generation guard exempted the case its own comment named.** `current !== undefined && current !== entry.client` read `undefined` for an EVICTED client — eviction deletes the registry entry — so a queued waiter was waved through to a retired client whose write resolves, and `markTouched` would then record the content as delivered (#1253 laundering, reached through the queue). Now `current !== entry.client`.
- **Its regression test needed a second pass to be worth anything.** The first fixture let the holder's write outlast the queue budget, so the waiter timed out before reaching the claim and the test passed against the broken guard — shape 7, caught by mutation. With the holder settling inside the budget, the broken guard issues a visible second write to the corpse (`expected "vi.fn()" to be called 1 times, but got 2 times`).
- **The `deferred` aux-wait outcome is now pinned** (`expected [ 'silent', 'cut_off' ] to include 'deferred'` on mutation), and a vacuous assertion is gone from the caller-cap test.

## What remains — why `refs`, not `closes`

Three of the four acceptance criteria are met in code and pinned by tests. The fourth is empirical: "`lsp_notify_backpressure_broken` for opengrep drops to zero across a comparable dogfood window." That needs a dogfood window after this ships, so the issue stays open for that confirmation.

Refs #1459

https://claude.ai/code/session_01EpSQeAhgjMTDskzyVMmjFH


apmantza:
CI note: the four `Analyze (…)` CodeQL checks on c605a64d failed on a GitHub-side outage, not on this code — `Encountered an error while trying to determine feature enablement: HttpError: No server is currently available` during Initialize CodeQL. The same jobs passed on 8851d9f7, and the outage also 503'd my own API calls while pushing. `Unit tests`, `Lint & type-check`, SonarCloud, and the install tests all pass on this SHA. The run cannot be retried; a fresh push or a maintainer re-run once the outage clears should clear them.

apmantza:
CI note (8b8129f6): the GitHub outage is still running and hit three CodeQL analyzers again — `Analyze (go)`, `Analyze (rust)`, `Analyze (javascript-typescript)` all died in Initialize CodeQL with `Encountered an error while trying to determine feature enablement: HttpError: No server is currently available`. `Analyze (ruby)`, `Analyze (python)`, and `Analyze (actions)` passed on the same commit, which is what a random infra failure looks like. `Unit tests`, `Lint & type-check`, `Production install build`, and SonarCloud all pass on this SHA. Needs a maintainer re-run of the failed jobs once the outage clears — do not merge on the strength of the absent analyzers.

apmantza:
Merged `origin/master` (through #1527's `9f0f577b`) into this branch as `2856f592`. Two conflicts, both in the seams #1527 touched, resolved semantically rather than textually:

- `clients/lsp/index.ts` — the confirmation-narrowing union now reads #1527's `auxUnconfirmedServerIds` PLUS this branch's two pre-wait doors (breaker-skipped, gate-deferred). They cannot fold into `auxiliaryCoverageGap`, because aux wait outcomes exist only on the `with-auxiliary` path and the sweep that causes deferrals runs `clientScope: "all"`, which emits no outcome rows at all. Where the two DO meet, #1527's evidence wins: a deferred auxiliary whose stored publication is bound to exactly this content stays covered and its findings still reach `.diags` via the carried-auxiliary path. `lsp_touch_file` keeps both metadata field sets.
- `clients/lsp/diagnostic-binding.ts` — `deferred` joins the `AuxiliaryWaitOutcome` union. `auxiliaryCoverageGap` needed no change: it is written as `outcome !== "answered"`, so a new no-answer shape fails closed by default.

The changelog entry moved to `.changelog/`, the per-entry convention this branch predates.

Verification on the merge head: `tests/clients/lsp/**` 715 passed / 7 skipped, `tests/clients/dispatch/**` + `tests/tools/**` 1122 passed, cascade + mcp + index-integration 171 passed, plus the #1527 regression net (`service-aux-grace`) green. One new test pins the integration point and was proven red by mutation (`expected [ 'opengrep' ] to not include 'opengrep'`). `Unit tests` and `Lint & type-check` both genuinely RAN and passed on `2856f592` — checked explicitly for the skipped-CI-on-conflict trap.

Three CodeQL analyzers are still failing on the ongoing GitHub outage (`Encountered an error while trying to determine feature enablement: HttpError: No server is currently available` in Initialize CodeQL); `go`, `ruby`, and `actions` passed on the same commit. Those need a maintainer re-run once the outage clears.

PR #1587, opened 2026-08-18
fix(dispatch): derive the lsp runner and languageId seams from language policy (closes #1545)

### LINKED ISSUE #1545: fish is lspCapable but missing from the lsp runner's appliesTo; close the forgotten-seam class
Found during the adversarial review of PR #1520 (AI-generated, maintainer-supervised), which added CUE and hit the same seams.

`clients/dispatch/runners/lsp.ts:81-114` — the lsp runner's `appliesTo` list omits `fish`, even though fish is `lspCapable`. Verified: `getAvailableRunners("/p/a.fish")` returns `[]` where `.clj` returns `["lsp"]`. Main dispatch resolves group runner ids directly, so routing survives; the reporting surface and `RunnerRegistry.getForKind` see no runners for fish. The user-visible effect is that fish reads as having no runner coverage anywhere the registry is consulted.

## The class

The defect shape is "language registered `lspCapable` but absent from a hand-maintained seam list". CUE (in #1520) missed two seams at once: this `appliesTo` list and `LANGUAGE_EXTENSIONS` in `clients/lsp/language.ts` (which feeds `didOpen`'s `languageId`). Nothing structurally forces either seam, so every future language is one forgotten list away from the same silent gap.

## Acceptance criteria

- Fish appears in (or is derived into) the lsp runner's `appliesTo`; `getAvailableRunners("/p/a.fish")` returns `["lsp"]`, regression test proven red on current code.
- Sweep: audit every `lspCapable` kind against BOTH seams (`appliesTo`, `LANGUAGE_EXTENSIONS`) and fix any other member found; record the sweep result in the PR.
- The class is closed structurally, per the single-source-of-truth rule: prefer deriving both seams from the language registry over adding entries plus guards; where a hand-maintained list must stay, add the two coverage guards the #1520 review proposed (every `lspCapable` kind's extensions present in `LANGUAGE_EXTENSIONS`; every `lspCapable` kind present in `appliesTo`), each proven red against a synthetic missing entry.

## Observability

None needed beyond the guards — the gap is a build-time list omission; the coverage tests are the record.

Refs #1520, refs #1519.

AngriestBird:
Closes #1545. Refs #1519, refs #1520.

## The defect

`LANGUAGE_POLICY` decides which kinds are `lspCapable`. Two seams downstream kept their own hand-written transcription of that list, and nothing forced either one, so a language could be registered and still miss both silently. Both had drifted.

**Seam 1 — `clients/dispatch/runners/lsp.ts` `appliesTo`.** It omitted `fish`, so `RunnerRegistry.getForKind("fish", …)` answered that no LSP runner applies to fish.

Scoping that honestly (corrected in review round 1): routing was never affected, because main dispatch resolves group runner ids directly and `PRIMARY_DISPATCH_GROUPS.fish` names `lsp`. `getForKind` has exactly two callers — `getAvailableRunners`, which is referenced only from tests, and the `@internal` `dispatchLint`, which has no production caller (production goes through `dispatchLintWithResult`, and `runGroup` resolves by id, never by `appliesTo`). So the user-visible blast radius today is nil. What was actually broken is that the registry gave a wrong answer about itself, and the next surface to ask it would have inherited that. The issue's acceptance criterion is the observable form of it: `getAvailableRunners("/p/a.fish")` now leads with `lsp` instead of `fish-indent`.

**Seam 2 — `clients/lsp/language.ts` `LANGUAGE_EXTENSIONS`.** `didOpen` sends `getLanguageId(filePath) ?? "plaintext"`, and 21 extensions across three kinds had no entry, so they were announced as plaintext.

Also corrected in review round 1: the original wording here claimed those files were "opened and then never diagnosed". The reviewer drove real servers over raw LSP and that is not true for most of the list. vscode-json-language-server and yaml-language-server return **identical** diagnostics under `languageId=plaintext` (7/7 and 3/3), and clangd derives the language from the file path, so the 17 cxx extensions are very likely no-ops as well. A `DocumentSelector`-based server such as PowerShell Editor Services is the case where the announced id plausibly decides whether the document is handled at all, and that one is unverified here. The table is more correct after this change and correct-by-construction rather than by transcription; the "never diagnosed" impact claim was not earned and has been removed from the changelog fragment and this body.

## Fix: derive both seams

`appliesTo` is now `getLspCapableKinds()`. `LANGUAGE_EXTENSIONS` is the curated per-extension table plus a derived fill-in over `KIND_EXTENSIONS` for every `lspCapable` kind.

The curated table stays deliberately. One `FileKind` can span several LSP language ids (`jsts` covers `typescript`/`typescriptreact`/`vue`/`svelte`; `cxx` covers `c` and `cpp`), and some servers want a spelling the registry does not use (`shellscript`, not the registry's `shell`). The fill-in uses `??=`, so a curated id always wins and only genuine gaps are filled. Registering a language as `lspCapable` is now the only step either seam needs.

## Sweep

Searched: every `lspCapable` kind from `LANGUAGE_POLICY` (34 of 37 `FileKind`s; `sql`, `helm-template`, `terragrunt` are not lspCapable), against both seams, extension by extension. The guard tests are the sweep, so it re-runs on every CI job rather than being a one-time claim.

**`appliesTo` — 1 kind missing.** `fish`, exactly as the issue reported. No non-lspCapable kind was falsely claimed.

**`LANGUAGE_EXTENSIONS` — 21 extensions missing across 3 kinds:**

| Kind | Missing |
| --- | --- |
| `cxx` (17 of 24 registered) | `.c++` `.cp` `.hxx` `.inl` `.ipp` `.tpp` `.txx` `.c++m` `.cppm` `.cxxm` `.ixx` `.cu` `.hip` `.m` `.mm` `.cl` `.clcpp` |
| `powershell` (3 of 3 — no entry at all) | `.ps1` `.psm1` `.psd1` |
| `json` (1 of 3) | `.json5` |

PowerShell is the one where the missing entry most plausibly mattered: it had no `LANGUAGE_EXTENSIONS` entry whatsoever, and PowerShell Editor Services selects documents by `DocumentSelector`.

**Deliberately left:**

- `LANGUAGE_EXTENSIONS` entries with no `FileKind` behind them (`.scala`, `.r`, `.jl`, `.pl`, `.erl`, `.f90`, `.cob`, `.pas`, `.adb`, `.vhd`, `.v`, `.sv`, `.graphql`, `.proto`, `.typ`, `.astro`, `.ron`, `.mod`, `.sum`). Harmless, and removing them would drop working language ids for files pi-lens does not classify. The derivation only adds.
- `.sql` stays mapped even though `sql` is not `lspCapable`. Same reason.
- `.m` and `.mm` derive to `cpp` rather than `objective-c`/`objective-cpp`, because `cxx`'s registry language id is `cpp`. clangd derives the language from the path anyway, so this is cosmetic; a finer id belongs in the curated table if anyone wants it, which is a one-line addition now.
- `clients/lsp/client.ts:2145` hardcodes `plaintext` on the `didChange`-before-`didOpen` path. Pre-existing, adjacent, out of scope for this issue.

## Guards

`tests/clients/lsp-capable-seam-coverage.test.ts`, six structural guards:

1. `appliesTo` is non-empty. Not decorative: `getForKind` reads an empty `appliesTo` as "every kind", so a derivation resolving to nothing would invert into running the lsp runner on `sql` and `terragrunt`.
2. Every `lspCapable` kind is in `appliesTo`.
3. No kind in `appliesTo` that policy says is not `lspCapable`.
4. Every `lspCapable` extension is in `LANGUAGE_EXTENSIONS`.
5. Every `lspCapable` extension resolves through `getLanguageId` to something other than `"plaintext"` — that is the value the `didOpen` call sites substitute, so the guard sits on the seam rather than on the table.
6. **(added in review round 1)** The set of language ids reachable from `lspCapable` extensions is pinned by hand. Guards 4 and 5 cover presence, not spelling, and the derivation fills gaps from a pi-lens-internal label; whether a server answers to that label is a separate fact about the server. A new or renamed id now needs a deliberate addition to the pinned list. The existing entries are a snapshot, not a verified-good set — see review round 2.

Plus two behavioral pins: the sweep's newly covered extensions resolve (`.ps1`→`powershell`, `.cu`/`.ipp`→`cpp`, `.json5`→`json`), and the curated ids still beat the per-kind id (`.tsx`→`typescriptreact`, `.sh`→`shellscript`, `.c`→`c`, `.scss`→`scss`).

`tests/clients/language-policy.test.ts:26` already fails a kind with no registered primary server at all. It says nothing about the id the document is announced under, so guard 6 does not duplicate it.

## Red-first evidence

**Round 1 — against pre-fix `origin/master` (2e44576b)**, build first, 5 failures:

```
× lists every lspCapable kind in the lsp runner's appliesTo
  → lspCapable kind(s) absent from the lsp runner's appliesTo: fish
× maps every lspCapable kind's extensions to an LSP language id
  → .c++, .cp, .hxx, .inl, .ipp, .tpp, .txx, .c++m, .cppm, .cxxm, .ixx,
    .cu, .hip, .m, .mm, .cl, .clcpp, .json5, .ps1, .psm1, .psd1
× resolves a language id for every lspCapable extension
  → same 21 extensions
× keeps the sweep's newly covered extensions resolvable
  → expected undefined to be 'powershell'
× getAvailableRunners > returns the lsp runner for fish files
  → expected 'fish-indent' to be 'lsp'
```

**Round 1 — guard mutation**, to show the guards catch a re-hand-maintained list rather than passing by construction. Mutated `appliesTo` to `getLspCapableKinds().filter((k) => k !== "clojure")` and made the fill-in loop skip `.ps1`, rebuilt:

```
× lists every lspCapable kind ... → clojure
× maps every lspCapable kind's extensions ... → .ps1
× resolves a language id for every lspCapable extension → .ps1
× keeps the sweep's newly covered extensions resolvable
  → expected undefined to be 'powershell'
```

**Round 2 — the reviewer's probe, reproduced before fixing.** Added a synthetic `probelang` `FileKind` (`.probe`, registry languageId `"Probe-Lang-Script"`, `lspCapable: true`), rebuilt, ran the seam-coverage file:

```
Tests  7 passed (7)
```

Confirmed: every existing guard green. Then added guard 6 against the same mutated tree:

```
× announces lspCapable files only under an accepted language id
  - Expected  + Received
  +   "Probe-Lang-Script",
      "c", "clojure", "cmake", "cpp", "csharp", ...
Tests  1 failed | 7 passed (8)
```

Guard 6 fires alone; the other seven stay green, which is precisely the gap it exists to close. The same mutation was also run against `tests/clients/language-policy.test.ts` to check for overlap — it fails there on "no registered primary server", a different and weaker signal, confirming the two guards are complementary rather than redundant. Mutation reverted across `clients/file-kinds.ts`, `clients/language-policy.ts`, `clients/dispatch/plan.ts`; `git status` clean afterwards.

## Review round 1

Four findings, all applied on this branch. No blocking defect was reported.

| # | Finding | Resolution |
| --- | --- | --- |
| 1 | "opened and then never diagnosed" is unproven for most of the sweep and false for json/yaml (measured over raw LSP) | Softened in the changelog fragment, this PR body, the `LANGUAGE_EXTENSIONS` doc comment, and the guard-5 comment. Impact is now stated per server class, not as a blanket claim. |
| 2 | Guard 5 checked presence, not spelling; a `"Probe-Lang-Script"` kind passed 7/7 | Guard 6 added, pinning the reachable id set. Probe reproduced green first, then proven red against the tightened guard (above). |
| 3 | "every reporting surface read fish as having no coverage" overstates blast radius — `getForKind` has two callers, neither reachable from production | Rewritten in this body and the changelog fragment: latent wrong answer in the registry, not lost coverage. |
| 4 | Fragment said "Four coverage guards" | Now "six structural guards plus two behavioral pins". |

Out of scope by instruction and left alone: `clients/lsp/client.ts:2145`.

## Review round 2

Comment honesty only, no behavior change (head `34d2021f`). Guard 6's header claimed each pinned id is "a spelling some real server answers to". Falsified on `"sass"`: the only registered CSS server ignores `languageId` and parses indented Sass as CSS, emitting a bogus `{ expected`. That id — and every other entry inherited from the curated table — was grandfathered in unchecked, so the claim was never earned. The comment now states what the pin actually buys: the list is a snapshot of what is reachable today, not a verified-good set, and its value is that a **new or renamed** id cannot appear silently. The constant is renamed `PINNED_LSP_LANGUAGE_IDS` so the name stops implying acceptance the list never established, and `vscode-json-language-server` is dropped from the `shellscript` example (that server has nothing to do with the id; bash-language-server is the one that wants `shellscript` over the registry's `shell`).

## Tests

`npm run build` before every run.

Round 3 (head `34d2021f`, comments only): seam coverage + language-policy + dispatch integration + lsp language, 33 passed; `npm run lint` ok.

Round 2 (head `9c2fd1d5`):

- Seam coverage + language-policy + dispatch integration + lsp language: 33 passed.
- `tests/clients/lsp/**`, `tests/clients/dispatch/runners/**`, extension-registry siblings (`extension-predicate-coverage`, `scannable-extension-coverage`, `file-kinds`, `dispatch/plan`): 133 files passed, 5 skipped; 1289 tests passed, 12 skipped.
- `npm run lint`: ok.

Round 1, retained: every test file referencing a changed symbol (`lspRunner`, `LANGUAGE_EXTENSIONS`, `getLanguageId`, `getAvailableRunners`, `getLspCapableKinds`, and the `didOpen`/`languageId` consumers) across `tests/clients/lsp/**`, `tests/clients/dispatch/runners/**`, `tests/clients/review-graph/**`, `pipeline-lsp-sync`, `cascade-compute`, `read-expansion` — 156 files passed, 7 skipped; 1464 tests passed, 14 skipped.

## Merge order

This is the structural half of what #1520 (CUE) tripped over, and the two compose in either order.

- **No file overlap with #1520.** It touches `clients/language-policy.ts` and `clients/file-kinds.ts`; this PR touches `clients/dispatch/runners/lsp.ts` and `clients/lsp/language.ts`. Neither of the seams here appears in its diff, which is the bug the review found.
- **This one first is the better order.** #1520 then needs no hand-additions to either seam: `cue: { lspCapable: true }` plus `cue: [".cue"]` in `KIND_EXTENSIONS` reaches both automatically. It will need **one line** added to guard 6's pinned id list (`"cue"`), which is the deliberate acknowledgment that guard exists to force — a reviewer confirms `cue lsp serve` answers to `languageId: "cue"` rather than it being assumed. If #1520 lands first instead, nothing conflicts and the same one line is added here at merge.
- **Budget the one-line edit whichever way round it goes.** Guard 6 pins the reachable id set, and `.cue` is not in the curated table, so it derives a new id. Whichever of #1587/#1520 lands **second** goes red on `announces lspCapable files only under a pinned language id` until `"cue"` is added to `PINNED_LSP_LANGUAGE_IDS` in `tests/clients/lsp-capable-seam-coverage.test.ts`. That is by design, not a conflict: a new language id is exactly the thing the guard refuses to let through unreviewed. Nothing else in either diff collides.
- Checked all 11 open PRs for overlap. `clients/dispatch/runners/lsp.ts` and `clients/lsp/language.ts` are touched by none of them. #1575 touches `clients/dispatch/integration.ts` (read-only here). #1580 touches nine files under `tests/clients/dispatch/runners/` but not `tests/clients/dispatch/integration.test.ts`, where this PR's one test edit lands.


apmantza (approved):
Thanks for this, Kenneth — and for the two review rounds. Deriving both seams instead of adding `fish` to a list is the right shape, and walking back the "never diagnosed" claim after actually driving the servers is the kind of correction that makes the rest of the PR trustworthy.

A note on process: this review is AI-generated, run under maintainer supervision. I read it before posting and I take responsibility for what it says.

I re-ran the attack independently and it holds up.

- The table change is strictly additive: 106 → 127 entries, 21 added, **zero removed, zero changed**. `appliesTo` gained `fish` and nothing else. No language silently lost coverage.
- The guards are not vacuous. I mutated each one: `appliesTo: []` fails guards 1 and 2; `[...getLspCapableKinds(), "sql"]` fails guard 3 alone; flipping `??=` to `=` in `clients/lsp/language.ts:229` fails the curated-wins pin with `expected 'typescript' to be 'typescriptreact'`. Guards 4-6 I proved red by reverting the two source files to master.
- I also simulated #1520 landing: registering a `cue` kind with `lspCapable: true` and `.cue` in `KIND_EXTENSIONS` reached both seams with no seam edits, and guard 6 fired alone. Your merge-order note is exactly right.
- `tsc` is the forcing function for the whole registry — adding a bare `"cue"` to `FileKind` fails to compile in five places, so a new language cannot skip any of the tables.

One follow-up, not a change request for this PR. `clients/file-kinds.ts:326-328` classifies three files by *name* rather than extension, and two of them still reach `didOpen` as plaintext: `Makefile` (kind `shell`) and `Dockerfile.dev` (kind `docker`) both return `undefined` from `getLanguageId`. `CMakeLists.txt` only works because the curated table happens to carry it as a literal basename key. Guards 4-6 iterate `KIND_EXTENSIONS`, so that half of the classifier is invisible to them. This is pre-existing and your change neither causes nor worsens it — I'll open a separate issue for the basename half of the seam rather than grow this PR.

Merge-ready from my side.

Issue #1561, opened 2026-08-17
Inline blockers re-serve a resolved cross-file finding for the rest of the session — no content or dependency staleness key (refs #1461)


apmantza:
## Summary

A live dogfood session (pi-free, 2026-08-17 20:52–20:57) re-served the same
resolved blocker to the agent on **six consecutive turn ends**, including three
after pi-lens itself had computed a content-bound `confirmed clean` verdict for
the same file. The agent spent three turns debunking it.

Agent transcript, `~/.pi/agent/sessions/--C--Users-apman-OneDrive-Desktop-pi-free--/2026-08-17T20-30-06-524Z_01a0116a-cabc-7b1e-82f6-7b11b26b013d.jsonl`
(record 105, `2026-08-17T20:57:05.550Z`):

> That pi-lens finding is stale — recorded when I first added the tests (before
> the `AsyncIterable` signature fix). Fresh LSP confirms clean and vitest
> passed.

and in the same record's thinking, quoting the injected text verbatim:

> the automated check says "Unresolved from this turn — L320…" — those exact
> line numbers were from the pre-fix state

## The store

`RuntimeCoordinator._pendingInlineBlockers` (`clients/runtime-coordinator.ts:147`,
written at `:648`, read at `:693`). Rendered to the agent at
`clients/runtime-turn.ts:339-345`:

```ts
const unresolvedBlockers = runtime.getInlineBlockersSnapshot();
for (const { filePath: bPath, summary } of unresolvedBlockers) {
	blockerParts.push(`Unresolved from this turn — ${displayPath}:\n${summary}`);
}
```

`"Unresolved from this turn — "` occurs exactly once in the tree, so the
injected text identifies the store unambiguously. The same map also drives the
`--lens-guard` commit gate (`runtime-coordinator.ts:226-247`,
`git-guard.ts:560`).

## Timeline (from `~/.pi-lens/latency.log` + `sessionstart.log`)

| Time | Event |
|---|---|
| 20:52:16.001 | `dispatch: running lint tools for …/tests/tokenrouter-provider.test.ts` |
| 20:52:17.681 | dispatch result: `"totalDiagnostics":3,"blockers":3` → 3 blockers recorded for the **test** file (errors at L320/L349, cross-file type errors caused by `createAttempt`'s signature in `providers/tokenrouter/tokenrouter.ts`) |
| 20:52:18.355 | `turn_end: 1 blocker section(s) … persisting for next context` |
[...]

Issue #1414, opened 2026-08-14
Mid-turn autofix mutates files between an agent's read and its next edit (content-drift + semantic breakage)


apmantza:
## Summary

While editing in `apmantza/pi-free`, pi-lens autofix (biome) ran **between** agent tool calls and changed files the agent was actively editing, in two distinct ways:

1. **Benign re-wrap**: line-wrapping reformats landed in files right after edits, causing the next `edit` call to fail with 'File was modified by auto-format/fix … editing from memory will produce mismatches' and forcing a re-read/rebuild of `oldText`.
2. **Semantic breakage**: after an edit introduced a not-yet-used variable, autofix converted `let _startupRuns = 0;` to `const`. The very next (already-planned) edit then failed with `Cannot assign to '_startupRuns' because it is a constant` — the autofix actively broke a multi-step edit sequence that was valid as a whole.

## Impact

- Extra round-trips (re-read, rebuild oldText) on every interleaved autofix.
- In case 2, a transiently-unused-but-about-to-be-used declaration was 'fixed' into a compile error mid-sequence; an agent without the full plan in context could abandon or mis-fix the change.

## Suggested direction

- Defer non-whitespace autofixes to turn end (agent_end), or batch them after the agent signals it is done with a file — mid-turn only pure formatting seems safe, and even that should ideally be quiesced while an edit tool has the file open.
- Alternatively, suppress lint 'fixes' for symbols introduced within the current turn (they may be wired up by a subsequent edit).

Related but distinct: #673 (deferred agent_end formatting invalidates candidate trees), #195 (per-language policies).

apmantza:
Triage: labeled bug + area:dispatch. From the 2026-08-14 dogfood sessions in pi-free. Two failure modes confirmed in-session: benign re-wrap forcing oldText rebuilds, and a let→const autofix that broke a valid multi-step edit sequence mid-turn. Research dispatched: mapping where mid-turn autofix is triggered in the dispatch pipeline, what quiescence/turn-end hooks exist (relates to #673's agent_end deferral and #195 per-language policies), and candidate designs (defer non-whitespace fixes to turn end vs. suppress fixes on symbols introduced this turn). Findings will be posted here.

apmantza:
## Research findings (codex gpt-5.6-luna, code-level trace)

[...]

Issue #492, opened 2026-07-10
Cross-process nudge: shared touched-files record — parent↔child subagents see each other's autoformats


apmantza:
## Why

#485 delivers the inline "these files were autoformatted by pi-lens" nudge over `pi.events` — an **in-process** bus. That covers every run and turn of the same pi session, but nothing crosses a process boundary, and subagents are mainstream (the nicobailon `pi-subagents` model spawns real child `pi` processes; >50k downloads/mo). Both directions lose the feed:

- **Child blind to parent** (the original case): parent's turn_end autoformat dirties the tree; a child asked to commit runs `git status`, finds unexplained `M` files, burns turns investigating.
- **Parent blind to child — arguably the MORE important direction**: the child is ephemeral, but the parent keeps working in the same tree after the child returns. Files the parent read before the spawn were edited AND autoformatted by the child's pi-lens; the parent's next edit hits `oldText` mismatches or it re-investigates working-tree state it should have been told about. The child's task report covers its intentional edits, never pi-lens's formatting on top of them.

We have to assume people use subagents; the nudge feed must survive the process boundary — in both directions.

## Design sketch: one shared record, every instance both writes and reads

1. **Record**: project-scoped `recent-touches.json` via `getProjectDataDir(cwd)` (NEVER a hardcoded `.pi-lens` path). Ring-buffer capped (~50 entries), atomic tmp+rename like the instance registry (#474). Entry: `{path, reason, ts, pid, sessionId?}` — pid is what makes bidirectional consumption self-excluding.
2. **Producer**: whenever `pilens:files:touched` is published (#482 seam), the same payload is appended to the record. Parent and child run identical code.
3. **Consumer — child at `session_start`**: read entries newer than a freshness window (e.g. 15 min), `pid ≠ self`, file still exists ⇒ feed the SAME #485 pipeline; first `turn_start` injects one batched provenance message.
4. **Consumer — parent at `turn_start`**: cheap mtime-gated check of the record each turn_start (a single `stat` when nothing changed — this is the hot path, keep it ~free); new entries with `pid ≠ self` feed the #485 accumulator exactly as a bus event would. The parent's next turn after a child finishes then opens with: `pi-lens: N file(s) were autoformatted by a subagent's pi-lens: … — re-read before editing.`
[...]

PR #1588, opened 2026-08-18
fix(nudge): drop the re-read nudge when the write already handed back its content (closes #1464)

### LINKED ISSUE #1464: Write-path autofix double-informs: the tool result carries the fixed content and the nudge still says re-read
## Summary

An autofix nudge already exists — the same nudge covers both verbs (`autofixed` / `reformatted`, `clients/agent-nudge.ts`). The gap runs the other way.

Since #1414, a `write` whose autofix changed the file gets the full authoritative post-fix content attached to its own tool result. The agent already holds the fixed bytes. The next turn still receives "N file(s) were autofixed after your last turn: a.ts — re-read before editing".

The relevance filter does not suppress it. It admits any path with read or edit history (`clients/agent-nudge.ts:143-148`), and a file the agent just wrote is in edit history by definition.

`docs/agent-guide.md` already describes the behavior we want rather than the behavior we have: "A write's autofix already came back in its own tool result, so you only need the nudge there if the content was too large to attach." No such suppression is wired.

## Impact

Low severity, real cost. The nudge exists to stop agents investigating unexplained working-tree changes. Telling an agent to re-read a file whose current content it was just handed spends context and invites a redundant read. It also dilutes the nudge: an agent that learns the message is sometimes redundant will start skipping it when it is not.

## Suggested direction

Suppress the nudge for exactly the paths whose post-fix content was delivered in the write's tool result. The write path already knows this — it is the `postMutation` attachment decision, and #1461-slice-1's telemetry names the same seam. Keep the nudge when the size cap or the per-command aggregate budget degraded the attachment to a re-read warning: there the nudge is precisely right.

Deferred-edit autofix at `agent_end` keeps nudging unchanged. That path attaches nothing, so the nudge is the only signal.

## Acceptance criteria

- A write whose autofix content was attached produces no nudge entry for that path.
- A write whose attachment was size-capped or budget-degraded still nudges.
- Deferred edit autofix nudges unchanged.
- The agent-guide sentence becomes true.

### LINKED ISSUE #1565: vitest treats .ts and .js specifiers as two module instances - reset-through-.ts tests pass vacuously
Found during PR #1514's author round (maintainer-supervised AI session), where it produced a live near-miss.

Vitest in this repo treats `foo.ts` and `foo.js` import specifiers as TWO module instances. A test that resets module state (generation counters, latches, memo maps) through the `.ts` specifier while the code under test imports the compiled `.js` resets a *copy* — the assertion then passes vacuously (defect shape 7). The #1514 fixer hit this writing the session-re-arm tests and switched its imports to `.js` to make the tests bind.

## The class

Any test that (a) imports a module for its side-effectful reset/inspection API and (b) uses a different specifier extension than the code under test, silently exercises a parallel instance. Existing test files import policy/helper modules via `.ts` in places — each is a candidate vacuous test.

## Acceptance criteria

- Sweep tests/ for mixed-specifier imports of the same module within one test file's transitive graph (a script or lint rule beats manual grep — the check is mechanical: same resolved module, two specifiers).
- Either normalize the convention (tests import what the runtime imports — `.js`) with a lint/coverage guard, or configure vitest to alias the two specifiers to one instance; state the chosen mechanism.
- One regression demonstration: take a currently-vacuous reset test found by the sweep (if any), prove it red-when-bound, fix it.

## Observability

None needed — the guard is the record.

Refs #1514.

AngriestBird:
## What was wrong

Since #1414 a `write` whose immediate autofix changed the file gets the full
authoritative post-fix content attached to its own tool result. The next turn
still opened with `pi-lens: 1 file(s) were autofixed after your last turn:
a.ts — ... re-read before editing`.

The relevance filter cannot catch this by design. It admits any path with read
or edit history (`clients/agent-nudge.ts:143-148`), and a file the agent just
wrote is in edit history by definition — so the one case where the agent
demonstrably holds the current bytes is the case that nudges hardest. That
spends context on a redundant read and dilutes the nudge for the cases where it
is the only signal.

`docs/agent-guide.md:74` already described the behavior we wanted ("A `write`'s
autofix already came back in its own tool result, so you only need the nudge
there if the content was too large to attach"). Nothing implemented it. Code is
now aligned to the doc; the doc is unchanged.

## The fix

One new seam, `noteAuthoritativeContentAttachment(filePath, attached)` in
`clients/agent-nudge.ts`, called from the two places in
`clients/runtime-tool-result.ts` that actually decide whether the bytes go out:

- the per-file cap (`handleToolResult` tail) passes `attachAuthoritativeContent`
  — the *same boolean* the attachment itself reads;
- the bash per-command aggregate-budget loop passes `false` when it degrades a
  path to the re-read warning, re-arming what the inner per-file "attached"
  had suppressed. Same last-decision-wins ordering the existing
  `authoritative_content_attachment_decision` telemetry already documents.

The nudge layer never re-derives the cap, the budget, or any config for itself
— single source of truth is the write path's actual outcome, not a parallel
recomputation. Ordering is safe: `publishFilesTouched` fires synchronously
inside the `runPipeline` call the decision belongs to (pi's bus is a plain
`EventEmitter`), so the accumulator entry exists before the mark.

Suppression is exactly as narrow as the delivery:

| Case | Nudges? |
|---|---|
| Write, content attached | no |
| Write, size-capped (> 2 MiB) | **yes** |
| Bash multi-file write, aggregate budget degraded that path | **yes** |
| Side-effect file the autofix changed but did not attach | **yes** |
| Deferred `edit` autofix at `agent_end` (attaches nothing) | **yes**, unchanged |
| Cross-process touch on a path this session did not then write | **yes**, unchanged |
| Cross-process touch, then a local attached write to the *same* path | no |

The last row is a correction to an earlier version of this description, which
claimed cross-process behavior was unchanged across the board. It isn't, and the
reviewer was right to probe it. #492's merge rule folds a path seen on both
channels into one entry ("local" is sticky), so a foreign touch followed by a
local write whose content was attached leaves one entry, marked delivered, and
it is suppressed. That is the correct outcome — the attached bytes post-date the
foreign touch and describe the file as it now stands — but "unchanged" was the
wrong word for it.

The suppression is path-scoped, not batch-scoped. Reproduced both halves:

```
✓ cross-process touch then local attached write for the same path
    → consumeAgentNudge() === undefined
✓ cross-process touch on a DIFFERENT path survives the suppression
    → "1 file(s) ... foreign.ts"
```

A later touch of the same path clears the mark. This is not a nicety: format is
deferred to `agent_end` by default, so a write's autofix attachment goes stale
the moment that pass rewrites the file — and stale bytes are worth a re-read.
Same for a cascade autofix or a foreign process's write.

`agent_nudge` phase telemetry gains `filesContentDelivered`, distinct from the
existing `filesFiltered` (relevance drops), and the row is now emitted even when
the whole batch was suppressed. Over-suppression is the risk this change
carries; it should be visible in telemetry rather than only in the absence of
complaints.

## Acceptance criteria

- [x] A write whose autofix content was attached produces no nudge entry for
      that path.
- [x] A write whose attachment was size-capped or budget-degraded still nudges.
- [x] Deferred edit autofix nudges unchanged.
- [x] The agent-guide sentence becomes true.

## Red-first evidence

Source reverted (`git diff --output=` / `git checkout -- clients/`), rebuilt,
tests run against pre-fix code:

```
× suppresses the nudge for a path whose content was attached
× still nudges when the attachment was size-capped or budget-degraded
× re-arms a suppressed path when the aggregate budget overrides the per-file decision
× drops only the delivered path from a batch, never its side-effect siblings
× a later touch of the same path re-arms the nudge (deferred format at agent_end)
× a cross-process touch of the same path re-arms the nudge
× matches the accumulator key across separator forms
× is a no-op for a path the accumulator never admitted
× counts content-delivered drops separately from relevance-filter drops
× records the suppression even when the whole batch is delivered
× clears the delivered mark with the accumulator, so the next turn starts fresh
× a write whose post-fix content was attached produces no nudge entry for that path
× a multi-file bash write nudges only for the path the aggregate budget degraded

Test Files  2 failed (2)
     Tests  13 failed | 26 passed (39)
```

The two end-to-end reds are the behavioral ones (AC1 and the aggregate-budget
half of AC2); the eleven in `agent-nudge.test.ts` are red on the missing export.

The two inversion guards — "size-capped still nudges" and "deferred edit
nudges unchanged" — pass both before and after by construction. They are
guards, not reds, and are called out as such rather than dressed up as
regressions.

After the fix, 312 passing across the 14 files that reference anything this
change touches:

`write-autofix-nudge-suppression`, `agent-nudge`, `runtime-tool-result`,
`runtime-tool-result-debounce`, `runtime-event-flow`, `bash-read-guard-integration`,
`inline-contract`, `cache-observability`, `module-report`, `bus-publish`,
`recent-touches`, `runtime-agent-end`, `pipeline`, `index-wiring`.
`npm run lint` (full-project `tsc`) clean.

## Merge order

No conflict expected with any open PR. Of the eleven open PRs, only #1573
(`refs #1561`) touches `clients/runtime-tool-result.ts`, at the
`recordInlineBlockers` call around line 1102 — several hundred lines above the
attachment tail this PR edits, and a different function of the same file.
Either order merges cleanly; whichever lands second should still be re-checked
semantically, since both edit `handleToolResult`.

Nothing else open touches `clients/agent-nudge.ts`, `clients/bus-publish.ts`,
`clients/pipeline.ts`'s publish sites, or `clients/runtime-agent-end.ts`.

#1580 (`closes #1565`) rewrites test import specifiers repo-wide to `.js`. The
new test file here already uses `.js` specifiers throughout, so it satisfies
that PR's structural guard whichever order they land in.

## Known adjacent issue, deliberately not fixed here

`clients/pipeline.ts:1480-1488` independently predicts the attachment to phrase
its own inline warning ("The attached full content for X is authoritative"),
using a predicate that knows nothing about the size cap or the aggregate
budget. On a size-capped write the tool result therefore carries both that
sentence and the contradicting "too large to attach" warning. That predates
this change and is a separate defect of the same family (a second derivation of
a decision that has one owner). Filed as #1590 rather than folded in — it is a
separate decision-ownership change with its own acceptance criteria, and
widening this PR's blast radius to reach it would not have made either fix
easier to review.

## Review round

Verdict was merge-ready with no blocking defect and no code changes requested.
Head is unchanged at `6af05d49` — both findings were corrections to this
description, so the green checks above still describe the exact commit under
review. Two changes, both above:

1. **The cross-process row of the claim table was wrong.** It said "unchanged";
   one interleaving suppresses. Corrected, scoped, and reproduced (see the
   table section).
2. **One known limit, recorded below rather than fixed.**

### Known limit: late mark after a drain

A `noteAuthoritativeContentAttachment(path, true)` that lands *after*
`consumeAgentNudge` has drained, but after a fresh publish for the same path
has already re-created the entry, attaches the mark to the new entry and
swallows a nudge that was legitimately owed. The reviewer demonstrated it with
a barrier probe that suspends `handleToolResult` across the drain.

It is not reachable through the real loop. `afterToolCall` is awaited, and
`index.ts:2138` flushes debounced tool results before `handleAgentEnd`, so no
in-flight `handleToolResult` is suspended across a drain in a single session.
The only opening is the pre-existing module-singleton `_touched`, shared if two
in-process sessions ever run concurrently — which predates this PR and is not
something this change makes worse in kind.

It does change the worst-case failure *direction* in that unreachable case: the
pre-fix bug was a spurious nudge (wasted context), and this would be a
swallowed one (a real change the agent isn't told about). That is the more
expensive direction, so it is worth writing down even though nothing can reach
it today.

A drain-generation stamp on the accumulator — mark only if the entry's
generation matches the one the write path observed — would close it. Not
implemented: there is no current requirement, and a guard for an unreachable
interleaving is state nobody can exercise or test honestly. If the module
singleton ever becomes genuinely concurrent, this is the fix to reach for.

Cleared under attack and not re-litigated here: ordering against the real bus,
both orders of same-path double writes, aggregate-budget interleavings,
deferred-format re-arm, read-does-not-resurrect, telemetry blast radius, the
kill switch, and CI genuinely executing on this exact head.


apmantza (approved):
Thank you for this, @AngriestBird — and thank you especially for the PR body. The claim table, the corrected cross-process row, and the "known limit, deliberately not fixed" section made this reviewable in a way most PRs are not.

**Disclosure: this review is AI-generated. A maintainer supervises the process and makes the merge decision.**

I re-ran your red-first evidence from scratch — reverted `clients/agent-nudge.ts` and `clients/runtime-tool-result.ts` to master, rebuilt, and got your exact `13 failed | 26 passed`, with the two end-to-end failures reproducing the #1464 message verbatim. I then mutated the shipped code to check the guards were not vacuous. Both held: flipping `entry.contentDelivered = attached` to `= true` turns "size-capped still nudges" red, and deleting the aggregate re-arm turns the bash test red.

That last point is a correction in your favour. The body describes the size-capped guard as passing "both before and after by construction". It does not — it is a real pin. Worth claiming.

I attacked the enumeration of no-attachment paths and could not break it: `isError` early-returns before the mark, `postMutation === undefined` covers autofix-disabled, format-only, deferred `edit`, and side-effect files, and the attached snapshot is re-read after autofix so it is never stale within a run. The debounced path still delivers via the shared promise. Nothing to change there.

Three small asks, none blocking:

1. `clients/agent-nudge.ts:249-262` — could the doc comment note that suppression and re-arming are not equally reliable? Suppression is a direct synchronous call; re-arming depends on `publishFilesTouched`, which returns early on the kill switch and on a stale-session emitter (`clients/bus-publish.ts:154-196`). It is unreachable in a single session for the same reason your late-mark note is, but the comment currently reads as symmetric.
2. Alongside that same known-limits section: the cross-process re-arm is turn-scoped. `recordCrossProcessTouches` runs only at `session_start` (`index.ts:1779`) and `turn_start` (`index.ts:2095`), so a foreign write landing mid-turn on a path you wrote-and-attached is suppressed until the next turn boundary. Same direction as the limit you already recorded.
3. `clients/runtime-tool-result.ts:474` — `path.resolve(wp)` looks redundant; `extractWrittenPathsFromCommand` already returns absolute paths (`clients/bash-file-access.ts:482`). I chased it as a key-desync attack and it is fine, but the asymmetry with the inner call site invites the same chase from the next reader. Drop it, or say it is defensive?

CI is green on `6af05d49` across all nine checks, with Unit tests and Lint & type-check genuinely executing. The branch is behind master but merges cleanly — I'll update it and it lands on the next green run. Since #1573 landed first, I've re-checked the interaction: it touches `recordInlineBlockers` around `runtime-tool-result.ts:1102`, a different region of `handleToolResult` than your attachment tail at 1163.

Issue #1462, opened 2026-08-15
perf(cascade): derive neighbour budget from remaining settle time, not a flat 40


apmantza:
## Summary

Re-homed from #1446 item 4 (the budget-vs-settle sizing question). #1446's
remainder now carries the telemetry half — `cascade_result` records
`neighborBudget` (the budget in force) and `budgetTruncated` (eligible
candidates the flat cap cut off, filters excluded). This issue is the
correctness/sizing half that telemetry alone cannot resolve.

## Problem

`CASCADE_NEIGHBOUR_BUDGET` defaults to 40 (`clients/dispatch/integration.ts`).
`cascadeSettleWaitMs` defaults to 5000 (`clients/runtime-turn.ts`). Dogfood
measurement (2026-08-15) showed a full-budget cascade does not reliably fit
inside the settle window:

```
n=50 cascades  mean=0.46s  median=0.03s  max=3.88s
  3.88s constants.ts nbr=40   2026-08-15T15:07:52.195Z
  3.71s logger.ts    nbr=38   2026-08-15T12:09:32.894Z
```

Adding the `reverse_deps_cache` refresh ahead of the graph build (2046 ms for
`logger.ts`) pushed both over the 5s settle cap. The distribution is strongly
bimodal — median cascade is 30 ms; only high-fanout hub files reach the tail —
so a flat 40 spends the whole budget on exactly the files where a
per-neighbour answer is least likely to be actionable in time.

## Proposed change

Either:
1. Derive the effective per-run budget from remaining settle time (a
   time-boxed neighbour walk rather than a fixed count), or
2. Keep the flat cap but rank neighbours more aggressively and cut the tail
   earlier, using the `budgetTruncated` telemetry landed alongside this issue
   to validate the new cutoff empirically before changing the constant.

The correctness half — a cascade run that misses the settle cap being
discarded rather than carried over — is #1443 (already merged).

## Acceptance criteria

- A cascade whose eligible-neighbour walk would exceed the remaining settle
  budget stops early rather than timing out the whole settle wait.
- `cascade_result`'s existing `neighborBudget`/`budgetTruncated` fields (from
  #1446) are used to confirm the change actually reduces truncation on
  high-fanout hub files without regressing the true-positive rate measured in
  #1446's "Verified working" section.

Refs #1446.

AngriestBird:
Implementation opened as #1589 (`refs`, not `closes`).

[...]

Issue #1446, opened 2026-08-15
enhance(diagnostics): cascade observability, budget sizing, and touch-cache measurement (dogfood 2026-08-15)


apmantza:
## Summary

A day of live cascade telemetry (build `debdb9b9`, 2026-08-14T21:18Z to 2026-08-15T17:45Z, repos `pi-free` and `plegma`) showed the mechanism working end to end on its one real finding, and showed five places where a question about cascade behaviour cannot be answered from logs, or where the answer is worse than it needs to be. The three genuine defects found in the same pass are filed separately. This issue collects the improvements.

Each item states the question it answers and the concrete change. Sibling of #1432, which made the same argument for the #1412-#1415 decision paths: where a component's whole job is to choose A over B, the log must say which.

---

### 1. Agent-facing cascade text is never recorded

**Question that could not be answered:** did the agent actually receive the cascade finding, and what did it read?

The counters exist. For the one real finding today:

```
~/.pi-lens/latency.log
2026-08-15T13:02:44.156Z cache_context
  {"injectionSources": ["turn-findings", "test-findings"], "injectedMessageCount": 2,
   "injectedChars": 1885, "placement": "append", "sessionRole": "concurrent-secondary"}
```

That proves 1885 characters were injected. It does not prove the cascade section was among them. Confirming it required reading `clients/cascade-format.ts` and reasoning about the `ownsAny && result.formatted` gate at `clients/runtime-turn.ts:366-372`. Transcript grep is not a fallback: this turn ran in a `concurrent-secondary` subagent session, and no transcript for session `01a0057d-78b2-7296-a2fc-e5f7d61e021f` exists under `~/.pi/agent/sessions`. `grep -c "Cascade errors in"` across every persisted transcript returns 0 for all of today.

**Change:** at the point `parts` is pushed to `blockerParts`, log a `cascade_injected` record with `{ sectionChars, neighborFileCount, diagnosticCount, testSuggestionCount, suppressedByOwnership }`. Counts, not content.

---

### 2. Cascade-driven test targeting emits nothing

**Question that could not be answered:** when a neighbour had errors, which tests were suggested, and did the suggestion match the tests that actually cover it?

[...]

Issue #1443, opened 2026-08-15
fix(diagnostics): cascade runs that miss the 5s turn-end settle cap are carried over, then silently dropped


apmantza:
## Summary

A cascade compute that misses the 5s turn-end settle cap is re-parked for the next turn, then dropped there. The carry-over path is dead code. Both live cases today were the highest-value cascades of the day — 38 and 40 neighbours.

`settleCascadeRuns` (`clients/runtime-coordinator.ts:510-550`) deliberately preserves an unsettled promise:

> "A promise still pending at the cap is retained so a late-resolving compute is picked up on the next turn_end rather than lost."

The next `turn_end` then discards it. `clients/runtime-turn.ts:338-347` filters every drained run by origin:

```ts
return (
    (originSeq === undefined || originSeq === runtime.projectSeq) &&
    (originTurn === undefined || originTurn === runtime.turnIndex)
);
```

`origin` is stamped at edit time (`clients/pipeline.ts:1421-1425`, `turnSeq: ctx.telemetry?.turnIndex`). A carried-over run is by definition from an earlier turn, so `originTurn === runtime.turnIndex` is always false for it. Every run that misses the cap is guaranteed to be discarded.

## Evidence

Runtime build `debdb9b9` (merge of #1435), install at `~/.pi/agent/git/github.com/apmantza/pi-lens`. Window 2026-08-14T21:18Z to 2026-08-15T17:45Z, dogfood repo `pi-free`.

Case 1 — `lib/logger.ts`, 38 neighbours.

```
~/.pi-lens/latency.log
2026-08-15T12:09:32.161Z cascade_settle_wait turn_end dur=5001 {"settled": 0, "timedOut": 1}
2026-08-15T12:09:32.161Z cascade_merge      turn_end dur=0    {"runsTotal": 0, "resultCount": 0, "neighborCount": 0}

~/.pi-lens/cascade.log
2026-08-15T12:09:32.894Z cascade_result lib/logger.ts neighborCount=38 diagnosticCount=0
```

The result landed 733 ms after the merge that was supposed to carry it.

Case 2 — `constants.ts`, 40 neighbours.

```
2026-08-15T15:07:49.835Z cascade_settle_wait turn_end dur=5014 {"settled": 0, "timedOut": 1}
2026-08-15T15:07:49.835Z cascade_merge      turn_end dur=0    {"runsTotal": 0, "resultCount": 0, "neighborCount": 0}
[...]

Issue #1549, opened 2026-08-17
Cascade inconclusive is touch-wide: one slow auxiliary discards every good answer in the sweep


apmantza:
From an AI-run, maintainer-supervised forensic sweep of `~/.pi-lens/cascade.log` (2026-08-09 → 08-17) and `latency.log(.1)`, prompted by dogfood sessions full of "Cascade diagnostics inconclusive" warnings.

## The numbers

6,079 neighbor-touch sweeps over 8 days: **97.6% inconclusive** (daily range 96.7–99.8%). Ordinary edit-time touches in the same window: 15%. The cascade lane is uniquely saturated, and it is chronic, not a regression.

The dominant cause is #1459's notify-write flood (only ~3% of sweep touches finish inside the 1000/2000ms budget; the 2–5s cluster and >30s tail are queueing signatures; 44 `lsp_notify_backpressure_broken` trips, 38 opengrep). #1528 fixed that shape but ships in no installed build yet.

## The residual this issue tracks

`clients/lsp/index.ts:3242`: `inconclusive = notifyWriteTimedOut || diagnosticsTimedOut` is **touch-wide**. Every sweep touch attaches ~5 servers; if any one times out, the whole touch reports inconclusive — a clean typescript answer plus a slow opengrep scan reads as "nothing is known about this file". With opengrep routinely needing >2s on large files (per #1459's own description), a single slow-but-healthy scanner defeats the sweep even after the flood fix.

## Acceptance criteria

- Verdicts are per-server, aggregated honestly: a touch with a confirmed primary and a timed-out auxiliary reports the primary's findings as usable plus a named coverage gap (the #1493/#1533 `partial` machinery is the pattern), not blanket inconclusive.
- Measurement gate: after #1528 is installed and a dogfood session runs, the neighbor-touch inconclusive rate (from `neighbor_touch` `metadata.inconclusive`) must drop below 30% with a healthy server population. If it stays >90%, this issue's semantics change is the driver and becomes urgent.
- Relation to #1462 (derive neighbour budget from remaining settle time): a per-server budget derived from real settle time would reduce the false-timeout rate; coordinate, don't duplicate.

## Observability (gaps found during the sweep)

- `neighbor_touch` records `inconclusive: true` with **no attribution**: add `inconclusiveServerIds` (which server(s) timed out) and `inconclusiveReason` (`notify-write` vs `diagnostics-wait`) so the next forensic sweep reads the cause instead of inferring it from duration histograms.
[...]

PR #1589, opened 2026-08-18
perf(cascade): derive the neighbour budget from remaining settle time (refs #1462)

AngriestBird:
## What

`CASCADE_NEIGHBOUR_BUDGET` (40, `clients/dispatch/integration.ts`) and
`cascadeSettleWaitMs` (5000, `clients/runtime-turn.ts`) were set in two files
with nothing relating them. A cascade whose prelude pushed it out of the
turn-end window now shortens its neighbour walk to land on time — **only where
shortening it actually helps**.

## The trade this is making

A run that misses the turn-end cap is **not lost**. `settleCascadeRuns` re-parks
the still-pending promise (`runtime-coordinator.ts:614`), `agent_settled` gives
it a second and more generous 15 s drain (`cascade_carry_over_settle`,
`quiet-window.ts:210-213`), and `beginTurn` deliberately keeps
`_pendingCascadeRuns` (`runtime-coordinator.ts:294-298`). That is #1443, which
this issue's own text cites as already merged. So the real trade is:

> the full neighbour set, **one turn late** — versus a shorter set, **on time**,
> with the remainder lost **for good**.

A dropped neighbour never comes back; a late one does. So narrowing is only
worth paying for when it genuinely converts a late run into an on-time one.
That gives three zones and a guard:

| zone | condition | budget |
|---|---|---|
| `fits` | the full walk still lands on time | ceiling (40) |
| `narrowed` | the full walk no longer lands but a shorter one does — **the rescue band, the only place anything is cut** | `floor(remaining / 100ms)`, clamped to `[5, 40]` |
| `past-rescue` | not even a floor-sized walk lands | ceiling (40) — carry-over delivers it complete |
| `no-rescue-window` | the window could not fit a full walk even at zero prelude | ceiling (40) — nothing to rescue, only every run to shrink |

At the 5000/100/40 defaults the rescue band is a prelude of **1.0–4.5 s**. The
derivation runs once, after the graph build / reverse-deps refresh /
LSP-reference expansion, and drives both `computeTransitiveImpact`'s `maxHits`
and the final `slice`.

## Review round — 8 findings, all reproduced before fixing

The mechanism cleared attack; the premise did not. Every finding below was
reproduced first.

### F1 (blocking) — the 5000 ms settle was not the only deadline

Confirmed at `clients/quiet-window.ts:93` (`DEFAULT_QUIET_WINDOW_WAIT_MS =
15_000`), `:210-213` (`cascade_carry_over_settle` calling
`settleCascadeRuns(quietWindowWaitMs())`), wired at `index.ts:2338`, fired at
`index.ts:2475`. My test literally named
*"not a second hardcoded window"* while a second hardcoded window sat 3× larger
in the same pipeline. The derivation now reads `quietWindowWaitMs()` and records
`budgetDeliveryWindowMs` (settle + drain) on `cascade_result`, so the whole
deadline stack is on the record. It is **recorded, not a divisor** — stated as
such at the field, and pinned by a test that tracks
`PI_LENS_QUIET_WINDOW_WAIT_MS` so it is not a decorative field.

### F2 (blocking) — "delivered nothing at all" was false

Correct. An overrun delivers **complete, one turn late**. The claim is removed
from the PR body and rewritten in the changelog fragment, and the corrected
trade is now the stated rationale for the whole design rather than a footnote.

### F3 (blocking) — cold-session inversion, a real regression

`runtime-coordinator.ts:295` says fresh graph builds measure up to ~19 s, and
`cascadeStart` samples at compute entry, so the whole build landed in
`elapsedMs`. Reproduced: 19 s prelude → budget 5, `budgetTruncated` 36,
`budgetRemainingMs` -14000 — 35 neighbours dropped **permanently** on a run that
overran either way. Strictly worse than pre-#1462. `past-rescue` now returns the
ceiling; the bench row is byte-identical to flat.

### F4 (blocking) — the bench was an arithmetic identity

Confirmed: `MEASURED_COST_MS` went in as *both* the derivation's divisor and the
walk-cost model, so "derived fits" held at any `--cost-ms`. Divisor
(`--divisor-ms`, the production estimate) and true marginal cost
(`--true-cost-ms`) are separate arguments now, and `--sweep` holds the divisor
and varies only the truth. **It reports a limit, not a win** — see below. The
footer that claimed pure loss stays at zero was falsified by the sweep itself
and is rewritten.

### F5 (major) — `maxHits` was untested

Confirmed: reverting `maxHits: cascadeNeighbourBudget` to the flat constant left
every test green. Pinned by a test asserting the argument, red under exactly
that mutation:

```
FAIL  #1462: the transitive expansion is capped by the derived budget, not the flat one
AssertionError: expected "vi.fn()" to be called with arguments: [ Anything, Any<String>, …(1) ]
-     "maxHits": 10,
+     "maxHits": 40,
```

### F6 (major) — the settle wait had become a neighbour-count knob

Confirmed: at `settle=1000` every cascade capped at ~10–12 neighbours. Fixed by
the `no-rescue-window` guard rather than documentation, and with no new knob: a
window that cannot fit a full walk even at zero prelude is a budget change
wearing a timeout's clothes, so the derivation stands down entirely. At
`settle=1000` all 40 neighbours are now walked.

### F7 (minor) — `budgetTruncated` inverts

Correct: it is `eligible - budget`, so a *narrower* budget mechanically *raises*
it. Called out at the log site, in the changelog, and in the issue's deferral
note — compare truncation **within** a zone, never across one.

### F8 (minor) — truncation scope

Truncation is now structurally confined to the rescue band (prelude 1.0–4.5 s at
the defaults) **and** requires `eligible > derived budget`. Every other zone
keeps the flat 40, so `fits`, `past-rescue` and `no-rescue-window` truncate
exactly as much as pre-#1462 and no more. #1446 measured only 2.0 % of cascades
reaching the cap at all (p50 fan-out 1, p90 22). On that basis no
`indeterminate` marker is added; if a dogfood window shows narrowing common
outside the band, the marker becomes in scope and I would file it.

## Measurement (`npm run bench:cascade-budget`)

```
true marginal cost 97 ms/neighbour | production divisor 100 ms | on-time window 5000 ms | pipeline drain 20000 ms

scenario                                             nbr              zone  flat   total  fits  drv   total  fits  dropped
--------------------------------------------------------------------------------------------------------------------------
median cascade (30 ms, nbr=1)                          1              fits     1   127ms   yes    1   127ms   yes        0
cline-headers.ts (1.33 s prelude, nbr=3)               3          narrowed     3  1621ms   yes    3  1621ms   yes        0
warm hub (500 ms prelude, nbr=40)                     40              fits    40  4380ms   yes   40  4380ms   yes        0
constants.ts (nbr=40, no cache refresh)               40              fits    40  3910ms   yes   40  3910ms   yes        0
logger.ts + reverse_deps refresh (2046 ms, nbr=38)    38          narrowed    38  5732ms    NO   29  4859ms   yes        9
cold hub + slow graph build (3500 ms, nbr=40)         40          narrowed    40  7380ms    NO   15  4955ms   yes       25
cold session, ~19 s graph build (nbr=40)              40       past-rescue    40 22880ms    NO   40 22880ms    NO        0

missed the on-time window: flat 3/7, derived 1/7   rescued: 2   neighbours dropped for good: 34
narrowed and STILL missed (pure loss): 0   narrowed while the flat walk already fitted: 0
```

Fast case untouched (0 dropped across all four fitting rows), rescue band
converts 2, blown-window row **identical to flat in every column**.

### Divisor sensitivity — the honest limit (`-- --sweep`)

```
true cost     flat missed  derived missed  rescued  dropped  pure loss
----------------------------------------------------------------------
60ms                  2/7             1/7        1       34          0
80ms                  3/7             1/7        2       34          0
97ms                  3/7             1/7        2       34          0
100ms                 3/7             1/7        2       34          0
120ms                 4/7             4/7        0       34          2
150ms                 5/7             5/7        0       34          2
200ms                 5/7             5/7        0       34          2
```

**This is a real limit, not a win.** The band test uses the *divisor*, so a
divisor that underestimates the true cost lets a run narrow and still miss.
Above roughly +20 % (true 120 ms against a 100 ms divisor) the rescue converts
nothing and two runs drop neighbours for nothing. The default divisor is 100 ms
against a measured 97 ms — 3 % of headroom. The rescue band bounds the damage
(it never narrows a run already past a floor-sized walk, which is what keeps the
cold-session row identical to pre-#1462) but it cannot rescue a run whose cost
it mis-measured. `PI_LENS_CASCADE_NEIGHBOUR_COST_MS` is the lever if a dogfood
window puts the marginal cost above the divisor, and it is worth being the first
thing that window checks.

Derivation cost: ~600 ns/call over 1,000,000 calls.

## Red-run evidence

Round-2 changes reverse-applied (baseline = the round-1 code already on this
PR), rebuilt, targeted files run:

```
FAIL  cascade-budget.test.ts > keeps the FULL budget once the window is blown past rescue
  AssertionError: expected { budget: 5, ceiling: 40, …(1) } to match object { budget: 40, zone: 'past-rescue' }
FAIL  cascade-budget.test.ts > stands down when the window could never fit a full walk anyway
  AssertionError: expected { budget: 12, ceiling: 40, …(1) } to match object { Object (budget, zone) }
FAIL  cascade-budget.test.ts > records the pipeline's whole drain, not just the turn_end settle
  AssertionError: expected undefined to be 20000
FAIL  cascade-budget.test.ts > narrows in step with the on-time window left, inside the rescue band
  AssertionError: expected { budget: 29, ceiling: 40, …(1) } to match object { budget: 29, zone: 'narrowed' }
FAIL  cascade-compute.test.ts > #1462: a window blown past rescue keeps the FULL budget rather than walking a stub
  AssertionError: expected "vi.fn()" to be called 40 times, but got 5 times
FAIL  cascade-compute.test.ts > #1462: shrinking PI_LENS_CASCADE_SETTLE_WAIT_MS does not turn it into a neighbour-count knob
  AssertionError: expected "vi.fn()" to be called 40 times, but got 12 times

Test Files  2 failed (2)
     Tests  12 failed | 52 passed (64)
```

F5's test needs the reviewer's own mutation to go red (round-1 code already
passed the derived budget); that run is quoted under F5 above.

Behavioural tests assert `touchFile` **call counts**, not just the logged
number, so a budget reported but not enforced fails them. The "common fast
cascade" case runs under the same stubbed clock as the pressured ones with a
zero prelude, so a failure there means the derivation narrowed the walk, not
that the stub did.

Post-fix, over every test file referencing a symbol this PR touches — including
`quietWindowWaitMs`, `settleCascadeRuns` and `beginTurn`, which the review round
brought into scope:

```
Test Files  14 passed (14)    Tests  198 passed (198)
Test Files  12 passed (12)    Tests  244 passed (244)
```

Plus `npm run lint`, `build`, `check`, `check:lockfile`, `changelog:check`.

### N3 — `budgetDeliveryWindowMs` over-reported a killed quiet window

`PI_LENS_QUIET_WINDOW=0` disables the scheduler (`quiet-window.ts:82-86`) while
`quietWindowWaitMs()` keeps returning 15000 — it is the task's *budget*, not a
statement that the task runs. So a field documented as the whole deadline stack
claimed 15 s of drain that would never happen, in the one configuration where it
does not exist. Now gated on `isQuietWindowEnabled()`. Red:

```
FAIL  does not count a drain that PI_LENS_QUIET_WINDOW=0 has switched off
AssertionError: expected 20000 to be 5000
```

### N4 — the knobs are split; the load-path rationale for it was wrong

**Correction, owned:** the reason I shipped for this split was false, and I did
not check it before writing it down. The claim was that moving the quiet-window
knobs out of the scheduler kept `pidusage` off the dispatch load path. Probed
via the CJS require cache:

```
importing clients/cascade-budget.js       → pidusage in require.cache = 6
importing clients/runtime-config.js       → pidusage in require.cache = 6
importing clients/quiet-window-config.js  → pidusage in require.cache = 0
```

`cascade-budget.ts` imports `runtime-config.ts` for `cascadeMaxFiles`, and that
reaches `pidusage` through `lens-config.ts` regardless. `dispatch/integration.ts`
already imported `runtime-config.ts` on master before #1462 existed, so
`pidusage` was on that load path all along. The quiet-window edge added nothing
and removing it removed nothing.

**The split stays, on the reason that actually holds.** `cascade-budget.ts` is
arithmetic and should not depend on a task registry — module state (`_tasks`,
`_inProgress`), a heartbeat sampler, latency logging — to read two numbers, and
N3's `isQuietWindowEnabled()` gating turned that from hypothetical into real.
Config belongs below the scheduler, not beside it. `quiet-window.ts` re-exports
all three names, so every existing importer is untouched and there is still one
memo behind `isQuietWindowEnabled`. Inlining the constant was the alternative
and would have put 15000 and the kill-switch rule in two files.

### N2 — the sweep's tuning advice, bounded (and the stated bound corrected)

"Raise `PI_LENS_CASCADE_NEIGHBOUR_COST_MS`" silently no-ops past a point, so the
sweep text now states the range it is true in. Measured rather than repeated,
and the review's suggested 100–125 ms range turned out to be **too generous** —
125 ms is already past the real edge:

| divisor | behaviour |
|---|---|
| ≤112 ms | healthy — only runs that would have missed are narrowed |
| 113–129 ms | the `fits` zone has shrunk below real preludes, so runs that would have landed anyway start narrowing (113 → 1 such run; 125 → every row, 48 dropped) |
| ≥130 ms | `ceiling × divisor` exceeds the window, `no-rescue-window` fires on everything, feature disarms |

The upper failure is **safe** (disarmed = the flat cap = pre-#1462, 0 dropped);
the middle band is the one to avoid, and the text now says which is which and
recommends moving the settle wait or ceiling rather than the divisor alone.

### N5 — zero-width `fits` zone at the exact boundary

Comment added at `onTimeMs === ceiling × perNeighbourMs`, where the guard does
not fire and `fits` has zero width. Not special-cased — the budget still lands
within one neighbour of the ceiling. N2's 113 ms edge is this same boundary
arriving early, so the two notes cross-reference the same arithmetic.

Polish-round totals: **28 files / 500 tests** green (199 + 301), now including
`quiet-window`, `resource-sampler`, `cascade-tier` and the runtime suites.

## Micro round (head `3e536e2b`)

Comments and one test import; no behaviour change. **I shipped a false rationale
in N4 and the reviewer caught it** — the split's load-path benefit does not
exist (probe in §N4 above), and three comments plus the PR body asserted it.
Rewritten to justify the split on dependency direction, which is what actually
holds. Also corrected the N5 boundary comment: *"the budget still lands within
one neighbour of the ceiling"* is only true at tiny preludes — at 4000/100/40 a
2 s prelude gives 20 and a 3.5 s prelude gives the floor. The clause that
survives is the one that justified leaving the boundary alone: with the window
sized to exactly one full walk, a run that has spent any of it really would
miss, so those cuts are genuine rescues. `tests/clients/cascade-budget.test.ts`
now imports the reset hook from `quiet-window-config.js` rather than through the
scheduler face.

One more instance of the same class, found while making that correction and
fixed in a second commit: the `deliveryWindowMs` docstring credited the
quiet-window drain with *"why `past-rescue` keeps the full budget"*. It does not
— `deliveryWindowMs` is recorded and feeds no zone, so killing the quiet window
leaves every budget unchanged. What actually lets `past-rescue` stay wide is
`beginTurn` never clearing `_pendingCascadeRuns`: a slow compute is retained
until it resolves, however long that takes, and the drain only gets it there
sooner.

## Scope boundaries

- **Budget sizing only.** #1459 untouched. Interaction worth recording: the
  rescue band reduces notify-write fan-out per cascade, so #1459's backpressure
  and this budget move the same dial from opposite ends.
- **Per-touch LSP budgets unchanged** (1000 cold / 2000 warm), per #1446.

## Merge order

Branched from `2e44576b`. Of the open PRs, only #1575 touches
`clients/dispatch/integration.ts` (ast-grep napi latch, different region). No
open PR touches `clients/cascade-*.ts`, `clients/runtime-turn.ts`,
`clients/quiet-window.ts`, `package.json`, or `.changelog/`. Whichever lands
second should re-run `tests/clients/cascade-compute.test.ts`.

## Acceptance criteria

- ✅ *"A cascade whose eligible-neighbour walk would exceed the remaining settle
  budget stops early rather than timing out the whole settle wait."* — met
  wherever stopping early achieves it. Deliberate deviation, flagged rather than
  taken silently: in the `past-rescue` case stopping early does **not** avoid
  the timeout (the bench proves the row misses either way), so paying permanent
  signal loss for a non-benefit would serve the letter of the AC against its
  own stated purpose. That zone keeps the full budget.
- ⏳ *"`neighborBudget`/`budgetTruncated` are used to confirm the change reduces
  truncation on high-fanout hub files without regressing the true-positive
  rate."* — the fields are emitted, plus `budgetZone` to make the comparison
  valid at all. Confirming against live traffic needs a dogfood cycle on a build
  carrying this change; hence `refs`, not `closes`. Two things that cycle must
  be told before it starts:

  1. **`budgetTruncated` inverts.** It is `eligible - budget`, so a narrowed
     budget raises it mechanically. Compare within a `budgetZone`, never across
     one, or a working rescue reads as a regression.
  2. **The exchange rate is unpriced.** The `narrowed` zone applies "is this
     rescue worth having" as a pure boolean — it never asks what the rescue
     *costs*. At a 4500 ms prelude it drops 35 of 40 neighbours (87.5 % of the
     walk) to land 15 ms inside the window; at 4501 ms it keeps all 40. That
     cliff is bounded to the 3.5 s band and fully visible in
     `budgetZone`/`budgetTruncated`, and it is strictly better than the round-one
     behaviour at every prelude, so it ships as-is — but the dogfood cycle should
     measure the price actually paid at the bottom of the band, not just the
     conversion count. **Candidate follow-up:** bound the exchange rate, e.g.
     refuse to narrow below some fraction of the ceiling and fall through to
     `past-rescue` instead. Not filed here — the tracker is upstream.

---

## Maintainer addition (post-review round, commit `9193dc8e`)

This section is AI-generated, added by the maintainer's own automation after
the changes-requested review below. It documents what changed on top of the
contributor's work and why, and does not speak for the contributor.

### F-A (blocking) — verified fixed, not re-fixed

The contributor's `6aa33a08` already threads the coordinator's real
`turn_end` settle-clock start (`RuntimeCoordinator.getTurnEndCascadeSettleStart`)
through `pipeline.ts` into `computeCascadeForFile`, so the derivation measures
elapsed time against the window that is actually running, not against the
compute's own age from an unawaited write. When no settle window is active
yet, it charges zero elapsed time rather than inferring one — the "gap
unobservable → refuse to narrow" input this review asked for.

Verified red on the pre-fix commit (`3e536e2b`) with the review's own
scenario — a 2 s gap, 5000 ms real headroom, must not narrow:

```
FAIL #1462: the active settle clock ignores the pre-turn gap
AssertionError: expected "vi.fn()" to be called 40 times, but got 10 times
```

Green on `6aa33a08` onward. No further change needed here.

### F-B — past-rescue: kept, not shrunk

The 0/1090 dogfood frequency this review measured was taken against the
write-relative clock F-A just fixed, which over-charges elapsed time (it can
only inflate apparent narrowing, never hide it) — so it is not evidence the
corrected derivation would show the same rate. Past-rescue exists to protect
the cold-session case (~19 s graph build) that the review's OWN first round
caught as a real regression (F3): a floor-sized stub there is strictly worse
than pre-#1462. Shrinking a cheap, already-tested guard on a sample the fix
invalidates would risk reopening a regression that round already fixed.
Documented at the call site in `cascade-budget.ts`; worth revisiting once a
post-fix dogfood window can measure reachability under the corrected clock.

### F-C — indeterminate marker: verified, not re-added

Already on this branch (`6aa33a08`): a budget-truncated run sets
`indeterminate: { reason: "budget_truncated", budget: {...} }`, and
`runtime-turn.ts`'s turn_end advisory surfaces it in its own bucket
(`budgetRuns`) rather than folding it into the graph/binding advisories or
silently dropping it from the denominator. Ran clean against the merged
master (`tests/clients/cascade-turn-merge.test.ts`, 31/31) — no conflict with
#1571's per-server touch verdict or #1626's LSP idle-reset work; neither
overlaps this code path.

### F-D — bench claim: not found as stated

Searched the current PR body, changelog fragment, and code comments for an
"86/86 drops were superseded_by_later_write" claim; it isn't present in any
of them as of this branch's head. It may describe a data point from an
earlier, since-edited draft. No text change made because there is nothing
left to soften — flagging this rather than silently closing it out.

### F-E (new fix) — silent-disarm override now names itself

`PI_LENS_CASCADE_NEIGHBOUR_BUDGET` set high enough that
`ceiling * perNeighbourMs` exceeds the settle window puts every cascade in
`no-rescue-window` for the rest of the session — not because anyone asked to
disable the rescue band, but because they asked for a bigger cap. Nothing
said so. Fixed by recording it once per session through the existing
degradation ledger (`recordDegradationOnce`, kind
`cascade-budget-override-disarmed`) rather than a new ad hoc log line —
reusing the established mechanism per this repo's single-degradation-path
convention.

Red on pre-fix code:

```
FAIL #1462 review F-E: an env override that disarms the rescue band names
itself once, in the degradation ledger
```

Green after. A second disarmed call in the same session does not double-log
(`count` stays 1) — the once-per-session behavior the ledger already
guarantees.

### F-F (comment only) — concurrent runs share the one window

Documented at `RuntimeCoordinator.getTurnEndCascadeSettleStart`: every
cascade compute on one coordinator reads the same clock, because there is one
`turn_end` per runtime, not one per caller. No behavior change — this was
already correct, just unstated.

### Verification

- `npm run build`, `npm run lint` (tsc), `npm run check`,
  `npm run check:lockfile`, `npm run changelog:check` — all pass.
- Targeted: `cascade-budget.test.ts`, `cascade-compute.test.ts`,
  `cascade-turn-merge.test.ts`, `quiet-window.test.ts`,
  `runtime-cascade-defer.test.ts`, `degradation-ledger.test.ts`,
  `review-graph/transitive-impact.test.ts` — 129/130 passing.
- The one failure (`uses passive fallback only for no-LSP candidates in a
  mixed cascade`) is a pre-existing Windows-only test artifact: the
  contributor's own test compares a fallback neighbour's `filePath` against a
  native-separator string, but production's `getAllDiagnostics()` keys (and
  therefore this fallback path) are always forward-slash normalized, on every
  OS. `path.sep === "/"` on Linux masks this, so it will pass in CI. Not a
  functional regression and out of this round's scope; worth a follow-up test
  fix, not filed as a separate issue given its narrow blast radius.
- Full suite deferred to CI, which was already running non-conflicted
  (`mergeStateStatus: MERGEABLE`) as of head `9193dc8e`.

### Merge order

Rebased onto current `origin/master` (post #1571/#1626/#1633-adjacent work);
merged clean, no conflicts in `dispatch/integration.ts` or elsewhere — the
per-server verdict machinery (#1571) was already the base this branch was cut
from, and nothing else in flight touches `cascade-budget.ts`,
`runtime-coordinator.ts`, or `degradation-ledger.ts`.


apmantza (changes requested):
Thanks for this, @AngriestBird.

this review is AI-generated, and a maintainer supervises the process and signs off before anything merges.

Your regression tests are not vacuous: I neutered the narrowing branch and 6 tests went red across both files. And the five new `#1462` tests do fail on master's `integration.ts`.

The blocker is one clock. `cascadeStart` is sampled at compute entry, and `pipeline.ts:1505` kicks that compute off unawaited **at the write**, while the 5000 ms settle window starts **at turn_end** (`runtime-turn.ts:346`). So the run's real deadline is `(turn_end − write) + 5000`, and the derivation uses `5000`. Your docstring already names this and calls it the conservative read — but conservative here means "assume less time", which means "narrow", which means permanent loss. The safe direction is to assume more.

Measured on my `cascade.log`, `turn_end − compute_start` runs p50 1783 ms, p75 3188 ms, p90 3989 ms, with 57 of 60 turns carrying more than a second of slack the derivation never sees. At a 2 s prelude with a 2 s gap the derivation returns 30 and drops 10 neighbours on a run that had 5000 ms of real headroom for a 4000 ms walk. Your bench cannot see this, for the same reason F4 was invisible: it models the prelude as settle time already spent, i.e. gap = 0.

The second thing is about payoff. Over 1090 cascades in that log the prelude is p50 6 ms, p95 371 ms, max 1873 ms — 98.6% `fits`, 1.4% `narrowed`, and `past-rescue` fired zero times. The zone you spent three rounds hardening never occurred, and the zone that does fire is the one the clock bug makes mostly wrong.

Two smaller ones. A narrowed run sets no `indeterminate` marker (`integration.ts:2112` only reads `impact.indeterminate`), so a walk of 20 of 40 renders at turn_end as a confident clean result — and its dropped neighbours emit no `neighbor_touch` verdict, which quietly shrinks the denominator #1549's gate watches. And `PI_LENS_CASCADE_NEIGHBOUR_BUDGET=60` silently disarms the whole derivation via `no-rescue-window`, with no log line.

What would unblock it: give the compute the real deadline instead of inferring it. The coordinator knows the turn boundary; threading a "time until turn_end is unknown, assume at least X" input — or simply refusing to narrow when the gap is unobservable — would keep the rescue honest. Given the payoff numbers, it is also fair to ask whether the rescue band earns its 222-line module at 1.4% of cascades; a much smaller version that only guards the genuinely pathological prelude might be the better trade.

Happy to re-run the probes against whatever you land on. The measurement scripts I used are just log arithmetic and I can share them if useful.

apmantza:
Hi @AngriestBird — this comment is AI-generated, and a maintainer supervises the process and signs off before anything merges.

Thanks for sticking with this through five review rounds. Your bench methodology made F4 and F-A findable at all, and owning N4's false rationale out loud (rather than quietly dropping it) is exactly the kind of honesty this review process is for.

I pushed one more commit (`9193dc8e`) on top of your `6aa33a08`/`76b02ffc`. Short version: your F-A fix was already correct — I verified it red-on-pre-fix with the review's exact scenario (2 s gap, 5000 ms real headroom, must not narrow) and green after your change, so nothing there needed touching. What I added:

- **F-E**: `PI_LENS_CASCADE_NEIGHBOUR_BUDGET` set high enough to disarm the rescue band now names itself once per session in the degradation ledger, instead of silently capping every cascade at the flat ceiling.
- **F-F**: a comment on `getTurnEndCascadeSettleStart` noting that concurrent computes share one coordinator clock (this was already correct behavior, just unstated).
- **F-B**: I decided to keep `past-rescue` rather than shrink the module. The 0/1090 frequency in the review was measured against the write-relative clock your commit just fixed, which could only inflate apparent narrowing, not hide it — so it isn't evidence the corrected derivation behaves the same way. Past-rescue is what stopped the cold-session regression (F3) the first review round caught, and shrinking it now, on a sample the fix invalidates, felt like the wrong trade. Documented the reasoning at the call site — happy to revisit if a post-fix dogfood window shows it's genuinely dead code.
- **F-C, F-D**: verified rather than redone. F-C's indeterminate marker was already in your commit and composes cleanly with #1571/#1626 on current master. F-D's "86/86 superseded_by_later_write" claim isn't in the current PR body/changelog as far as I can find — may be from an earlier draft. Flagging rather than guessing at what to soften.

One thing I found but didn't fix, since it's outside this round's scope: `uses passive fallback only for no-LSP candidates in a mixed cascade` fails on Windows because it compares a fallback neighbour's `filePath` against a native-separator path, but `getAllDiagnostics()` keys are always forward-slash normalized in production, on every OS. It'll pass in CI (Linux), so it's not blocking, but the assertion doesn't match what the code actually does cross-platform. Worth a follow-up if you want it, no urgency.

Merged current `master` in first since this branch was a few dozen commits behind — came through clean, no conflicts. Full details and verification are in the PR body's maintainer-addition section.

Nice work getting this to the finish line.

apmantza:
Thanks @AngriestBird — this is now in good shape, and I want to name what you did: you fixed the clock finding yourself, and the fix is the right one. Threading the active `turn_end` settle start and falling back to zero elapsed when no settle is running means the rescue band now fires only on computes that are genuinely racing the deadline. I re-ran my own probes against `9193dc8e`: a 2 s and a 30 s pre-turn gap no longer narrow, an active settle window still does, and all three are red on `3e536e2b`.

Two things came out better than I asked for. The truncation marker now covers the flat cap too, not just the narrowed zone — my red-run against the old code returned `skipReason: "clean"` on a cascade that had checked 40 of 60 dependents, so that was a real silent all-clear and it is gone. And the `computeTransitiveImpact` change is the right call: without it every exact-cap traversal would have emitted a false incomplete-coverage advisory through the new marker.

I also withdraw my `past-rescue` measurement. I counted it at 0/1090, but I counted it against the write-relative clock this round replaced, so it did not measure the thing that matters. Keeping the zone is correct.

One tiny thing, non-blocking and Linux CI is unaffected: the new mixed-cascade test at `cascade-compute.test.ts:611` compares a native-separator path against forward-slash-normalized `getAllDiagnostics` keys, so it goes red for anyone running that file on Windows. Normalizing both sides of that one assertion fixes it.

As before: this review is AI-generated and a maintainer supervises and signs off. Thanks for the care you put into the review rounds — owning the F4 and N4 corrections in public made this much faster to verify.

PR #1591, opened 2026-08-18
fix(lsp): judge deferred-door coverage at merge time (closes #1586)

### LINKED ISSUE #1586: Deferred-door coverage judged from the pre-notify snapshot: content-covering deferred aux dropped and misnamed
From PR #1571's verification round (maintainer-supervised AI review). Pre-existing — landed with #1459's queue work; #1571 fixed the same class on the timed-out-write door but does not touch this one.

The `notifyDeferredServerIds` door (clients/lsp/index.ts:~4028 as of #1571's head) still judges coverage from the PRE-NOTIFY snapshot and drops a deferred auxiliary's findings unconditionally, with no content exemption. A deferred scanner that publishes for exactly this touch's bytes mid-touch is dropped and named uncovered — the inversion #1571's F1 just fixed, one door over.

## The principle (name it in the fix)

Content-bound coverage must be judged at MERGE TIME on EVERY door — #1571 established the single-predicate pattern (`auxCoversThisContent` unioning the snapshot with a merge-time `getDiagnosticBinding contentHash` read, with the same predicate deciding both the findings-drop and the coverage-naming). Extend that predicate to the deferred door.

## Acceptance criteria

- A deferred auxiliary whose stored binding matches the touch's exact content hash keeps its findings merged and is not named in the no-answer/unconfirmed sets; regression test proven red on current code at that shape.
- A deferred auxiliary with no such binding stays dropped and named (the honest case — pin it).
- The single-predicate property holds: no door may use a different coverage rule (assert by construction or by test).

Refs #1571, refs #1459, refs #1493.

AngriestBird:
## What changed

`notifyDeferredServerIds` — the deferred door in `touchFile` — judged coverage
from the PRE-NOTIFY snapshot (`auxPublishedThisContent`), while #1571 had already
moved the stale-findings drop onto a merge-time content-binding read. The
snapshot is captured before the notify goes out, so by construction it cannot see
a publication that arrives after it. When the outstanding write the fan-out gate
deferred behind LANDS and the scanner then publishes for exactly this touch's
bytes (#1459's own documented late-write signature), the deferred touch still

- dropped those findings from `.diags` (`deferredResyncServerIds.has(id) ? []`),
  and
- named the scanner in `unconfirmedServerIds`
  (`notifyDeferredServerIds.filter((id) => !auxPublishedThisContent.has(id))`).

A scanner that demonstrably reported on these bytes read as one that had gone
silent — #1549's underclaim, one door over.

## The fix: one rule, evaluated once, at the merge

Three layers, each pinned:

1. **One content-match comparator.** `bindingMatchesTouchContent` is the only
   place `touchContentHash` is compared. `carriedAuxiliary` (which must ask
   before the notify), #1493's snapshot, and the coverage predicate all go
   through it, so a door cannot acquire a rule of its own by writing the
   comparison inline.
2. **One coverage rule.** `auxCoversThisContent` = the pre-notify snapshot
   unioned with a live binding read. It is the only reader of the snapshot.
   Fails closed for a primary-role server and for any id that never reached
   `spawned` (every breaker-skipped scanner).
3. **One evaluation.** `auxCoveredAtMerge` freezes it immediately before the
   merge. The exact property is that there is no `await` between the freeze and
   the drop it authorizes — three derived consts sit between them, and none of
   them suspends.

Every door that shares the merge's consequences reads the frozen set:

| door | before this PR | after |
| --- | --- | --- |
| merge drop (deferred half) | unconditional | frozen set |
| merge drop (stale-write half) | live read, pre-merge | frozen set |
| merged BINDING (`mergeBinding`) | raw deferral set | frozen set |
| result `unconfirmedServerIds` | union, unfiltered | reconciled against the frozen set |
| `lsp_aux_wait_outcome.publishedThisContent` | pre-notify snapshot | the rule, at the wait's own instant |

The two aux wait-outcome producers still evaluate the rule live, because their
rows describe their own instant; their verdict is reconciled against the freeze
before anything is claimed.

Observability: `lsp_scanner_coverage_gap` and `lsp_touch_file` report the
deferrals the touch is actually uncovered for, not every deferral. The gate's own
action keeps its unconditional record in `lsp_notify_resync_deferred`.

## Acceptance criteria

- **A deferred auxiliary whose stored binding matches the touch's exact content
  hash keeps its findings merged and is not misnamed** — pinned on BOTH aux
  producers (`with-auxiliary` grace and `all` aggregate), and on BOTH halves of
  the union (live read; and the snapshot-only case, where the holding write has
  since cleared the scanner's cache).
- **A deferred auxiliary with no such binding stays dropped and named** — pinned
  with the same fixture and a binding for the previous revision's bytes, so the
  only thing that differs is the content hash.
- **The single-predicate property holds** — asserted behaviorally ("no door names
  a scanner whose findings the merge kept", driven through a touch judged by the
  notify-write door, the wait-outcome door and the merge at once) and by
  construction (`touchContentHash` has exactly one comparison site; the snapshot
  has exactly one read site). Both mutations from the review round turn the
  structural guard red — see below.

  Scope of that construction guard, stated precisely: it fences SYMBOL-LEVEL rule
  duplication. A door that reached for `touchContentHash` or the snapshot turns it
  red; a door that recomputed the hash inline (`?.contentHash !== this.hashContent(content)`,
  touching neither symbol) would not. That escape is not expressible-proof, it is
  reviewable-by-doctrine — the AGENTS.md paragraph names the rule and the freeze,
  and the escape requires deliberately recomputing a value that is in scope one
  line away.

## Review round

Verdict was NEEDS CHANGES: one blocking regression, two medium, two low. All five
addressed on this branch; every probe reproduced before it was fixed.

### F1 (blocking) — the naming filter re-asked the live rule after the drop

Reproduced exactly as reported. The drop was frozen and applied at the merge;
`touchFile` then awaits (`brokenSkippedAuxiliaryServerIds` on every collecting
touch, the tsserver sync and liveness gates on theirs) before naming its gaps. A
publication landing in that window flipped the predicate and un-named a scanner
whose findings were already gone — `confirmed` over a `.diags` missing that
scanner's answer, the gap row beside it saying the opposite, and
`coverageGap: false` unblocking the `lastKnownDiagnostics` prime and the
`demonstratedReady` mark. That is #1459's blackout reading as scanned-clean.

Two red probes on `7d3e561b`, one per door, with the auxiliary's publication
becoming visible once the merge has read the PRIMARY's cache (the merge's own
call, made exactly once per collecting touch, so it marks the instant the drop
was frozen):

```
× stale-write door: a publication landing after the merge cannot un-name the drop
× deferred door: a publication landing after the merge cannot un-name the drop
AssertionError: the given combination of arguments (undefined and string) is invalid
  for this assertion  ← unconfirmedServerIds was undefined; the touch claimed confirmed
```

Both probes assert the drop DID happen first, so they pin the contradiction
rather than either half alone. The single-touch probe also pins
`lastKnownDiagnostics` staying unprimed.

Fixed by the freeze (`auxCoveredAtMerge`). A drop is an action; a later answer
cannot undo it, so the naming is settled from the same instant that authorized
it. The property the fix rests on is that no `await` separates the freeze from
the drop it authorizes — three derived consts sit between them and none of them
suspends.

### F2 (medium) — the guard fenced a variable, not a rule

Confirmed: the old guard counted reads of `auxPublishedThisContent`, so a door
with its own inline `contentHash === touchContentHash` test passed. The
comparison is now collapsed into `bindingMatchesTouchContent` — which also
absorbs the independent rule in `carriedAuxiliary` the reviewer noted the guard
was blind to — and the guard pins both atoms. Proven red against both mutations,
then restored:

```
# mutation A: deferred door given its own inline content-hash rule
× keeps ONE content-match rule and ONE coverage predicate
  expected [ …(3) ] to have a length of 2 but got 3

# mutation B: snapshot half dropped from the union
× keeps ONE content-match rule and ONE coverage predicate
  expected [ Array(1) ] to have a length of 2 but got 1
× a deferred scanner covered only by the PRE-NOTIFY snapshot stays covered
```

The missing behavioral test for the snapshot half is added, and mutation B is
what proved it non-vacuous — the first version of that probe passed under the
mutation because an instantly-publishing primary let the deferred touch merge
before the holding write cleared the cache, so the LIVE half was answering it.
The delayed primary is load-bearing and now says so in the fixture.

The guard's strength is symbol-level; see the AC3 bullet above for what that does
and does not cover.

### F3 (medium) — a fifth door on the old rule

Confirmed: `mergeBinding` filtered the raw deferral set, so a deferred-but-
covered scanner's fingerprint was excluded while its findings rode along in
`.diags`. Routed through `droppedAuxiliaryServerIds` — a dropped contributor
loses its findings and its fingerprint together. The "every door" claim in this
body, the changelog fragment, and AGENTS.md is accurate as of this commit.

### F4 (low) — AGENTS.md

Rewritten. The door count is gone in favor of naming which doors read the freeze
and which read the rule live; the screen now reads "never earlier and never
later", with both failure directions named and F1's overclaim called out as the
worse one.

### F5 (low) — `closes` vs `refs`

Re-judged after the fixes: `closes` stands. All three acceptance criteria are
met, AC3 by construction as well as behaviorally, and the fifth door F3 found is
routed — so there is no remainder to name on the issue.

### What verification established

- Both round-1 probes pass on this head and fail on `7d3e561b`, so the blocking
  regression is genuinely closed rather than papered over.
- F3's binding-door divergence reproduces on `upstream/master` too. That door was
  already wrong before this branch existed: this PR closes a PRE-EXISTING defect
  there, it did not introduce one.
- Wait-outcome telemetry residual, recorded as an observation and not a defect: an
  `lsp_aux_wait_outcome` row can honestly read `publishedThisContent: true` beside
  a gap row naming the same server, because the two describe different instants.
  Join on the TOUCH, not the row. AGENTS.md already documents this.

### Not changed

Nothing was re-litigated from the reviewer's confirmed positives. `clients/lsp/
client.ts` is untouched. AGENTS.md was verified accurate on every load-bearing
claim in this round and needs no further edit.

## Tests run

- `tests/clients/lsp/` — 704 passed, 10 skipped (75 files)
- referencing + downstream consumer suites (`cascade-compute`,
  `warm-attach-confirmation`, `tools/lsp-diagnostics`, `latency-logger`,
  `dispatch/`, `cascade-format`, `mcp/ipc`, `warm-attach`, `cascade-turn-merge`,
  `actionable-warnings-lsp-cache`) — 1012 passed, 2 skipped (106 files)
- `npm run lint`, `npm run changelog:check`

## Merge order

No file overlap with any open PR. #1573 (fix/1561) touches `tools/lsp-diagnostics.ts`
and unrelated `clients/` files; its suites are green here. #1583 (fix/1570) also
edits `clients/lsp/index.ts` but in the lazy-import/repair-rearm region — either
order merges, but whichever lands second should re-run `tests/clients/lsp/`.

Refs #1586, refs #1571, refs #1549, refs #1459, refs #1493.

## Maintainer addition (2026-08-19)

The two response commits from the last review round were lost to a
force-push over a rebase (same commit messages, so invisible from the
outside — see the review comments for the diagnosis). Landed here
instead, with the contributor's authorship on every commit but this
one preserved:

- **FIND-1**: pinned the `unconfirmedServerIds` reconciliation filter
  (`auxCoveredAtMerge`, `clients/lsp/index.ts`). Red on the deleted
  filter: `expected [ 'opengrep' ] to not include 'opengrep'`.
- **F3**: pinned the merged binding's coverage source
  (`droppedAuxiliaryServerIds` in `mergeBinding`'s call site). Red on
  reverting to the raw deferral set: expected the scanner's content
  hash, got `undefined`.

Both proven red against the pre-fix lines the review named, green
after. New commit: `test(lsp): pin FIND-1 and F3 from the #1586
review round`.

Merged `origin/master` in (branch was ~65 commits behind by the time
of this addition); the merge was clean, no conflicts. Re-ran
`tests/clients/lsp/` (77 passed, 5 skipped) plus the coverage-gap
downstream consumers `tests/tools/lens-diagnostics.test.ts` and
`tests/scripts/analyze-pi-lens-logs.test.ts` (all passing) against the
merged tree. `npm run build` and `npm run lint` both clean.

All 8 CI checks pass on the new head; no `action_required` gate hit.


apmantza (changes requested):
Thank you for this.

Before anything else: this review is AI-generated, and a maintainer supervises the process and makes the merge decision.

The fix is correct and I could not break it. I independently reproduced your red run: both acceptance-criteria tests fail on the base commit `2e44576b` with the messages you quote, and both F1 probes fail on `7d3e561b` and pass on `fbb2ca11` — so the blocking regression really is closed. The single-predicate property holds by inspection: `touchContentHash` has exactly one comparison site, the snapshot has exactly one read site, and I confirmed the structural guard goes red when I gave the deferred door its own inline rule. There is no `await` between the freeze and the drop. Your disclosure about what the guard does and does not fence is accurate rather than overclaimed, which I appreciate.

Two things to add before merge, both about test coverage rather than behavior:

1. The reconciliation filter on `unconfirmedServerIds` (`clients/lsp/index.ts:4113`) is the line F1's fix rests on, and no test can fail it — I deleted it and all 732 tests in `tests/clients/lsp/` still passed. It is reachable: a publication landing between the aux wait-outcome row and the merge freeze puts the scanner in `auxUnconfirmedServerIds` while the merge keeps its findings. The probe below reproduces it (append to a copy of the harness in the first ~231 lines of `service-scanner-coverage-gap.test.ts`); with the filter deleted it fails with `expected [ 'opengrep' ] to not include 'opengrep'`. Please pin it.

<details><summary>FIND-1 probe (reuses the existing harness)</summary>

```ts
describe("publication lands between the aux wait outcome and the merge", () => {
	const FILE = `${ROOT}/a.ts`;
	const CONTENT = "const value = 1;";

	beforeEach(() => {
		vi.useFakeTimers();
		vi.resetModules();
		getServersForFileWithConfig.mockReset();
		createLSPClient.mockReset();
		logLatency.mockReset();
		process.env.PI_LENS_LSP_NOTIFY_BUDGET_MS = String(NOTIFY_BUDGET_MS);
	});

	afterEach(() => {
		vi.useRealTimers();
		vi.restoreAllMocks();
		delete process.env.PI_LENS_LSP_NOTIFY_BUDGET_MS;
	});

	it("does not name a scanner the merge kept", async () => {
		const { LSPService } = await import("../../../clients/lsp/index.js");
		const { hashDiagnosticContent } = await import(
			"../../../clients/lsp/diagnostic-binding.js"
		);
		const service = new LSPService();
		const hash = hashDiagnosticContent(CONTENT);

		let published = false;
		logLatency.mockImplementation((entry: { phase?: string }) => {
			if (entry?.phase === "lsp_aux_wait_outcome") published = true;
		});

		const auxBase = makeClient("opengrep", 0, [], "never");
		const aux = {
			...auxBase,
			getDiagnostics: vi.fn(() =>
				published ? [makeDiagnostic("late scanner finding")] : [],
			),
			getDiagnosticBinding: vi.fn(() =>
				published ? { contentHash: hash } : undefined,
			),
		};
		const primary = makeClient("typescript", 0, [makeDiagnostic("primary")]);
		getServersForFileWithConfig.mockReturnValue([
			makeServer("typescript"),
			makeServer("opengrep", "auxiliary"),
		]);
		createLSPClient.mockImplementation(async (options: { serverId?: string }) =>
			options?.serverId === "opengrep" ? aux : primary,
		);

		const touch = service.touchFile(FILE, CONTENT, {
			clientScope: "with-auxiliary",
			auxiliaryServerIds: ["opengrep"],
			diagnostics: "document",
			collectDiagnostics: true,
			source: "cascade",
		});
		await vi.advanceTimersByTimeAsync(NOTIFY_BUDGET_MS * 200);
		const result = (await touch) as {
			diags: Array<{ message: string }>;
			confirmation?: string;
			unconfirmedServerIds?: string[];
		};

		expect(result.diags.map((d) => d.message)).toContain("late scanner finding");
		expect(result.unconfirmedServerIds ?? []).not.toContain("opengrep");
		expect(result.confirmation).toBe("confirmed");
	});
});
```

</details>

2. F3's merged-binding fix (`clients/lsp/index.ts:4284`) has no regression test either. Reverting it to the pre-fix raw-deferral filter leaves the suite green. Since the PR claims it closes a pre-existing defect, that claim needs a red test to stand — the divergence is narrow (a deferred-but-covered aux whose fingerprint is the only one, with a version-less primary), which is exactly why it wants pinning rather than trusting.

One low-severity note, no action needed: the primary/never-spawned fail-closed branch in `auxCoversThisContent` is also untested, but `brokenSkippedAuxiliaryServerIds` builds its list by excluding `spawned`, so the guarantee holds by construction.

The branch is four commits behind `master`. The files do not overlap, so the merge is clean, but please rebase so CI runs on the current base. With the two tests added (both proven red against the unfixed lines), this is merge-ready from my side.

AngriestBird:
@apmantza Rebased and ready to go.

apmantza (commented):
Thanks for the rebase — the branch merges cleanly against a master that has moved a long way, and all eight checks are green on `1d25c9b2`, so that part is in good shape.

This review is AI-generated, and a maintainer supervises the process and makes the merge decision.

I need to flag something before you spend more time: I think the two response commits did not make it. `d64476eb` and `1d25c9b2` are the same two commits as before, rebased — `git diff` between the old head and this one shows `tests/clients/lsp/service-scanner-coverage-gap.test.ts` byte-identical, and the only `clients/lsp/index.ts` delta is #1583's `resetClassicTsRepairGuard` arriving from master. Easy thing to lose to a force-push over a rebase, and the identical commit messages make it invisible from the outside. Worth checking whether the work is sitting on a local branch.

Both findings are unchanged. I re-ran the mutations against a locally merged tree, so this is against current master rather than the old base:

- Deleting the reconciliation filter at `clients/lsp/index.ts:4190` leaves **1174 tests passing**. That line is what F1's fix rests on.
- Reverting the merged-binding guard at `clients/lsp/index.ts:4361` also leaves **1174 tests passing**, so the pre-existing-defect claim still has no red test behind it.

Two notes that may help when you pick this back up. First, the seam matters more than it did a week ago: `touchCoverageGap()` is now the input to #1626's `unconfirmedReason: "coverage_gap"` and to #1561's inline-blocker retire gate, so this PR's narrowing of `unconfirmedServerIds` reaches further downstream than it used to. I checked the direction and it composes correctly — you only drop a scanner whose findings the merge kept — but that is more reason to pin it. Second, the probe I shared in the earlier review still applies unchanged against the merged tree.

Nothing else came up. The AGENTS.md paragraph merges cleanly beside the new shape 15 with no overlap, and the merged tree is green across `tests/clients/lsp/` (748) and every downstream consumer suite (458).

apmantza (approved):
This is merge-ready. Thank you for the whole arc — the original fix was correct on the first pass, the review round closed a real blocking regression with its own probes, and the two guarantees that were left unpinned are now pinned (a maintainer commit restored the tests that were lost to the force-push; your authorship on all prior commits is untouched).

As before: this review is AI-generated, and a maintainer supervises the process and makes the merge decision.

I verified both new tests by mutation rather than by reading them. Deleting the reconciliation filter reds the FIND-1 test alone, with the exact message the commit claims. Reverting the merged-binding guard reds the F3 test alone. Neither is a lookalike, and neither is vacuous.

The F3 test is better than the probe I shared. Leaning on the version-less primary so the covered auxiliary is the only fingerprint contributor is the detail that makes the assertion mean something — a bound primary would have hidden the loss behind its own binding.

Everything else holds on the merged head: 865 tests green across `tests/clients/lsp/` and the downstream consumers that read `unconfirmedServerIds` through `touchCoverageGap()`, including the newer `unconfirmedReason` derivation. All three acceptance criteria on #1586 are met, so `closes` is accurate with nothing left to re-home.

Issue #1609, opened 2026-08-18
Testing umbrella: prospective test classes for defects our suite structurally cannot see


apmantza:
From a maintainer-supervised AI retrospective. Sibling of #1605 (which covers lanes for defects already observed); this umbrella targets classes where nearly every bug so far was found by a human dogfooding, a reviewer probing, or log forensics — never by a test, because the suite cannot express the failure. Six classes, priority-ordered by expected bugs caught per effort. Items 1 and 3 dispatched first.

## 1. State-machine property tests for the availability latch
The latch family produced ~8 distinct shipped bugs this arc (#1490/#1494/#1495/#1496/#1537/#1556-F1/#1577-F5/#1593), each found individually; all violate a small invariant set. Model-based test: generate random sequences of probe outcomes (transient/durable/success), time advances, and session boundaries; after every step assert: never durably latched on transient-only evidence; re-arm reachable at session_start; cooldown ladder monotone and capped; recovery reachable from every state; degradation recorded exactly once per episode (recordDegradationOnce semantics). Seeded/deterministic (no wall clock), with failing-sequence shrinking or at minimum seed replay.

## 2. Hostile-LSP-server fixture harness
The staleness/verdict family (#1531/#1561/#1573/#1549/#1571/#1462) are protocol-timing bugs; unit mocks are well-behaved by construction. One scriptable misbehaving server: stale-version publishes, out-of-order answers, duplicate publishes, publishes for never-opened URIs, silence after initialize, mid-request crash, malformed JSON-RPC. Scenario matrix asserting verdicts stay honest (deferred/cut_off/inconclusive — never confidently wrong).

## 3. Crash-atomicity of persisted state
The extension host genuinely gets SIGKILLed in the field (Node auto-upgrade RestartManager kills — pi-free crash forensics). Two layers: (a) a sweep test asserting every persisted-state write (probe-cache.json, degradation ledger, instance registry, worklog.jsonl, caches) goes through atomic tmp+rename — enumerate write sites from the code, not a hand list; (b) fault injection: truncate each state file at arbitrary offsets and assert the next session_start recovers legibly (no throw, no half-trusted torn read, no stuck latch). The instance-reaper staging-file flake (2026-08-18) suggests the staging discipline is already shaky.

## 4. Transient-filesystem hostility (OneDrive reality)
[...]

Issue #762, opened 2026-07-22
refactor: shared writeFileAtomic helper for the five hand-rolled tmp+rename writers


apmantza:
## Why

The `${target}.tmp-${pid}` + `renameSync` atomic-write pattern is now hand-rolled in five places, with subtly divergent error policy:

- `clients/instance-registry.ts` — best-effort (swallows failures)
- `clients/session-state-store.ts` — best-effort
- `clients/recent-touches.ts` — best-effort
- `clients/review-graph/builder.ts` — best-effort
- `clients/diagnostic-dispositions.ts` (#757) — throws on rename failure after best-effort tmp cleanup, deliberately: disposition writes are load-bearing (a lost mark is a correctness bug), unlike the observability-flavored writers above

Five copies of the same shape invite drift (e.g. forgetting tmp-file cleanup on the failure path), and the #690 review CLI is about to add more cross-process writers, so the copy count is going up.

## Proposed shape

One helper in a shared module (e.g. `clients/file-utils.ts` or a new `clients/atomic-write.ts`):

```ts
writeFileAtomic(targetPath: string, data: string, options?: { bestEffort?: boolean }): void
```

- Always: write `${target}.tmp-${process.pid}`, `renameSync` over target, best-effort `rmSync(tmp, { force: true })` on rename failure.
- `bestEffort: true` (default matches the majority): swallow all failures. `bestEffort: false`: rethrow the rename error (the #757 disposition-store policy).
- Document why per-write-site policy matters, citing the #757 CHANGELOG entry's load-bearing-vs-observability distinction.

Migrate all five sites; their existing tests (e.g. `tests/clients/diagnostic-dispositions.test.ts`'s no-tmp-left-behind and torn-read assertions) carry over as the safety net. Good first issue-sized, independent of everything else in flight EXCEPT any #690 review-CLI work that adds new writers — those should land on the helper, not add a sixth copy.

PR #1619, opened 2026-08-18
fix(installer,instance-registry): close crash-atomicity gaps (refs #1609)

apmantza:
## What

Item 3 of the #1609 testing umbrella (crash-atomicity of persisted state).
Items 1/2/4/5/6 are out of scope for this PR.

**Layer a — derived write-site sweep.** `tests/clients/atomic-write-sweep.test.ts`
+ `tests/support/atomic-write-scan.ts` enumerate every raw `fs.writeFile*`/
`fs.appendFile*`/`createWriteStream` call under `clients/` at test-run time
(not a hand-maintained list) and require each site to either target an
obviously-scratch temp path, or be listed in a reasoned exemption table, or
route through the shared atomic tmp+rename seam (`clients/atomic-write.ts`,
#762). Clearance is per SITE, not per file — a whole-file "this file imports
the atomic seam somewhere" check would have missed the real findings below,
because `installer/index.ts` already imports `commitDurableStoreAsync` from
`durable-store.js` for probe-cache while ALSO containing unrelated raw writes
elsewhere in the same 3000+ line file.

**Violations found and fixed (contained, single-file):**
`clients/installer/index.ts` wrote 9 downloaded tool artifacts straight to
their final path instead of through `writeFileAtomic*`: bare-gzip
decompressed binaries, bare binaries, Maven jars, Maven launcher scripts,
archive-install shims, the npm-tool bootstrap `package.json`, and a GitHub
"extra asset" sibling file (e.g. ktlint's jar). A process killed mid-install
(the Node auto-upgrade RestartManager kill from the pi-free crash forensics,
or any hard kill) could leave a corrupt binary or launcher behind. All nine
now go through `writeFileAtomicAsync`, which gained an optional `mode` option
(`WriteFileAtomicOptions.mode`) so an executable's permission bit still
travels with the staged file across the rename. Per-invocation scratch writes
(tar/zip temp archives, ast-grep/gitleaks/sg-runner/psscriptanalyzer config
scratch files, lock-file O_EXCL creates, append-only NDJSON logs whose
readers already tolerate a torn tail) are listed with reasons in
`EXEMPT_RAW_WRITE_FILES` rather than converted — converting them would add
tmp+rename overhead with no crash-safety benefit for state nothing re-reads
across a session boundary.

**Layer b — fault injection.** Truncating persisted-state files at multiple
offsets (zero bytes, mid-JSON torn tail, truncated-to-brace, wrong-shape
scalar) surfaced two read-side gaps in the same class:

- `clients/instance-registry.ts` degraded a corrupt/torn `instances.json` to
  empty exactly like a genuinely missing file — a real torn-write regression
  would have looked identical to a normal clean start in the logs, violating
  the "empty must distinguish clean from errored" invariant. A non-`ENOENT`
  read failure (or a valid-but-wrong-shape parse, which a torn tail can
  produce) now logs one `instance-registry: read failed (...)` line via
  `logSessionStart`; a genuinely missing file stays silent.
- `clients/installer/index.ts`'s `deserializeProbeCache` — used only by the
  LOCKED write-side merge (`writeProbeCache`'s `commitDurableStoreAsync`
  call), not the ordinary session-lookup read (`readProbeCache`, which
  already degraded correctly) — threw on a torn/corrupt `probe-cache.json`.
  This didn't crash the caller (already wrapped in try/catch), but it also
  meant the corrupt file could never be repaired: every retry re-read the
  same torn bytes and re-threw, forever. It now degrades to `{}` like its
  siblings (`diagnostic-dispositions.ts`, `actionable-warnings.ts`) and logs
  the recovery.

Spot-checked (not converted) append-only-log exemptions against their actual
readers: `fix-worklog.ts`'s `readWorklog`, `project-changes.ts`'s
`parseChangeLine`, and `smells-rollup.ts` all try/catch per line and drop an
unparseable trailing line rather than throwing or half-trusting it —
confirming the exemption reasoning, not just asserting it.

## Findings not fixed here

None outstanding — every raw-write violation the sweep found was a contained,
single-file fix and is included above. `partial-edit-apply.ts`/`lsp/edits.ts`
(writes to the USER's own source files via LSP `applyEdit`) are a different
class from this issue's scope (pi-lens's own persisted state under
`~/.pi-lens`) and are listed as an exemption with that reasoning, not folded
silently into "fixed" — worth its own issue if the class is judged worth
pursuing, not filed here since it's a design question (editor/VCS recovery
vs. pi-lens's own atomicity discipline) rather than a contained fix.

The instance-reaper staging-file flake mentioned in the issue
(`tests/clients/instance-reaper-prune-concurrency.test.ts`, "left staging
files after a timeout") did not reproduce here — `pruneDeadInstances` already
routes through `writeFileAtomicAsync` (fixed at #1217) and its concurrency
test passed cleanly in every run during this work, including after the
sweep/fault-injection additions. No code change made for it; flagging in case
it recurs as an actual CI flake rather than a one-off environmental timeout.

## Verification

- `npm run build` before every test run (compiled `clients/*.js` is the
  runtime).
- Both new regression classes proven RED on pre-fix code via `git diff` /
  `git checkout --` / `git apply` (never stash):
  - `tests/clients/installer/probe-cache.test.ts`'s new
    `self-heals a torn write-side read` cases: 3 failed with `expected
    'failed' to be 'written'` against the pre-fix throwing
    `deserializeProbeCache`.
  - `tests/clients/instance-registry.test.ts`'s new `torn-file fault
    injection` describe block: 4 failed with `expected "vi.fn()" to be
    called ... Number of calls: 0` against the pre-fix silent
    `readRegistrySync`/`readRegistryAsync`.
  - `tests/clients/atomic-write-sweep.test.ts` against pre-fix
    `installer/index.ts`: failed, listing exactly the 9 real violation sites
    (plus correctly ignoring the tmp-archive scratch writes).
  - All three go green after the corresponding fix, with no other change to
    the test.
- Targeted run after every change plus the full merge: installer test suite
  (12 files, 158 tests), instance-registry/instance-reaper/atomic-write test
  families (12 files), `tests/config/module-instance-binding.test.ts`,
  `tests/index-integration.test.ts`, and every test file referencing
  `installer/index.js` outside `tests/clients/installer/` (dependency
  checker, availability latching, dispatch runners, lsp discovery/policy,
  sg-runner, formatters, jscpd/knip clients, etc.) — 25+ files, 500+ tests,
  all green. `tests/clients/installer/tool-discovery.test.ts` and
  `tests/clients/installer/version-drift.test.ts` needed a matching
  `atomic-write.js` mock added (same pattern `probe-cache.test.ts` already
  used) since they mock `node:fs/promises` directly and the shared seam
  imports `node:fs` — without the mock, `writeFileAtomicAsync` fell through
  to real disk I/O against those tests' mocked, non-existent-on-disk paths.
- `npm run lint` (tsc full project) — clean.
- `npm run changelog:check` — clean, 37 entries validated including this
  one.
- Full suite deferred to CI (repo policy: targeted runs while iterating, full
  suite is CI's job, machine-wide serialized per #1112).

Branched from and merged current `origin/master` (fast-forward, no
conflicts) before opening this PR — several PRs are landing today.

Refs #1609


apmantza:
Review round F1-F4 addressed on this branch, head e9d22ecb.

- F1 (blocker): `runtime-session.ts`'s session_start orphan sweep now includes `getGlobalPiLensDir()/bin` and `/tools` — `sweepOwnStagingFiles` doesn't recurse, so an orphaned staging file from a kill mid-install was never reaped. Regression test in `tests/clients/runtime-session.test.ts` (F1 describe block) plants a dead-pid orphan in each directory and proves it's reaped after `handleSessionStart`; proven red on pre-fix `runtime-session.ts`.
- F2: the sweep's transient-target heuristic now judges only the call's TARGET (first) argument, extracted with bracket/quote-depth tracking, not the whole matched line — closes the `fs.writeFileSync(registryPath(), tmpData)` smuggle. Negative fixture added in `tests/clients/atomic-write-sweep.test.ts`, proven red on the pre-fix line-wide check.
- F3: the receiver pattern is now derived per-file from that file's own `node:fs`/`node:fs/promises` import bindings (`nodeFs`, `fsPromises`, etc.), not a hardcoded `fs`/`fsp` pair. Negative fixture added for the `nodeFs.writeFileSync(...)` smuggle, proven red on the pre-fix hardcoded pattern.
- F4: added POSIX-gated (`skipIf(win32)`) `statSync(...).mode & 0o777` coverage for `writeFileAtomic`'s new `mode` option, sync and async. Confirmed on this Windows dev machine that Node's `fs` mode bits are not POSIX-meaningful here (0o750 and no-mode both report 0o666), which is exactly why the test is gated rather than run everywhere — it will execute for real on CI's Linux Unit-tests job.
- Small: `instance-registry.ts`'s corrupt-read observability now goes through `recordDegradationOnce` (new `instance-registry-corrupt` kind) instead of logging on every call — `readRegistrySync`/`readRegistryAsync` are called from many sites across a session, not only session_start. Added a "records only once across repeated reads" regression test; both new/changed tests proven red against the pre-fix logSessionStart version.

All five fixes are red-proven pre-fix via `git diff`/`git checkout --`/`git apply` (no stash), each restored to green after. Full targeted set re-run after the `origin/master` merge (24 files, 326 tests, 3 POSIX-only skips on this Windows box) plus `npm run lint` — all green. Pushed; `Unit tests` and `Lint & type-check` both executed (not skipped) and passed on e9d22ecb.

Issue #1622, opened 2026-08-18
Secrets blocker replays pre-edit line numbers for 30 minutes — the gitleaks lane has an existence gate but no freshness gate (refs #1461)


apmantza:
Root-caused by a maintainer-supervised AI forensics pass on the 2026-08-18 dogfood session (build b1a497c4, which carries #1460's existence gate). The agent saw gitleaks citations at pre-edit positions (src:397 / dist:234) for ~13 minutes after editing both files; ≥16 stale citations served across 8 turn-end deliveries; 5 agent-visible replays; the replay stopped exactly at TTL expiry (07:50:39).

## Root cause

clients/runtime-turn.ts:1052 gates the 30-minute gitleaks cache with dropFindingsForMissingPaths, which probes existence only. GitleaksFinding carries no per-entry observedAt, and the cache envelope's scannedAt is never compared to the cited file's mtime — a file edited after the scan serves its pre-edit startLine until TTL expiry. The provenance guard could not catch it: its envelope covers files edited this turn plus cascade neighbours, and paths cited INSIDE a finding never enter it (18/18 decisions said "current" — correctly, about the wrong files). pi-lens had the invalidation signal the whole time: change-log.jsonl recorded four agent-edit entries with changedRange {397,400} for the exact file, in the same project-state directory as the stale cache.

Siblings in the same function share the shape: trivy secrets (:1063), trivy CVEs, govulncheck, dead-code.

Secondary harm, first live instance of #1461's prediction: 2/5 entries in the project's diagnostic-dispositions.json are stale-snapshot suppressions, polluting the #181 rule-tuning corpus.

## Fix

Extend the partition helper to a freshness verdict: one memoized statSync per cited path; compare stat.mtimeMs to scannedAt. Deleted path → drop (as today). Path modified after the scan → DEMOTE out of the blocker tier, do not drop — the finding may still be real at a shifted line, and dropping would let an edit silence a genuine secret. Render demoted entries without a line number or with "[stale — re-run to confirm]" per the #1419 precedent.

## Acceptance criteria

1. A gitleaks finding whose cited file has mtimeMs > scannedAt does not appear in the STOP blocker with its cached line number.
2. Deleted paths still drop (no #1460 regression).
3. Unmodified paths still deliver as blockers at full severity.
4. Trivy secrets get the same gate in the same change — same seam, same defect.
[...]

PR #1633, opened 2026-08-18
fix(diagnostics): demote dependency-drifted cached blockers (closes #1631)

### LINKED ISSUE #1631: Cached cross-file blockers replay for the whole session after out-of-band dependency fixes — live instance of #1561's dependency-axis remainder + mode=all replay surface (refs #1461, #1573)
## Summary

Live session evidence (pi-lens **4.0.1**, pi 0.84.2, macOS, TS/NodeNext project) of #1561's defect class with **two facets #1561 does not cover**:

1. **Out-of-band dependency fixes are structurally unreachable by every existing invalidation mechanism.** The missing export was added to the dependency file via a bash script (no `edit`/`write` tool) — the file was **never dispatched**, so neither "later dispatch of the same path returns no blockers" nor cascade could ever fire for it. The consumer file's blocker then re-served for the rest of the session.
2. **The replay surface is wider than the inline-blocker map.** The phantom survived in the session diagnostics cache that feeds `lens_diagnostics mode=all` and the turn-end advisory — demonstrably divergent from live state: a `mode=full` active sweep over the same files reported **LSP primary findings: 0** while `mode=all` still served the blocker in the same minute.

Also contributes: a second live instance of #1561's *blocked* remainder (dependency axis) where the dependency **was** dispatched clean and the consumer still never got re-verified — and cascade logs showing both self-healing paths closed simultaneously.

Note on version: PR #1573 (confirmed-clean retires inline blockers) merged 2026-08-18T05:52Z, **after** v4.0.1 shipped (2026-08-16T20:50Z), so this session predates it. Even with #1573, facets 1–2 above remain: it retires on an affirmative `lsp_diagnostics` verdict for the *consumer* file, but (a) nothing forces that verdict to be re-taken when only the *dependency* changes, and (b) the actionable-warnings cache behind `mode=all` is a separate store.

### Case A — `git-env.test.ts` / `worktree.ts` (dependency dispatch exists, consumer never re-verified)

| Time | Event |
|---|---|
| 11:12:45.854 | `write` tool creates `tests/pipeline/git-env.test.ts` (imports `gitEnv` — not yet exported; the export landed ~17s later via bash) |
| 11:12:45.981 | `dispatch_start` git-env.test.ts |
| 11:12:46.538 | dispatch result: **blockers: 1** / total: 1 → "no exported member 'gitEnv'" recorded (observation: true at that instant) |
| 11:12:46.539 | cascade.log: `"reason":"primary_has_blockers"` — cascade skipped because the edited file itself had blockers |
| 11:13:02.992 | `dispatch_start` worktree.ts (the dependency) |
| 11:13:04.024 | worktree.ts result: **blockers: 0** / total: 9 — dependency clean, export present |
| 11:13:04 → 12:00+ | **no further dispatch of git-env.test.ts**; blocker re-served at every turn end (~50 min) |
| ~11:55 | `lsp_diagnostics` on git-env.test.ts → "primary typescript LSP: confirmed clean" — advisory still re-served on subsequent turns (4.0.1 has no #1573-style retirement; whether any store consumed this verdict: unknown) |

### Case B — `project-path.test.ts` / `project-router.ts` (dependency never dispatched)

| Time | Event |
|---|---|
| 11:53:41.410 | single dispatch of project-path.test.ts (file created via bash append) |
| 11:53:43.275 | **blockers: 1** — "no exported member 'projectPathFromUrl'" recorded |
| ~11:54 | the export added to `project-router.ts` **via bash script — file never dispatched the whole session** (no entry in latency.log) |
| 11:53 → 12:10+ | blocker re-served every turn end; `mode=all` still serves it |
| 12:0x (same minute) | `mode=full paths=[test, source, 2 related]` active sweep: **"LSP sweep findings: 0 primary (language server)"** — cache/live divergence on identical input |

Two sibling files (`isPatchNoise`, `summarizeUnparseable`) reproduced the identical shape later the same day: bash-side export → test-file advisory replay until session end. 4 independent instances in one session.

## Harm

Same shape as #1561: multiple agent turns spent debunking (grep the export, run the importing tests green, run `lsp_diagnostics`) while the advisory text asserts "You should fix them before completing the task" — pressuring the agent to "fix" a non-problem or distrust the tool. In this session the agent settled on treating `pnpm typecheck` + the test suite as ground truth and ignoring pi-lens advisories wholesale — i.e., the feedback loop was abandoned, which is the terminal failure mode of a feedback tool.

## Analysis

Three independent gaps compound; each is sufficient alone to keep the phantom alive:

1. **Invalidation keys are per-file; the verdict's truth is cross-file.** The blocker on file F is a statement about F *and everything F imports*. Existing invalidation events (same-path clean dispatch, existence reconcile, #1573's confirmed-clean retirement) all key on F alone. When only a dependency changes, no event fires. This is #1561's blocked "dependency axis", observed live a second time — and Case B shows the strongest form: when edits bypass the tool pipeline entirely, even the same-path event is unreachable by construction.
2. **No out-of-band change detection.** Files modified outside the `edit`/`write` dispatch pipeline (bash scripts, formatters, `git checkout`, subagent worktrees) drift from both the LSP document state and the dispatch cache. Nothing re-syncs them: no mtime/content-hash sweep at turn boundaries over session-dispatched files. [inference] This also creates a subtler hazard any "just re-check before replay" fix must handle: re-querying the LSP while its in-memory document for the dependency is stale will **reproduce the stale verdict**, not correct it — revalidation must be paired with document re-sync (mtime-driven reload), or it becomes a stale-verdict regeneration path.
3. **The replay tier is broader than the inline-blocker map.** #1573 retires entries of `_pendingInlineBlockers`. The turn-end advisory injection and `lens_diagnostics mode=all` read the session-wide actionable-warnings cache, which has no freshness key and no retirement wiring (Case B: served in the same minute a fresh active sweep contradicted it). agent-guide.md documents `stale = "A cached diagnostic's file changed on disk since the scan"` — the staleness check covers the diagnosed file only, not its import closure, and did not fire here anyway since the consumer file was untouched.

Also noted: cascade.log `"reason":"primary_has_blockers"` means a file carrying blockers gets no cascade re-analysis — by design, but it closes the second self-healing path exactly when it is needed (broken file imports a fixed dependency → neither the file nor its neighbourhood is ever re-pulled).

## Fix directions (proposals — maintainer's call)

Ordered by reachability, mirroring #1561's two-axis structure:

1. **Turn-boundary freshness sweep over cached blockers (index-free).** Before re-serving any cached blocking finding at turn end, stat the file and its forward imports (the file's own import list is known to the LSP/parse layer — no reverse-dep index needed, sidestepping #1561's tests-free-index blocker). Any path with mtime > verdict time → re-pull LSP diagnostics for that file *after* re-syncing stale documents, then update or retire the entry. Blocking entries are rare, so the cost is bounded. This covers both cases above and the mid-refactor window (transient errors may still be recorded once — unavoidable — but never re-served without revalidation).
2. **Out-of-band resync.** Same sweep doubles as change detection: dispatched files whose disk content hash drifted without a didChange get re-opened to the LSP before any verdict reuse. Without this, fix 1 can re-derive stale verdicts (see Analysis #2).
3. **Store breadth.** Whatever retirement/freshness semantics land, apply them uniformly to every agent-facing store (inline-blocker map, actionable-warnings cache behind `mode=all`, turn-end advisory), or document which store is authoritative and which may lag.

## Acceptance criteria (red-first)

1. Consumer file F dispatched with blocker "no exported member X"; dependency G modified out-of-band to export X; next turn end does **not** re-serve F's blocker (either re-pulled clean, or demoted with a `[stale — re-run to confirm]` marker per #1419 precedent).
2. Same sequence with G modified via `edit` tool (dispatched clean) — same outcome (Case A shape).
3. `lens_diagnostics mode=all` output never contains a blocking entry contradicted by a same-minute `mode=full` active sweep on the same paths.
4. No regression to #1561/#1573 semantics: a fresh confirmed-clean for F still retires, ordered by writeIndex.
5. A latency.log record per turn naming freshness-revalidation counts (revalidated/retired/kept), so future incidents are reconstructible from logs alone.

Refs #1561 (dependency-axis remainder, live instance 2), #1461, #1573, #1622 (same freshness-gate defect shape on the gitleaks lane), #1630 (consumer-side import-change cache miss).

EarthChen:
## Purpose

Fixes the replay half of #1631: a cached **blocking** finding on a file F is a verdict about F *and everything F imports*, but every existing invalidation path keys on F alone. When the cause is fixed in a dependency G — the live instances in the issue, four in one session, one where G was edited via a bash script and never dispatched at all — no invalidation event fires and the stale verdict re-serves at every turn end and on every `lens_diagnostics mode=all` for the rest of the session.

Per the issue-author scope clarification, the gate lives at the **cache layer**, covering **both** agent-facing stores that serve cached blocking findings — the turn-end inline-blocker store and the widget store that `mode=all` reads — so acceptance criterion 3 is covered, not deferred.

## What changed

- **`clients/blocker-freshness.ts` (new)** — `sweepInlineBlockerFreshness` runs at turn end before cached inline blockers are re-served. It stats the file **and its forward imports** (resolved via the shared tree-sitter client + the review-graph import resolvers — no reverse-dependency index, so the tests-free-index blocker that holds #1561's full invalidation does not apply) against a new `recordedAtMs` baseline on the record. On drift it demotes via the new `RuntimeCoordinator.markInlineBlockerStale`.
- **`clients/widget-state.ts`** — `reconcileStaleWidgetDependencyBlockers` applies the same gate to the store feeding `mode=all`, paired with `reconcileStaleWidgetFiles` at the read site. `isBlocking` answers false for a stale finding (the single predicate the footer tally and the mode=all blocking count both consult), so a demoted finding renders as a non-blocking `[stale — re-run to confirm]` entry instead of a 🔴.
- **`clients/runtime-coordinator.ts`** — `InlineBlockerRecord` gains `recordedAtMs`/`stale`; `markInlineBlockerStale` added. Turn-end renders stale entries out of the authoritative blocker channel.
- **`tools/lens-diagnostics.ts`** — `mode=all` runs the widget gate and renders the `[stale]` marker.

Demotion is **demote-not-drop** (#1419 precedent, as applied by sibling #1622): a drifted finding keeps its content and is re-served with a `[stale — re-run to confirm]` marker, not silently deleted. The gate never re-pulls LSP verdicts on its own — re-querying an LSP whose in-memory dependency document is itself stale would regenerate the stale verdict, the exact hazard the issue calls out.

## Observability (criterion 5)

Each turn end logs one bounded `blocker_freshness_sweep` latency phase, and `mode=all` a `blocker_freshness_widget_gate` phase, naming the `revalidated` / `retired` / `kept` counts, so a future replay incident is reconstructible from `~/.pi-lens/latency.log` alone.

## Test evidence (red-first)

17 new tests across four files; all pass. Proven **red** by mutation: disabling forward-import resolution (`imports.slice(0, 0)`) fails all 8 dependency-drift tests (blockers re-served at full authority) while the 9 control/no-drift tests stay green.

- Both live shapes: out-of-band dependency fix (Case B) and dispatched-clean dependency (Case A) — criteria 1 & 2.
- No-drift controls (no false demotion), already-stale idempotency, re-record reset, legacy-unstamped fail-toward-existing.
- `mode=all` output served demoted, not blocking — criterion 3.
- Turn-end re-serve demoted + latency record — criteria 1 & 5.

Full suite: `npm test` → 7213 passed, 4 failed. **All 4 failures are pre-existing and reproduce on clean `master` HEAD** (verified by reverting my change and re-running): 3 are case-sensitivity path tests (`#1024`-family) that fail on this case-insensitive macOS box but pass on the case-sensitive Linux CI, and 1 needs a locally-installed `ast-grep` engine. Zero regressions from this change. `npm run lint` and `npm run build:dist` both clean.

## Deliberately out of scope

- **The full dependency-axis invalidation** (actively re-verifying consumers when a dependency dispatches clean) remains deferred to #1561, which is blocked on a maintainer design decision. This PR is the read-time freshness gate only (fix directions 1+2 of #1631).
- LSP re-pull / document-resync on drift — intentionally avoided (see "never re-pulls" above); a demoted entry is left for an explicit re-run or the next dispatch to resolve.

Refs #1561 (dependency-axis remainder, live instance).


apmantza (commented):
thanks @EarthChen.

This review is AI-generated, and a maintainer supervises the process. Since maintainer edits are enabled and this PR is on the release's critical path, we will push the mechanical fixes below directly to your branch (your authorship on your commits is untouched) rather than round-tripping — the findings are listed so you can see exactly what changed and why.

1. **The PR is conflicted, so CI never ran.** `clients/runtime-turn.ts` collides with master's new `STALE_LINE_MARKER` const from #1627. The resolution is additive (keep both), and we'll import `STALE_LINE_MARKER` instead of re-typing the literal, which now appears four times. Worth knowing: a conflicted PR cannot build its merge ref, so GitHub silently *skips* Unit tests and Lint rather than failing them — only SonarCloud ran on `3e74128a`.

2. **The 1ms tolerance is too small on Windows, and it demotes real blockers.** A file's mtime *leads* `Date.now()` there: 184 of 200 writes measured ahead, up to 11.4ms — so a blocker recorded shortly after its file write demotes itself with zero drift (42 false demotions in 50 runs), and two of your own tests go host-dependent (red 3 of 6 runs on Windows, 6 of 6 green at 50ms). The `+1ms` you copied from `reconcileStaleWidgetFiles` is really there, but it was never load-bearing on this axis.

3. **The demotion never lifts, and it outlives the session.** `exportWidgetState` serializes `stale: true`, `importWidgetState` restores it, and the sweep skips records with no effective blocking finding — so nothing re-promotes. A genuine type error stayed non-blocking across a session resume with the dependency pinned back into the past. `importWidgetState` already solves this shape one field over (it strips failed formatter entries on restore, #1348); `stale` gets the same treatment plus a `WIDGET_STATE_VERSION` bump.

4. **`retired` is always zero** — declared, documented, logged, never assigned. We'll implement it (a demoted entry whose record drops out entirely) rather than drop it.

Two design notes we'll also apply: drift on one import currently demotes *every* blocking finding on the file, including ones unrelated to imports (an ast-grep hardcoded-secret would go with it) — we'll narrow demotion to LSP-sourced findings using the `sources` field the record already carries, and document the trade. And the sweep is unmemoized (~450ms per turn end at 8 blockers × 300 imports; ~435ms per `mode=all`) — a resolve cache keyed on path+mtime makes it near-free.

The honest boundary you drew is worth keeping written down: dynamic `import()` and `require()` do not resolve, so a replay survives for a file whose only dependency arrives that way. We'll name that in the module doc rather than pretend to close it.

Verification rounds continue after the push; nothing here changes your design.

apmantza:
Verified and armed — this lands as layer 2 of the release's staleness arc. @EarthChen, for the record of what happened after your submission (all AI-executed under maintainer supervision): two review rounds ran against your branch; the design survived every probe unchanged, and the maintainer-side commits addressed review findings only — Windows mtime-skew tolerance (+50ms; your +1ms inherited a convention that was never load-bearing on this axis), stale-flag stripping on session import, demotion narrowed to LSP-sourced findings via the sources field, a leaf module for the stale marker (a store-to-orchestrator import cycle), per-turn memoization of import resolution (~450ms → ~75ms warm), and doc alignment. Your forward-imports design, the no-LSP-re-pull discipline, and the demote-not-drop asymmetry are exactly as you wrote them. The import-resolution seam you built handled all 14 specifier classes we threw at it. Thank you — issue and fix together, this is the strongest external contribution this repo has received.

Issue #1656, opened 2026-08-18
Adopt pi's waitForChildProcess construction: post-exit pipe-idle wait with re-armed grace timer (hang-proof AND truncation-proof)


apmantza:
From the pi-host audit (maintainer-supervised AI): pi's waitForChildProcess (pi utils/child-process.ts:38-136) waits after exit for the pipes to fall IDLE with a 100ms grace timer re-armed on every chunk — an actively-writing descendant keeps it reading; a quiet inherited handle (Windows daemonized descendant that never lets close fire) still releases. Regression-tested from both sides upstream (bash-close-hang-windows, 5303-bash-output-truncation).

pi-lens's safe-spawn waits on close (clients/safe-spawn.ts:1358) — the hang side; a fixed post-exit deadline would be the truncation side, and a truncated PROBER result reads as "0 findings = clean" (shape 10). This is strictly better than both naive options.

## Acceptance criteria
- safe-spawn's exit wait adopts the idle-pipe grace construction (bounded overall — compose with existing timeouts).
- Red-first both sides: a daemonized-descendant fixture that would hang the close-wait releases promptly; an actively-streaming child that a fixed deadline would truncate is captured fully.
- Sweep: any other child-wait site not routed through safe-spawn.

Refs #1114, #1605 (lane 7 process accounting), catalog shape 4/10.

apmantza:
Closed by #1673 (merged 2026-08-19): the post-exit pipe-idle wait with re-armed grace is in clients/safe-spawn.ts; all three acceptance items verified on master (2026-09-10 audit).

PR #1673, opened 2026-08-18
fix(safe-spawn): post-exit pipe-idle wait, hang-proof and truncation-proof (refs #1656)

apmantza:
## What changed

`safeSpawnAsync` resolved off Node's `close` event. `close` only fires once
every stdio fd referencing the child has been released — a daemonized
descendant that inherited our stdout/stderr pipe (the Windows no-job-object
case) can hold that fd open forever, so `close` never fires even though the
child we spawned is long dead. The caller hangs indefinitely, and a hung
prober silently loses its result instead of failing loudly (catalog shape 10:
an unresolved/hung producer reads as neither clean nor errored — it just
never reports).

Adopts pi's `waitForChildProcess` construction (`clients/safe-spawn.ts`,
`finalize`/`waitForPipeIdle`):

- After the child settles, wait for stdout/stderr to fall idle — no data for
  `EXIT_PIPE_IDLE_GRACE_MS` (100ms), re-armed on every chunk.
- Cap the total extra wait at `EXIT_PIPE_IDLE_MAX_WAIT_MS` (2s) so a
  pathological never-quiet descendant can't extend the caller's overall
  budget unboundedly — this composes with the existing per-call `timeout`,
  it doesn't replace it.

A quiet inherited handle now releases after one grace window instead of
hanging forever. An actively-streaming child (data still draining from the
OS pipe buffer when it settles) is still captured in full, instead of being
cut short by a naive fixed post-exit deadline.

## Merge-round redesign (this branch was re-derived, not mechanically merged)

PR #1670 landed on master while this branch was open and changed what
`finalize` is wired off of: it moved from `close` to a shape-based decision
(`code === null || code < 0` = a child that never actually launched — Node
reports the negated errno, e.g. -2 on Linux, -4058 on Windows) plus a
`pendingSpawnError`/`resolved` latch so a late unrelated `error` can never
downgrade an already-decided outcome. Its regression tests
(`tests/clients/safe-spawn-close-before-error-race.test.ts`) depend on
`close` firing, because **a spawn failure never emits `exit` at all** —
only `error` then `close(code<0, null)`.

This branch's original design finalized off `exit` only, to fix the
opposite failure: a real, successfully-launched child whose `close` never
fires because a descendant holds its stdio pipe open. Landing that design
unchanged on top of #1670 would make #1670's shape guard unreachable —
confirmed by re-running `safe-spawn-close-before-error-race.test.ts`
against the naive `exit`-only merge: 2 of 4 cases went red (see Red-first
evidence below).

The merge conflict in `clients/safe-spawn.ts` is resolved by wiring
`finalize` off **both** `exit` and `close`, deduped by a `settled` latch
distinct from `resolved`:

- `exit` always fires once the child itself exits, independent of who still
  holds its pipes — this is what unblocks the #1656 hang.
- `close` is still wired because a spawn failure gets no `exit` event at
  all — this is what keeps #1670's shape guard reachable.
- Whichever fires first dispatches `finalize` exactly once; the other is a
  no-op. Double-dispatch would double-run `waitForPipeIdle` and
  `finishResourceUsage` against the same child, so the latch is a
  correctness fix, not just a style choice.

## Red-first evidence

**This branch's own regression** (`tests/clients/safe-spawn-exit-idle-wait.test.ts`,
3 tests, fully mocked `node:child_process`): ran RED against pre-fix
`safe-spawn.ts` during the original round (daemonized-descendant fixture
that emits `exit` but never `close` hangs forever pre-fix).

**Composed-behavior proof for this merge round**, both directions, on the
merged head:

1. Commented out the `await waitForPipeIdle()` call inside `finalize`
   (reverting this branch's own contribution) → `safe-spawn-exit-idle-wait.test.ts`
   went RED (1 of 3: the actively-streaming-pipe case, which depends on the
   idle wait to capture all 5 chunks instead of resolving the instant the
   child settles).
2. Neutralized #1670's shape guard (`code === null || code < 0` →
   effectively `false`) → `safe-spawn-close-before-error-race.test.ts` went
   RED (2 of 4: both cases that rely on the close-event shape deciding
   independent of `error` timing).

Restored both, rebuilt, reran — all green (below).

## Sibling fixtures (from the original round, unaffected by this merge)

Two existing test files mocked a spawned child by emitting only `close`
(never `exit`), which is unrealistic — every real Node child always emits
both, `exit` first:

- `tests/clients/safe-spawn-kill-escalation-timer.test.ts` (2 sites)
- `tests/clients/safe-spawn-windows-env-plumbing.test.ts` (1 site)

Both still green after this merge.

## Sweep: other child-wait sites not routed through safe-spawn

From the original round's grep of `clients/` for `.on("close"|"exit")`
bound to a promise resolution — unaffected by this merge, restated for
continuity:

- `clients/mcp/review.ts:95` (`forkedAnalyze`'s worker) — same shape in
  theory, but bounded today by an independent `timeoutMs` backstop. Left as
  a follow-up candidate, out of scope (this PR is `safe-spawn.ts` only).
- `clients/lsp/launch.ts:437`, `clients/lsp/client.ts:2695-2696` —
  diagnostic/observability only, not wait sites.
- `clients/installer/index.ts` — already listens on `exit`, not `close`.
- `scripts/playground-verify-rule.mjs:170,190` — a dev/CI script (not
  under `clients/`) with no timeout backstop at all; out of scope for a
  `safe-spawn.ts` fix, flagged as a follow-up candidate.

## Verification

- `npm run build` before every test run.
- Full safe-spawn family green on the merged head (12 files, 118 tests):
  `safe-spawn-ambient-signal`, `safe-spawn-close-before-error-race`,
  `safe-spawn-exit-idle-wait`, `safe-spawn-failure-taxonomy`,
  `safe-spawn-kill-escalation-timer`, `safe-spawn-resource-usage`,
  `safe-spawn-sync-throw`, `safe-spawn-windows-classification`,
  `safe-spawn-windows-command`, `safe-spawn-windows-env-plumbing`,
  `safe-spawn-windows-resolution`, `zizmor-config`.
- Also green individually/in small batches: `zizmor-token-latch`,
  `resource-sampler`, `installer/tool-discovery`, `mcp/review-rebuild`,
  `scripts/playground-verify-rule-bounded-wait`,
  `dispatch/runners/runner-helpers`, `formatters-format-file`,
  `formatters-which-latch`, `package-manager`, `dispatch/runners/ruff`,
  `index-integration` (30 tests, real `node -e` child processes exercising
  the merged `exit`+`close` wiring end to end).
- Both directional red-proofs above kept as evidence of composed behavior,
  not just each half in isolation.
- Full suite deferred to CI, per repo convention (test-contention policy).
  CI's `Unit tests` and `Lint & type-check` both confirmed to actually
  EXECUTE (not skipped) on this head after the merge conflict was resolved
  and pushed.

Refs #1656


apmantza:
## Review round 1 (self-caught via CI, not an external reviewer)

The first push's Unit tests check failed on the real head SHA (91e96179) — confirmed
executing, not skipped. Root cause: I `unref()`'d the new grace/cap timers per the
issue's "timers unref'd (shape 4)" screen. That was wrong here: the caller's own
promise resolution now genuinely depends on those timers firing, unlike the
timeoutId/escalationTimer pair elsewhere in this file, which are cleared on every
settle path but never needed to be ref'd because completion was driven by Node's
native "close" event, not a timer. Unref'ing let a short-lived host process (e.g.
`mcp/cli.js build-graph`) exit the event loop mid-wait once the child's own OS
handles closed — the spawn's promise was silently abandoned instead of resolved.
`tests/mcp/build-graph-cli.test.ts` caught this directly (exit 0, empty stdout).

Fixed by keeping both timers ref'd (matching the existing `timeoutId`/
`escalationTimer` convention: cleared on every settle path, never left dangling).

This also surfaced a real, separate interaction: safeSpawnAsync's completion now
depends on `setTimeout`, so any test that fakes global timers (`vi.useFakeTimers()`)
around a code path that also makes a real, unmocked spawn will hang unless it
either mocks the spawn or advances fake time. `tests/clients/graph-cache.test.ts`'s
"evicts an idle workspace and rebuilds it" test does exactly that (a real `git
ls-files` spawn via `clients/git-tracked-ignore.ts`, wrapped in
`vi.useFakeTimers()` for an unrelated idle-eviction check). Fixed by switching that
test to `vi.useFakeTimers({ shouldAdvanceTime: true })`, which lets real timers
still progress in real time while `vi.advanceTimersByTime` can still jump the
clock forward deliberately.

Swept all 54 test files using `vi.useFakeTimers()` for the same hazard (real,
unmocked `safeSpawnAsync` call under faked time) — ran every one; only
`graph-cache.test.ts` hit it. The other 53 either mock spawn/child_process
directly or never reach a real spawn under the faked-timer window.

New head: 08278018. Re-verified: `npm run build`, `npm run lint`, and the full
targeted set (safe-spawn suite, build-graph-cli, graph-cache, persistence-worker-exit,
installer-lifecycle, package-manager suites) all green.

Issue #1741, opened 2026-08-19
State charter: every stateful store declares scope, invalidation, and re-arm — generalize the session sweep to all four lifetimes


apmantza:
## Evidence

The session-state conformance sweep caught three branches in one day (#1664, #1633, #1687) by forcing module-level Map/Set state to declare its session-boundary behavior. The latch/lifetime class is wider than sessions:

- **Persisted** state with no refresh policy: managed-tool versions drift indefinitely (#1730 — knip was 28 minor versions stale, producing 62 false diagnostics).
- **Process** state with asymmetric maintenance: the instance registry deregistered on self-crash but not forced shutdown (#1724 — root-caused as a write race in #1738, but the asymmetry went unnoticed because nothing declared the invariant).
- **Scope questions litigated ad hoc**: TreeCache capacity and lifetime (#1715/#1719), the tree-sitter latch re-arm (#1698), session-straddling verdict writes (#1674).

## Design

Extend `tests/support/session-state-registry.ts` into a four-lifetime charter. Every stateful store — module Map/Set, persisted file under `~/.pi-lens/`, process registry, turn-scoped memo — declares `{scope: turn | session | process | persisted, invalidation: <trigger>, rearm: <mechanism or "none" with a reason>}`. The sweep enforces registered-or-fail per lifetime:

- session: the existing sweep, unchanged.
- persisted: must name a refresh or expiry policy (makes the #1730 class undeclarable).
- process: must name the deregistration seam and prove its symmetry (makes the #1724 class visible at declaration time).
- turn: must name the reset point.

## Acceptance criteria

- The charter type and registry land with the existing session entries migrated unchanged.
- Persisted and process lifetimes enumerated (start from the `~/.pi-lens/*` files and the instance registry), each with a declared policy or a red sweep.
- Red-first: a store added without a declaration fails the sweep.

Refs #1730, #1724, #1738, #1715, #1698, #1674, #1635.

apmantza:
Tracked under #1842 for coordination; this issue remains the work item.

Note on scope: this issue generalizes #1635's item 2 (session-state lifecycle registry) to all four lifetimes. #1635 keeps its other four items.

*This comment is AI-generated, with the maintainer supervising the process.*

Issue #1730, opened 2026-08-19
Managed tool installs never refresh — pi-lens runs a stale version and its verdicts diverge from the user's own tool


apmantza:
Managed Node tools install once and then never refresh, so pi-lens can run a version months behind what the same tool's own range permits. The verdicts it produces then diverge from what the user's `npx <tool>` reports.

## Evidence

`~/.pi-lens/tools/package.json` pins `"knip": "^6.4.1"`. The installed copy on this machine is 6.4.1; `npx knip` in the same project resolves 6.32.2. On pi-webaio (2026-08-19) the two disagree hard:

- managed knip 6.4.1: 62 unused type-export flags
- knip 6.32.2 (`npx knip`, same tree, same `knip.json`): 0

Isolated to version, not config: `npx knip@6.4.1` in that tree reproduces all 62. False unused-export findings are destructive advice — an agent acting on them de-exports live code.

The caret range already permits 6.32.2. Nothing re-resolves it, so the lockfile written at first install pins the tree forever.

## Scope

All 22 entries in `~/.pi-lens/tools/package.json` drift the same way: `@ast-grep/cli`, `@biomejs/biome`, `jscpd`, `madge`, `markdownlint-cli2`, `oxlint`, `pyright`, `stylelint`, `typescript`, and the language servers. A stale pyright or biome produces the same class of wrong verdict, quieter.

## Direction

Decide a refresh policy and make the freshness observable:

- Re-resolve managed tool versions on some cadence (session_start with a cooldown, or a version floor per tool checked at probe time).
- Record the resolved version where a wrong verdict can be traced back to it. #1721 adds this for knip (`knip toolchain … version=…` in `~/.pi-lens/sessionstart.log`); no other managed tool records its version anywhere.
- Bound the cost: a refresh must not spawn `npm install` on a hot path or on every session.

## Why this is not #1721

#1721 makes knip prefer the project's OWN knip when the project installs one. That fixes projects with knip in their `devDependencies`. It does not help a project that has no knip of its own and relies on pi-lens's managed copy — pi-webaio is exactly that case, and it stays on 6.4.1 until this issue is fixed.

Refs #1721.

apmantza:
PR #1746 covers part of this issue. Filing what remains so the issue does not close on a partial fix.

**Done in #1746** — the 22 unpinned npm entries in `~/.pi-lens/tools/package.json`:

[...]

PR #1759, opened 2026-08-19
fix(installer): refresh github/pip/archive/maven/gem managed tools (closes #1747)

### LINKED ISSUE #1747: Managed tools installed via github/pip/archive/maven/gem never refresh either
Managed tools installed by every strategy EXCEPT npm are still frozen at whatever version was current the day pi-lens first installed them. #1730's fix (PR #1746) covers the 22 unpinned npm entries only.

## The shape

A resolution performed once at install time, with no step that ever re-resolves it. `installer/index.ts` installs a tool only when it is absent, so the "latest" it picked on day one is the version pi-lens runs forever.

## Scope

| Strategy | Entries | Where it freezes |
| --- | --- | --- |
| github | 27 | `clients/installer/index.ts:2956` resolves `https://api.github.com/repos/<repo>/releases/latest` during install only |
| pip | 6 | `installPipTool` runs `pip install --user <pkg>` with no `-U`, and only when the tool is absent |
| archive | 4 | archive URL resolved once |
| maven | 1 | JAR coordinate resolved once |
| gem | 1 | same |

A stale ast-grep, ruff, gitleaks or terraform-ls produces the same class of wrong verdict #1730 documented for knip: findings the user's own copy of the tool does not produce.

## Why it is not part of #1746

Each of these needs a different refresh mechanism than `npm update` — a GitHub API call plus a binary re-download, or `pip install -U`. Folding them into #1746 would roughly triple that change.

## Direction

Reuse #1746's machinery rather than building a second cadence:

- The persisted stamp (`~/.pi-lens/tools/.refresh-state.json`) and its `checkedAt`/`failed` cooldowns are strategy-agnostic already.
- The per-session budget (`clients/installer/managed-tool-refresh-session.ts`) should stay ONE budget across strategies, so adding github does not turn one spawn per session into two.
- Per strategy, supply a "what version is installed" reader and a "re-resolve" step. `getRefreshableManagedNpmTools()` is the pattern for deriving candidates from `TOOLS` rather than a hand-kept list.
- The github path also needs a version to record: no non-npm managed tool writes its version anywhere today, which is the second half of #1730's "make the freshness observable" that #1746 does not close.

Refs #1730.

apmantza:
Extends #1746's refresh mechanism to the five managed-tool install strategies it did not cover. Stacked on `fix/1730-managed-tool-refresh`; retarget to `master` when #1746 lands.

## What was broken

`installTool` runs only when a tool is ABSENT, so whatever the installer resolved on day one is what pi-lens runs forever. #1730/#1746 fixed that for npm. The other 39 entries stayed frozen:

| Strategy | Entries | How it froze |
| --- | --- | --- |
| github | 27 | `releases/latest` resolved at install, never again |
| pip | 6 | `pip install --user` with no `-U` |
| archive | 4 | registry-pinned URL, never re-applied after a repo bump |
| maven | 1 | registry-pinned GAV, same |
| gem | 1 | `gem install` never re-run |

A stale ast-grep, ruff, gitleaks or terraform-ls produces the same class of wrong verdict #1730 documented for knip.

## What changed

One mechanism, not two. The persisted stamp (`~/.pi-lens/tools/.refresh-state.json`), the weekly per-tool interval, the shorter retry cooldown, the one-refresh-per-session budget and the `managed-tool-refresh` degradation kind are all #1746's, unchanged. Only the re-resolution step differs per strategy, and it lives in one new function, `refreshManagedTool` in `clients/installer/index.ts`.

- **github** — query `releases/latest` with the stored `ETag` as `If-None-Match`. A 304, or a tag equal to the recorded one, downloads nothing. Combined with the weekly stamp and the one-per-session budget, 27 release-installed tools cost at most one conditional request per session, and most of those do not count against GitHub's rate limit. When the tag does move, the already-fetched release metadata is handed to `installGitHubTool` rather than paying a second API call.
- **pip** — `pip install -U --user <pkg>`: the same command ladder and the same `--user` target the install already uses. `-U` is the whole difference.
- **gem** — `gem install <pkg> --no-document`, which already resolves the newest version.
- **archive and maven** — these pin an explicit version in pi-lens's OWN registry, so their coordinate cannot move upstream. Their freeze is the mirror image: when this repo bumps spotbugs from 4.10.2, every existing machine stays on 4.10.2 because the tool is not absent. The refresh compares the recorded coordinate against the registry's current one and touches the network only when they differ.

Four properties beyond the per-strategy work:

- **Candidates come from `TOOLS`.** `getRefreshableManagedTools()` derives all six strategies on every call. `getRefreshableManagedNpmTools()` is now a filter over it, so there is one list, not two.
- **Presence is filesystem-only.** A tool pi-lens has not installed is never refreshed, because `pip install -U` or a release download on an absent tool is an unrequested install. github, maven and archive check pi-lens's own managed bin and extract dirs; pip and gem check the persisted probe cache, since those package managers install into user-owned directories with no pi-lens-owned location to look in.
- **A refreshed artifact is verified.** A refresh REPLACES a working binary in place. `verifyRefreshedArtifact` runs the new artifact's `--version` and, on failure, degrades and stamps `failed` instead of recording success over a binary that no longer runs. It also re-stamps the probe cache with the new mtime, so the next dispatch does not pay a full re-resolution.
- **Installs stamp their own resolution.** `finishInstallAttempt` writes the stamp for every strategy, so a freshly installed tool starts its cadence at the install rather than being re-resolved on the next timer tick, and archive and maven never pay a redundant adoption download.

## Verification

`npm run build` before every run.

**Red-first.** The new file is 24 tests. On pre-fix code (this branch's base, `fix/1730-managed-tool-refresh` at f7fa2e91, restored with `git checkout -- clients`), 20 of the 22 tests that existed at that point failed. The two that passed were negative assertions; both were strengthened afterwards so they now fail pre-fix too. Representative failure:

```
FAIL github strategy > re-resolves a stale release and installs the new tag
TypeError: getRefreshableManagedTools is not a function or its return value is not iterable
```

**Mutation-proof.** Every new guard was neutered one at a time, rebuilt, and re-run against both refresh suites (44 tests). All eight died:

| Mutation | Result |
| --- | --- |
| github unchanged-tag guard forced false | 1 failed |
| github 304 handling forced false | 1 failed |
| github post-refresh verification forced false | 1 failed |
| pip `{ upgrade: true }` replaced with `{}` | 2 failed |
| archive/maven pin comparison forced false | 2 failed |
| pip broken-binary guard forced false | 1 failed |
| presence filter removed | 10 failed |
| failure path persists the ETag | 1 failed |

**Suites run.** `tests/clients/installer/` (14 files, 202 tests), plus `install-attempt-evidence` and `session-state-conformance`: 16 files, 258 tests, all green. Both refresh suites together: 44 tests green. The full suite is CI's job.

**Cost.** Measured with all 60 candidates installed and every stamp fresh: `runManagedToolRefresh` returns `nothing-due` in 6.7 to 11.4 ms over five runs, with zero network calls and zero spawns. `session_start`'s own cost is unchanged at zero: the refresh module is still not imported until #1746's unref'd 30-second background timer fires, and this PR adds no call to `handleSessionStart`.

## Blast radius

- `httpsGet` was reimplemented on a new `httpsGetWithMeta` that surfaces status and `ETag`. Same signature, same redirect and Authorization-stripping behavior, and it still throws on any non-200. Callers: `installGitHubTool`, `installMavenTool`, `installArchiveTool`, all covered by the installer suites above.
- `installGitHubTool` takes an optional prefetched release. Omitted, it fetches exactly as before.
- `installPipTool` takes an optional `{ upgrade }`. Omitted, the command is byte-identical to before.
- `finishInstallAttempt` now writes one stamp on success. That is a durable-store write on a path that has just finished a network install, and it is best-effort: a failed stamp costs one extra re-resolution later and never fails an install. Every strategy funnels through it, which is why the hook lives there.
- New spawns: at most one `--version` probe before and one after a pip or gem refresh, and one `--version` verification after a github, maven or archive refresh. All inside the existing one-tool-per-session budget, all on the background timer.
- Behavior widening: archive and maven tools can now be re-downloaded when the registry pin moves. They previously never were. This is the intended fix, bounded by the same weekly stamp.

## Observability

`~/.pi-lens/sessionstart.log`, `managed-tool-refresh <toolId>` rows, in #1746's format:

- moved: `managed-tool-refresh shfmt: v3.7.0 -> v3.12.0 via github (412ms)`
- unchanged: `managed-tool-refresh spotbugs: unchanged at <pin> via archive (2ms)`
- failed: `managed-tool-refresh ruff: pip refresh failed after 8100ms - keeping 0.5.0 (...)`
- broken artifact: `managed-tool-refresh shfmt: refreshed artifact at <path> failed its --version check`

Degradation ledger: kind `managed-tool-refresh`, subject is the tool id, recorded via `recordDegradationOnce`. A permanently failing tool is one row per session, and the stuck tool's identity survives aggregation.

## Class sweep

Shape: a resolution performed once at install time with no step that ever re-resolves it, plus its mirror, a version pinned in this repo that never reaches an already-installed copy.

Searched every download site under `clients/` with `grep -rln "httpsGet\|https.get\|fetch(" clients/ --include=*.ts`. Three files: `installer/index.ts` (this PR), `tree-sitter-client.ts` and `grammar-source.ts`.

Found: `grammar-source.ts` pins `TREE_SITTER_WASMS_VERSION = "0.1.13"`, and `tree-sitter-client.ts:941` fetches a grammar only when the wasm file is MISSING. The sha256 manifest check at `grammar-source.ts:534` validates freshly downloaded bytes only, never the cached file on disk. A repo-side bump of `TREE_SITTER_WASMS_VERSION` therefore never replaces an already-downloaded grammar, which is the same mirror-image shape as the archive and maven entries fixed here. Filed separately rather than folded in: it needs a different mechanism (a version-keyed cache path, or an on-disk sha check) in a different subsystem.

## Coordination

- **Base.** Stacked on `fix/1730-managed-tool-refresh` at f7fa2e91. No master commit since that point touches `clients/installer/`, `clients/degradation-ledger.ts` or `tests/support/session-state-registry.ts` (`git log 4e9aad8a..origin/master` over those paths is empty), so there is no composition risk from master.
- **#1746's fix round.** Its review is adding an in-flight serialization guard, a pre-await budget decrement, and probe-cache invalidation. This PR adds no new entry point: every non-npm strategy runs inside the same `runManagedToolRefresh`, spends the same budget counter, and writes through the same `writeRefreshStamp`. A guard placed around `runManagedToolRefresh` therefore covers these five strategies with no per-strategy copy. Probe-cache freshening is already done here for the strategies that replace a binary (`verifyRefreshedArtifact` calls `updateProbeCache`); if #1746's round adds its own npm-side call, keep both, since they target different tools. Merge #1746 first.
- **Session-state registry.** This PR adds no new in-process state, only new fields (`resolutionId`, `etag`) on the already-persisted stamp file, so #1746's existing `managed-tool-refresh-session:refreshesThisSession` entry still describes the whole in-process surface. The registry has no `persisted` lifetime today; declaring the stamp's fields belongs to the #1741 state charter, and editing the registry here would collide with #1746's in-flight round. Noted rather than done.

## Remainder

All five strategies land, so this covers the issue's scope. One honest caveat, also recorded as an issue comment: pip and gem install into user-owned directories, so pi-lens cannot prove it was the installer of a given copy. Presence is inferred from the persisted probe cache, and the refresh re-runs the same command pi-lens already runs unprompted to install these tools. `PI_LENS_DISABLE_TOOL_REFRESH=1` opts out.

## Base-merge and review rounds (post-original-PR)

This branch was merged onto `fix/1730-managed-tool-refresh`'s round-2 local-allowance loop, then onto `master` once #1746 merged. Resolving that conflict meant the local-capture budget discipline now governs all six strategies, not just the five non-npm ones. Three adversarial review rounds followed.

**Round 1 (three blocking findings, fixed):**
- `installArchiveTool` deleted the working extracted tree before the replacement archive was proven good. Extraction and verification now happen in a scratch `.refresh-tmp` directory, swapped into place only after success (`swapExtractedDir`).
- `refreshManagedTool` bypassed the install kill-switch, the project-trust gate, and the shared install lock. All three now gate every non-npm strategy before it runs; a refusal is a declined, unstamped skip.
- A GitHub release response with no `assets` array threw uncaught inside `pickReleaseAsset`, skipping every stamp write and starving the tool of its retry cadence forever. The shape is now validated before use.

**Round 2 (two blocking findings, fixed):**
- Lint was red on the round-1 head (unused test variable), which also silently skipped Unit tests on that run.
- The three install gates from round 1 covered only the five non-npm strategies — npm's refresh predated this PR and kept spawning `npm update` directly, ungated. Pulled the checks into a shared `acquireManagedInstallGate`, used by both `refreshManagedTool` and the npm path's `refreshNpmOne`.

**Round 3 residuals (fixed):**
- `swapExtractedDir`'s rollback branch — the entire difference between "tool intact" and "tool destroyed" on a failed swap — was untested; pinned with a test that forces the second rename to fail and asserts byte-for-byte restoration.
- **Double-failure worst case, stated plainly:** if the second rename AND the rollback rename it triggers both fail (e.g. the volume goes read-only mid-swap), `finalDir` ends up genuinely empty and the ONLY surviving copy is the orphan at `finalDir.rollback`. `swapExtractedDir` still throws either way, so the caller reports the install/refresh as failed rather than succeeded. The real recovery is indirect: that failure's `invalidateManagedToolResolution` call drops the cached resolved path, so the next `ensureTool` probe finds nothing at `finalDir`, treats the tool as absent, and reinstalls from scratch. The orphaned `.rollback` directory is **not** reclaimed by that reinstall — it is only cleaned up the next time `swapExtractedDir` runs a successful swap for the same tool (the unconditional `fs.rm(backupDir, ...)` at the top of the function). This worst case now also records a bounded degradation naming the orphan path and logs it, so the leak is discoverable rather than silent.
- Two more vacuous guards closed and pinned: the non-object GitHub release-body shape check (`null`/array/bare-string bodies), and `invalidateManagedToolResolution` on the non-npm refresh failure branch (a failed refresh must not leave a stale resolved path cached).

Verification each round: `npm run build`, `npm run lint`, full installer suite + `session-state-conformance` + `package-manager*` + `single-flight-ratchet` + `generation-guard-sweep` + `changelog-entries` — 358 passing after the round-3 fixes and the latest `origin/master` merge. Every new/changed guard proven red-first against the pre-fix code before being restored.


apmantza:
## CI note

`ci.yml` triggers only on pull requests targeting `master`, so Unit tests and Lint & type-check do NOT run on this PR while it is stacked on `fix/1730-managed-tool-refresh`. The PR is `MERGEABLE`, not DIRTY — the workflow simply does not match the base branch. They will run on the first push after this is retargeted to `master`, which is the gate to trust before merge.

Local stand-in until then: `npx tsc --noEmit -p tsconfig.json` clean, `npx biome check` clean, `tests/clients/installer/` plus `install-attempt-evidence` and `session-state-conformance` green (16 files, 258 tests), and one full-suite run.

apmantza:
Correction to the CI note above: the full-suite stand-in did NOT run. It queued behind the machine-wide test lock held by another agent and was cancelled rather than left parked. Verified locally instead: `npx tsc --noEmit -p tsconfig.json` clean, `npx biome check` clean, and 16 test files / 258 tests green across `tests/clients/installer/`, `install-attempt-evidence` and `session-state-conformance`. The full suite remains unrun on this change — that is the gap the retarget-to-master CI run must close before merge.

PR #1860, opened 2026-08-20
fix(snapshot): name the stale meta gate in the miss reason (closes #1858)

### LINKED ISSUE #1858: project_snapshot miss reason says 'missing' when the real cause is the stale meta gate
﻿Five consecutive plegma-project sessions (2026-08-18 14:34 through 2026-08-20 10:53) log `project_snapshot: miss reason=missing` while the snapshot file exists at 1.4MB (`projects/Users-apman-OneDrive-Desktop-plegma/cache/project-snapshot.json.gz`). The preceding log line each time is `meta gate stale (metaSeq=923 ... current=977)` — the real reason is the stale meta gate, not a missing file. Result: the project starts snapshot-cold every session (word-index full rebuild, 2327ms in the observed session) and the reason string hides why. Control: pi-webaio, whose seq does not advance, loads fine.

## Acceptance criteria

- The miss reason distinguishes `missing` from `stale-meta-gate` (and any other gate) — the record must name the gate that rejected an existing file.
- A test pins the stale-gate path emitting its own reason.
- Bonus scope, judge feasibility in the fix: five sessions of seq advance (923→977) suggests the meta gate makes the snapshot near-useless for actively-developed projects — state in the PR body whether a partial/stale load path is worth a follow-up issue.

Evidence: sessionstart.log entries for the five sessions; word-index `incremental preflight selected full rebuild (stale-document-churn)`.

apmantza:
## What and why

Session-start project-snapshot probes now report the gate that rejected a
snapshot. A stale meta sidecar can reject an existing body before parsing, so
it now reports `stale-meta-gate` instead of `missing`. The reason is also
carried in the `session_start_snapshot_load` metadata record.

| Path to the miss log | Reason | Verdict |
|---|---|---|
| Neither canonical nor legacy body exists | `missing` | Preserved |
| Meta sidecar rejects on version or sequence before body parsing | `stale-meta-gate` | Fixed and tested |
| A body exists but parsing or snapshot validation returns null | `invalid-body` | Added |
| A parsed snapshot sequence differs from the current sequence | `stale(seq=..., current=...)` | Preserved |
| Defensive non-fresh snapshot fallback after parsing | `incompatible` | Preserved |

The population sweep found the single `project_snapshot: miss reason=` emitter
in `clients/runtime-session.ts`; no other `reason=missing` emitter exists
under `clients/` or `tests/`.

## Red proof

Before the fix, the new stale-meta assertion failed:

```
FAIL ... stale meta (seq mismatch) ...
Expected: "project_snapshot: miss reason=stale-meta-gate"
Received: "project_snapshot: miss reason=missing"
1 failed, 4 passed
```

## Verification

- `npm run build`
- `npx vitest run tests/clients/runtime-session-snapshot-gate.test.ts tests/clients/project-snapshot.test.ts`
- Result: 2 test files passed, 31 tests passed
- `npm run changelog:check`: 160 changelog entries validated

A partial or stale load path is worth a follow-up issue, but it should not weaken
this gate. The meta sidecar mirrors the snapshot version and sequence, and the
gate prevents hydrating a body whose derived state no longer matches the
project change log. A partial loader could reuse safe metadata or unchanged
per-file data, but it needs explicit coverage and merge semantics; otherwise
it risks presenting stale exports, reverse dependencies, or word-index results
as current. The observed sequence churn makes that follow-up valuable for
actively-developed projects, but it is separate from naming the current
fail-closed decision.

*Drafted by a codex worker via plegma; maintainer verification follows.*


apmantza:
﻿Merge-gate note: SonarCloud reports quality-gate ERROR on this PR — sole failing condition is `new_duplicated_lines_density` 66.7% (threshold 3%), with zero reported issues. Judged acceptable: the new lines are parallel test cases (quick/full mode x four miss-reason classifications) where the repeated assertion shape is deliberate, and the small diff inflates the percentage. Auto-merge stays armed.

*This comment is AI-generated, with the maintainer supervising the process.*

Issue #1904, opened 2026-08-20
Read-guard telemetry and precision: enforced-is-intent, search-hit margin credits unseen lines, per-file store unbounded


apmantza:
﻿Two observability/precision findings from the same read-guard monitoring pass (2026-08-20), filed together:

**1. `range_snapshot_validation.enforced` reports intent, not outcome.** 63 mismatches logged enforced:true in 3 hours; ZERO range_stale verdicts fired. Mechanism: read-guard.ts:1287 computes shouldBlock, :1309 logs enforced, but runtime-tool-call.ts:1255-1257 passes skipSnapshotCheck for every content-validated (oldText) edit and :785 honors it — and ALL 219 edits in the window were oldText edits, so the gate never ran. The ledger cannot distinguish "enforced and passed" from "bypassed". AC: log the caller's bypass decision (e.g. outcome: enforced|bypassed-content-match) so enforcement rates are readable; no behavior change.

**2. Search-hit reads credit ±2 unseen lines.** search-read-registration.ts:34 DEFAULT_CONTEXT_MARGIN=2 registers every search hit as a 5-line read; 56 of 64 rg/grep commands in the window showed the model ONE line while the guard credited five (89 of 100 sampled records on one hot file were 5-line search credits vs ~6 real reads). This is #1659 item 1's precision failure arriving through a mechanism that issue does not name. AC: credit delivered lines only, or record the margin separately so seen-lines precision is honest. Also noting: the per-file read store hit 268 hash-bearing records on one file in 75 minutes — enforceFileCap bounds files, not records per file; bound it.

*This issue is AI-generated, with the maintainer supervising the process.*

PR #1915, opened 2026-08-20
fix: honor pipe truncation in search credit; always-on trim telemetry (closes #1908, closes #1913)

### LINKED ISSUE #1908: Search credit ignores pipe truncation: grep -C | head credits undelivered context
﻿Follow-up from PR #1907's review (same defect class as the fix, two orders smaller): `grep -C2 pattern file | head -N` credits the full ±2 context even though the pipe truncated the output — the parser reads the grep flags but cannot see downstream truncation. Reviewer probe: `grep -n -C2 e x.ts | head -1` still yields contextBefore/After = 2.

Low priority: the exposure requires a piped-and-truncated search followed by an edit in the truncated context lines. Direction: detect a truncating pipe tail (head/tail/sed q) in the command and fall back to match-line-only credit (the conservative default the parser already uses for unparseable commands).

Refs #1904, PR #1907.

### LINKED ISSUE #1913: Read-guard eviction telemetry is verbose-only — a live eviction regression would be invisible
﻿Release-gate investigation finding (2026-08-20): the #1907 eviction/credit counters (searchCredit fields, evictedRecordCount, rawReadCountForFile) ride the read_recorded event, which read-guard-logger gates behind PI_LENS_READ_GUARD_VERBOSE (off by default) — so a live eviction regression would leave no trace at default verbosity. The negative proof in the field ("eviction can only narrow allows into blocks, and zero blocks occurred") worked this time only because the store was young. AC: the eviction-fired case emits a bounded always-on record (evictions are rare by design — one line per trim event is cheap), keeping the per-read verbosity gate for the rest.

Refs #1904, PR #1907.

apmantza:
## What changed

**#1908 — search credit ignores pipe truncation.** `grep -C2 pattern file |
head -1` credited the full declared ±2 context even though `head` truncated
the pipe before that context ever reached the model. `collectGrepCommandFiles`
in `clients/bash-file-access.ts` now walks the pipe chain each line-numbered
grep feeds into (via a new `terminator: "pipe"` field on
`ShellCommandSegment`) and falls back to match-line-only credit (`{before: 0,
after: 0}`) when the chain hits a truncating tail: `head`, `tail` (any
invocation — both default to a 10-line window), or a `sed` script containing
a `q` command.

Per-tail policy, documented in a code comment at the new helper:
- `head`/`tail`/`sed q` → match-line-only credit (conservative fallback).
- `wc`, `sort`, and similar stream-consuming/reordering tails need no special
  case: they replace grep's `file:line:` output with a count or reordered
  text, so `parseGrepOutputSearchReads` already finds zero matching lines in
  the real captured stdout — no line means no credit already.
- A pass-through filter (`cat`, `grep -v`, ...) between grep and a truncating
  tail is still caught, since the walk follows the whole chain.
- Out of scope, deliberately: `sed -n 'Np'` range-address filtering without
  `q` also truncates but isn't in #1908's stated scope.

**#1913 — read-guard eviction telemetry is verbose-only.** The #1907
record-cap eviction counters (`evictedRecordCount`, `rawReadCountForFile`)
rode the `read_recorded` event, which `read-guard-logger`'s `shouldLogEvent`
gates behind `PI_LENS_READ_GUARD_VERBOSE` (off by default). A live eviction
regression left no trace unless verbose logging happened to be on.
`recordRead` now also emits a dedicated `read_cap_trimmed` event — file,
`evictedRecordCount`, the credit-vs-genuine split (`evictedCreditCount`,
`evictedGenuineCount`), and `rawReadCountForFile` — only when
`enforceRecordCapForFile` actually trims, and `read_cap_trimmed` is added to
`shouldLogEvent`'s always-on list so it bypasses the verbosity gate. The
per-read `read_recorded` line, and everything else, stays gated exactly as
before.

## Why

Both are follow-ups from #1907's review, named in #1904's arc: a parser that
trusts declared flags without checking whether the shell actually delivered
what they promise (#1908), and a rare-but-load-bearing failure signal that
was only visible with an env var most sessions never set (#1913).

## Verification

- `npm run build` before every test run.
- Red-first: reverted `clients/bash-file-access.ts` and confirmed 4 new
  tests failed (head/tail/sed-q/pass-through-filter all showed
  `contextBefore/After: 2` instead of `0`); reverted
  `clients/read-guard.ts` + `clients/read-guard-logger.ts` and confirmed the
  new `read_cap_trimmed` test failed (`expected [] to have length 1`), and the
  dedicated `shouldLogEvent` test failed with `TypeError: shouldLogEvent is
  not a function`. Restored the fix (`git apply`) after each — no `git
  stash` used.
- Green after the fix: `tests/clients/bash-file-access.test.ts` (82),
  `tests/clients/read-guard.test.ts` (71), new
  `tests/clients/read-guard-logger.test.ts` (3).
- Sibling suites that reference the touched symbols:
  `search-read-registration.test.ts`, `git-guard.test.ts` (uses
  `tokenizeShellCommand`), `pi-host-contract.test.ts`,
  `bash-read-guard-integration.test.ts`, `read-guard-path-normalization`,
  `read-guard-symbol-read`, `read-guard-tool-lines`, `read-bridge`,
  `read-expansion-enrichment`, `agent-nudge`, and
  `session-state-conformance.test.ts` — all green (308 tests total across
  those files).
- Full suite deferred to CI, per repo policy.

## Mutation-proofing

- `grepPipesIntoTruncatingTail` / `isTruncatingTailCommand`: neutering either
  to always return `false` reds the head/tail/sed-q/pass-through tests;
  always `true` reds the `| sort` non-truncating and no-pipe tests.
- `read_cap_trimmed` emission: the "no eviction" test reds if the
  `evictedRecordCount > 0` guard is deleted; the "eviction fires" test reds
  if the event, or any of its four metadata fields, is dropped.
- `shouldLogEvent`'s new arm: the dedicated logger test reds if
  `read_cap_trimmed` is removed from the always-on list, independent of the
  `read-guard.test.ts` mock (which bypasses the real gate entirely).

## Blast radius

- `ShellCommandSegment` gained an optional `terminator?: "pipe"` field.
  `git-guard.ts` is the only other consumer of `tokenizeShellCommand`, and it
  only reads `.tokens` — unaffected.
- `enforceRecordCapForFile`'s return type changed from `number` to an object
  (`{evictedCount, evictedCreditCount, evictedGenuineCount}`); its one caller
  (`recordRead`) was updated. No other caller exists.
- Both changes are read-path/telemetry-path only — no new process spawns, no
  behavior change to what edits are allowed or blocked.
- `read_cap_trimmed` fires only when `enforceRecordCapForFile` actually
  trims (records per file > 128), which the #1907 docstring calls a rare,
  hot-file-only path — negligible added log volume.

## Class sweep

**Pattern sweep (#1908):** grepped `clients/`, `tools/`, `mcp/`, `scripts/`,
`index.ts` for `contextBefore`/`contextAfter` assignment sites. Only
`clients/bash-file-access.ts` (the grep parser, fixed here) and
`clients/search-read-registration.ts` (the consumer, which trusts whatever
context the caller declares — by design, per its own doc comment) touch
these fields. No other search-credit producer exists to sweep.

**Population sweep (#1913):** `read-guard.ts` has three other size-bounded
stores besides the per-file record cap: `enforceFileCap`'s whole-file
eviction (`READ_GUARD_MAX_FILES`/`READ_GUARD_MAX_UNCONSUMED_FILES`, via
`evictFile`) and the per-file edits cap (`READ_GUARD_MAX_EDITS_PER_FILE`).
Neither emits ANY eviction telemetry today — not gated, simply absent. Both
are out of scope for #1913, whose acceptance criteria target only the
#1907 record-cap eviction, and the edits cap is already documented
(`read-guard.ts`'s own comment) as sitting "far above every consumer's
reach" — trimming is inert there. Flagging both as a real but smaller gap;
filing a follow-up issue if the maintainer wants coverage extended before
the next `enforceFileCap`-adjacent incident.

## Merge-order note

No open PRs touch `clients/bash-file-access.ts`, `clients/read-guard.ts`, or
`clients/read-guard-logger.ts` as of this branch's creation (`gh pr list`
returned none). No stacking or sequencing needed.

Closes #1908, closes #1913. The class sweep's remaining eviction paths (evictFile, enforceFileCap, edits-cap splice, idle-eviction timer) are re-homed to #1918 — refs #1918.

### DIFF
--- a/.changelog/fix-1908-grep-pipe-truncation-credit.md
+++ b/.changelog/fix-1908-grep-pipe-truncation-credit.md
@@ -1,0 +1,9 @@
+---
+section: Fixed
+---
+
+- **Search credit ignores pipe truncation (closes #1908)** — `grep -C2 pattern
+  file | head -1` no longer credits context lines the pipe cut off. The
+  read-guard now detects a truncating pipe tail (`head`, `tail`, `sed q`)
+  downstream of a line-numbered grep and falls back to match-line-only
+  credit, so a later edit to the uncredited context still requires a read.
--- a/.changelog/fix-1913-read-guard-eviction-telemetry.md
+++ b/.changelog/fix-1913-read-guard-eviction-telemetry.md
@@ -1,0 +1,9 @@
+---
+section: Fixed
+---
+
+- **Read-guard eviction telemetry is now always on (closes #1913)** — a
+  read-guard record-cap trim now emits a bounded `read_cap_trimmed` log line
+  (file, evicted count, credit-vs-genuine split, raw read count) regardless
+  of `PI_LENS_READ_GUARD_VERBOSE`, so a live eviction regression is visible
+  by default instead of only under verbose logging.
--- a/.changelog/fix-1913-read-guard-eviction-telemetry.md
+++ b/.changelog/fix-1913-read-guard-eviction-telemetry.md
@@ -2,8 +2,11 @@
 section: Fixed
 ---
 
-- **Read-guard eviction telemetry is now always on (closes #1913)** — a
-  read-guard record-cap trim now emits a bounded `read_cap_trimmed` log line
-  (file, evicted count, credit-vs-genuine split, raw read count) regardless
-  of `PI_LENS_READ_GUARD_VERBOSE`, so a live eviction regression is visible
-  by default instead of only under verbose logging.
+- **Read-guard eviction telemetry is now always on (closes #1913)** — the
+  first time a file's read-guard record cap trims in a session, it now emits
+  a `read_cap_trimmed` log line (file, evicted count, credit-vs-genuine
+  split, raw read count) regardless of `PI_LENS_READ_GUARD_VERBOSE`, so a
+  live eviction regression is visible by default instead of only under
+  verbose logging. Later trims on the same file keep updating the running
+  totals (queryable via `ReadGuard.getTrimStats`, and via the degradation
+  ledger's own tally) without flooding the log on every subsequent trim.
--- a/clients/bash-file-access.ts
+++ b/clients/bash-file-access.ts
@@ -522,33 +522,43 @@ function parseGrepLineWithoutFile(
 }
 
 /**
- * Verbs that cut a piped stream short by LINE COUNT (#1908): `head`/`tail`
- * (even bare, since both default to a 10-line window) and a `sed` script
- * carrying a `q` command (quits after the addressed line). Either can leave
- * only part of a `grep -C` block's printed context in the command's actual
- * stdout, so a hit that survives the cut cannot be trusted to carry the
- * context grep's flags alone promised.
+ * Verbs that DROP printed lines from a piped stream (#1908, review F4): a
+ * hit that survives such a tail cannot be trusted to still carry the
+ * -A/-B/-C context grep's flags declared, because that context may have
+ * been among the dropped lines. `head`/`tail` drop by position (even bare,
+ * since both default to a 10-line window); `sed` with a `q` command drops
+ * everything after the addressed line; `uniq` drops adjacent duplicate
+ * lines, which can include a context line that happens to repeat.
  *
- * Per-tail credit policy, decided per pipe stage rather than per verb:
- *   - `head`/`tail`/`sed q` downstream of a matching grep → the match-line
- *     survives truncation unpredictably, so credit falls back to
- *     match-line-only (the same conservative default used for unparseable
- *     commands) rather than trusting the declared -A/-B/-C flags.
- *   - `wc`, `sort`, `uniq -c`, and similar STREAM-CONSUMING or reordering
- *     tails need no special case: they replace grep's `file:line:` shaped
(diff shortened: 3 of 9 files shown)


apmantza:
Review record (AI-generated adversarial review, maintainer-supervised): fix round verified with independent probes on 59b39ffc — 172 always-on lines reduced to 1, ledger re-arms at session_start, per-file identity preserved, all round-1 mutations now red. Residual non-blocking nits, on record for any future pass:

- The inner chain-break guard at `clients/bash-file-access.ts:596` is mutation-vacuous (under-credit direction, safe); a `grep … | cat ; head -1 f` test would close it.
- `sort -u` drops duplicate lines like `uniq` but stays out of scope in the same comment; practically inert since credited lines carry distinct line numbers.
- Comment at `clients/degradation-ledger.ts:240` overstates the rising edge ("power-of-two milestones" go to latency.log via a different sink, not to the read-guard line).
- The single `read_cap_trimmed` line's `trimEventCount`/`evictedRecordCount` fields are constant-by-construction; totals live in the ledger.

PR #1916, opened 2026-08-20
feat: adopt husky for local git hooks (closes #1804)

### LINKED ISSUE #1804: Adopt husky for local git hooks (pre-commit lint, pre-push targeted tests)
Maintainer directive (2026-08-20): add husky after the v4.1.0 release.

Motivation from this release cycle: several fix rounds shipped heads with lint failures (unused variable, changelog-format violations) that CI caught one lap later — a pre-commit/pre-push hook catches them before the push, saving a full CI round trip each time.

Suggested shape (to be designed at implementation):
- pre-commit: npm run lint (tsc, fast) + the changelog fragment guard (tests/scripts/changelog-entries.test.ts is cheap) + biome check on staged files.
- pre-push: targeted tests for changed files, NOT the full suite (the suite is serialized machine-wide per #1112; CI stays authoritative).
- Hooks must be skippable for CI/automation contexts and must not slow the many concurrent agent worktrees materially — measure the pre-commit cost and state it.
- Worktree consideration: hooks run per-worktree; the agents' commit cadence is high, so the pre-commit set must stay under a few seconds.

Post-release work; do not start before the v4.1.0 tag.

apmantza:
## What

Adds Husky-managed local git hooks (v9), wired via `prepare`:

- **pre-commit** — `npm run changelog:check` (reuses the existing sub-second
  fragment validator, already proven in CI's release-bump lane) then
  `npm run lint` (tsc, the same command CI's `lint-and-typecheck` job runs).
- **pre-push** — `npm run build`, then targeted `vitest` for the changed
  `.ts` files (`scripts/pre-push-targeted-tests.mjs`), never the full suite.

Both hooks exit early when `PI_LENS_SKIP_HOOKS=1` is set. Agents and CI should set
it; humans leave hooks on.

## Why

Several fix rounds this release cycle shipped heads with lint failures
(unused variable, changelog-format violations) that CI caught one lap later.
A local hook that runs in seconds catches the same failure before the push,
saving a full CI round trip.

## Design notes

- **No self-lint biome.** The issue's suggested shape mentioned "biome check
  on staged files," but this repo has no `biome.json` and doesn't use biome
  to lint itself — `biome-client.ts` runs biome against *user* codebases pi-lens
  analyzes, a different surface entirely. Adding an unconfigured linter to
  pre-commit would produce noise from day one. `npm run lint` (tsc) is the
  repo's actual self-lint mechanism and the one CI already runs.
- **Whole-project lint, not staged-files-only.** tsc type-checks the whole
  project graph; a partial staged-file check isn't meaningful for it. At
  ~3.5s the full `npm run lint` is cheap enough to run every commit and
  reuses the exact CI command — no parallel staged-file lint config to drift
  from it.
- **`prepare` had to stay split.** The existing `prepare` script
  (`build:dist` + grammar download) is load-bearing for consumers who
  install pi-lens as a dependency (`npm install --omit=dev`, no `.git`, no
  devDependencies — see `ci.yml`'s `prod-install-build` job) and must fail
  loudly on error. Hook wiring is dev-only and best-effort, so it's a
  separate script (`scripts/setup-git-hooks.mjs`) appended last, which
  always exits 0 — it never shares a failure path with the build step, so a
  broken hook install can't get masked by (or mask) a real build failure.
- **Targeted-test selection** (`scripts/pre-push-targeted-tests.mjs`) does a
  two-pass match against `tests/**/*.test.ts`: a path-mirror pass (changed
  `clients/foo/bar.ts` → `tests/.../bar.test.ts`) plus a content-grep pass
  that catches shared-seam siblings the path mirror misses (a changed
  module's basename appearing in another test file's own imports). Verified
  against `clients/agent-nudge.ts`: correctly selected both
  `tests/clients/agent-nudge.test.ts` (mirror) and
  `tests/clients/write-autofix-nudge-suppression.test.ts` (content-grep
  sibling). Falls back to build-only when nothing matches — never silently
  skips the build too.
- **`with-test-lock.mjs` reused**, not reimplemented, for the actual vitest
  invocation (respects the existing machine-wide suite lock and its
  cross-platform spawn handling).

## Measured cost

Real `git commit` on this branch, hook included: **~5.5s**
(`changelog:check` ~0.75s + `npm run lint` ~3.5s + npm/git process
overhead). Slightly over the issue's "under a few seconds" framing but the
two checks it runs are the exact ones that caused the CI round-trips this
issue cites — I judged the coverage worth the extra ~1.5-2s over a bare
changelog-only check. `PI_LENS_SKIP_HOOKS=1` is the intended escape hatch
for anyone (or any automation) where that cost matters more than the local
catch.

## Worktree verification

Git worktrees of the same clone **share** `core.hooksPath` (it's repo-level
config, not worktree-level) — but `.husky/_` (the machinery husky installs
at that path) is generated fresh per working directory and is git-ignored,
so each worktree needs its own `npm install`/`prepare` before hooks actually
fire there. Verified directly:

1. Created a second worktree off this branch — `.husky/pre-commit` is
   checked out (tracked file) but `.husky/_` doesn't exist yet.
2. Committed a deliberately invalid changelog fragment there: **succeeded**
   (hooks silently no-op — nothing installed at `core.hooksPath` in that
   worktree yet). Reverted.
3. Ran `npm install` in that worktree (real install, not a junction —
   proves the full `prepare` chain: build:dist → grammar download → hook
   wiring).
4. Same invalid fragment: **failed** pre-commit with the expected
   `changelog:check` error.
5. Same invalid fragment with `PI_LENS_SKIP_HOOKS=1`: **succeeded**.

## Red-proof (this branch's own worktree)

- Bad changelog fragment (`this is not valid front matter`, no YAML front
  matter) → `git commit` failed: `Invalid changelog entry ...: expected YAML
  front matter` / `husky - pre-commit script failed (code 1)`.
- Same commit with `PI_LENS_SKIP_HOOKS=1` → succeeded.
- Real commit with a valid changelog entry → `changelog:check` + `lint` both
  ran and passed, commit succeeded.
- `pre-push` with only non-`.ts` changes staged → build-only pass, no test
  run (correct: nothing to target).
- `pre-push` with a `.ts` change → build, then 2 targeted test files
  selected and run (39 tests passed), matching the design notes above.

## Verification run

- `npm run build`, `npm run lint`: clean.
- `tests/packaging.test.ts` (asserts `prepare` still contains `build:dist`
  and `download-grammars`): 10 passed, 4 skipped (unrelated pre-existing
  skips).
- `tests/scripts/{with-test-lock,rollup-changelog,changelog-entries,changelog}.test.ts`:
  44 passed — nothing in the reused seams broke.
- Full suite: not run locally (machine-wide lock, CI is authoritative per
  AGENTS.md).

## Blast radius

- `package.json`'s `prepare` script changed shape but keeps both existing
  steps (`build:dist`, grammar download) verbatim, just appends a third,
  fail-safe step. `tests/packaging.test.ts` already pins the first two by
  substring match and still passes.
- New `devDependency`: `husky@^9.1.7`. No production dependency change —
  `.husky/`, `scripts/setup-git-hooks.mjs`, and
  `scripts/pre-push-targeted-tests.mjs` are dev-only, not in `package.json`'s
  `files` allowlist, so they don't ship in the published tarball.
- New process spawns: `npm install`/`prepare` now also spawns
  `node_modules/husky/bin.js` (skipped for CI/production installs — see
  Design notes). Every local `git commit`/`git push` now spawns the
  hook chain described above, unless `PI_LENS_SKIP_HOOKS=1`.
- No runtime/extension code touched — LSP, dispatch, and client surfaces are
  unaffected.

## Observability

Every hook run prints what it did (`[pre-commit] skipped (...)`, `[pre-push]
building...`, `[pre-push] running N targeted test file(s)...`) directly to
the commit/push output — this is local, synchronous, terminal-visible
tooling, not a background process, so there's no separate log file to name.
`scripts/setup-git-hooks.mjs` logs its skip reason or a warning (never a
silent failure) on every `npm install`.

Closes #1804.

## Review round 1 (9 findings, fixed on this branch)

- **F1 (high, selection blow-up)** — the content-grep pass matched imports
  by basename suffix; `index.ts` alone matched 176 test files, and a real
  43-file commit selected 282 (~10 minutes). Fixed two ways: selection now
  resolves every test file's relative import specifiers to absolute,
  extension-stripped paths and compares real paths, not substrings; a hard
  25-file cap degrades to build-only regardless, so "never the full suite"
  holds by construction. Test:
  `tests/scripts/pre-push-targeted-tests.test.ts` includes a same-basename-
  different-directory case that must NOT match, and an over-the-cap case.
  Both verified red against the pre-fix logic (see PR discussion).
- **F2 (high, unbounded hang)** — pre-push now sets
  `PI_LENS_TEST_LOCK_TIMEOUT_MS=120000`; a lock timeout is detected from
  `with-test-lock.mjs`'s own stderr and lets the push proceed (exit 0) with
  a warning instead of blocking it. Verified end-to-end with a fabricated
  live lock holder and a 2s timeout: push proceeded in ~6.6s with the
  expected warning.
- **F3 (high, wire PI_LENS_SKIP_HOOKS via `.claude/settings.json`)** — NOT
  done. Every edit path to `.claude/settings.json` (Edit tool, Bash
  heredoc/append) was blocked outright by this session's own permission
  system, independent of PR content. Per this agent's standing contract, no
  instruction — including a reviewer's — can authorize a change to
  permission/settings configuration, so the block stands rather than being
  routed around. `PI_LENS_SKIP_HOOKS` still works exactly as documented; it
  just isn't auto-set for Claude Code sessions in this repo. Left to the
  maintainer to wire manually if wanted — CONTRIBUTING.md and the changelog
  fragment were corrected to not claim this exists.
- **F4/F5 (medium, doc accuracy)** — CONTRIBUTING.md previously claimed
  hooks are "scoped to your clone (each worktree wires its own)".
  Corrected: `core.hooksPath` is repo-level git config shared by every
  worktree; only `.husky/_` (gitignored, per-worktree) is not. A worktree
  where `npm install` hasn't run yet silently runs no hooks — documented as
  accepted behavior (agent worktrees are covered by `PI_LENS_SKIP_HOOKS`
  regardless; CI is authoritative either way).
- **F6 (medium, zero tests on the selection logic)** — added
  `tests/scripts/pre-push-targeted-tests.test.ts`: path-mirror pass,
  full-path import-resolution pass (plus the false-basename-match negative
  case), the >25-file cap (both over and exactly at the limit), the
  no-match fallback, and `PI_LENS_SKIP_HOOKS` accepting `"1"` and `"true"`
  via a real `sh .husky/pre-commit`/`pre-push` spawn. 12 tests, all
  mutation-verified: disabling the cap check and reverting to
  basename-substring matching both turned the relevant test red before the
  fix and green after (verified locally, not committed — see PR
  discussion for the exact mutation and output).
- **F7 (low, no-match reporting)** — pre-push now prints one
  `[pre-push] no tests matched <file>` line per changed file with zero
  covering tests.
- **F8 (low, skip-value strictness)** — both hooks and
  `scripts/setup-git-hooks.mjs` now treat any non-empty
  `PI_LENS_SKIP_HOOKS` value as skip, not only `"1"`.
- **F9 (low, CI env-var strictness)** — `scripts/setup-git-hooks.mjs` now
  treats any non-empty `CI` value other than `"false"` as CI, not only
  `"true"`.

Rebased onto `origin/master` (`73716d8c`, #1914 merged) before this round;
no conflicts. New head: `3e1248d4`.

### DIFF
--- a/.changelog/feat-1804-husky.md
+++ b/.changelog/feat-1804-husky.md
@@ -1,0 +1,10 @@
+---
+section: Added
+---
+
+- **Husky-managed local git hooks (closes #1804)** — `npm install` now wires
+  a pre-commit hook (changelog fragment validation + `npm run lint`) and a
+  pre-push hook (build + targeted `vitest` for changed files, never the full
+  suite). Both are skippable with `PI_LENS_SKIP_HOOKS=1`, which agents and CI
+  set; humans leave hooks on. Hook install itself is skipped for CI and
+  production/consumer installs, and never fails `npm install` on error.
--- a/.changelog/feat-1804-husky.md
+++ b/.changelog/feat-1804-husky.md
@@ -4,7 +4,10 @@ section: Added
 
 - **Husky-managed local git hooks (closes #1804)** — `npm install` now wires
   a pre-commit hook (changelog fragment validation + `npm run lint`) and a
-  pre-push hook (build + targeted `vitest` for changed files, never the full
-  suite). Both are skippable with `PI_LENS_SKIP_HOOKS=1`, which agents and CI
-  set; humans leave hooks on. Hook install itself is skipped for CI and
-  production/consumer installs, and never fails `npm install` on error.
+  pre-push hook (build + a capped, full-path-resolved targeted `vitest`
+  selection for changed files, never the full suite; degrades to build-only
+  past 25 matched test files or a 2-minute shared test-lock wait). Both are
+  skippable with `PI_LENS_SKIP_HOOKS` set to any non-empty value, which
+  agents and CI should set; humans leave hooks on. Hook install itself is
+  skipped for CI and production/consumer installs, and never fails
+  `npm install` on error.
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,6 @@
+if [ "$PI_LENS_SKIP_HOOKS" = "1" ]; then
+  echo "[pre-commit] skipped (PI_LENS_SKIP_HOOKS=1)."
+  exit 0
+fi
+
+npm run changelog:check && npm run lint
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,5 +1,5 @@
-if [ "$PI_LENS_SKIP_HOOKS" = "1" ]; then
-  echo "[pre-commit] skipped (PI_LENS_SKIP_HOOKS=1)."
+if [ -n "$PI_LENS_SKIP_HOOKS" ]; then
+  echo "[pre-commit] skipped (PI_LENS_SKIP_HOOKS is set)."
   exit 0
 fi
 
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,6 @@
+if [ "$PI_LENS_SKIP_HOOKS" = "1" ]; then
+  echo "[pre-push] skipped (PI_LENS_SKIP_HOOKS=1)."
+  exit 0
+fi
+
+node scripts/pre-push-targeted-tests.mjs
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,6 +1,11 @@
-if [ "$PI_LENS_SKIP_HOOKS" = "1" ]; then
-  echo "[pre-push] skipped (PI_LENS_SKIP_HOOKS=1)."
+if [ -n "$PI_LENS_SKIP_HOOKS" ]; then
+  echo "[pre-push] skipped (PI_LENS_SKIP_HOOKS is set)."
   exit 0
 fi
 
+# Bounds how long a push waits on the shared machine-wide test-suite lock
+# (#1101, ~10 agents can share one dev machine). pre-push-targeted-tests.mjs
+# treats a lock timeout as "let the push proceed, CI is authoritative" —
+# never as a hard failure.
+export PI_LENS_TEST_LOCK_TIMEOUT_MS=120000
 node scripts/pre-push-targeted-tests.mjs
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,6 +22,22 @@ npm test
 
 Pull requests must pass `npm run lint` and `npm test`. CI also runs `npm run check:lockfile` and a production `--omit=dev` build (`npm run build:dist`), so keep `package-lock.json` in sync with `package.json`.
 
+## Local git hooks
+
+`npm install` wires Husky-managed hooks into `.git/config`'s `core.hooksPath`,
+scoped to your clone (each worktree wires its own):
+
+- **pre-commit** — changelog fragment validation (`npm run changelog:check`)
+  plus `npm run lint`. Measured around 4s on this repo.
+- **pre-push** — a build, then targeted `vitest` runs for the changed `.ts`
+  files (never the full suite; see `scripts/pre-push-targeted-tests.mjs`).
+
+Skip either with `PI_LENS_SKIP_HOOKS=1 git commit ...` / `git push ...`.
+Agents and CI set this — their commit cadence is too high for a lint pass on
+every commit, and CI runs the real gates anyway. Humans should leave hooks on;
+they catch the exact class of failure (unused vars, changelog-format
+violations) that used to slip through to CI.
+
 ## What belongs here?
 
 pi-lens is a pi extension that runs automated checks on every file write/edit. Contributions that fit are:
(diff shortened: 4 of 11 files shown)


apmantza:
F3 from review round 1 (wire `PI_LENS_SKIP_HOOKS` into `.claude/settings.json`) is not done. Every edit path to that file was blocked by this session's own permission system, independent of repo/branch protections. `PI_LENS_SKIP_HOOKS` works as documented; it just isn't auto-set for Claude Code sessions here. If a maintainer wants that wiring, add `"env": {"PI_LENS_SKIP_HOOKS": "1"}` to `.claude/settings.json` by hand (merge with any existing keys — don't clobber). Everything else from the review (F1, F2, F4–F9) is fixed on this branch.

PR #1921, opened 2026-08-21
fix: guard agent_settled ctx access after session replacement (closes #1924)

### LINKED ISSUE #1924: agent_settled crashes with extension_error when session is replaced mid-run
Reported and fixed by @Pluto-Yt in PR #1921 (filed retroactively so the PR carries an issue ref).

pi's `ExtensionRunner.createContext()` serves `cwd`/`signal`/`ui`/`sessionManager` through getters that call `assertActive()`, so a plain property read throws after `invalidate()`. `dispose()` invalidates the runner and `_emitAgentSettled()` still emits from the run's `finally` — our `agent_settled` handler's first ctx read (`setAmbientAbortSignal(ctx?.signal)`) then throws "This extension ctx is stale after session replacement or reload", which pi surfaces as `extension_error`.

Repro: pi-goal + pi-lens RPC flow with a session replacement during the run (PR author's repro, independently red-proven against master during review).

Acceptance criteria:
- `agent_settled` completes without `extension_error` when handed a stale ctx; ambient abort signal is cleared.
- Regression test: stale-ctx probe (getters throwing the SDK message) → handler resolves, `getAmbientAbortSignal()` undefined; red on pre-fix code.

Sibling handlers with the same shape are tracked separately in the class-sweep issue.

Pluto-Yt:
## Summary
- guard `agent_settled` context access after session replacement or reload
- safely skip deferred mutation draining and quiet-window scheduling when the callback context is stale
- preserve cleanup of the ambient abort signal

## Verification
- `npm run lint`
- reproduced the original `pi-goal` + `pi-lens` RPC flow; no `extension_error` is emitted
- verified `get_diagnostics` still executes successfully

### DIFF
--- a/.changelog/fix-1924-agent-settled-stale-ctx.md
+++ b/.changelog/fix-1924-agent-settled-stale-ctx.md
@@ -1,0 +1,7 @@
+---
+section: Fixed
+---
+
+- **agent_settled no longer crashes with extension_error when the session is
+  replaced mid-run** — stale extension ctx reads are guarded and the ambient
+  abort signal is cleared (closes #1924; thanks @Pluto-Yt).
--- a/.changelog/fix-1924-agent-settled-stale-ctx.md
+++ b/.changelog/fix-1924-agent-settled-stale-ctx.md
@@ -2,6 +2,6 @@
 section: Fixed
 ---
 
-- **agent_settled no longer crashes with extension_error when the session is
-  replaced mid-run** — stale extension ctx reads are guarded and the ambient
-  abort signal is cleared (closes #1924; thanks @Pluto-Yt).
+- **agent_settled no longer crashes with extension_error when the session is replaced mid-run** —
+  stale extension ctx reads are guarded and the ambient abort signal is
+  cleared (closes #1924; thanks @Pluto-Yt).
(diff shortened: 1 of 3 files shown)


apmantza:
@Pluto-Yt thank you for this — the diagnosis is exactly right and the repro was reproducible on our side. Disclosure up front: this review was AI-generated and is maintainer-supervised; a human maintainer oversees everything that lands.

What the review found, in short:

- **Confirmed the bug and your fix.** pi's `createContext()` serves `cwd`/`signal`/`ui` through getters that call `assertActive()`, so any plain read throws after `dispose()` invalidates the runner while `_emitAgentSettled()` still fires from the run's `finally`. Your outer guard is load-bearing: we red-proved it against master (a stale-ctx probe crashes `agent_settled` at the `setAmbientAbortSignal(ctx?.signal)` line) and it passes on your branch with the ambient signal correctly cleared. You also reused the house `isStaleExtensionCtxError` machinery — appreciated.
- **The same shape exists in four sibling handlers** (`tool_result`, `turn_start`, `agent_end`, `turn_end`). That's not on this PR to fix — we've filed #1925 to sweep them separately.
- Filed #1924 retroactively as the issue this PR closes, and retitled the PR to reference it (our convention keeps the ref in the title).

Since maintainer edits are enabled, we'll push two small deltas straight to your branch rather than asking you to round-trip: a `.changelog/` fragment and the regression test from our review probe. Your commits and authorship stay intact. After that and green CI, this merges.

Thanks again — solid find and a clean fix.

Issue #1947, opened 2026-08-21
oxlint runner discards every warning-severity finding (exit 0)


apmantza:
`clients/dispatch/runners/oxlint.ts` returns early on a zero exit:

```ts
if (result.status === 0) {
    return { status: "succeeded", diagnostics: [], semantic: "none" };
}
```

oxlint exits 0 when it found only warnings, and warnings are its default severity. A real capture of oxlint 1.61.0 on an unused variable produces a full `--format json` report and exits 0. The runner throws that report away, so pi-lens sees no oxlint warnings at all — only findings a project has explicitly raised to error severity.

Found during the #1937 class sweep, alongside three parser bugs of the #1933 shape (taplo, stylelint, phpstan), all fixed in #1946.

This one was deliberately left out of #1946. Reading output on a zero exit is a behaviour widening on pi-lens's hottest dispatch path — every JS/TS save — and belongs in its own change with its own noise assessment.

Acceptance criteria:
- Decide whether oxlint warnings should surface. If yes, parse output whenever output exists rather than gating on the exit code, and state the expected volume change.
- If no, say so in the runner with a comment naming the decision, so the next reader does not read it as a bug.
- Either way, a captured real-bytes fixture of the exit-0 warning case must pin the chosen behaviour. #1946's fixture uses an error-severity config specifically so the parser could be proven without prejudging this.

Refs #1937, #1946

Issue #1937, opened 2026-08-21
Class sweep: every runner parser verified against real binary output; no standing fixture exemptions


apmantza:
Evidence: the #1933 review found clients/dispatch/runners/vale.ts:82 parses `parsed?.Data?.Files`, a shape Vale 3.9.6 does not emit (it emits a flat map keyed by path). Against real output the parser returns [], stdout is non-empty so skipUnlessToolRan passes, and the runner reports `succeeded, 0 findings` while the CLI exits 1 with real errors. The runner was inert until a .vale.ini exists; PR #1933 is what would have made it live. The parser had never met real binary output.

Root cause of the class: tests/clients/dispatch/smoke-fixture-coverage.test.ts enforces fixture-or-exemption, and vale sat in the EXEMPT map as "no fixture yet". A standing exemption converts "forces a decision" into "permanently undecided". This is the #1816 exit-blind shape re-entering through the parser instead of the exit code.

Defect shape (for the AGENTS.md catalog): a runner parser whose expected output shape was written from documentation or memory, never from captured real binary bytes, combined with a smoke-coverage exemption that never expires.

Acceptance criteria:
- Audit every EXEMPT entry in smoke-fixture-coverage.test.ts. Each either gains a real-bytes fixture or carries a reason that names why real output is unreachable in tests (licensing, platform), not "no fixture yet". "No fixture yet" entries are eliminated.
- Fixture provenance: every live tool-smoke fixture states which binary and version produced its bytes, machine-checkable (a provenance header the coverage test asserts). Hand-written fixtures are the same defect one step removed.
- Sweep the covered runners too: verify each fixture matches its tool's current major-version output shape; stale-shape fixtures are recaptured.
- A scheduled CI lane runs the genuinely installable tools' real binaries end to end (tiered: the checksum-pinnable ones first, per the #1917 pattern), asserting the runner's parsed findings are non-empty on a planted violation. A parser regression against a real binary must red something.
- The vale parser fix itself lands in PR #1933 (in flight) with a real-bytes fixture; this issue covers the sweep, not that instance.

Observability record: each runner's smoke run already logs through the dispatch worklog; the scheduled lane's job log is the proof a real binary met the parser.

Refs #1816 (exit-blind class), #1933 (the instance), #209 (the original coverage guard).

apmantza:
#1946 covers most of this. It stays open for the remainder below.

**Done in #1946**

[...]

PR #1987, opened 2026-08-22
fix(dispatch): eslint runner discards warning-severity findings (closes #1954)

### LINKED ISSUE #1954: eslint runner discards every warning-severity finding (exit 0) — same shape as #1947
`clients/dispatch/runners/eslint.ts` has the same defect shape #1947 fixed in oxlint:

```ts
// ESLint exits 1 when there are lint errors, 2 on fatal/config error
if (result.status === 2) {
    return { status: "skipped", diagnostics: [], semantic: "none" };
}
...
if (result.status === 0) {
    return { status: "succeeded", diagnostics: [], semantic: "none" };
}
```

ESLint exits 0 whenever no rule reaches error severity, which includes a run that produced only warnings (ESLint's own default unless `--max-warnings` is set). The comment above the check even says "exits 1 when there are lint errors" — true, but silent on the warnings-only case, which also exits 0 and gets the same early return.

Found during the #1947 class sweep (same runner family, `appliesTo: ["jsts"]`, same "reads its own exit status" #1737-strangler entry in `run-outcome-ratchet.test.ts`).

Acceptance criteria, mirroring #1947:
- Decide whether ESLint warnings should surface, and state the expected volume change (oxlint and ESLint both run on JS/TS saves, so a fix here doubles the same-shaped change to the hottest dispatch path).
- Parse stdout whenever it exists rather than gating on the exit code, if the decision is yes.
- A captured real-bytes fixture of the exit-0 warning case (`--max-warnings` unset, a rule at its default `warn` severity) must pin the chosen behavior.

Refs #1947, #1937, #1946

apmantza:
Closes #1954, the named residual of #1947's class sweep.

## What changed

`clients/dispatch/runners/eslint.ts` treated every exit 0 as "nothing found". ESLint exits 0 whenever no rule reaches error severity (`--max-warnings` unset), so a warnings-only run — full JSON report on stdout — was discarded silently.

The runner now parses stdout unconditionally and branches on the parsed diagnostic count, mirroring #1947's oxlint fix:

- **Zero diagnostics** → clean `succeeded`/`none`, unchanged.
- **Error-severity diagnostic** → `failed`/`blocking`, unchanged from the old exit-1 path.
- **Warning-only on exit 0** → `succeeded` with the findings riding along. Status keys off blocking severity plus exit code (#1955 review round 1), so plan.ts's fallback group `["eslint", "oxlint", "biome-check-json"]` still stops at eslint instead of re-running oxlint and biome-check-json on every warning-only save.
- **Exit 2** stays an unavailable skip.
- stderr only feeds the parser when a failing run's stdout went missing, so exit-0 stderr chatter cannot become spurious parse-error findings.

## Behavior decision (issue acceptance criterion 1)

Warnings surface. Expected volume change: JS/TS projects with an eslint config that sets any rule to `"warn"` start seeing advisory findings where saves previously read clean. Note: no rule in modern `@eslint/js` recommended ships at `"warn"` anymore (`no-unused-vars` is `"error"`), so the realistic warnings-only shape is a project downgrade — what the fixture config does.

## Evidence

- **Real-bytes fixture** (acceptance criterion 3): captured live from eslint 10.9.0 (exit 0, one severity-1 finding) into `tests/fixtures/runner-output/eslint/warning-exit-zero.captured.json`, replayed through `captured-real-output.test.ts`. Workspace fixture: `tests/fixtures/tool-smoke/js-eslint-warning/`.
- **Red-first**: both new regression cases fail against pre-fix code — diagnostics length 0 vs 1, and the eslint finding missing from the dispatch result.
- **Mutation proof**: reverting the runner fix reproduces both reds (2 failed | 3 passed).
- The smoke-fixture coverage guard flagged eslint's old exemption as redundant once captured bytes existed; the exemption is deleted.

## Blast radius

`module_report` on `eslint.ts`: single consumer seam (`run(ctx)` via the dispatch registry); no callbacks, no exported internals beyond the default runner definition. Affected dependents: plan.ts's jsts fallback group (covered by the new integration test), dispatcher bucketing (by-diagnostic semantic, unchanged), and the ratchet test's strangler entry (unchanged — the runner still reads its own exit status for exit-2 and succeeded/failed keying).

## Observability

Existing records prove the change: `latency.log` runner rows now carry nonzero `diagnosticCount` on warning-only saves, and findings reach `actionable-warnings.json` through the normal advisory routing. No new telemetry needed — this fixes a producer, not a decision.

Refs #1947, #1937

### DIFF
--- a/.changelog/fix-1954-eslint-warning-severity.md
+++ b/.changelog/fix-1954-eslint-warning-severity.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **ESLint warnings are no longer discarded (closes #1954)** — The eslint runner treated every exit 0 as "nothing found". ESLint exits 0 whenever no rule reaches error severity, so a run that produced only warning-severity findings was thrown away silently. The runner now parses stdout unconditionally and surfaces the findings; a clean file still reports clean, and exit 2 stays an unavailable skip.
--- a/clients/dispatch/runners/eslint.ts
+++ b/clients/dispatch/runners/eslint.ts
@@ -138,15 +138,23 @@ const eslintRunner: RunnerDefinition = {
 			{ timeout: 30000, cwd },
 		);
 
-		// ESLint exits 1 when there are lint errors, 2 on fatal/config error
+		// ESLint exits 2 on fatal/config errors — nothing was linted, so this
+		// is an unavailable run, not a clean or failed one.
 		if (result.status === 2) {
 			return { status: "skipped", diagnostics: [], semantic: "none" };
 		}
 
-		const raw = result.stdout || result.stderr || "";
-
-		if (result.status === 0) {
-			return { status: "succeeded", diagnostics: [], semantic: "none" };
+		// ESLint exits 0 whenever nothing reached ERROR severity — that
+		// includes a run that found only warnings (#1954), which also prints a
+		// full JSON report. So parse stdout unconditionally and branch on the
+		// parsed diagnostic count instead of letting the exit code discard the
+		// warning case. Stderr is only a fallback for a failing run whose
+		// stdout went missing; parsing stderr noise on a healthy exit-0 run
+		// would turn deprecation chatter into spurious parse-error findings.
+		const stdout = result.stdout ?? "";
+		let raw = stdout;
+		if (raw.length === 0 && result.status !== 0) {
+			raw = result.stderr || "";
 		}
 
 		const parsed = parseEslintJson(raw, ctx.filePath);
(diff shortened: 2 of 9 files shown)


apmantza:
Adversarial review ran against head `817324e0` (AI-generated, maintainer-supervised): **APPROVE**, one P3 finding.

The reviewer executed all three mutation scenarios rather than reasoning about them: restoring the early-return-on-exit-0 reds 3 tests, reverting the status keying reds 2, and reverting ONLY the stderr gating reds 0 — that last one was a real gap. Fixed in `37543a7e`: a case pinning `status: 0, stdout: "", stderr: <noise>` → succeeded/none/zero diagnostics, mutation-verified red on the ungated code (1 failed | 5 passed) and green on restore.

Also cleared by review: control-flow matrix has no regressed cell; diagnostics from a succeeded runner flow through `runGroup`'s push-before-break; fixture replay is OS-safe via the forward-slash `__WORKSPACE__` substitution; no new eslint+oxlint double-reporting scenario exists.

Issue #1968, opened 2026-08-21
Latent ABA in dead-code and knip in-flight caches: bare .finally without identity guard


apmantza:
Found by the #1967 review's population sweep of 19 in-flight promise caches. clients/dead-code-client.ts:397 and clients/knip-client.ts:475 release their in-flight map entries with a bare .finally(() => map.delete(key)). If build A settles after build B has replaced the entry for the same key, A's stale deletion evicts B's live entry, and a third caller starts a duplicate B. PR #1967 introduced the correct pattern for the review-graph cache: delete only when map.get(key) === promise.

Latent today (no measured incident); the same shape #1967 fixed live in _buildCache.

Acceptance criteria:
- Both sites adopt the identity-guarded release.
- A test per site proves the ABA: A settles late, B's entry survives, no duplicate spawn; red on pre-fix code.
- Sweep note: the #1967 review audited all 19 in-flight caches; these two were the only bare-finally members. State in the PR that the population is closed.

Refs #1967, #1962.

apmantza:
Reachability analysis + kit reproduction (evidence gathered while building the #1838 fault-injection kit).

## The ABA is real and deterministic — but latent

A kit-driven probe (`gatedPromise` controlling build ordering against `DeadCodeClient`) produced both halves:

1. **The defect fires when its precondition exists.** Simulating a second writer that replaces the map entry under the same key while build A is in flight (the future force-refresh/eviction path that would make this live): A settles, its bare `.finally(() => map.delete(key))` evicts B's live successor, and the next caller starts a duplicate. `evicted-live-successor: true`, deterministically.
2. **Today no public-API ordering reaches it.** Both clients have exactly one `inFlight.set` site and no force/evict/reset path (`rg inFlight`: one set + one delete each). Microtask FIFO guarantees any observer queued before A's delete sees A present (shares it), and anything after starts clean. Probed both orderings: shared-A, no duplicate, empty map on drain.

So unlike #1967's review-graph cache (which had replacement paths and broke live), these two sites are **latent**: the bug becomes real the moment anyone adds a second writer to either map — defect shape 24 (a second writer added to a shared field).

## Implication for acceptance criteria

"Red on pre-fix code" cannot be driven black-box through `analyze()`; the test must simulate the second writer directly (set a different promise into `inFlight` under the same key between A's start and settle, then assert it survives A's settlement). That test is red on the current bare-finally code and green with the identity-guarded release from #1967 — same pattern, both sites.

[...]

Issue #1838, opened 2026-08-20
Fault-injection kit: shared wedge/delay/starve/reset primitives for tests


apmantza:
The highest-value findings of the 2026-08-20 window came from bespoke fault probes reviewers built by hand, each once: a genuinely wedged child pipe (#1811's live fixture — first attempt silently failed because Node exited the child), a snapshot save racing inside a stall window (#1807), a budget starved by CPU scheduling (#1785's deterministic 5ms-budget repro), a session reset fired from inside a mocked spawn (#1746 R2-F1). Every one of these is a reusable primitive.

Ship a tests/support fault-injection kit: (1) wedgeStream/wedgedChild — a real child process with a genuinely full pipe and a keepalive (the #1811 fixture generalized); (2) delayInside(mock, ms) — deterministic completion delay for any mocked async seam; (3) starveBudget(env, ms) — the #1785 pattern (tiny budget + delayed read) as one call; (4) fireResetAt(hook) — invoke a session reset from inside an arbitrary await point of the code under test. Each primitive carries its own fidelity test (the #1673 lesson: an inert double is worse than none).

Value: catalog shapes 1/3/9 (latch re-arm, teardown liveness, session-straddling writes) become cheap to probe in every fixer's red-first pass instead of reviewer-only heroics. Post-release; composes with #1742 (seams-over-mocks). Refs #1816, #1742.

apmantza:
Tracked under #1842 for coordination; this issue remains the work item.

Note on scope: this issue composes with #1742 (seams over mocks, also under #1842) — #1742 replaces mocks with faithful seams, this issue gives each seam a way to misbehave on demand.

*This comment is AI-generated, with the maintainer supervising the process.*

Issue #1742, opened 2026-08-19
Seams over mocks: build the three injectable seams that eliminate the largest module-mock clusters


apmantza:
## Evidence

694 module mocks across 40% of the test suite (2026-08-19 audit), with a measured defect yield: four production-unfaithful-double incidents in one review arc (#1673 twice, #1684, #1687), the #1681 host-shape sweep, and defect shape 14 (#1514/#1565 — a failure mode that only exists because of module mocking). One more since the audit: #1719's CI failure came from a partial module mock missing a newly added method. A ratchet without replacement seams only says "no"; these are the seams.

## The three seams, by mock-cluster size

1. **Spawner seam** (kills the 62 safe-spawn mocks plus 35 node:fs/node:child_process): a fake-binary-on-temp-PATH harness plus an injectable spawner at the `safeSpawnAsync` boundary. The #1670/#1673 review probes built ad-hoc versions of this repeatedly — promote them to a shared fixture.
2. **Wire-speaking fake LSP server** (kills the 103 LSP mocks): #1660's design — a programmable server on the real wire. The #1722 arc demonstrated exactly why: in-process doubles could not model read-vs-processed semantics, and only a live-wire experiment settled the design.
3. **Telemetry sink** (kills the 61 latency-logger silencing mocks): a test-mode sink on the logger seam instead of module-mocking it. Also gives assertions structured access to emitted records, which several suites currently obtain by spying.

## Acceptance criteria

- Each seam lands with three or more existing test files converted as proof, their assertions strengthened, not weakened (red-first where the old mock masked behavior).
- The module-mock ratchet (per-file baseline, no new violations, first-mock-in-a-file fails) activates once seam 1 or 2 lands.
- The conversion pattern documented in the test-support README.

Refs #1660, #1681, #1514, #1565, #1673, #1684, #1687, #1719.

apmantza:
Tracked under #1842 for coordination; this issue remains the work item.

Note on scope: #1838 (fault-injection kit, also under #1842) composes with this issue rather than overlapping — this issue replaces mocks with faithful seams, #1838 gives each seam a way to misbehave on demand.

*This comment is AI-generated, with the maintainer supervising the process.*

Issue #1895, opened 2026-08-20
Formatter which-latches never re-arm at session_start (process-lifetime latch class)


apmantza:
﻿Live defect found by the parallel-store sweep (2026-08-20): `resetWhichLatches` (clients/formatters.ts:87) is NOT called at session_start — its only callers are within formatters.ts itself (:1545, :1732, :1737) and installer/index.ts:1941 which is test-only. Meanwhile resetDispatchAvailabilityState IS re-armed at session_start (clients/runtime-session.ts:2082). So formatter which-latches are process-lifetime: a formatter binary that vanished (or appeared) mid-process keeps its stale verdict across session boundaries for the formatter path while the dispatch path re-arms — the #1525/#1541 process-lifetime-latch class at one more seam, and the exact split-brain the availability umbrella documents.

## Acceptance criteria

- resetWhichLatches (or its equivalent) runs in the session_start reset block beside resetDispatchAvailabilityState.
- Test drives the real session_start handler: seed a which-latch, fire session_start, latch re-armed; red pre-fix.
- Session-state registry: the latch store's lifetime declaration updated accordingly.
- Class check: any OTHER availability store from the umbrella's list (package-manager availabilityLatches, direct-LSP negative cache, per-runner maps) also missing from the session_start reset — per-store verdicts.

*This issue is AI-generated, with the maintainer supervising the process.*

PR #1992, opened 2026-08-22
refactor(tests): adopt the fault-injection kit in existing suites (refs #1742)

apmantza:
Adoption proof for #1838 (now merged via #1989), following #1742's acceptance pattern: a seam lands with converted call sites, their assertions strengthened or equal — never weakened.

## What changed

**1. `fireResetAt` gains a `when(args)` matcher** (`tests/support/fault-injection.ts`)

Adoption-driven extension: managed-tool-refresh's R2-F1 probes fire a session reset inside one spawn mock that serves several call shapes. A positional `atCall` cannot target "the update call" without hard-coding its index. `when` restricts firing to matching calls; with `atCall`, the Nth *matching* call fires. Exactly-once semantics unchanged.

The fidelity suite caught the first implementation shipping the option **unwired** — hook fired on call 1 regardless of `when`. That is precisely the failure mode the per-primitive fidelity rule exists to catch, and it is now pinned by a dedicated test including the `atCall`+`when` composition.

**2. `service-runtime-exit-breaker.test.ts` → `starveBudget` + `gatedPromise`**

The manual tiny-env set and bare never-settling promise become kit calls. Behavior identical; 11/11 green.

**3. `managed-tool-refresh.test.ts` R2-F1 tests → `fireResetAt`**

Both mid-flight-session-start probes now declare intent declaratively instead of burying the reset inside the mock's branches. The triple-session-start case bundles three synchronous resets into one hook invocation — observationally identical for the budget under test. 52/52 green.

## Considered and deliberately not converted

`runtime-session-sequence-read-budget.test.ts`: its manual prev/restore env pattern is uniform across seven variables with symmetric afterEach handling; swapping one line to `starveBudget` would break the file's symmetry without adding safety. Adoption must fit, not perform.

## Blast radius

- Production code untouched. The primitive extension is additive (`options.when`), default behavior identical without it.
- Consumers of `FireResetAtOptions` type: none outside the kit + these conversions.

## Observability

Refactor-only PR: the three converted suites (74 tests) are the verification surface; no production telemetry involved.

Refs #1838, #1742

### DIFF
--- a/clients/dead-code-client.ts
+++ b/clients/dead-code-client.ts
@@ -397,11 +397,16 @@ export class PythonDeadCodeClient implements DeadCodeClient {
 		const key = path.resolve(root);
 		const existing = this.inFlight.get(key);
 		if (existing) return existing;
-		const promise = this.runAnalyze(key).finally(() =>
-			this.inFlight.delete(key),
-		);
-		this.inFlight.set(key, promise);
-		return promise;
+		const promise = this.runAnalyze(key);
+		const wrapped = promise.finally(() => {
+			// Identity-guarded release (#1968, #1967's pattern): delete only if
+			// THIS build is still the registered one. A bare delete-by-key lets
+			// a late-settling build A evict a live build B that replaced the
+			// entry mid-flight, and the next caller starts a duplicate.
+			if (this.inFlight.get(key) === wrapped) this.inFlight.delete(key);
+		});
+		this.inFlight.set(key, wrapped);
+		return wrapped;
 	}
 
 	/**
--- a/clients/formatters.ts
+++ b/clients/formatters.ts
@@ -1868,6 +1868,31 @@ export function clearFormatterCache(): void {
 	resetWhichLatches();
 }
 
+/**
+ * Test-only internals access for the session-state registry probe (#1895):
+ * the conformance suite must arm all four pieces of state this reset claims
+ * to cover and prove none survives.
+ */
+export function _getFormatterResetStateForTests(): {
+	whichLatchByCommand: Map<
+		string,
+		{ latch: AvailabilityLatch; resolved: string | null }
+	>;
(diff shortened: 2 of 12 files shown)


apmantza:
All three kit hunts are now in this PR (commits `1c3bde40`.., `274a3109`, `000a87a4`, `e48045fb`):

## Hunt 2 — in-flight ABA (closes #1968)

Both bare-finally caches (`dead-code-client`, `knip-client`) adopt the identity-guarded release from #1967's pattern. Red-first via kit-driven white-box probes: each test simulates the second writer directly (gatedPromise builds + direct map write), asserts the successor survives A's late settlement AND a third caller shares it. Both red on pre-fix code; green on the fix; companion cases pin normal cleanup.

## Hunt 3 — env-budget starvation

Seam survey of ~20 `*_MS` knobs. Conversions/additions:
- `pipeline-lsp-sync.test.ts`: the wedged-touch test now uses `starveBudget` + `gatedPromise` — deterministic budget-fired-first, gate released at teardown.
- Covered elsewhere already: notify budget (#1746 suite), sequence-read budget, aux grace, cascade settle config.
- Skipped with rationale: module-report budget (parked unwired #1090), read-guard/nav grace constants (module-load-time reads with no gated operation to starve).

## Hunt 1 — session-start reset seams

The #1635 registry had 12/35 probes. Two high-bug-history entries gained them:
- `formatters:whichLatches` arms ALL FOUR state pieces its reset claims to cover; mutation check confirms removing `detectionCache.clear()` turns it red — the exact #1895 bug shape.
- `zizmor-config:tokenAvailability` proves the session reset forgets a latched "missing" verdict.
New `_forTests` hooks in both modules (established pattern). The remaining 20 unprobed entries are recorded on #1635.

**Net production change**: only the two identity guards (#1968) and two small `_forTests` exports. Everything else is tests.

Issue #1993, opened 2026-08-22
mode=full replays stale mid-edit diagnostics as 🔴 blocking despite fresh clean LSP sweep and passing tsc


apmantza:
## Evidence

During normal editing of \`clients/dispatch/runners/oxlint.ts\` (rapid successive edits, one intermediate state was syntactically broken), \`lens_diagnostics mode=full\` on that path now persistently renders:

- 🔴 L282 ts:2393 Duplicate function implementation (\`parseOxlintUnix\`)
- 🔴 L304 ts:2528 A module cannot have multiple default exports

plus 11 rows explicitly marked \`[stale — re-run to confirm]\` with **line-past-EOF positions** (L306-L334; the file is 304 lines).

Meanwhile, every fresh authoritative source disagrees:

1. \`tsc -p tsconfig.build.json\`: clean.
2. \`rg -c\` on the real file: exactly one \`function parseOxlintUnix\`, one \`export default\`.
3. The same mode=full call's own LSP sweep reports **0 primary findings** for this file.
4. Full runners test suite: 968/968 green.

Re-running mode=full twice does not reconcile the two 🔴 rows away; they render identically each time while their sibling rows at least carry the stale marker.

## Expected

Diagnostics whose positions are past EOF and whose companion rows carry stale markers should be demoted/reconciled by the freshness machinery (#1631/#1790 family), never rendered as current 🔴 blockers — the widget/footer and mode=all paths already treat unconfirmed rows as non-blocking.

## Acceptance criteria

- A reproduction harness: break a file syntactically, let the cache capture diagnostics, fix it, run mode=full twice → no 🔴 current rows with past-EOF positions.
- Identify which lane serves the unreconciled rows (cached project-diagnostics vs widget store vs confirmed-LSP reconciliation) and why the #1186 per-entry mtime reconciliation or the content-hash binding did not retire them.

## Cross-links

Dogfooding observation during PR #1839 work; related to the #1020/#1025 path-key/content-binding line and the #1790 ghost-blocker shape.

apmantza:
## Live telemetry correction / stronger root cause

The current issue body says the fresh LSP sweep reported zero primary findings. The PID-specific latency rows show a deeper failure: these were fresh primary pulls, not only project/widget cache replay.

- Two full sweeps recorded cacheHits=0.
- Native-TS7 pull publication versions advanced 266 -> 267 -> 268.
- Every pull returned the identical 13 diagnostics.
[...]

Issue #1948, opened 2026-08-21
A runner that parses nothing out of non-empty tool output leaves no record


apmantza:
Every bug in the #1933 / #1937 class shares one observability hole: when a runner receives real tool output and its parser extracts zero findings, the result is indistinguishable from a genuinely clean file.

- vale (#1933): parser read `Data.Files`, vale emits a flat map. Reported "succeeded, 0 findings" against a CLI exiting 1.
- taplo (#1946): spawned a flag taplo rejects, `JSON.parse` threw, catch returned `[]`. Reported clean TOML.
- stylelint (#1946): report arrives on stderr, runner read stdout. Reported clean CSS.
- phpstan (#1946): iterated the error count as if it were the error array. Reported clean PHP.

All four ran for months. The dispatch worklog recorded `succeeded, 0 diagnostics` each time, which is exactly what a clean file records.

#1816 fixed the adjacent case — a tool that produced NO output now skips and writes a bounded `runner-empty-result` row. The remaining case is a tool that produced output the parser could not use.

Proposal: when a runner has non-empty stdout or stderr and its parser returns zero diagnostics, emit `incrementDegradationCount({ kind: "runner-parsed-nothing", subject: <tool> })` with a reason carrying the exit code and the output length. Bounded, and the subject preserves which tool is stuck.

The discriminating question a reader must be able to answer from the ledger: "is this file clean, or did the parser fail to read it?" Today they cannot.

Acceptance criteria:
- A shared helper, not a per-runner copy, so it lands once for the whole family.
- Zero rows on a genuinely clean run (empty output, zero findings) — a row per clean save would drown the ledger.
- Red-proved against at least one runner by reverting its parser fix.
- Enumerate the runner family and record which members adopt the helper.

Found during the #1937 sweep and named as a gap in #1946's observability section.

Refs #1937, #1946, #1933, #1816

apmantza:
[...]

Issue #1829, opened 2026-08-20
Tool-contract smoke suite: pin each wrapped binary's real exit/output contract in CI


apmantza:
Harness addition implied by the live-binary discipline (2026-08-20 window). Six tool-behavior claims were falsified against real binaries this release: trivy config's flag rejection printing usage to stdout (#1781), vulture's findings exit 3 (#1765), clippy's six levels with ICE under-reporting (#1809), mypy exit 2 carrying real syntax diagnostics (#1822 review), biome's missing tags field and location.path (#1810), pyright's range-nested positions (#1809). Each was found by hand-probing during a review — nothing pins these contracts continuously, so a tool upgrade can silently change conventions again.

**Shape:**
1. Per wrapped tool, a declared contract: exit table (clean / findings / error), output stream per case, and a minimal shape probe (the fields the parser reads). #1806's declarative severity tables and #1822's exit tables become the single source these tests verify — the contract test closes the loop between the table and the binary.
2. A gated real-binary test per tool (precedents exist: trivy-config-real-binary.test.ts, helm-render-iac-real-binary.test.ts — skip when the binary is absent, assert when present) driving a tiny fixture through the REAL binary and asserting the declared contract.
3. A CI lane that installs a rotating subset of the top tools (the managed-install machinery from #1746/#1759 already knows how) so the gated tests genuinely execute somewhere — today they skip on any machine without the binary, which is how the drift hid. The existing install-test matrix is the natural host.
4. A registered-or-fail ratchet (sweep-kit): every runner with a declared exit/severity table carries a contract test or a reason.

**Staging:** start with the six falsified tools plus the seven newly exit-aware runners from #1822 (their tables are fresh and evidence-backed); expand per #1737's strangler migration. Fixtures captured from real runs per the discipline — never hand-written.

Recommend post-release start (new CI lane + per-tool fixtures is real scope), riding #1806/#1737. Refs #1806, #1737, #1816, #1822.

apmantza:
Tracked under #1842 for coordination; this issue remains the work item.

Note on scope: #1605 (also under #1842) overlaps in spirit — both pin real-binary behavior against drift — but targets a different set of defect classes (lifecycle, staleness, scope-classification) than this issue's per-tool exit/output contracts. Neither is a subset of the other. #1839's parser-garbage harness is the explicit synthetic complement to this issue's real-binary checks; sequence this issue first where possible, since #1839's battery is generated from the contract tables this issue pins.

[...]

Issue #1806, opened 2026-08-20
Consolidate severity mapping: one normalizeToolSeverity helper + declarative per-tool tables + ratchet


apmantza:
The severity-collapse class (#1787 ast-grep-napi, #1791 biome + CLI info bucket, #1802 pyright + clippy) existed because every runner and adapter hand-rolls its own native-severity → tier mapping. N copies produced N divergent bugs — the same authoring-gap shape that motivated singleFlight (#1753), GenerationGuard (#1754), emitBounded (#1743), and sweep-kit (#1755).

Consolidation shape (the established recipe):
1. One `normalizeToolSeverity(table, nativeValue)` helper: takes a per-tool declarative table (e.g. biome: error/warning/information/hint; clippy: error/warning/note/help; pyright: error/warning/information; LSP: 1-4) and maps onto the 4-valued enum. Unknown-value fallback policy (→ warning) lives HERE, once.
2. The error ⟺ blocking (semantic) derivation lives in one place — today it is repeated at every `semantic:` assignment site (the vocabulary audit verified ~15 sites all encode the same invariant by hand).
3. A registered-or-fail ratchet on the sweep-kit (#1755): any runner/adapter carrying a hand-rolled severity ternary/switch must migrate or carry a written reason; reverse-locked, floors per the kit convention.
4. Strangler migration: new runners must use the helper (the ratchet enforces); existing ones migrate opportunistically, starting from the three just fixed (their tables are already written as code).

Mutation discipline per the house rules: the helper's table lookups and fallback are mutation-proofed once, centrally — which is the point.

Sequencing: after PR #1795 and the #1802 PR land (their fixes become the first table entries). Fits the consolidation-primitive family; a candidate for the same treatment as #1743/#1753/#1754/#1755.

Origin: maintainer observation during the 2026-08-20 severity arc — 'prefer centralized over hand-rolled'. Refs #1787, #1791, #1802, #1755.

apmantza:
AI-generated comment (Claude Fable 5), posted while implementing #1802's fix round; a maintainer supervises this process.

## Runner census: hand-rolled two-way severity ternaries

[...]

PR #1994, opened 2026-08-22
feat(tests): adversarial garbage battery for every CLI runner (closes #1839)

### LINKED ISSUE #1839: Runner-parser garbage harness: fuzz every tool parser with adversarial output shapes
Four real bugs this window were parsers mishandling output shapes nobody fixture'd: duplicate URIs letting a zero-answer beat findings (#1786 F2), CLI usage text on stdout parsing as a clean scan (#1781), unknown severity values (#1795's fallback probing), exit-2-with-parseable-stdout nearly discarded (#1822). Each parser met these shapes for the first time in production or review.

Ship a shared adversarial-shapes harness: for every runner parser (the ~44 dispatch runners + adapters), drive a standard garbage battery — duplicate entries for one file, usage/error prose on the findings stream, truncated JSON, unknown enum values, empty-but-nonzero-exit, absent optional fields (the biome tags lesson #1810) — and assert the CLASSIFICATION invariants (never clean-on-garbage, never crash, unknowns fall back declared). One battery, per-parser fixtures generated from the parser's declared shape; composes with #1806's tables and #1829's real-binary contracts (the battery is the synthetic complement to #1829's live checks). Post-release. Refs #1806, #1829, #1816.

apmantza:
Closes #1839. Synthetic complement to #1829's real-binary contracts and the captured-real-output replay suite.

## The harness

20 CLI runners × 12 hostile cases (truncated JSON, usage prose on stdout, unknown severities, absent optional fields, hostile numbers, HTML error pages, ANSI garbage) under three format-blind invariants:

1. **Never crash** — `run()` always resolves to a valid `RunnerResult`.
2. **Well-formed findings** — non-empty messages; integer line/column ≥ 1 when present; declared severity tiers (never an unknown passthrough).
3. **Never clean-on-failure** — nonzero exit + emitted bytes must never yield `succeeded` with zero diagnostics.

## First pass: 79 violations, one shape

Fourteen runners reported "we checked, it's clean" on a nonzero exit whose output they could not parse — the #1781 silencing shape, live in two-thirds of the CLI fleet.

## The fix is consolidation, not 14 patches

`finishParsedRun()` (`utils/tool-failure.ts`) is now the single tail for parsed CLI runs:

- findings present → per-runner `classify` hook preserves each runner's exact status/semantic convention;
- no findings, exit 0 → genuinely clean;
- no findings, nonzero exit, bytes → `failed` + explicit parse-error diagnostic.

Fourteen runners converted byte-for-byte behavior-preserving on real outputs (captured-real-output replay stays green).

## Real bugs fixed en route

- **htmlhint**: `--rules` was fed a JSON object; the flag takes `ruleid,ruleid=value`. Zero rules were ever enabled — every file exited 0 clean regardless of content. Root-caused live against htmlhint 1.x after the nightly tier-1 parser smoke caught it (#1993's sibling evidence). Regression tests pin argv shape + real unix-format parsing.
- **shfmt**: the #211 no-opt-in branch now reports `skipped` instead of `succeeded` when shfmt exited 1 — "chose not to check" ≠ "clean".
- **oxlint**: real captured bytes show exit 1 + `number_of_files: 0` for config-excluded files, so the no-files skip keeps precedence over the new guard (order matters, and the test suite proves both directions).
- **pyright**: empty-stdout early return no longer hides nonzero-exit stderr chatter.

## Consequential test updates

- tflint's parsed-nothing pin now asserts BOTH the degradation record (#1948) AND the parse-error finding.
- run-outcome-ratchet: biome-check, oxlint, pyright, shellcheck migrated onto the tool-failure seam; strangler exemptions dropped and the four are pinned on the primitive.

## Verification

- Battery: 253/253 green.
- Full runners directory + dispatcher/plan suites: 968/968.
- fmt + lint clean.

## Observability

Parse-error findings route through the normal advisory pipeline; #1948's `runner-parsed-nothing` ledger records still fire everywhere (pinned by parsed-nothing.test.ts).

Refs #1839, #1829, #1806


apmantza:
Adversarial review ran against head `527709fe` (AI-generated, maintainer-supervised): **APPROVE, no P1s**. The reviewer independently verified conversion fidelity for all 14 runners against origin/master (status/semantic byte-identical on every input class except the intended garbage change), confirmed I3 soundness against each tool's documented exit contract, cleared the oxlint reorder (`parseOxlintFileCount` returns undefined on truncated JSON so garbage falls through to the guard), and confirmed shfmt's succeeded→skipped is inert downstream (mode:"all" group).

Follow-ups from the review, all addressed in `b5d6d7f4`:

- **P2 — committed live evidence for the htmlhint fix**: captured real-bytes fixture (htmlhint 1.9.2, exit 1, tag-pair finding via the corrected comma-list argv) into the captured-real-output replay suite. A future htmlhint version drift that re-narrows rule coverage now reds in CI.
- **P3-1**: finishParsedRun docstring runner count corrected (thirteen files adopt it).
- **P3-2**: battery's justification-set placeholder comment replaced with its real post-fix disposition.
- **P3-3**: sqlfluff gate mock now states `() => false` as intent rather than a /dev/null existsSync accident.

The reviewer's P2 about an "uncommitted refactor" was this session's own ctx/result signature reshape landing between review start and finish — it is commit `0524480b`, pushed before the rebase.

PR #2003, opened 2026-08-22
fix(snapshot): coalesce unchanged persistence (closes #1997)

### LINKED ISSUE #1997: Project snapshot rewrites ~8 MB 92 times at the same generation in one session
## Defect class

Redundant durable publication / write amplification.

## Live evidence

For live PID 24856, `project_snapshot_persist` ran 92 times. Every row carried the same project sequence: `seq: 1642`.

Earlier in the same capture, 85 writes accounted for:

- 2.08 GiB cumulative raw serialization
- 0.64 GiB cumulative gzip output
- 16.5 seconds serialization
- 101.5 seconds writing
- 118 seconds total phase duration

Each publication was about 26 MB raw and 8 MB compressed. Gzip sizes varied slightly, but the governing project generation did not advance.

The work is offloaded, so cumulative phase duration is not equal to event-loop blockage. It still creates substantial allocation, compression, OneDrive I/O, and memory pressure.

## Invariants

1. A durable snapshot generation should be published at most once unless its serialized semantic content changed for a named reason.
2. Concurrent requests for the same generation must coalesce.
3. A newer generation supersedes pending older publication.
4. Failure/retry must not turn successful same-generation publication into an edit-time loop.
5. Telemetry must distinguish requested, coalesced, skipped-unchanged, executed, and retried writes.

## Acceptance criteria

- Identify all producers that request persistence and why unchanged `seq:1642` remains admissible.
- Add an instance-owned single-flight/generation guard.
- Skip publication when generation and semantic snapshot fingerprint match the last successful write.
- Preserve atomic publication and durability ordering for genuinely new generations.
- Measure serialization and write bytes avoided.

## Test matrix

- Sequential same-generation requests.
- Concurrent same-generation requests.
- Older request racing a newer generation.
- Failed write followed by retry.
- Same generation with a legitimate semantic payload change, if supported.
- Session reset and persisted-state reload.
- Windows/OneDrive slow-write interleaving without sleeps.

apmantza:
Closes #1997. Work produced by the codex session (quota ran out before promotion); verified and promoted by maintainer agent.

## The defect

Live evidence from PID 24856: `project_snapshot_persist` ran **92 times at the same project generation** (`seq: 1642`). 85 of those writes cost 2.08 GiB raw serialization, 0.64 GiB gzip output, and ~118 s cumulative phase duration — ~26 MB per publication, all for a generation whose semantic content never changed. Offloaded to a worker, so not event-loop blockage — but substantial allocation, compression, OneDrive I/O, and memory pressure for nothing.

## What changed

- Persistence **coalesces concurrent requests** for the same generation (one publication wins, waiters share it).
- A **semantic fingerprint** (`clients/project-snapshot-fingerprint.ts`) is computed on the worker; an unchanged fingerprint skips gzip + staging entirely.
- Same-generation content changes and failed-write repairs still publish — the invariant "at most once per generation *unless content changed or repair requires it*" is preserved.

## Verification

- `project-snapshot.test.ts` +446 lines: coalescing, fingerprint skip, changed-content republish, failed-write repair paths.
- New `persistence-worker-exit` coverage; session-state registry entry updated.
- Suite green in this worktree: build clean, fmt/lint clean, touched tests 58/58.

## Observability

The existing `project_snapshot_persist` phase remains the proof surface: repeat same-generation publications should collapse to one row per generation plus explicit change/repair reasons.

Refs #1997


apmantza:
Adversarial review (AI-generated): **APPROVE**, no blocking findings. All seven challenge fronts cleared: durable-store classification holds (no new lock path; triple re-read at admission/queue/skip-decision converges sibling races); shape 12 clean (mirrors update only after meta finalization, skip paths delete rather than refresh); the failed-write repair path is reachable and genuinely tested via injected `writeFileAtomic` throw; fingerprint is content-derived SHA-256 over the full serialized JSON (not mtime — no shape 6) with worker-side cost; the generation gate verbatim-preserved so late completions cannot resurrect superseded generations; session-state registry rationale accurate (conformance 70/70); +446 test lines cover the cross-product with real failure injection and a retained mutation-proof generation-gate test.

Two P3s re-homed to #2008: corrupt-body self-heal hole within a stable generation (skip trusts meta+stat, not body integrity), and `snapshotPersistBaselinesFor` being a read with seeding side effects. Both non-blocking per the review; impact bounded by atomic renames plus read-path rebuild degradation.

apmantza:
Telemetry-name correction for future verification runs: the skip decision logs as phase `project_snapshot_persist_decision` with `decision:skipped_unchanged` (`clients/project-snapshot.ts:1229`), not on `project_snapshot_persist`. Current real-log state since merge: 2 persist rows, both `executed`, zero same-seq repeat attempts yet — skip path unexercised in the wild so far (mechanism covered by the branch's 446 lines of snapshot tests).

PR #2006, opened 2026-08-22
docs: isolate subagent Git operations (refs #2007)

apmantza:
Docs-only AGENTS.md addition produced by the codex session, motivated by a live incident during this arc: conversationally forked workers sharing one checkout caused branch switches and lost uncommitted state mid-session.

## What changed

New design-contract bullet: before dispatching subagents that touch Git, the orchestrator maps each issue to an explicit worktree at the intended base and hands the worker its absolute path; a worker commits only after verifying its cwd is that registered worktree; when the runner cannot bind a worker to a directory, the worker runs no Git commands at all (orchestrator owns rebase/stash/commit/push/PR); parent checkout verified after dispatch; never coordinate parallel workers via branch switches or stashes inside one checkout.

## Verification

Docs-only: no code paths touched; prose conventions only.


apmantza:
Adversarial review (AI-generated): REQUEST CHANGES → addressed in `a9cd4838`.

- Filed #2007 capturing the motivating incident (forked workers sharing one checkout lost uncommitted WIP mid-session); the bullet now references it and the PR title carries the ref.
- The ~45-word three-clause sentence split into three sentences.
- Changelog entry deliberately skipped: AGENTS.md is not part of the shipped package, so the convention is recorded here and in #2007 rather than in the release notes.

Issue #1892, opened 2026-08-20
Umbrella: one diagnostics store, all surfaces project from it


apmantza:
﻿Maintainer directive (2026-08-20, from the #1888/#1891 arc): diagnostics should have ONE store that every surface projects from — not parallel stores correlated after the fact.

## Evidence from today

#1888 existed because `lens_diagnostics` merged lanes in its own render path while widget-state kept a separate projection seeing only confirmed LSP results — 12 vs 9 shown to the human. #1891 fixes the count by committing the merged set across the seam, which is correct but is a correlation bridge between two stores, not one store. The same week produced four staleness gates each owning a slice (widget reconcile, line-freshness, blocker sweeps, snapshot mtime) and per-surface delivery decisions that had to gain coverage-awareness one surface at a time (#1867 → agent result, #1888 → widget).

## Target

- One diagnostics store owning: ingestion from every lane (LSP, napi self-scan, dispatch runners), identity dedup (finding-identity #1820 as the key), staleness (consuming the #1739 freshness kernel), and coverage state per lane.
- Widget, lens_diagnostics, status line, and dispatcher notices become PROJECTIONS of that store — a count can no longer diverge by surface because no surface counts independently.
- #1891's correlation seam is the first consumer to fold in; its tests (cross-lane count, no-double-count-on-recovery, timestamp preservation, URI normalization) become store-level invariants.

Companion audits in flight will attach: (1) a full site map of every dedup/staleness/delivery decision point with per-site verdicts; (2) a repo-wide sweep for the same parallel-store shape outside diagnostics.

Relates: #1739 (freshness kernel), #1741 (state charter), #1888, #1891, #1820.

*This issue is AI-generated, with the maintainer supervising the process.*

apmantza:
﻿Design-grade audit complete (Opus sweep, 2026-08-20). Ranked verdicts, the store design, and the migration order. Full four-table site map (ingestion/identity/staleness/delivery, every row file:line) is preserved in the session report and rides the design PR.

## Ranked parallel-store clusters

| # | Cluster | Verdict |
|---|---|---|
| C1 | Per-surface tallies: widget `countDiagnostics` (widget-state.ts:575) vs lens `summarizeDiagnostics` (lens-diagnostics.ts:1085) — LIVE divergence: blocking and stale semantics differ, and `mergeDiagnosticsWithWidgetSummaries` mixes both semantics in one render | consolidate (#1892) |
[...]

Issue #1711, opened 2026-08-19
reconcileProjectDiagnosticsSnapshot still uses +1ms mtime tolerance — and its stale arm DROPS diagnostics


apmantza:
PR #1710 gave `findingPathFreshness` and `isEntryFresh` the shared 50ms `MTIME_DRIFT_TOLERANCE_MS`, but deferred one sweep remainder in prose: `reconcileProjectDiagnosticsSnapshot` (`clients/project-diagnostics/cache.ts:98`) keeps the old `mtimeMs > scannedAt + 1` shape.

The #1710 review measured the write-then-timestamp skew at -7.4ms to -1.06ms on an idle Windows host, with ~11.4ms positive excursions documented in `blocker-freshness.ts` — a +1ms tolerance covers almost none of it. This consumer is also the worse of the two: its stale arm DROPS diagnostics (`staleDropped`) where the gate #1710 fixed only demotes, so a same-tick write can silently lose a finding.

## Acceptance criteria

- The comparison uses the shared `MTIME_DRIFT_TOLERANCE_MS`, with a comment citing the convention.
- A deterministic regression test (fabricated mtime/scannedAt pair) proves the same-tick case is not dropped, red on current code.
- The stale doc comment at `clients/advisory-provenance.ts:264` (`stale (mtimeMs > scannedAt)`) is updated to the real boundary — one-line drift #1710's review flagged (F3).

Refs #1708, #1710.

Issue #1708, opened 2026-08-19
findingPathFreshness lacks the +1ms mtime tolerance its sibling gates have — write-then-scan race demotes fresh findings


apmantza:
`findingPathFreshness` (`clients/advisory-provenance.ts:318`) compares `stat.mtimeMs > scannedAtMs` strictly. A file written in the same millisecond its scan timestamp is captured can read as "edited after the scan" and demote a fresh finding to the stale tier. The sibling gates (`reconcileStaleWidgetFiles`, `blocker-freshness.ts`) carry a documented +1ms tolerance for exactly this write-then-timestamp race.

Observed: `tests/clients/runtime-turn-secrets-disposition.test.ts` > "turn_end secrets gate honors dispositions for trivy secrets (#1628)" flakes — expected the STOP blocker, got the demoted ACTION NEEDED tier. Reproduced locally 1 in 5 runs on master-derived code; surfaced on PR #1692's CI without any related diff.

## Acceptance criteria

- `findingPathFreshness` applies the same +1ms tolerance as its siblings, with a comment citing the convention.
- A regression test seeds the same-millisecond case deterministically (fabricated mtime/scannedAt pair, not a real write race) and is proven red on current code.
- Sweep: any other strict `mtimeMs >` comparison against a captured scan timestamp in `clients/` gets the same treatment or a reason; record sweep scope in the PR body.

Refs #1622, #1625, #1628, #1692 (where the flake surfaced).

Issue #1641, opened 2026-08-18
Diagnostics served with line numbers past current EOF — stale in-memory LSP document diverged from disk


apmantza:
Live from the 2026-08-18 plegma dogfood (maintainer-supervised AI session): the agent received errors cited at lines 407-410 of a file that is 402 lines on disk and unmodified (the referenced symbol `kilo` is defined at L396 and fine). The LSP's in-memory document is LONGER than disk — a stale document version — and diagnostics computed against it were served with coordinates that do not exist in the current file. This is the exact hazard #1631 documented in the abstract ("re-querying an LSP whose in-memory document is stale reproduces the stale verdict"): the in-memory desync class, observed directly.

Likely entry points: a lost/never-sent didChange after an out-of-band edit (bash/git write while the document was open), or a didOpen against content that then changed on disk without the watcher catching it. The #1095/#1531 binding machinery guards per-touch content binding; whatever surface served these (workspace sweep pull, cached widget state) either bypassed the binding check or the binding was computed against the same stale in-memory content (self-consistent but wrong vs disk).

## Acceptance criteria

- A cheap validity gate at every serving surface: a diagnostic whose start line exceeds the file's current on-disk line count is provably stale — demote with the stale marker and emit a `diagnostic_past_eof` record naming file, cited line, actual line count, and serving surface (the observability that makes the next instance attributable).
- Document resync: when the gate fires, trigger a didClose/didOpen (or didChange full-sync) for that file so the desync heals rather than regenerating (the #1631 resync-before-revalidate rule).
- Root-cause instrumentation: log document version + content-length at didOpen/didChange send time vs disk length at serve time, so the DIVERGENCE moment is reconstructible (today we only see the symptom).
- Red-first: open a doc, mutate the file on disk out-of-band to be shorter, serve cached diagnostics citing past-EOF lines → gate must demote; proven red on current code.
- This is #1635's truth-vs-render class 4 made concrete — the "cited line must exist in current content" metamorphic test should land with it.

Refs #1631, #1633, #1635, #1095, #1531.

apmantza:
PR #1664 (merged) covers most of this issue. The past-EOF gate runs at widget-state, `lens_diagnostics` (both `mode=all` and `mode=full`), and the TUI blocking-render loop. Demotion resyncs the document, and `diagnostic_past_eof` plus `lsp_document_send` records make the divergence moment reconstructible. Red-first proven against the live bug shape.

[...]

PR #2013, opened 2026-08-23
feat(freshness): kernel comparator + verdict type; migrate six stores (closes #1739)

### LINKED ISSUE #1739: Freshness kernel: one comparator, one verdict type — the same staleness bug was fixed three times in one day
## Evidence

- The mtime-tolerance defect was fixed three times in 24 hours in three files: `findingPathFreshness` and `isEntryFresh` (#1710), then `reconcileProjectDiagnosticsSnapshot` (#1711) — each an independent reimplementation of the same comparison.
- The #1631/#1641 `stale` collision cost three maintainer conflict rounds because two gates independently owned one field; the `staleReason` discriminator was retrofitted at merge time.
- The mtime-only cache-key collision (#1664, 207/300 wrong cycles) recurred the same day in #1633's import memo — two hand-rolled freshness keys, one lesson.

## Design

One `clients/freshness.ts` module owning:

- The comparator: mtime + size + shared `MTIME_DRIFT_TOLERANCE_MS`, one implementation.
- The `FreshnessVerdict` type with `staleReason` (per-writer semantics, each gate heals only its own — the #1631/#1641 resolution made structural).
- The shared stat and line-count memos (mtime+size-keyed).

Every gate — past-EOF (#1641), dependency-drift (#1631), the delivery gates (#1692's registry enumerates the consumers), reconcile (#1711), workspace-cache epochs (#1687) — imports instead of reimplementing. The #1692 registry's evidence needles then point at kernel calls, which strengthens its guard.

## Acceptance criteria

- The kernel lands with at least the three #1710/#1711 sites migrated, behavior-identical (existing suites green unchanged).
- A registered-or-fail sweep: any `mtimeMs >`-shaped comparison against a scan timestamp outside the kernel fails, with a reason required.
- Red-first on the sweep (current out-of-kernel sites red until migrated or excused).

Refs #1710, #1711, #1708, #1664, #1631, #1641, #1692, #1687.

apmantza:
Closes #1739 — the freshness kernel primitive that enables staleness-store consolidation.

## The kernel

`clients/freshness.ts` owns the single mtime-vs-reference comparison and drift tolerance that at least six stores had independently reimplemented (three carrying the identical #1710/#1711 tolerance defect):

- `freshnessFromMtime({mtimeMs, referenceMs})` → explicit verdict arms: `fresh` / `stale: modified-after-reference` / **`indeterminate: no-mtime-evidence`**. The indeterminate arm exists because callers legitimately differ on stat-failure policy (diagnostics cache drops; the drift detector skips) — the kernel shares the *comparison*, not the *policy*.
- `MTIME_DRIFT_TOLERANCE_MS` ownership moves here (50ms, #1710's measured Windows-skew evidence); `blocker-freshness.ts` re-exports for its existing importers.

## Migrated behavior-identically (148 existing tests green unchanged)

| Store | Notes |
| --- | --- |
| blocker-freshness `detectDrift` | own-file + forward-import mtimes; indeterminate → skip (unchanged) |
| advisory-provenance finding-path freshness | no-reference → live (preserved); stat-fail → missing (preserved) |
| project-diagnostics snapshot reconcile [#1711] | indeterminate → drop (preserved) |
| widget-state record gate | per-record touchedAt comparison |
| widget-state per-entry #1186 gate | per-entry observedAt comparison |
| widget-state dependency-drift #1631 gate | import-mtimes any-stale |
| lsp/workspace-diagnostics-cache deps | fail-closed on unreadable preserved via separate catch |

## Registered-or-fail sweep

`tests/clients/freshness-sweep.test.ts`: every `.mtimeMs > identifier` comparison in `clients/` must carry a descriptor or exemption with a reason. Red-first: written first against the unmigrated tree, it listed all six sites before migration.

Exempted with reasons: age-based TTL checks (`bounded-pid-file-lock`, `installer/index.ts`) are expiry-not-freshness; `lombok.ts` is max-selection; `read-guard.ts` is a session-start membership gate.

## What this enables next

The #1993 render-side fix can reconcile per-entry through this kernel comparator, and the staleness-store umbrella (#1892 direction) gets a single verdict vocabulary instead of per-store booleans.

Refs #1739, #1710, #1711, #1708, #1664, #1631, #1641, #1692, #1687

### DIFF
--- a/.changelog/refactor-1739-freshness-kernel.md
+++ b/.changelog/refactor-1739-freshness-kernel.md
@@ -1,0 +1,5 @@
+---
+section: Added
+---
+
+- **Freshness kernel: one comparator, one verdict type for staleness gates (closes #1739)** — `clients/freshness.ts` owns the mtime-vs-reference comparison and shared drift tolerance that six stores had independently reimplemented (three copies carried the identical #1710/#1711 tolerance defect). `freshnessFromMtime` returns an explicit verdict (`fresh` / `stale: modified-after-reference` / `indeterminate: no-mtime-evidence`) so each caller keeps its own no-evidence policy while sharing the comparison. A registered-or-fail sweep fails on any new out-of-kernel mtime-vs-reference comparison in `clients/`.
--- a/clients/advisory-provenance.ts
+++ b/clients/advisory-provenance.ts
@@ -5,7 +5,7 @@ import type { RuntimeCoordinator } from "./runtime-coordinator.js";
 import { logLatency } from "./latency-logger.js";
 import { normalizeMapKey, toProjectRelativePath } from "./path-utils.js";
 import { resolveRunnerPath } from "./dispatch/runner-context.js";
-import { MTIME_DRIFT_TOLERANCE_MS } from "./blocker-freshness.js";
+import { freshnessFromMtime } from "./freshness.js";
 
 export type AdvisoryFileRole = "source" | "test" | "affected";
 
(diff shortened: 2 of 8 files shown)


apmantza:
Deep adversarial review (AI-generated): REQUEST CHANGES → all findings addressed.

- **P1** (deleted-file ENOENT crash in project-diagnostics reconcile): restored — stat failure now maps to undefined mtime → kernel indeterminate arm → drop, exactly the pre-migration policy. The reviewer proved it red via the existing deleted-file test; green after fix.
- **P2** (drift content): restored pushing the drifted dependency's own path.
- **P3**: sweep exemptions match rel-path only (basename matching would have exempted every index.ts in clients/); lombok key corrected to `lsp/lombok.ts`. Known missed shapes (destructured comparisons, reversed operands, aliased locals) documented in the sweep header as the accepted floor.
- Reviewer's cosmetic-worktree note: discarded.

Per the re-verification convention: touched suites re-run on the fix SHA — freshness-sweep + conformance + all six migrated stores = 117 green; lint/fmt clean.

Issue #2011, opened 2026-08-23
Bash/script writes that static extraction misses bypass dispatch entirely (sed -i.bak, python, node, perl, PowerShell)


apmantza:
## Evidence

Dispatch runs only when a bash tool result yields statically-extracted written paths (`clients/bash-file-access.ts:727`, routed via `runtime-tool-result.ts:546` and `pipeline.ts:1422`). Live-verified dispatch coverage:

| Command | Dispatched? |
| --- | --- |
| `echo x > file.ts` | Yes |
| `tee file.ts` / `sed -i ...` | Yes |
| `sed -i.bak ...` / `sed --in-place ...` | **No** (`bash-file-access.ts:744` matches only the exact token `-i`) |
| `python rewrite.py` / `python -c "open(...).write(...)"` | **No** |
| `node rewrite.js` / `perl -i ...` / PowerShell `Set-Content` | **No** |
| `python generate.py > file.ts` | redirect target only |
| `pnpm exec prettier --write file.ts` | **No** (`npx` yes) |
| bare `cargo fmt` / bare `dotnet format` | No (explicit-target forms yes) |

Live corroboration from this session: an agent edited `clients/*.ts` repeatedly via \`python - <<EOF\` heredoc rewrites. None triggered dispatch; files carried stale diagnostics until a later TS-tool edit re-touched them. Additional known gaps: nested `cd`, variable-held paths, globs, unrecognized wrappers, `cp -t`, commands writing successfully but exiting nonzero (`runtime-tool-result.ts:670` returns early).

## Root cause

Static extraction of explicit written paths is pattern-based and can never be exhaustive. There is no general post-command filesystem-diff fallback, so anything a script writes internally bypasses dispatch completely — no LSP sync, no lint, no format, no cooldown evidence (#1995 seams included).

## Fix direction (needs design assessment)

Run pi-lens over all actually-touched files after a successful bash command, instead of only hook-extracted paths. Two candidate shapes:

1. **Extend static extraction** — cover the enumerated table. Whack-a-mole: every script language, every flag spelling, plus the un-enumerable classes (variables, globs, nested cd, arbitrary interpreters).
[...]

Issue #2000, opened 2026-08-22
Opaque Python/Node/PowerShell writes bypass dispatch and read-guard authorship


apmantza:
## Defect class

Mutation attribution and read-guard provenance blind spot at the shell/tool-result boundary.

## Evidence

Pi-lens synthesizes post-edit write events only when a successful bash tool result contains a command form that extractWrittenPathsFromCommand() recognizes. Recognized explicit paths do two coupled things in clients/runtime-tool-result.ts:

1. readGuard.recordWritten(path) marks the bytes as agent-authored;
2. a synthetic write event runs LSP sync, dispatch, fileSeq, and change-log refresh.

Files changed internally by arbitrary programs receive neither half:

- python rewrite.py
- python -c with Path.write_text/open(...).write(...)
- node scripts/rewrite.mjs
- perl -i
- PowerShell Set-Content / Out-File

There is no general post-command filesystem-diff recovery. An outer redirect such as python generate.py > generated.ts attributes only that explicit redirect destination.

### Read-guard read-across

This is not merely missed lint dispatch.

A recognized bash write calls recordWritten(), so a later Edit is allowed as session-authored content even without a separate Read. An opaque script write does not. Depending on prior state, the later Edit can therefore:

- be blocked as zero-read even though the agent authored the file;
- be blocked as file-modified-since-read with provenance indistinguishable from an external actor;
- lack the synthetic creation/read stamp used by normal Write/bash-write paths;
- miss mutation receipt, post-edit summary, and partial-apply correlation.

The guard is correctly failing closed with the evidence it has; the defect is that the mutation producer never supplies the evidence.

Focused verification is green for supported forms: 130 tests passed across bash-file-access, bash read-guard integration, and synthetic runtime-tool-result routing.

Key seams:

- clients/bash-file-access.ts: command extraction
- clients/runtime-tool-result.ts: recordWritten plus synthetic write routing
- clients/read-guard.ts: session-authored and modified-since-read decisions
- clients/pipeline.ts: post-edit dispatch
- clients/recent-touches.ts: not a general filesystem watcher

## Invariants

1. A host-observable command that changes in-project source must either produce the exact changed-file set or an explicit mutation-coverage-unknown verdict.
2. Read-guard authorship and post-edit dispatch consume the same frozen changed-path set; they must not diverge.
3. An opaque command must never silently imply that no source changed.
4. Ignored/vendor/out-of-project paths remain excluded.
[...]

Issue #936, opened 2026-07-30
Review graph should actually build and persist on large monorepos (enablement, follow-up to #919)


apmantza:
#919/#933 made review-graph build failures *observable* (review-graph.log + surfaced last-attempt reason) and #921/#930 made the file-cap message honest. This issue tracks actually making the graph WORK on large repos, where three structural limits remain:

1. **Persist circuit-breaker** (`GRAPH_PERSIST_MAX_ELEMENTS` default 200K): a repo over the cap builds successfully in-memory and then never persists — every session restarts cold. The #919 reporter's monorepo measures ~208K elements, right over the line. Options: raise the default (measure real load/parse cost first), shard the persisted snapshot, or persist a capped-but-useful subgraph (hot files first) with an honest coverage marker.
2. **Per-session restart, no resume**: the background build starts from scratch each session; on a ~5-6k-file repo with short-lived sessions it may never complete. Options: checkpoint the walk/extraction so a later session resumes, or lower per-file cost enough that a full build fits in a few minutes reliably.
3. **Out-of-band build** (#924): a CLI warm command for CI/cron sidesteps both, but only if the persist cap (1) is also addressed — an out-of-band build that can't persist is useless.

Suggested order: land #933's logging → get the reporter's review-graph.log to confirm which limit actually bites (likely 1) → then size the fix. Related: #919 (observability, closing via #933), #921 (cap honesty), #924 (CLI).

apmantza:
Cost-model assessment (benchmarked against the real compiled builder on Windows/x64, synthetic fixtures at the #919 reporter's NestJS density, ~42 elements/file):

| Files | Elements | Cold build | RSS delta | JSON size | Load cost |
|---|---|---|---|---|---|
| 1,000 | 42,008 | 7.7s | +111MiB | — | — |
| 5,000 | 210,008 | 22.1s | +316MiB | 59.8MiB | read 119ms + parse 297ms + index rebuild 199ms |
| 10,000 | 420,008 | 45.9s | +384MiB | 120.6MiB | (not separately timed) |

Key conclusions:

[...]

Issue #451, opened 2026-07-08
perf: seq-guard review-graph freshness — replace the per-build O(project) walk+stat sweep with RuntimeCoordinator sequence state


apmantza:
## Split out of #450 (cascade deferral)

Every `buildOrUpdateGraph` call runs `getGraphSourceFiles(cwd)` (full directory walk) + `sourceSignatureMapAsync` (stat per file) just to confirm the cached graph is still valid — **157ms p50 across 1,487 reused builds** on this machine, O(project files) on large repos.

RuntimeCoordinator already owns the invalidation signal (`projectSeq` + per-file bumps + change-log with an `external` source), and the project *snapshot* is already seq-guarded — the review graph isn't.

## Sketch

- Coordinator: add a `file → projectSeq-at-last-bump` map (non-breaking; per-file `fileSeq` counters stay as-is) + `getFilesChangedSince(seq)`
- Builder: workspace cache entry records `builtAtProjectSeq`; when a seq hint is provided and all files changed since the cached seq are covered by the incremental-update set, skip the walk+stat sweep and go straight to incremental
- **Staleness guard:** external edits (user IDE, git ops) don't bump seq. Keep a periodic full sweep (every Nth fast-path build or age-based) + rely on the #202 content-hash confirm inside incremental rebuild. Same accepted-staleness class as the snapshot's seq freshness.

## Priority note

After #450 defers cascade off the write hot path, this sweep no longer affects perceived edit latency — it's background CPU/IO contention only (competes with the next edit's dispatch). Worth doing, not worth rushing.

## Broader thread

The seq/change-log machinery is an underused invalidation bus: snapshot uses it; review-graph, reverse-deps (rewrites its snapshot inside every cascade), madge, and word-index each have bespoke freshness schemes. A small shared "seq-guarded cache" primitive would consolidate them (same species as the shared TreeSitterClient seam, #416).

PR #2014, opened 2026-08-23
feat(runtime): one mutation seam with attributed receipts (refs #2000)

apmantza:
Phase 1 of #2000, applying the ONE-store principle to touched-file derivation.

## Design

The seq store (`bumpFileSeq`/`getFilesChangedSince`) was already the canonical "which files changed" derivation — what cascade freshness (#451) consumes. But the bump+change-log pairing existed as **three hand-copied implementations** (`runtime-tool-result.ts`, `runtime-agent-end.ts`, `lsp-mutation.ts`). Each copy is a place where a future mutation class gets dropped — #2011's opaque-write bypass is that exact disease.

This PR consolidates onto **`RuntimeCoordinator.recordProjectMutation({filePath, source, cwd?, changedRange?, onAppendError?})`**, which does three things once:

1. bumps the seq store (semantics unchanged);
2. appends a bounded attributed receipt — `MutationReceipt {seq, filePath, source, turnIndex, ts}` in a 512-entry ring; oldest evicted; `droppedMutationReceiptCount` surfaced so a truncated window is never silently incomplete (shape 9 + #936 honesty);
3. appends the durable change-log entry when `cwd` is provided; append failures route to `onAppendError`, never break dispatch.

Receipt paths use the same `normalizeMapKey`+resolve form as `getFilesChangedSince`, so both views join without a third normalizer. New query: `getMutationsSince(seq)`.

Phase 2 (next PR) adds the bounded pre/post snapshot producer for opaque bash commands feeding this same seam with an explicit source.

Ecosystem note: codex's `TurnDiffTracker` independently validates this shape — one turn-scoped ledger fed by every mutation producer, consumers derive. opencode's `FileMutation` service is the same convergence for owned writes only.

## Tests

**New: `tests/clients/mutation-seam.test.ts` (8 cases)** — pins the seam contract on a REAL `RuntimeCoordinator`: receipt attribution + normalized path form; strict-after-seq filtering of `getMutationsSince`; ring cap evicts oldest with dropped-count > 0 (never silently incomplete); session reset clears receipts alongside the seq store; append-failure isolation (receipt survives, error routed); no-cwd path records receipts without change-log I/O; and a real-sink happy path asserting the actual `change-log.jsonl` bytes (seq/source/filePath round-trip) rather than a logger mock.

**Edited: `tests/clients/lsp/edits.test.ts`** — the mixed-operations test's fake runtime (`bumpFileSeq` spy array) is replaced by a REAL `RuntimeCoordinator` whose receipts and changed-set are asserted directly (`getMutationsSince(0)` order + source attribution, `getFilesChangedSince(0)` normalized keys). This is the point of the seam: lsp mutations demonstrably land in the one store. No other existing tests edited; all prior behavior covered by unchanged suites (197 green across tool-result/agent-end/pipeline/coordinator/conformance).

## Verification

197 tests green across the eight touched suites; build + lint + fmt clean; conformance sweep green (new coordinator state clears in `resetForSession`).

Refs #2000, #2011, #451


apmantza:
Deep adversarial review (AI-generated): REQUEST CHANGES → all findings addressed.

- **P1** (vacuous append-failure test): confirmed and fixed with a REAL filesystem fault — the project data dir is pre-created as a regular file, so the append's recursive mkdir genuinely throws EEXIST. The test now asserts `onAppendError` fired exactly once with an Error AND the receipt survived. Mutation-proofed per convention: neutering the try/catch (`if (false)` guard) turns exactly this test red; restored → green. Reviewer's empirical onAppendError-count-0 observation is now impossible to reproduce silently.
- **P3-1**: `getMutationsSince` returns shallow copies.
- **P3-2**: disappears with the P1 fix (no stray hermetic write).
- All cleared suspicions noted; no action needed.

Per the re-verification convention: mutation proof run (red on mutant, green restored), touched suites re-run on fix SHA — 91 green + lint/fmt clean.

apmantza:
**Blast radius + hot-path measurement (the sections this PR's own new template now requires — retro-applied per the convention discussion):**

**Blast radius** (`module_report blastRadius: true`, post-edit):
- `clients/runtime-coordinator.ts`: 12+ direct dependents (pipeline, runtime-agent-end, git-guard, blocker-freshness, project-snapshot, quiet-window, mcp/session, advisory-provenance, project-diagnostics/fresh-fetch, …). The seam ADDS surface; no existing export changed shape except `bumpFileSeq`'s additive `key` return field (destructuring callers unaffected). Sole `MutationReceipt` writer (shape-24 clear, repo-grepped).
- `runtime-tool-result.ts` / `runtime-agent-end.ts`: delegation-only wrappers — no new symbols or edges; import graphs unchanged.
- `lsp-mutation.ts`: `LspMutationRuntime` gained an optional member — additive, existing fakes typecheck.

**Hot-path cost delta (measured, not assumed):**
- First measurement caught a REAL tax: the seam paid `normalizeMapKey` twice per mutation (~200µs/call realpath on Windows/OneDrive) → **346µs/mutation over bare bumpFileSeq**.
- Fixed in d7f027e8: `bumpFileSeq` returns its key; the seam reuses it. Re-measured: **1.2µs/mutation** (ring push + allocation). At realistic mutation rates this is noise; at the pre-fix rate it was a genuine #1673-class silent tax.

The measurement discipline earned its keep on its first outing.

PR #2018, opened 2026-08-23
feat(runtime): opaque-write recovery via bounded stat diff (refs #2000)

apmantza:
Phase 2 PR-A of #2000 — closes the opaque-write bypass (#2011's defect).

## Mechanism

A bash command whose writes `extractWrittenPathsFromCommand` does not recognize is an opaque candidate. No mutating-verb heuristics (shape 16: hand-classification of tool behavior is how silent gaps ship).

1. **Pre-snapshot** (`handleToolCall`): stat-only `{mtimeMs,size}` over the bounded project source universe (cap 2000, cooperative budget). Awaiting is bounded work on an already-second-scale bash path; fire-and-forget would race the child.
2. **Store**: ONE pending snapshot per cwd (`OpaqueSnapshotStore`) — a newer bash call replaces an unconsumed one with a counted eviction. Bounded by construction, never grows with session length.
3. **Post-diff** (`handleToolResult` bash block, recognized-empty branch): re-stat, diff → frozen changed set feeds BOTH consumers from one pass: `readGuard.recordWritten` + synthetic-write dispatch through the existing per-file loop (receipts flow via the #2014 seam, source attribution preserved).
4. **Coverage-unknown**: missing/evicted snapshot or cap-exceeded emits bounded `opaque_mutation_coverage_unknown` latency rows. Never a clean claim (invariant 3).
5. **Failure atomicity** (invariant 5): partial writes landing before a nonzero exit ARE attributed — the files changed and the agent authored them; skipping leaves ghost staleness. Deliberately diverges from the recognized path's `isError` filter (restore semantics), commented at both sites.

## Tests

**New: `tests/clients/opaque-mutation-scan.test.ts` (4 cases)** — real temp-tree filesystem: snapshot keys normalized + size/mtime captured; file-cap-exceeded verdict above 2000 entries (unbounded-walk proof); real rewrite+add+delete diff (modified/added detected, deletion and unchanged not reported); store single-slot-per-cwd with eviction counting and consume-on-take.

**No existing tests edited.** The handler wiring is thin delegation over this seam; the existing 39 tool-result tests (including the change-log append test CI caught earlier today) pass unmodified.

## Blast radius

`runtime-tool-call.ts` / `runtime-tool-result.ts`: additive branches on the bash path only; non-bash tools untouched. New module has no dependents yet beyond these two sites. Hot-path note: the pre-snapshot awaits only on **bash** tool calls (already seconds-scale); write/edit paths unchanged.

## Known scope for PR-B

AGENTS.md supported/unsupported command matrix; end-to-end handler tests with real `node -e`/`python -c` children (module-level coverage here; wiring covered by the unmodified existing suite).

Refs #2000, #2011

### DIFF
--- a/.changelog/feat-2000-opaque-write-recovery.md
+++ b/.changelog/feat-2000-opaque-write-recovery.md
@@ -1,0 +1,5 @@
+---
+section: Added
+---
+
+- **Opaque-write recovery for bash commands (refs #2000 phase 2)** — commands the path extractor does not recognize (python/node/perl/PowerShell internal writes) now get a bounded pre/post stat diff of the project source universe. Recovered files are attributed to the read guard as agent-authored and dispatched through the mutation seam, with explicit coverage-unknown telemetry instead of silent gaps.
--- a/clients/opaque-mutation-scan.ts
+++ b/clients/opaque-mutation-scan.ts
@@ -1,0 +1,125 @@
+/**
+ * Opaque-mutation recovery (#2000 phase 2, PR-A).
+ *
+ * A bash command whose writes are NOT recognized by
+ * `extractWrittenPathsFromCommand` (python/node/perl/PowerShell internal
+ * writes, restores) previously bypassed dispatch AND read-guard authorship.
+ * This module provides the bounded observation seam:
+ *
+ * - `captureFileStats` snapshots {mtimeMs, size} over the bounded project
+ *   source universe BEFORE the command runs (stat-only; cooperative budget);
+ * - `diffFileStats` produces the changed-path set AFTER;
+ * - `OpaqueSnapshotStore` holds at most ONE pending pre-snapshot per cwd —
+ *   a newer bash call replaces an unconsumed one (counted), so the store is
+ *   bounded by construction and never grows with session length (shape 9).
+ *
+ * Budget exhaustion or a missing snapshot yields an explicit UNKNOWN verdict
+ * to the caller — never a clean claim (issue invariant 3). All stat keys use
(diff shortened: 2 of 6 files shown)


apmantza:
Deep adversarial review (AI-generated): REQUEST CHANGES → dispositions. Review fixes in 3c76fd98.

- **P1 (args/deps compile error)**: already fixed and committed as a8f26d8d before the reviewer's snapshot of the WIP state — the review ran against mid-flight uncommitted work plus stale compiled JS. Current head compiles clean; 68 tests green.
- **P2 entryBudgetExceeded** (the real catch): fixed — captureFileStats now uses collectSourceFilesWithBudgetAsync and maps the flag to a new `entry-budget-exceeded` unknown verdict. A partial universe can never read as a confident empty diff (invariant 3 restored).
- **P2 cross-session swap**: snapshot keys are now `cwd:sessionGeneration`; a concurrent-secondary overwrite degrades to no-pending-snapshot (coverage-unknown) instead of misattributing another session's writes.
- **P2 test-honesty claims**: corrected — module docstring now states handler wiring coverage is PR-B scope, not covered here. The PR body's 'wiring covered by the unmodified existing suite' sentence is withdrawn (passing-unmodified proves non-coverage of new branches).
- **P3s**: same-ms/same-size collision documented in module header (hash confirm = PR-B). Partial-recognition commands skipping telemetry, budgetMs semantics doc, synthetic isError re-entry test, prioritizeCodeKinds — all dispositioned to PR-B with this comment as the record.

Per the re-verification convention: touched suites re-run on fix SHA — 68 green across five suites; build/lint clean.

Issue #1635, opened 2026-08-18
Testing umbrella: catch state-evolution and truth-vs-render bugs at source instead of in dogfooding


apmantza:
From a maintainer-directed AI retrospective (2026-08-18). Sibling of #1605 (smoke lanes) and #1609 (property/atomicity classes). Those cover single-shot behaviors; this umbrella targets the two families dogfooding still catches exclusively: multi-turn state evolution and truth-vs-render divergence. Every staleness incident this week (gitleaks/knip replays, #1631's dependency drift, the advisory-provenance blind spot) was a bug the current suite structurally cannot express.

## 1. Turn-simulation harness (flagship)
A deterministic harness with a fake clock, fake FS, and scripted turn sequences (edit → dispatch → turn_end → edit → turn_end …), asserting what the agent SEES at each turn end. Converts every dogfood complaint into a ten-line replayable scenario. Seed with the four incidents from 2026-08-18 as its first scenarios: gitleaks stale replay (#1622), knip stale cache (#1630), dependency-drift replay (#1631), advisory-provenance file-set blind spot. Build during the v4.1.0 staleness arc — #1633's gate and #1634's unification are its first consumers.

## 2. Session-state lifecycle registry
Generalize #1613: every piece of session-scoped state (latches, holds, memos, dedupe sets, caches) registers with a declared reset policy; one conformance test drives session_start/reset and asserts each re-arms; a derived sweep (atomic-write-scan pattern) flags unregistered state. Kills the process-lifetime-latch class permanently (8 bugs this arc; #1626's sweep-hold nearly the 9th).

## 3. Log-contract tests
Every emitted phase record validated against a registered schema plus completeness rules: no decision without a cause; recovery paths must emit the compensating record (#1606/#1612); teardown paths must emit their proving record even on failure (#1620). Producers enumerated by sweep, not hand-list.

## 4. Truth-vs-render metamorphic tests
Per agent-facing surface: re-derive ground truth independently and assert rendered claims hold — cited line numbers exist in current file content; a header promising N items shows N (#1627-H1); cause labels match actual causes (#1618, #1632-F1); no NaN renders (#1632-F4).

## 5. Renderer round-trips
Parse the composed turn-end message back; assert tier membership equals underlying data — blocking findings in the STOP tier only, demoted entries with identity intact, suppressed ones traced. Guards the gate-stacking seams (#1622/#1625/#1627/#1633).

## Acceptance criteria (per class)
[...]

PR #2019, opened 2026-08-23
feat(runtime): git-first opaque recovery, no universe cap (refs #2000)

apmantza:
Phase 2 follow-up to #2018, driven by live dogfooding.

## The finding

On pi-lens itself (>2000 sources), **every** prescan degraded to `file-cap-exceeded` (15/15 telemetry rows in one session) — recovery never fired on its own repo. The cap existed only because the design required a full-universe pre-snapshot.

## Redesign: timestamp + git window

- **Pre side records only `startedAt`** when inside a git worktree (cached `rev-parse` probe). No walk.
- **Post side**: `git status --porcelain -z` → keep paths with mtime in `[startedAt − 150ms, now]` → same ignore/vendor filters as recognized writes. Handles renames; spawn failures yield explicit `git-failed`; content-identical rewrites correctly unreported.
- **No file-universe bound anywhere on this path** — works at 10 files or 5000.
- Non-git trees keep the bounded stat-diff fallback unchanged.
- Store slot is now `{startedAt, strategy, stats?}`; session-stamped keys unchanged.

## Tests (real git repo, real child processes)

- modified+added detection inside the mtime window
- exclusion of writes older than the window floor
- **end-to-end**: `handleToolCall` → real `node <script>` child writes → `handleToolResult` asserts BOTH that the durable change log records source `opaque-script` for the recovered file AND that `runPipeline` was invoked with `ctx.filePath` = that file — dispatch participation is pinned by assertion, not claimed
- Child scripts execute via temp `.cjs` invoked as `node "<file>"` — invoking the bare path went through the Windows file association and opened the maintainer's IDE instead of executing (found live; fixed and documented)

7/7 suite green; build/lint/fmt clean.

Refs #2000, #2011

### DIFF
--- a/.changelog/feat-2000-opaque-git-recovery.md
+++ b/.changelog/feat-2000-opaque-git-recovery.md
@@ -1,0 +1,5 @@
+---
+section: Changed
+---
+
+- **Opaque-write recovery goes git-first (refs #2000)** — inside a git worktree the pre side records only a timestamp and `git status --porcelain` plus an mtime window answers what changed, with no file-universe cap: recovery now works on any repo size, including large monorepos where the stat-walk previously degraded every command to coverage-unknown. The bounded stat-diff path remains for small non-git trees.
--- a/clients/opaque-mutation-scan.ts
+++ b/clients/opaque-mutation-scan.ts
@@ -21,10 +21,16 @@
  * never a clean claim (issue invariant 3). All path keys use
  * normalizeMapKey+resolve, joining the mutation-seam's key form.
  *
- * KNOWN LIMITATION (shape 6): identity is mtime-window (+size for the
- * stat-diff path). A rewrite landing in the same mtime tick AND the same
- * byte length is undetected on the stat path; the git path detects any
- * content change regardless. A content-hash confirm is future scope.
+ * KNOWN LIMITATIONS:
+ * - shape 6: stat-diff identity is size+mtimeMs - a rewrite landing in the
+ *   same mtime tick AND the same byte length is undetected there. A content
+ *   hash confirm is future scope.
+ * - INVARIANT 4 (per issue): ignored/vendor paths are EXCLUDED. Under the
+ *   git strategy this means writes landing ONLY in .gitignore'd locations
+ *   are never reported - an all-ignored write set reads as an empty (clean)
+ *   recovery BY SPEC, not by oversight. Codegen targeting dist/build-style
+ *   outputs should use explicit redirect destinations so the extractor
+ *   recognizes them.
  */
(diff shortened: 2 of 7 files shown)


apmantza:
Deep adversarial review (AI-generated): REQUEST CHANGES → per-finding dispositions. Reviewer probed against head 8852b31f + uncommitted WIP; fixes landed in a7981278, 6214d47b, ff16c338.

- **P1 uncommitted drift**: resolved — committed deliberately (6214d47b) with kernel migration + session reset + registry entry. Kernel verdict semantics verified directly against `clients/freshness.ts` (`stale` = mtime > reference+50ms; the effective window floor becomes startedAt−100ms — documented).
- **P1 gitignored writes read clean**: disposition = **per-spec**. Issue invariant 4 requires ignored/vendor paths EXCLUDED; an all-ignored write set reading as empty recovery is by design, now documented in the module header with the codegen-redirect guidance. AGENTS.md matrix (PR-B) will carry it.
- **P2 red-first for headline scale bug**: accepted as scoped-out — the git strategy has no file-universe cap *by construction* (recovery cost is git's, not ours); a >2000-file fixture would test git itself. Noted in this comment as the record.
- **P2 blast-radius/measurement**: measured on this repo (>2000 sources): prescan = memoized rev-parse (~218ms cold first probe, then ~0); post-scan `git status` recovery = **~61ms vs 117–193ms for the old stat walk** — the redesign is also a perf win, not just a correctness one.
- **P2 nondeterministic red**: root-caused to stale compiled-JS races during active rebuilds in the main checkout (same class hit three times today); clean clone passed 53/53 at review time, current tree passes repeatedly, CI green twice.
- **P3 copy-status framing**: fixed (ff16c338).
- **P3 window doc/code mismatch**: fixed (ff16c338).
- **P3 gitRepoMemo staleness**: fixed earlier (a7981278 — resetOpaqueMutationState wired into session_start, #1635-registered).

Per re-verification convention: touched suites green on final head; required checks verified on SHA before merge.

apmantza:
**Post-merge telemetry observation (one day of live use):** recovery works (`prescan`×419 all git-strategy, `_recovered`×25), but `_coverage_unknown` ×38 are **all** `no-pending-snapshot`, including many clearly read-only commands (`gh pr view`, `grep`, `ls`). Two possibilities worth a look when we tune this: (a) interleaved bash calls evicting the one-slot baseline more often than expected, or (b) commands whose `tool_call` prescan didn't run at all (subagent children / light-mode sessions share this log). Either way the row can't distinguish "baseline evicted" from "never recorded" — consider a reason split before treating these counts as signal.

PR #2020, opened 2026-08-23
fix(diagnostics): apply strict-FP dispositions immediately via shared cached seam

Don-Yin:
## Summary

`lens_diagnostics mode=all` now applies strict false positive dispositions immediately. It retains the weak no I/O fallback when a cached file cannot be read. One focused regression test covers the reported stale finding.

## Verification

`npm run build` passed.

`npm run lint` passed.

`npm test -- tests/tools/lens-diagnostics.test.ts` passed all 104 tests.

TypeScript LSP reported zero errors in the changed files.

## Note

The repository wide `npm test` currently has unrelated environment dependent failures. Those failures cover absent managed tool binaries and existing installer and LSP cache assertions. The frozen focused gate runs the complete 104 test file for the changed tool.


apmantza (changes requested):
Thank you for the contribution - this targets a real staleness defect, and the focused red-first test is exactly right: we verified empirically that reverting only `tools/lens-diagnostics.ts` (test kept) reproduces the stale-finding failure you are fixing.

**Disclosure: this review is AI-generated; a maintainer supervises the process.**

Verdict: **REQUEST CHANGES** - one substantive gap and two smaller asks.

### P1 - The fix's own goal is unmet in `mode=delta` (class sweep)

The stated goal is "cached false positive marks take effect immediately," but three sibling cache-only sites still serve stale strict-FP marks via `applyWeakDispositions`:

- `filterDeltaReportDispositions` (~L618)
- delta's `visibleWarningFiles` (~L742)
- the carried-over block in the empty-delta text (~L811)

A strict-FP mark therefore still does not take effect immediately in delta mode - same defect class, demonstrably not size 1. Please sweep these sites with the same strict-immediate treatment (same read-failure fallback to weak), or explain why delta is exempt.

### P2 - Unmeasured I/O regression + missing Blast radius section

The change replaces a deliberately zero-I/O filter with `fsSync.readFileSync(s.filePath)` per cached summary on every delta/all query. Precedent: #1625 F2 measured per-file sync reads at ~270ms/run in `clients/diagnostic-dispositions.ts` and hoisted them out of this exact path; `applyWeakDispositions`' doc comment promises a ZERO-I/O instant contract. Please measure the delta (bounded to session-cached widget files, so likely modest) and add the PR-template Blast radius section with numbers.

### P3 - Stale-content edge worth a code comment

If a file changes after dispatch but survives mtime-granularity reconciliation, strict anchors hash NEW content so an existing FP mark silently stops matching and the finding reappears. Fail-open direction - acceptable - but a brief comment at the read site would prevent future confusion.

Everything else checked out cleanly: fallback retained verbatim, blocking-droppable invariant intact, no freshness-sweep violations, delivery registry already covers this surface, tests are real-store and OS-agnostic, composition with master tip is clean.

apmantza:
Round-1 adversarial-review dispositions — pushed directly to your branch as `81cfd3bd` (maintainer edit; your authorship preserved for the original fix).

**P1 class sweep**: ✅ all three sibling sites now use `applyCachedDispositions` — a single shared function extracted from your fix so every cache-only path applies strict-FP marks uniformly. Verified by grep: mode=all summaries, `filterDeltaReportDispositions`, `visibleWarningFiles`, and empty-delta carried-over note all call it.

**P2 measurement**: measured with a 200-file benchmark through the real mode=all rendering path:
| Scenario | Cost |
|---|---|
| No marks (common case) | **~35 ms** (guard skips all I/O) |
| Weak-only store | ~137 ms |
| ≥1 strict FP mark, first query | ~200 ms |
| Repeat queries | ~215–230 ms |

Your original approach was ~330 ms unconditionally even with zero marks. The guards restore the instant contract for common cases while keeping immediate strict convergence when marks exist.

**P3 stale-content edge**: code comment added at the read site.

One test failure in our round-1 verification was a cwd-resolution mismatch in the delta-report FP test (strict anchor landed in the wrong store); fixed by building a cwd-local mark tool.

108/108 tests green; build/lint/fmt clean.

PR #2021, opened 2026-08-23
feat(runtime): close partial-recognition gap + command matrix (refs #2000)

apmantza:
PR-B items for #2000 phase 2, plus the budgeted content-hash confirm (added at maintainer request during review).

## Partial-recognition gap closed

Recovery runs for EVERY bash command with a pending baseline (~60ms git-first). Already-recognized paths are subtracted using the SURVIVING written set — so a failed mixed command attributes its redirect target as `opaque-script` instead of dropping it (failure atomicity, invariant 5). Synthetic opaque events strip `isError` because the main path early-returns on failed host results before attribution. Explicit unknowns: `partial-recognition-no-baseline`, git-hiccup-on-partial.

## Content-hash confirm

Stat-diff identity upgrades to content when both capture sides hashed a file: sha256, bounded by an 8MB per-capture budget; files past it (or asymmetric budget shifts between captures) degrade to mtime+size, documented.

**Measured cost** (round-2 review): hashed capture on a 9MB/300-file tree ≈ 302ms vs 66ms plain (~5×), applied at both captures in NON-GIT trees only → up to ~600ms per bash command there. Git-first repos pay zero (no hashing). Bounded by design; tradeoff accepted for correctness on small non-git trees.

## Tests (real repos, real children, real sinks)

Mixed-command subtraction (redirect=agent-write + internal=opaque-script) · no-baseline telemetry via DELTA read-back of latency.log (append-only-safe) · failed-mixed attribution of BOTH writes · hash-confirm red-first via utimesSync-aligned collision · budget-degrade fallback. 12/12 in suite; all touched suites green.

## AGENTS.md

Supported/unsupported command matrix, recovery window semantics, failure-atomicity source flip, hash-budget asymmetry, telemetry rows.

Refs #2000


apmantza:
Round-2 adversarial review (AI-generated): REQUEST CHANGES → dispositions on 04ae18a2.

- **P2 permanently-green telemetry test**: fixed with a DELTA assertion — lines this run appended vs before-snapshot — so regression re-reds regardless of history.
- **P2 stale body / unmeasured hash cost**: PR body rewritten with the reviewer's own numbers credited (~5× hashed capture, non-git only); budget-asymmetry documented in module header + AGENTS.md.
- **P3 Sonar sha1**: already fixed in 2f56458b after your snapshot (sha256 — not a security context, but not worth the debate); duplicate-branch S1871 collapsed too.
- **P3 source-flip doc**: added to AGENTS.md failure-atomicity row.
- All cleared suspicions noted; red-first empirically verified by you against both a9cd808b and origin/master — thank you, that's exactly the proof the convention demands.

Issue #2023, opened 2026-08-23
safe-spawn Windows .cmd wrapper hard-fails when PATH lacks System32 (chcp coupling)


apmantza:
## Evidence

Found during #2015 test migration (fix-installer-tests agent, 2026-08-23):

`clients/safe-spawn.ts` \`buildWindowsShellCommand\` prefixes every cmd.exe spawn with \`chcp 65001 && …\`. When the child environment's PATH cannot resolve \`chcp.com\` (System32), the \&\& short-circuits and **every** \`.cmd\`/\`.bat\` spawn exits 1 with empty output — silently, since wrapper output is swallowed. Reproduced by stripping a test env's PATH to Node's dir only: all verify spawns failed until \`%SystemRoot%\System32\` was appended.

## Impact

Any embedding environment with a minimal PATH gets silent total failure of all Windows shim launches through safe-spawn. Production installs usually have System32 on PATH, so exposure is edge-case — but it is a hard-fail-with-no-signal design.

## Fix directions

1. Pin chcp via absolute path: \`%SystemRoot%\System32\chcp.com\` (no PATH dependency).
2. Or use \`&\` sequencing so a chcp failure doesn't kill the real command (weaker: codepage then unapplied).
3. Or drop chcp when the target is known-ASCII-safe.

Also noted while measuring: \`getToolPath\` miss-path cost scales strongly with PATH length (~7ms → ~900ms when System32 present; npm-global candidate probing spawns .cmd wrappers). Separate but adjacent.

*This issue is AI-generated, with the maintainer supervising the process.*

Issue #1569, opened 2026-08-17
probe-cache.json persists a transient-degraded tool tier for 24h across processes


apmantza:
From PR #1563's class sweep (maintainer-supervised AI session). The cross-process variant of the family.

`clients/installer/index.ts` (`getToolPath`) persists the selected tool tier to `~/.pi-lens/probe-cache.json` with a 24h TTL — so a degraded-tier selection made during a transient stall survives up to 24 hours ACROSS PROCESSES. Worse than the in-process cases: no session reset touches it.

Acceptance: the persisted entry records whether any candidate tier was transient at selection time; a transient-tainted entry is not served past the transient cooldown (or not persisted at all); red-first with a probe simulating the stall + process restart. Refs #1537, refs #1539.

PR #2024, opened 2026-08-23
fix(installer): tree-killed verifies; transient probes keep installs (closes #2015)

### LINKED ISSUE #2015: markdownlint-cli2 auto-install verify loop: SIGTERM'd verifications starve the runner into 30s timeouts on every large .md edit
## Evidence

Meta-dogfooding log audit (2026-08-23), real `~/.pi-lens/` data:

- Two runner timeouts today on `pi-lens/AGENTS.md`:
  `{"runnerId":"markdownlint","durationMs":30081,"status":"failed","failureKind":"timeout"}` (06:22:05Z) and `30018`ms (09:27:45Z).
- Install telemetry shows the cause chain: **2 × "auto-install markdownlint-cli2: installed but verification failed, cleaning up"** plus **23 × "auto-install verify failed (signal=SIGTERM)"** — an install/verify kill loop. The managed install never survives verification, so per-edit runs fall back to a slow/absent binary and charge the full 30s runner timeout.
- Blast radius: `slow-hook-path: 66` today; worst tool totals ~60s on `.changelog/*.md` edits (59906ms on one file) — the markdownlint timeout inflates whole per-edit pipelines.
- Anomalous siblings yesterday: two `lsp_document_drift` phases on `runners/markdownlint.ts`, 214829ms (`resynced`) and 70229ms (`deferred`).
- #2012's teardown evidence and #2009's spawn-timeout cooldown do not touch this lane: the kill happens inside installer VERIFICATION, not at spawn or runner teardown.

Refs #1995 (closed by #2009 for the cooldown seam; this is the remaining upstream defect that made those timeouts appear in the first place).

## Root cause hypothesis

The installer's post-install verification of markdownlint-cli2 is repeatedly SIGTERM'd (23 kills). Either the verify child exceeds an internal budget on Windows (`npm exec`-style cold start of the `.cmd` shim), or the ambient abort signal is reaching the verify spawn (#197's ambient signal is captured at spawn time — a turn abort mid-verify would kill it and the cleanup path then re-installs next time, looping).

## Acceptance criteria

1. Verify failures record WHY they were killed (timeout vs signal vs exit) with a bounded row — today the loop is only visible as repeated identical lines.
2. A verify failure must not silently retry-install on the next edit within the same session without backoff; N consecutive verify failures cool the lane down like other availability latches (#1500 classification rules apply: derive the outcome from `classifyProbeFailure`, never hardcode `missing`).
3. After fix: no 30s markdownlint runner timeout rows on this machine for AGENTS.md-sized files when the managed install is healthy; if the binary genuinely cannot verify, the runner reports unavailable quickly instead of charging the full timeout.
4. The existing `tool-install-noise` count drops from its current 33/day baseline.

## Test matrix

- Verify passes first try → single install, no retries.
- Verify killed by timeout → typed outcome + backoff, no same-session reinstall loop.
- Verify killed by external abort (turn interrupt) → distinct from timeout, does not poison the latch.
- Healthy install after previous failures → recovery visible in telemetry.

apmantza:
Closes #2015 — root-causes and breaks the markdownlint-cli2 install/verify/reinstall loop.

## Root cause (empirical)

`verifyToolBinary` spawned via raw `node:child_process` with `shell:true` on Windows. The 10s timeout SIGTERM reached **only cmd.exe** — the grandchild node process was orphaned, kept scanning, held handles against the cleanup `rm`, and the freshly installed binary was deleted anyway. One day's telemetry: 17 failed verifies (15 SIGTERMs), 4 cleanups, one 42.8s reinstall burn, worst runner timeout 44.2s. Baseline recorded on the issue.

## Three coupled changes

1. **`verifyToolBinary` → `safeSpawnAsync`**: tree-kill via lifetimeCoupled semantics, typed `result.signal`/`spawnFailure`, failure logs carry explicit `kind=killed-signal=<sig> | spawn-error | exit-nonzero`.
2. **Transient-keep policy** in `installNpmTool`: when the last verify attempt was transient (killed/spawn-boundary), keep the installation and return undefined — the next `ensureTool` re-probes the existing binary via discovery instead of reinstalling (#1569 semantics: a killed prober is not a verdict). This is what actually ends the loop; a backoff latch alone would still churn.
3. **Nonzero exit stays a real verdict**: cleanup path unchanged for binaries that genuinely cannot run.

## Blast radius

- `verifyToolBinary` callers: install paths + managed-tool-refresh + probe-cache discovery — all consume the boolean contract, unchanged. `onTransient` semantics preserved (now also fires on typed spawn-boundary failures per #1569).
- Hot-path cost: verify spawns unchanged in count; each gains safe-spawn wrapper overhead (~ms) against eliminating multi-second orphan cleanup contention. No new I/O.

## Tests

**New: `tests/clients/installer/verify-binary-semantics.test.ts`** — real shims, platform-branched: exit-0 verifies + forwards version; nonzero exit = real verdict (NOT transient); timeout kill fires transient callback with bounded elapsed (tree-kill proof).

**Adapted (5 suites)**: re-pointed from the dead raw-spawn seam to safe-spawn mocks. `managed-tool-refresh.test.ts` uses a hoisted fixture-exit-code map answering --version probes centrally so scenario-controlled spawnMocks keep working. `strategies` uses an overridable verifyProbe slot preserving the pip test's stateful probe counter. `installer-lifecycle` gained System32-on-PATH in its stripped test env (safe-spawn's chcp prefix requires it — gap filed as #2023).

253/253 across all 15 installer suites (4 consecutive runs); build/lint clean.

Closes #2015

### DIFF
--- a/.changelog/fix-2015-installer-verify-loop.md
+++ b/.changelog/fix-2015-installer-verify-loop.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Installer verify loop broken at the root (closes #2015)** — `verifyToolBinary` now spawns through `safeSpawnAsync` (tree-kill on timeout, typed kill-reason) instead of raw spawn whose SIGTERM orphaned grandchild node processes on Windows `.cmd` shims. A killed/inconclusive prober no longer counts as a verdict: the freshly installed binary is KEPT for cheap re-probe instead of deleted, ending the install/verify/reinstall churn (23 SIGTERMs and 4 cleanups observed in one day).
--- a/clients/installer/index.ts
+++ b/clients/installer/index.ts
@@ -2124,89 +2124,67 @@ export async function verifyToolBinary(
 	 */
 	timeoutMs = 10000,
 ): Promise<boolean> {
-	return new Promise((resolve) => {
-		const isWindows = installerPlatform() === "win32";
-		const hasKnownWindowsExt = /\.(cmd|exe|ps1)$/i.test(binPath);
+	// #2015: safeSpawnAsync instead of raw spawn. Raw spawn's timeout
+	// SIGTERMed only cmd.exe on Windows (.cmd shims run shell:true), orphaning
+	// the grandchild node process - which kept scanning, held handles against
+	// the cleanup rm, and turned one slow cold-start verify into an
+	// install/verify/reinstall loop (23 SIGTERMs observed in one day).
+	// lifetimeCoupled tree-kill prevents orphans; result.signal gives a typed
+	// kill-reason for transient classification.
+	const isWindows = installerPlatform() === "win32";
+	const hasKnownWindowsExt = /\.(cmd|exe|ps1)$/i.test(binPath);
 
-		// On Windows, resolve the best executable path:
-		// - extensionless → prefer .cmd (cmd.exe-safe)
-		// - .ps1 → prefer .cmd sibling to avoid PowerShell execution-policy hangs
-		// - .cmd / .exe → use as-is
-		let execPath =
-			isWindows && !hasKnownWindowsExt ? `${binPath}.cmd` : binPath;
-		let useShell = isWindows && /\.(cmd|bat)$/i.test(execPath);
+	// On Windows, resolve the best executable path:
+	// - extensionless → prefer .cmd (cmd.exe-safe)
+	// - .ps1 → prefer .cmd sibling to avoid PowerShell execution-policy hangs
+	// - .cmd / .exe → use as-is
+	let execPath = isWindows && !hasKnownWindowsExt ? `${binPath}.cmd` : binPath;
 
-		if (isWindows && /\.ps1$/i.test(execPath)) {
-			const cmdSibling = `${execPath.slice(0, -4)}.cmd`;
-			if (require("node:fs").existsSync(cmdSibling)) {
-				execPath = cmdSibling;
-				useShell = true;
-			} else {
-				// Fall back to running without shell — cmd.exe can't run .ps1
-				useShell = false;
-			}
+	if (isWindows && /\.ps1$/i.test(execPath)) {
+		const cmdSibling = `${execPath.slice(0, -4)}.cmd`;
+		if (require("node:fs").existsSync(cmdSibling)) {
+			execPath = cmdSibling;
 		}
+	}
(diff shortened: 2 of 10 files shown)


apmantza:
Round-1 adversarial review (AI-generated): REQUEST CHANGES → dispositions on 1db2e968.

- **P2 red-first/tree-kill proof**: rebuilt the timeout test — cmd.exe now launches a DETACHED node grandchild writing a marker at +6s while cmd holds past the 1.5s budget. Pre-fix raw spawn orphans the grandchild → marker appears → **RED on origin/master (empirically verified via checkout of master's installer/index.ts)**; safeSpawnAsync tree-kills it → green. The suite is no longer non-red on master.
- **P2 class sweep**: python user-base probe (raw spawn, NO timeout — a hung python hung the candidate walk forever) migrated to safeSpawnAsync with 5s budget. `lsp/launch.ts` shell:true exempted with inline reason (long-lived server owned by LSP lifecycle).
- **P2 Windows CI visibility**: acknowledged as a gap — unit-tests job runs ubuntu-only. Disposition: the .cmd path is exercised by local dev + nightly tool-smoke; #2023 tracks the adjacent chcp coupling. A windows vitest job is worth filing separately.
- **P3 ENOENT/EACCES false-keep**: real catch — only kill-class failures (signal, kind=timeout|killed) are transient now; kind="spawn" (child never launched) remains a cleanup-verdict.
- **P3 eternal-reprobe**: bounded per call (~150s worst) and availability suppression caps caller retries; a cross-session latch deferred — noted here as the record.
- **P3 shimExitCodes leak**: cleared in beforeEach.
- **P3 measurement**: end-to-end verify cost dominated by child runtime (unchanged); wrapper overhead measured in-suite (<15s worst-case budget assertion on a 30s child = tree-kill latency proof).

253/253 installer suites green on final head; build/lint clean.

apmantza:
Round-1 addendum (from CI): the strengthened grandchild test red'd on Linux — root cause is a REAL safe-spawn parity gap: `killPidTreeSync` on POSIX kills only the direct child (`process.kill(pid)`, no group walk), so grandchildren survive timeouts. Filed **#2026** (fix = POSIX detached+group-kill; touches every POSIX spawn through safe-spawn, so it warrants its own reviewed PR rather than a drive-by here). The Windows tree-kill assertion — #2015's actual platform and defect home — remains active; POSIX skips with a pointer to #2026.

Issue #571, opened 2026-07-12
lens_diagnostics_full scan results never reconcile into the footer/widget-state cache


apmantza:
## Symptom (observed live on pi-drykiss, 2026-07-12)

The TUI footer kept showing a TypeScript error (`typescript:2353`, "correlationId does not exist") on `src/review-scope.ts:162` well after the underlying issue was fixed (a `correlationId` field was added to `ScoutOptions` in `src/scout.ts`). Independent verification confirmed the error was genuinely gone:

- `npx tsc --noEmit` (bypasses pi-lens's LSP client entirely) — clean, no errors.
- Grep confirmed `correlationId` exists in both `ResolveReviewScopeOptions` (`review-scope.ts:74`) and `ScoutOptions` (`scout.ts:47`).
- A `lens_diagnostics_full` scan had already run and would have reported `review-scope.ts` as clean (real diagnostics for the file were fetched fresh during that scan).

Yet the footer never updated.

## Root cause

`recordDiagnostics` (`clients/widget-state.ts:238`) is the **only** writer to the footer's `allDiagnostics` store, and it has exactly one caller: `clients/pipeline.ts:1215`, inside the **per-edit dispatch pipeline**. It only runs when a file is actually edited through pi-lens's own edit tool.

`lens_diagnostics_full` (and the `lsp_workspace_diagnostics` scan underneath it) is a separate, read-only reporting path — grepped and confirmed it never calls `recordDiagnostics`. So even though a full scan does the work of fetching fresh, authoritative diagnostics for every file in the project (including ones the footer has stale data for), that fresh data is reported back to the caller and then discarded — never reconciled into `widget-state`.

Practical consequence: if file A's diagnostics only became invalid/valid because of a change in file B (a shared interface, a type, etc.), and A itself is never directly re-edited, the footer for A can be stale indefinitely — a full scan proves the fresher truth but has no path to correct the footer.

## Proposed fix

Have `lens_diagnostics_full` (or the workspace-diagnostics scan it wraps) call `recordDiagnostics(filePath, diagnostics, writeIndex?)` for each file it freshly checks, using the same write-ordering-guarded path `clients/pipeline.ts` already uses (`clients/write-ordering-guard.ts`'s `WriteOrderingGuard`, wired in #560) so a concurrent per-edit write during the scan can't be clobbered by an older scan result or vice versa.

Guardrails:
- Must not regress the write-ordering protection #555/#560 already built — reuse the same guard, don't bypass it.
[...]

PR #2025, opened 2026-08-23
fix(diagnostics): authoritative LSP sweep retires stale widget state (refs #1993)

apmantza:
Fixes the #1993 defect class: mode=full replaying stale mid-edit blockers as current 🔴 beside a clean sweep.

## Root cause

`mergeDiagnosticsWithWidgetSummaries` was additive-only: widget-store diagnostics seeded `byFile` first, fresh LSP findings merely dedup-merged beside them. A **confirmed, fully-covered clean sweep therefore never retired** the file's stale entries. Observed originally on `clients/dispatch/runners/oxlint.ts`: ts:2393/ts:2528 ghost blockers with past-EOF positions persisting across repeated mode=full calls while tsc, rg, and the sweep itself all reported clean — and re-running didn't reconcile them, because nothing in the merge ever removed anything.

## Fix

`mergeDiagnosticsWithWidgetSummaries` gains an `authoritativeLspFiles` set (resolved paths with a confirmed, fully-covered result, computed in `formatFullMode`). Widget state for those files is not seeded; the fresh sweep's own diagnostics stand. Files without full coverage keep today's additive behavior — fail-open by design: an unconfirmed or timed-out sweep must never retire widget state it did not actually re-check.

## Tests (red-first verified by stashing the production change)

1. **Updated** dedup test pins the inversion: fresh workspace-scan copy renders, cached dispatch-time copy retired — still exactly one row (dedup-by-file/line/rule guarantee unchanged).
2. **New #1993 test**: a clean fully-covered sweep retires the stale blocker (`totalBlocking` absent-at-zero; text clean), AND the fail-open guard — with no sweep result for the file, the stored finding stays visible.

104/104 in the suite; 177 green across touched suites; build/lint/fmt clean.

## Blast radius

Single call site (`formatFullMode`); new optional parameter — additive signature. Non-bash/diagnostic surfaces untouched. The footer-reconcile loop already full-replaced per fully-covered result (#571); this brings the rendered summary to the same semantics.

## Tests section (template)

- New test "a clean fully-covered sweep RETIRES stale widget blockers" — regression proof for #1993 + fail-open guard assertion.
- Edited test "deduplicates LSP diagnostics..." — survivor flipped to the fresh sweep copy per authoritative semantics; duplicate-row guarantee still pinned.

Refs #1993

### DIFF
--- a/.changelog/fix-1993-full-mode-stale-blockers.md
+++ b/.changelog/fix-1993-full-mode-stale-blockers.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **mode=full no longer replays stale mid-edit blockers (refs #1993)** — a confirmed, fully-covered LSP sweep is now authoritative for its files: widget-store diagnostics captured from a since-fixed broken intermediate state are retired instead of rendering as current 🔴 blockers beside a clean sweep. Unconfirmed or timed-out sweeps keep the fail-open behavior.
--- a/tests/tools/lens-diagnostics.test.ts
+++ b/tests/tools/lens-diagnostics.test.ts
@@ -2094,9 +2094,89 @@ describe("lens_diagnostics mode=full", () => {
 
 		const result = await run(makeTool({}, lspService), { mode: "full" });
 		const text = String(result.content[0].text);
-		expect(text).toContain("cached dispatch message");
-		expect(text).not.toContain("same diagnostic from workspace scan");
+		// #1993: a CONFIRMED, fully-covered sweep is AUTHORITATIVE for the file -
+		// the fresh workspace-scan copy renders and the cached dispatch-time
+		// copy is retired instead of the reverse. Still exactly ONE row (the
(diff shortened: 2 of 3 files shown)


apmantza:
Round-1 adversarial review (AI-generated): **APPROVE** with three P3s → all dispositioned on HEAD.

- **P3 retirement observability**: `lsp_authoritative_widget_retire` bounded latency row now fires when files' widget diagnostics are retired (count only, no content) - a falsely-authoritative regression can no longer vanish silently.
- **P3 weak-anchor nuance**: documented at the skip site per your suggestion.
- **P3 class-sweep statement**: added to the PR body - merge fn has exactly one production caller (formatFullMode), footer full-replaces (#571), delta/all cache-only by contract. Your grep proof incorporated.
- Red-first independently re-verified by you against real master after correcting the clone-HEAD mistake - appreciated; that's exactly the rigor the convention demands.

104/104 on final head; build/lint clean.

PR #2027, opened 2026-08-24
fix(spawn): POSIX process-group kill reaches grandchildren (closes #2026)

### LINKED ISSUE #2026: safe-spawn POSIX killPidTreeSync kills only the direct child - grandchildren survive timeouts
## Evidence

Empirically reproduced today by the #2015 tree-kill regression test (tests/clients/installer/verify-binary-semantics.test.ts, \"grandchild\" case):

- Shim spawns a detached node grandchild that writes a marker file at +6s, then holds past verifyToolBinary's 1.5s budget.
- On Windows: taskkill /F /T kills the whole tree -> no marker (green).
- On Linux CI: \`process.kill(pid, \"SIGKILL\")\` (clients/safe-spawn.ts:392, killPidTreeSync) terminates ONLY the direct child (sh). The backgrounded node grandchild survives and writes the marker (red).

## Impact

Every POSIX timeout through safeSpawnAsync orphans grandchildren - long-running children spawned by killed tools keep running (the exact #2015 defect class, POSIX flavor). The Windows branch has taskkill /T; POSIX has no equivalent.

## Fix directions

1. Spawn POSIX children with \`detached: true\` so each child leads its own process group, and kill the GROUP (\`process.kill(-pgid, SIGKILL)\`) - grandchildren inherit the group, so one signal reaches the whole tree. Caveat: detached changes terminal-signal delivery semantics; needs a deliberate review of every caller.
2. Or walk /proc/<pid>/task/*/children recursively (Linux-only; macOS needs pgid approach anyway) - option 1 is the portable one.

Blast radius note: changing default detachment affects ALL POSIX spawns through safe-spawn (signal delivery, terminal coupling). Needs its own PR with per-caller review, not a drive-by.

apmantza:
Closes #2026 — POSIX safe-spawn timeouts now kill the whole process tree.

## Root cause (empirical)

`killPidTreeSync` on POSIX did `process.kill(pid, SIGKILL)` — **direct child only**. Windows had `taskkill /F /T`; POSIX had nothing equivalent, so any tool spawning its own children (npm→node, sh→sleep, pip→python) survived timeouts as orphaned grandchildren. Proven by the #2015 grandchild regression test: green on Windows, red on Linux CI (marker file written by the surviving grandchild — evidence recorded in #2026).

## Design

- **POSIX children spawn `detached: true`** → each child leads its own process group (pgid == pid).
- **`killTree` signals the GROUP**: SIGTERM(−pgid), escalating to SIGKILL via the existing 1s timer; ESRCH falls back to a direct pid attempt.
- **`killPidTreeSync` does the same** for lifetime cleanup (process exit + SIGINT/SIGTERM/SIGHUP forwarding).
- **Detached children stop receiving terminal signals directly**, so EVERY POSIX pid registers for lifetime cleanup — pi's own signal/exit handlers forward the kill, preserving die-with-host semantics for coupled AND uncoupled spawns.
- Windows path untouched (taskkill /T was already correct). LSP servers unaffected — they spawn through their own launch seam by design.

## Blast radius (the honest part)

Every POSIX spawn through safe-spawn changes detachment semantics. Mitigations verified:
- Signal forwarding preserved via lifetime registration (all POSIX pids now registered, not just lifetimeCoupled).
- Piped stdio works identically detached.
- Full installer + fault-injection-kit + spawn-semantics suites green: **394 tests**, including the kit's wedged-child and teardown-evidence tests that exercise these exact paths.

## Tests

The #2015 grandchild assertion is **un-gated**: red on pre-fix Linux CI (#2026's evidence), green here. Plus 394-suite regression sweep.

## AGENTS.md

Runner-process-model section documents the POSIX group semantics.

Closes #2026


apmantza:
Round-1 adversarial review: REQUEST CHANGES → dispositions on HEAD.

- **P2 escalation gate**: fixed — group-mode SIGKILL escalation now probes GROUP liveness (`process.kill(pgid, 0)`) instead of gating on direct-child death, stays REF'D past resolve as that group's only remaining backstop (finalize deletes the pid from lifetimeState), and emits a bounded `spawn_group_kill_escalation` row when it fires (**P3-3 observability** also closed). The writer test now IGNORES SIGTERM — surviving soft TERM proves the hard-escalation path; red by construction on the pre-fix ungated timer.
- **P2 dead discriminator**: collapsed to a single documented fallback (any failure → direct-child attempt).
- **P3-1 residual detachment blast radius**: acknowledged in the PR's Blast radius section — handler-doesn't-run scenarios (sync-blocked pi, hard kill) orphan children where master's same-group children would have died with the terminal. No caller relying on die-without-pi-exit found; accepted with documentation.
- **P3-2 class sweep**: added to PR body — merge-class is size-1 at the safe-spawn chokepoint for bounded tools; named sibling `clients/mcp/review.ts:61` (own worker, no known forks) deferred with reason rather than silently unswept.
- **P3 shimExitCodes leak** from round-1 of #2024 lineage: already cleared.

280 tests green across installer + fault-injection + spawn-semantics suites on final head; build/lint clean.

Issue #2028, opened 2026-08-24
Edit-time advisory replays stale diagnostics for script-modified and deleted files long after they are fixed or removed


apmantza:
## Evidence

During 2026-08-23/24 dogfooding, the per-edit advisory (the \"🔴 STOP\" block injected on tool use) repeatedly cited diagnostics for \`tests/clients/safe-spawn-kill-escalation-timer.test.ts\` that were verifiably already resolved:

- Multiple turns in a row reported \`L105/L106 Unterminated string literal\` and duplicate-import errors at line numbers whose content had been rewritten. Ground truth each time: \`npm run lint\` = 0 errors, suite green (verified by rerun immediately before and after).
- After the file was **deleted entirely** (superseded by another suite; deletion committed), subsequent check blocks continued to reference its content (\"Test file not found\" phrasing plus stale line diagnostics) for at least two more turns.
- Same replay pattern appeared for \`clients/runtime-tool-call.ts\` and \`tests/clients/mutation-seam.test.ts\` mid-edit states: errors cited line numbers whose content had been replaced, persisting past verification.

Each stale claim was individually disproven with fresh greps/builds/lint runs recorded in-session; none were phantom reports about CURRENT content.

## Why this happens (hypothesis, ranked)

1. **Script-modified files skip fresh LSP sync.** Writes made through bash/python heredocs are opaque to the pre-#2019 host; without recovery (this repo exceeded the old stat-diff cap all day: 419/419 prescans capped), no synthetic dispatch ran, so no lane re-checked the file - yet the advisory still rendered cached diagnostics for it on later touches.
2. **Deleted files lack an existence-drop on this surface.** \`dropFindingsForMissingPaths\` (#1461) gates some delivery surfaces; whether the inline edit-advisory is among \`DELIVERY_SURFACES\` gated entries needs verification - today's behavior suggests it served cached content for a deleted path.
3. Post-restart, git-first recovery + authoritative retirement (#2025) should close most of this - but the deleted-file case may still leak if the advisory reads summaries without an existence probe.

## Acceptance criteria

- [ ] Identify which store/lane feeds the per-edit advisory block and whether \`DELIVERY_SURFACES\` registers it gated or labeled.
- [ ] Deleted file: its cached findings must be dropped (or explicitly labeled deleted-path) on the next advisory, never re-cited. Add a red-first test driving mark -> delete -> next-edit advisory.
[...]

PR #2030, opened 2026-08-24
feat(delivery-gate): register remaining agent-facing surfaces (refs #2028)

apmantza:
Closes the #2028 class sweep: every agent-facing delivery surface now registered in `DELIVERY_SURFACES`.

## Gated (real staleness wiring)

- **`tool-call:stop-blocker`** (`clients/pipeline.ts`) — the 🔴 STOP advisory now routes its blockers through `dropFindingsForMissingPaths({ store: "stop-blocker", ... })` before rendering. Deleted files' blockers stop rendering; when all cite deleted paths, no ghost `🔴 STOP — 0 issue(s)` header appears at all.
- **`lsp-diagnostics:tool-output`** (`tools/lsp-diagnostics.ts`) — gates cite the tool's existing staleness stack: `createWorkspaceDiagnosticsCacheContext` entry + `cacheCtx.lookup`, whose cache module performs own-file mtime, reverse-dependency mtime, and #1095 content-binding validation.

## Labeled (live — nothing stored to go stale)

- `git-guard:commit-blocked` — synchronous preflight rejection returned inline with the failed git command.
- `read-guard-tool-lines:preflight-errors` — hashline BLOCKED / RE-READ REQUIRED computed fresh per edit attempt.
- `agent-behavior:thrashing-notice` — ephemeral in-result notice computed at detection time.

## Honest scope note

The stop-blocker gate changes RENDERING only: `result.hasBlockers`, inline-blocker summary, and git-guard turn-state records still reflect the raw set. Changing those semantics was out of scope; noted here as the record.

## Tests

Two new integration tests in `tests/clients/pipeline.test.ts` (existing mocked-dispatch harness): deleted-file blocker dropped from rendered output; mixed live/deleted set renders survivors only. **Red-first verified by stashing only the pipeline wiring** — both failed on pre-fix code with the deleted-file blocker rendering. Registry enumeration (`EXPECTED_SURFACE_IDS`) updated for +5.

124 tests green across finding-delivery-gate / pipeline / lsp-diagnostics suites; build/lint clean.

Refs #2028

### DIFF
--- a/.changelog/fix-2028-delivery-gate-surfaces.md
+++ b/.changelog/fix-2028-delivery-gate-surfaces.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Register remaining agent-facing delivery surfaces (#2028)** — The per-edit 🔴 STOP block now drops blockers whose cited file no longer exists, so a deleted file's stale blocker no longer re-blocks the agent with no remediation. All five previously unregistered surfaces (stop blocker, `lsp_diagnostics` output, git-guard verdicts, read-guard preflight errors, thrashing notices) are registered in the finding-delivery gate registry.
--- a/clients/finding-delivery-gate.ts
+++ b/clients/finding-delivery-gate.ts
@@ -505,6 +505,57 @@ export const DELIVERY_SURFACES: Record<string, DeliverySurfaceEntry> = {
 		["reconcileProjectDiagnosticsSnapshot"],
 		["reconcileProjectDiagnosticsSnapshot("],
 	),
+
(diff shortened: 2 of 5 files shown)


apmantza:
**CI note:** the Unit tests failure on head e56872e3 was an OOM kill (job duration 115s vs ~4.5min norm, exit 137, zero test failures logged). The identical code passed Unit tests on the immediately-prior head, and local runs are green (124 tests across touched suites). Re-run succeeded for the previous occurrence of this pattern earlier today. Master's own CI re-validates on merge.

PR #2036, opened 2026-08-24
feat(lsp): collect-later delivery for slow auxiliary LSPs (closes #2001, #2002, refs #2026)

### LINKED ISSUE #2001: Opengrep declares 3.5s but the 2s auxiliary ceiling cuts off most live scans
## Defect class

Declared server budget is systematically truncated below observed completion time.

## Live evidence

Opengrep's wait strategy declares aggregateWaitMs: 3500 because warm scans are documented around 1.3s and cold/rule-load scans can take about 3.5s.

Both auxiliary paths apply a default global ceiling of 2000ms through PI_LENS_AUX_GRACE_MS policy. In one 13-hour dogfood session:

- 23 opengrep waits answered
- 46 were cut off
- 3 were silent
- answered p50: 867ms
- answered p95: 1259ms
- two notify writes landed late at 2327ms and 2548ms

Thus roughly 64% of classified opengrep waits hit the 2s ceiling, and there is direct evidence that at least some useful work lands shortly after it.

This is distinct from verdict correctness: current code correctly reports partial coverage and names opengrep. The problem is that the security scanner often fails to contribute to the edit that triggered it.

## Invariants

1. Security-scanner coverage gaps remain explicit; a longer wait must never turn silence into clean.
2. Interactive latency has a declared ceiling and is measured against coverage gained.
3. A queued/wedged scanner must not charge its full cold budget on every edit.
4. Late content-bound answers remain reusable and cannot be double-counted.
5. Per-server declared budgets and global ceilings must not contradict silently.

## Acceptance criteria

- Run a controlled dogfood A/B with PI_LENS_AUX_GRACE_MS at 2000, 2500, and 3500.
- Measure opengrep answered/cut-off/silent rates, edit p50/p95, notify queue depth, breaker activity, and matching late publications.
- Decide policy from the curve:
  - raise the default if coverage gain is substantial and latency bounded;
  - otherwise use adaptive grace based on warm history/queue state;
  - or detach opengrep into collect-later delivery while keeping the current edit responsive.
- Record effective declared budget, global ceiling, and the winning limiter in lsp_aux_wait_outcome.
- Update stale comments/tests that describe a 2500ms caller cap while both current call sites default to 2000ms.

## Test matrix

- Warm answer at 1.3s, late answer at 2.3-2.6s, cold answer at 3.5s, and permanent silence.
- Healthy queue versus notify-inflight contention.
- Primary clean/nonempty outcomes.
- Subsequent unchanged-content read consuming a late bound answer.
- Global override below/equal/above the server declaration.
- Multiple auxiliaries with different declared budgets.

Refs #1458, #1549, and #1884.

### LINKED ISSUE #2002: typos consumes ~1.43s of every auxiliary grace despite a documented sub-100ms scan
## Defect class

Steady-state auxiliary latency contradicts the declared wait-policy model and sets the interactive edit floor.

## Live evidence

In one 13-hour dogfood session, typos produced:

- 80 answered outcomes
- 1 silent
- 1 deferred
- answered p50: 1431ms
- answered p95: 1491ms
- maximum: 1526ms

The strategy declares aggregateWaitMs: 1500 and documents a native single-file scan as sub-100ms whose seedFirstPush should arrive near-instantly. Live behavior instead consumes almost the complete budget on ordinary edits.

Because the auxiliary grace completes only when the slowest admitted auxiliary settles, typos regularly defines a roughly 1.4-1.5s lower bound even when TypeScript, ast-grep, and opengrep answer earlier.

This is separate from #2001: opengrep is often cut off by too small a ceiling; typos answers reliably but unexpectedly late.

## Invariants

1. A seedFirstPush scanner documented as near-instant must either meet that contract or use a policy matching measured behavior.
2. Spell-check coverage gaps remain explicit; shortening the synchronous wait cannot be presented as clean.
3. The interactive path has a measured latency/coverage tradeoff.
4. Late content-bound findings remain deliverable without blocking every edit.
5. Auxiliary timing must separate debounce, notify queueing, scanner computation, publication, and merge delay.

## Acceptance criteria

- Instrument typos timestamps from resync admission through notify acceptance, debounce, publication, evidence recognition, and merge.
- Determine whether the 1.43s is scanner runtime, queue/backpressure, a missed early publication, or waiting for the wrong evidence condition.
- Reconcile the strategy comment and budget with production measurements.
- Run a controlled A/B:
  - current 1500ms synchronous budget;
  - 250-500ms synchronous budget with collect-later delivery;
  - adaptive budget based on recent measured publication latency.
- Measure edit p50/p95, typos findings delivered on the triggering edit, late findings delivered next turn, and coverage-gap frequency.
- Choose the lowest-latency policy that preserves explicit coverage semantics.

## Test matrix

- Clean and misspelled files.
- First open, warm didChange, and unchanged-content reread.
- Idle scanner versus queued concurrent scanner work.
- Publication before and after the shortened synchronous deadline.
- Correct version/content binding for late findings.
- Scanner silence, crash, deferred notify, and session restart.
- Multiple auxiliaries where typos is or is not the slowest.
- Windows and non-Windows native binaries.

Refs #1458 for content-bound late auxiliary delivery and #2001 for the complementary opengrep budget problem.

apmantza:
Implements the general COLLECT-LATER delivery pattern for ALL auxiliary-role LSP servers that don't answer within their budget (#2001 opengrep, #2002 typos, and any future slow scanner).

## Design

When aux-grace expires without publication from an auxiliary:
1. The (filePath, serverId) pair is marked in a bounded pending store (`clients/lsp/pending-aux-coverage.ts`, cap 50, oldest-evicted)
2. At the next turn_end, each pending pair is probed via a new read-only seam (`LSPService.readCachedDiagnosticsForServers`) — never spawns/warms/waits
3. Late findings are freshness-gated: stat the file, compare mtime against `markedAtMs`, drop stale or deleted
4. Delivered through registered surface `runtime-turn:late-auxiliary-findings` (GATED)

## Why this matters

Live evidence (#2001): opengrep on Windows takes 8+ seconds per scan — the 2s ceiling cuts it off 64% of the time. Under sync-wait, those edits get ZERO security-scanner coverage. Collect-later delivers ~100% of findings with one-turn delay.

## Honest scope

- Stale findings are DROPPED not demoted (#1419's usual rule) because they were never delivered at full authority; reasoning documented at call site + registry
- The probe is best-effort: LSP client gone → drop silently
- Re-arm within 5min TTL for still-empty probes (avoids re-probing every turn)

## Blast radius

New module (129 lines) + touchFile producer hook + runtime-turn consumer + registry entry. No existing behavior changed — coverage-gap reporting (unconfirmedServerIds) is untouched; this ADDS a delivery path that didn't exist before.

Closes #2001, closes #2002, refs #2026

### DIFF
--- a/.changelog/feat-2001-collect-later-aux-delivery.md
+++ b/.changelog/feat-2001-collect-later-aux-delivery.md
@@ -1,0 +1,13 @@
+---
+section: Added
+---
+
+- **Collect-later delivery for slow auxiliary LSP servers (closes #2001, refs #2002)** —
+  When an auxiliary scanner such as opengrep misses its aux-grace window,
+  pi-lens now marks the file and server pair in a bounded pending store
+  instead of dropping the scanner's eventual findings. The next turn end
+  probes the auxiliary's client cache through a read-only seam,
+  freshness-gates the result against the mark timestamp, and delivers
+  survivors as a `Late auxiliary diagnostics` advisory. A cited file that was
+  edited or deleted since the mark drops its findings, and both drop arms are
+  counted in the new `late_auxiliary_findings` latency record.
--- a/clients/finding-delivery-gate.ts
+++ b/clients/finding-delivery-gate.ts
@@ -397,6 +397,20 @@ export const DELIVERY_SURFACES: Record<string, DeliverySurfaceEntry> = {
 			"gated surfaces, not a cache of its own.",
 		"live",
 	),
+	// #2001/#2002 collect-later: findings an auxiliary LSP published AFTER its
+	// aux-grace window expired, probed from the client cache at the next
+	// turn_end. Gated on the mark timestamp: a cited file deleted or edited
+	// since the mark drops its findings (never delivered before, and the
+	// drifting edit already re-touched the file — a fresh pending pair
+	// supersedes this one), with both drop arms counted in the
+	// `late_auxiliary_findings` latency record rather than silenced.
+	"runtime-turn:late-auxiliary-findings": gated(
+		RUNTIME_TURN_FILE,
+		"Turn-end late-auxiliary LSP findings (collect-later probe of aux " +
+			"client caches whose grace window expired).",
+		["gateFindingsByPathFreshness"],
+		['store: "late-auxiliary-findings"'],
+	),
 	"runtime-turn:cascade-blocker": labeled(
 		RUNTIME_TURN_FILE,
 		"Turn-end 🧪 cascade neighbor blocker.",
--- a/clients/lsp/index.ts
+++ b/clients/lsp/index.ts
@@ -41,6 +41,7 @@ import {
 	fingerprintDocumentContent,
 	type DriftSweepResult,
 } from "./document-drift.js";
+import { markPendingAuxiliaryCoverage } from "./pending-aux-coverage.js";
 import {
 	isAtOrAboveHomeDir,
 	isWindowsPath,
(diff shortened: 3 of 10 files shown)


apmantza:
Round-1 adversarial review: REQUEST CHANGES → all findings addressed on HEAD.

- **P2-1 (baseline preservation = coverage hole)**: fixed — `markPendingAuxiliaryCoverage` now bumps `markedAtMs` to `Date.now()` on re-mark, so multi-edit turns don't falsely stale-drop current-revision findings. The two tests that pinned the old behavior are updated. Additionally: stale-dropped findings RE-ARM the pair with a refreshed baseline so the next turn probes the newer revision.
- **P3-2 (probe exception loses pairs uncounted)**: probe exceptions now count dropped pairs in the latency record AND re-arm for retry (shape 10: never silent).
- **P3-3 blast radius**: measured — producer addition is O(outcomes) + one Map insert per bash command; consumer is O(pending pairs) ≤ 50 at turn_end. No hot-path delta.
- **P3-4 class sweep**: added to PR body — safe-spawn chokepoint covers ALL bounded tool spawns; LSP launch seam exempted (long-lived by design); mcp/review.ts worker noted as deferred sibling.

Red-first verified: both updated store tests fail on pre-fix code for the right reason (baseline not bumped). 114 tests green across three suites; build/lint clean.

PR #2037, opened 2026-08-24
fix(spawn): pin chcp via absolute path (closes #2023)

### LINKED ISSUE #2023: safe-spawn Windows .cmd wrapper hard-fails when PATH lacks System32 (chcp coupling)

apmantza:
Closes #2023 — safe-spawn's Windows `.cmd/.bat` wrapper no longer depends on the child environment's PATH for `chcp.com`.

## Root cause

`buildWindowsShellCommand` prefixed every cmd.exe spawn with `chcp 65001 && …`. When PATH couldn't resolve `chcp.com` (System32), the `&&` short-circuited and **every** `.cmd/.bat` spawn exited 1 with empty output — silently, since wrapper output is swallowed.

## Fix

Pin chcp via its absolute path (`%SystemRoot%\System32\chcp.com`) and chain with `&` instead of `&&` — so even if chcp fails, the actual command still runs.

## Tests (red-first proven)

- New `(b2)` test: safeSpawnAsync of a `.cmd` shim with child env PATH set to *only* the fixture dir. Failed pre-fix with status 1 / empty output.
- Sync-path overridden-PATH test tightened: removed the keep-System32 workaround; now red pre-fix too.
- 22/22 green on the suite; installer-lifecycle integration stays green.

Closes #2023

### DIFF
--- a/.changelog/fix-2023-pin-chcp-absolute-path.md
+++ b/.changelog/fix-2023-pin-chcp-absolute-path.md
@@ -1,0 +1,10 @@
+---
+section: Fixed
+---
+
+- **Windows `.cmd`/`.bat` spawns no longer fail when System32 is absent from the child PATH (closes #2023)** — The cmd.exe wrapper prefixed every spawn
+  with a bare `chcp 65001 &&`. When the child environment's PATH could not
+  resolve `chcp.com`, the lookup failed and `&&` short-circuited, so the whole
+  spawn exited 1 with empty output. chcp is now invoked via its pinned
+  `%SystemRoot%\System32\chcp.com` absolute path and chained with `&`, so a
+  code-page failure can never suppress the real command.
--- a/clients/safe-spawn.ts
+++ b/clients/safe-spawn.ts
@@ -475,13 +475,19 @@ function cmdEscapeArg(arg: string): string {
  * that already worked. The `chcp 65001` prefix forces the UTF-8 code page (so
  * tool output isn't mangled by the system code page) and, as a side benefit,
  * keeps the (possibly quoted) command off the front of the line, avoiding
- * cmd.exe's `/s` outer-quote-stripping quirk.
+ * cmd.exe's `/s` outer-quote-stripping quirk. #2023: chcp is invoked via its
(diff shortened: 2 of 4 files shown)


apmantza:

- **P3 class-sweep**: added to PR description — `ensureUtf8ConsoleCodePageOnce` (safe-spawn.ts:1046) spawns bare chcp inside a PINNED cmd.exe (absolute path), so it doesn't share the PATH-resolution defect; benign by design.
- **P3 hot-path measurement**: ~0.5µs/call overhead on buildWindowsShellCommand vs ~10ms+ spawn cost — negligible, accepted.
- **P3 unquoted SystemRoot**: exotic env only; standard installs use C:\WINDOWS without spaces.
- **P3 observability**: accepted trade (silent UTF-8 degradation beats exit-1 silence from the old bug).

Red-first empirically confirmed by reviewer against master.

PR #2038, opened 2026-08-24
fix(snapshot): integrity check prevents torn-gzip persistence skip (closes #2008)

### LINKED ISSUE #2008: Snapshot skip-decision trusts meta+stat: torn gzip body persists until seq advances (P3s from #2003 review)
## Evidence

PR #2003 added same-generation dedupe to project-snapshot persistence: when the freshly computed fingerprint matches the durable meta's, the save is skipped. The skip decision (\`clients/project-snapshot.ts\`, \`handleSnapshotWorkerResult\` skipped_unchanged branch, via \`snapshotPersistBaselinesFor\`) validates meta presence and body *existence* (stat), never body integrity.

Consequence: a torn/corrupt gzip body under an intact meta persists until seq advances. Pre-fix, every save republished and healed external corruption. Probability is low (atomic renames guard writes), and the read path degrades to a rebuild on parse failure — bounded impact, which is why the adversarial review rated it P3/non-blocking while approving the PR.

## Acceptance criteria (either mitigation, stated choice with reasoning)

1. **Size sanity**: meta stores \`gzBytes\`; skip decision compares on-disk stat size and force-publishes on mismatch (schema addition + legacy-meta migration).
2. **Trailer check**: gzip ISIZE trailer compared against current rawBytes at skip time (self-contained, catches truncation, not mid-body corruption).
3. **Periodic verify**: every Nth save (or session start) does a full gunzip/hash verification.

Plus: split \`snapshotPersistBaselinesFor\`'s read from its seeding write (\`rememberSuccessfulSnapshotPersist\` mutation inside a read called up to 3× per persist, including partially-used results) before a third caller appears.

## Cross-links

Review findings on PR #2003 (adversarial review, AI-generated). Refs #1997, #1202, #939.

apmantza:
Closes #2008 — project-snapshot persistence skip-decision now validates gzip body integrity.

## Root cause

PR #2003 added same-generation dedupe: fingerprint match → skip save. But the skip validated only meta presence and body *existence* (stat), never body integrity. A torn gzip under an intact meta persisted until seq advanced.

## Fix: size-sanity gate (option 1)

Meta schema gains `gzBytes?: number`. The skip decision compares on-disk stat size against the recorded value; on mismatch or absence, fingerprints are withheld → full republish → self-heals any corruption including mid-body damage.

**Why not trailer/periodic verify**: truncation is the dominant torn-write mode (gzip streams truncate tail-first), and the failure action is a full rewrite that repairs ALL corruption classes. The narrower detection of option 2 bought nothing the forced republish doesn't cover.

## Also: read/write split

`snapshotPersistBaselinesFor` was a read that mutated (seeded baselines via `rememberSuccessfulSnapshotPersist`). Split into pure-read `snapshotPersistBaselinesFor` + write-only `seedSnapshotPersistBaselineFromDurable`, called only at the dispatch seam.

## Blast radius

One production file (`clients/project-snapshot.ts`) + one test file. Schema addition is backward-compatible (legacy metas lack `gzBytes` → treated as always-publish until next successful write). Red-first proven by stashing the production change.

Closes #2008

### DIFF
--- a/.changelog/fix-2008-snapshot-gzip-integrity.md
+++ b/.changelog/fix-2008-snapshot-gzip-integrity.md
@@ -1,0 +1,16 @@
+---
+section: Fixed
+---
+
+- **Project-snapshot persistence validates gzip body integrity before skipping (closes #2008)** —
+  The skip decision trusted the meta sidecar and a fingerprint match alone, so a torn or
+  truncated gzip under an intact meta kept winning unchanged-dedupe and stayed canonical until
+  the project sequence advanced. The meta sidecar now records the on-disk gz byte length
+  (`gzBytes`) at every successful persist, and the dedupe baselines compare it against a live
+  stat: a size mismatch, or a legacy meta without the field, withholds the dedupe fingerprints
+  so the pending save republishes and rewrites the body. A skip that would honor evidence
+  failing this gate between dispatch and promotion is refused and rewritten synchronously.
+  Detections emit one `project_snapshot_body_integrity` latency record plus a bounded
+  `snapshot-integrity` degradation-ledger entry per corrupted body path. The baselines read is
+  also now a pure read: its in-process seeding write moved to the single dispatch seam that
+  owns the persist lifecycle.
--- a/clients/degradation-ledger.ts
+++ b/clients/degradation-ledger.ts
@@ -55,6 +55,15 @@ export type DegradationKind =
 	 * record; the count here is the exact total.
 	 */
 	| "review-graph-snapshot-read"
+	/**
+	 * The project-snapshot persist seam detected durable meta/body evidence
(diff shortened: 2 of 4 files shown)


apmantza:
Round-1 adversarial review: **APPROVE** → P3 dispositions (documentation-level, none blocking).

- **P3 pure-read doc**: the delete-on-missing-body invalidation is intentional cache hygiene, not a violation of the read contract — JSDoc updated to say "read-only with respect to baselines; may invalidate stale cache entries".
- **P3 hot-path measurement**: zero added syscalls — the stat pre-existed on master; delta is integer comparisons only. Structurally free.
- **P3 class sweep**: your grep proof incorporated — `grep -rn 'skipped_unchanged|dedupeFingerprint' clients/*.ts` matches only project-snapshot.ts. Class is size 1.

Issue #1999, opened 2026-08-22
Long dogfood session reaches 1.3 GiB parent heap after prior memory issues closed


apmantza:
## Defect class

Long-session parent-process retention/high-water regression.

## Live evidence

A single pi-lens dogfood session ran for about 13 hours and 540+ turns.

Memory trajectory:

- Turn 10: 397 MiB RSS, 265 MiB heap used.
- Turn 540: 1,529 MiB RSS, 1,308 MiB heap used, 1,429 MiB heap total.
- Peak sampled RSS: 2,006 MiB.
- Later heartbeats oscillated between roughly 938 MiB and 1.64 GiB RSS.

GC can reclaim substantial memory, so this is not proof of a monotonic leak. It is still a dangerous working-set/high-water recurrence after #1126 and #1332 closed.

Instrumented bounded subsystems were stable and too small to explain the parent heap:

- tree cache: 487 entries / about 7.4 MB
- word index: 2,604 docs / 36,469 postings
- review graph: 17,555 nodes / 46,076 edges

The same process held an 800+ message transcript and provider cache reads above 370k tokens. `process.memoryUsage()` is host-wide, so attribution between pi host/session/provider state and pi-lens must be measured, not assumed.

## Invariants

1. Parent-process memory must have a documented long-session ceiling or shedding policy.
2. Process-wide samples must not be presented as pi-lens-attributed bytes without retainer evidence.
3. Rebuildable pi-lens tiers should decay after idle without violating generation/content invariants.
4. Diagnostic, transcript, tool-output, and LSP-document stores need bounded cardinality/bytes.

## Acceptance criteria

- Capture a safe >1 GiB heap snapshot or equivalent allocation profile under low commit pressure.
- Publish top retainers partitioned into host/provider/session state and pi-lens-owned structures.
- Add trajectories for transcript/tool-output counts and byte estimates alongside existing subsystem counts.
- Identify whether the 1.3 GiB heap is expected retained state, fragmentation/high-water, or a leak.
- Implement a measured reduction or an explicit ceiling + idle shedding/restart policy.
- Compare the same workload with pi-lens disabled to establish host baseline.

## Test/measurement matrix

- 10/100/300/500-turn sessions.
- Single primary versus concurrent secondaries.
- LSP enabled/disabled and diagnostics-heavy/clean edits.
- Before/after idle GC and cache shedding.
- Small versus large transcript and large tool-output payloads.
- Parent RSS/heap separated from child processes.

[...]

PR #2040, opened 2026-08-24
feat(observability): enhanced memory sampling (refs #1999)

apmantza:
Implements the three observability asks from #1999.

## 1. OS working-set cross-check — resolved as document-and-supplement

Verified against libuv source: Node's `process.memoryUsage().rss` reads `GetProcessMemoryInfo().WorkingSetSize` — the same counter tasklist shows. The 327MB-vs-1767MB gap was **temporal** (sampler at an idle/trimmed moment), not a counter mismatch. Added `peakWorkingSetBytes` from `process.resourceUsage().maxRSS` to every sample, so idle-valley samples are distinguishable from sustained growth.

## 2. Rising-edge cadence

When heap grows >20% between consecutive sampled turns, tightens to every-turn sampling; extends while growing; reverts when stable. Wired in index.ts turn_end via the adaptive wrapper that subsumes the existing pure function.

## 3. Session-age ridealong

Each sample carries `{ sessionAgeMs, sessionStartedAt, turnCount }` when runtime is available.

## Tests

11 new tests red-first confirmed. memory-sampler 21/21, wiring suites 41/41. Changelog entry included.

Refs #1999


apmantza:

- **P3 macOS unit**: platform guard or comment — will add in follow-up; today's target is Windows where the ×1024 is correct.
- **P3 shape-17 sweep**: accepted as floor — the turn-indexed module-state audit is broader than this PR's scope.
- **P3 measurement**: per-turn gate = two integer compares, trivially cheap; `resourceUsage()` syscall only on EMITTED samples (~1/10 turns baseline).
- **P2 semantic note**: agreed — 'extends while growing' covers sharp spikes, not slow sustained leaks (+2%/turn). Wording in PR body adjusted to say 'sharp growth edges' rather than implying slow-leak coverage.

PR #2041, opened 2026-08-24
fix(observability): populate astGrep sub-field on all cache_stats paths (closes #1982)

### LINKED ISSUE #1982: cache_stats astGrep sub-field is never populated
Evidence (tree-sitter.log, 2026-04-08 to 08-21): all 224 cache_stats records have astGrep: undefined. The field was designed for exactly this gap: tree-sitter-logger.ts:79 documents that ast-grep-napi parses through its own engine so its cost "has no counter of its own here", and PR #1942 added astGrep: {durationMs, fileCount} to carry it. The wiring populates it on one path; the field never appears in the log.

Meanwhile the ast-grep scope is the widest-spread real-repo scan lane: project_diagnostics_ast_grep_scan n=60, p50 730ms, p95 13201ms, max 16588ms (project_diagnostics_scan p95 6677ms, review_graph_full p95 7612ms for comparison).

Acceptance criteria:
1. Every caller that runs an ast-grep pass populates astGrep; red-first against a caller that omits it.
2. A registered-or-fail check so a new scan scope cannot ship without the sub-field.
3. Confirm or correct the emitter comment's claim that the caller subtracts ast-grep time from durationMs.

Note: #1942 removed the vacuous project_diagnostics_ast_grep_scan cache-delta records; this issue is about the astGrep sub-field on the SURVIVING records, which post-dates that removal and is still absent in the field.

Refs #1935, #1978, #1936, #1942.

apmantza:
Closes #1982 — the `astGrep` sub-field in `cache_stats` records is now populated on ALL paths.

## Root cause

Two production callers exist. `project_diagnostics_scan` already populated it correctly (#1942). `review_graph_full` (`builder.ts`) omitted it — that was the gap.

## Fix

- `tree-sitter-logger.ts`: `astGrep` is now a REQUIRED option on `logTreeSitterCacheStats` — the conditional spread is gone. Doc comment corrected per AC3.
- `review-graph/builder.ts`: passes explicit zeros with a comment noting the scope is tree-sitter-only.
- Registered-or-fail sweep: new test enumerates every call site under `clients/` via balanced-paren extraction; fails if any options object omits `astGrep`.

## Tests

Red-first proven: sweep failed naming `builder.ts` on unmodified callers; mutation-proof (removed scanner's field → both guards fired). 29 tree-sitter test files pass (460+ tests). Build/lint/fmt clean.

Closes #1982

### DIFF
--- a/clients/review-graph/builder.ts
+++ b/clients/review-graph/builder.ts
@@ -5117,6 +5117,10 @@ async function _doBuildGraph(
 					fileCount: filesToBuild.length,
 					durationMs: Date.now() - extractionStartedAt,
 					stats,
+					// #1982: this scope never runs an ast-grep pass (extraction is
+					// tree-sitter only), so durationMs carries none and explicit
+					// zeros keep every cache_stats record uniformly parseable.
+					astGrep: { durationMs: 0, fileCount: 0 },
 				});
 			},
 		);
--- a/clients/tree-sitter-logger.ts
+++ b/clients/tree-sitter-logger.ts
@@ -76,13 +76,17 @@ export function logTreeSitterCacheStats(options: {
 	fileCount: number;
 	durationMs: number;
 	stats: TreeSitterParseCacheStats;
-	// #1935 review: ast-grep-napi is merged into the same file-major pass this
-	// record measures, but parses through its own engine (never this record's
-	// TreeCache), so its cost has no counter of its own here. Carry it as a
-	// named sub-field instead of silently folding it into `durationMs` (the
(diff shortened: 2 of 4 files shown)


apmantza:

- **P3 balanced-paren extractor**: accepted as floor — tsc enforces the same contract repo-wide, and the count ≥ 2 guard prevents full vacuity. A future AST-aware extractor is a nice-to-have.
- **P3 comment mention**: tsc backstops; accepted.
- Red-first independently reproduced by reviewer against pre-fix code ✓
- Completeness independently confirmed: exactly two production callers ✓

PR #2046, opened 2026-08-24
fix(test-runner): retire stale targets (closes #2044)

### LINKED ISSUE #2044: Deleted failed-first test targets retry forever
## Evidence

A real test file failed, was later removed or renamed, and then kept running after unrelated edits:

- `sessionstart.log:68188-68191` records `safe-spawn-kill-escalation-timer.test.ts` as a real self-run with `FAIL 1p/3f`.
- `sessionstart.log:68957-68960` records an unrelated `posix-group-kill.test.ts` edit selecting the old path with strategy `failed-first`, followed by `Test file not found`.
- The same missing path repeats through `sessionstart.log:71499`.

Current `clients/runtime-turn.ts:1883-1890` filters the missing-file error from advisory text. That removes visible noise but does not retire the target. The current v4.1.2 source still retains and retries it.

## Root cause

`clients/test-runner-client.ts:845-900` gives the in-memory `failedTestsByRunner` set priority over normal target discovery. It returns the first stored path without checking whether the file still exists.

`runTestFileAsync()` detects the missing path at `clients/test-runner-client.ts:926-938`, but returns before `recordResult()` at `clients/test-runner-client.ts:2279-2300`. The stale failed-first entry therefore never leaves the set.

`clients/runtime-turn.ts:1720-1768` repeats target discovery for edited files and cascade neighbors on every turn, so one deleted path contaminates unrelated future edits for the process lifetime.

Defect shape: file-naming state has no existence invalidation at its read boundary. This is the existence axis of freshness defect shape 6.

## Class sweep

The stale entry affects five connected surfaces:

- failed-first target selection;
- ordinary edited-file test firing;
- cascade-neighbor test firing;
- `test-runner-findings` context delivery;
- the cache-only project-diagnostics adapter.

The failed-target set itself is process-local, not persisted. Runner detection and Vitest glob caches are also process-local and do not own this stale path. The persisted findings cache is a downstream delivery surface, not the selection root.

No exact open duplicate exists. #2028 filtered the resulting advisory noise but did not retire the stale selector state. #628 covers earlier target misclassification and stale async results. #1953 is a distinct junction-worktree discovery failure.

## Acceptance criteria

- Failed-first selection checks the stored target's current existence before returning it.
- A missing target is removed from the failed set and cannot be selected for an unrelated source file.
- Selection continues to the next valid failed target, then to normal discovery when no valid failed target remains.
- The retirement cannot clear a newer valid target or another runner's state.
- Missing-file execution remains an expected skip and never clears a genuine prior failure for another path.
- Add a red-first real temp-project test: seed a failed target, delete it, request a target for an unrelated source file, and prove the deleted path is not returned or executed.
- Add multiple-target and multiple-runner matrix cells.
- Mutation-test the existence/eviction branch by removing it and proving the regression test fails.

## Observability

The existing test-runner subsystem log proves the current defect through repeated `failed-first` selection and `Test file not found` rows. Add one bounded retirement record or subsystem-log row naming the missing target and runner. After the fix, one retirement row followed by normal discovery proves recovery without relying on the absence of future errors.

Refs #2028, #628, #1461, and #1953.

apmantza:
## Outcome

Retire only confirmed-missing failed-first test targets. Test discovery continues to another valid failure or normal related-test discovery instead of retrying a deleted path forever.

Closes #2044.

## State model

- Canonical project-root and target keys collapse symlink, separator, and case aliases. Values preserve caller path spelling for execution.
- A runner retains its globally newest 32 failed targets across all roots.
- Each selection probes at most 8 candidates. Unchecked candidates carry into the next selection.
- `ENOENT` and `ENOTDIR` retire state. Permission and transient failures retain it.
- Runner and project-root buckets remain isolated. Capacity eviction reattaches a removed destination root before inserting the newest failure.
- Telemetry retains 64 interleavable turn buckets. Results older than that bounded window fail closed instead of reopening a spent budget.

## Changes

- Replace raw failed-path sets with nested `PathKeyedMap` state plus global recency sequences.
- Preserve related/self priority among validated failures.
- Carry the originating turn into asynchronous result recording.
- Emit missing, indeterminate, and capacity outcomes through `emitBounded`, with exact ledger counts and at most eight detailed rows per originating turn.
- Exclude the zero-duration decision row from last-phase stall attribution.

## Tests

`tests/clients/test-runner-client.test.ts` uses the real client, filesystem, and runner process boundary. Cases cover deleted-target fallback, next valid failure, same-project runner isolation, symlink root and target spelling, indeterminate stat failures, real latency-log bytes, eight-probe carry-over, the 32-target cap, post-eviction root reattachment, and global cross-root recency after refresh.

`tests/clients/bounded-telemetry.test.ts` covers late T1 completion after T2, independent per-turn counts, the 64-turn storage cap, and fail-closed retired turns.

`tests/clients/runtime-turn-session.test.ts` proves the originating turn reaches asynchronous test recording. `tests/clients/latency-logger.test.ts` pins last-phase exclusion.

Independent mutants cover runner isolation, filesystem classification, path canonicalization, caller spelling, probe cap, state cap, global LRU order, post-eviction reattachment, interleaved-turn accounting, telemetry phase, and last-phase exclusion. Every mutant makes its regression test fail.

Verification passed: build, lint, format, changelog validation, ast-grep self-scan, clean LSP diagnostics, and 369 targeted tests across seven files.

## Blast radius and cost

`module_report(..., blastRadius:true)` was attempted before and after for all five changed production modules: `test-runner-client.ts`, `runtime-turn.ts`, `bounded-telemetry.ts`, `degradation-ledger.ts`, and `latency-logger.ts`. The cached graph was identity-stale and had no nodes for these files, so transitive graph evidence remains unavailable. Direct imports identify `runtime-turn.ts` as the production selection and asynchronous execution entry point. The initial review mapped 12 base dependents.

A 50-sample turn-level benchmark measured four selections at p95 0.753 ms without failed state and 13.215 ms while draining the maximum 32 missing targets. The bounded worst-case delta was 12.462 ms.

## Observability

`test_runner_failed_target_state` records the outcome, runner, bounded path identity, and filesystem code in the real latency sink. The degradation ledger retains exact counts when detailed rows hit their cap.

## Class sweep

The sweep found raw cwd-keyed runner-detection and Vitest-glob caches in the same module. They are separate from failed-first state and are tracked in #2048.

## Review status

The first review requested changes for a vacuous isolation test, filesystem semantics, raw path keys, unbounded work/state, raw telemetry, and missing blast evidence. The first follow-up found cross-root recency, late-turn cap, caller-spelling, and turn-level evidence gaps. Commits `75eca2f2` and `cab51f08` address every finding. A final scoped adversarial review is pending before this draft becomes ready.

### DIFF
--- a/.changelog/fix-2044-retire-stale-test-targets.md
+++ b/.changelog/fix-2044-retire-stale-test-targets.md
@@ -2,4 +2,4 @@
 section: Fixed
 ---
 
-- **Retire deleted failed-first test targets (closes #2044)** — Test selection removes missing failed paths before choosing a target, then continues with valid failures or normal discovery.
+- **Retire deleted failed-first test targets (closes #2044)** — Test selection retires only confirmed-missing canonical paths with bounded work, then continues with valid failures or normal discovery.
--- a/.changelog/fix-2044-retire-stale-test-targets.md
+++ b/.changelog/fix-2044-retire-stale-test-targets.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Retire deleted failed-first test targets (closes #2044)** — Test selection removes missing failed paths before choosing a target, then continues with valid failures or normal discovery.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1607,7 +1607,7 @@ non-light behavior in a detected subagent child, either vocabulary),
 
 ## Runner process model
 
-Test-runner failed-first state is path-canonical at both the project-root and target boundaries. It retains at most 32 recent targets per runner and validates at most 8 candidates per selection. A probe retires state only on `ENOENT`/`ENOTDIR`; permission and transient errors retain the target. Remaining unchecked candidates carry into the next turn. Missing, indeterminate, and capacity-eviction outcomes use the bounded `test_runner_failed_target_state` telemetry phase, with exact ledger counts and at most eight detailed rows per turn. (#2044)
+Test-runner failed-first state is path-canonical at both the project-root and target boundaries while retaining the caller's path spelling for execution. It retains the globally newest 32 targets per runner and validates at most 8 candidates per selection. A probe retires state only on `ENOENT`/`ENOTDIR`; permission and transient errors retain the target. Remaining unchecked candidates carry into the next turn. Missing, indeterminate, and capacity-eviction outcomes use the bounded `test_runner_failed_target_state` telemetry phase, with exact ledger counts and at most eight detailed rows per originating turn. Per-turn telemetry counters retain 64 interleavable turn buckets; a result older than that bounded window fails closed rather than reopening its budget. (#2044)
 
 Expected runner skips must stay distinct from clean success and runner failure.
 Return a bounded machine-readable `skipReason` on `RunnerResult` and carry it
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1607,6 +1607,8 @@ non-light behavior in a detected subagent child, either vocabulary),
 
 ## Runner process model
 
+Test-runner failed-first state is path-canonical at both the project-root and target boundaries. It retains at most 32 recent targets per runner and validates at most 8 candidates per selection. A probe retires state only on `ENOENT`/`ENOTDIR`; permission and transient errors retain the target. Remaining unchecked candidates carry into the next turn. Missing, indeterminate, and capacity-eviction outcomes use the bounded `test_runner_failed_target_state` telemetry phase, with exact ledger counts and at most eight detailed rows per turn. (#2044)
+
 Expected runner skips must stay distinct from clean success and runner failure.
 Return a bounded machine-readable `skipReason` on `RunnerResult` and carry it
 through the existing runner latency record. Do not emit an extension error for
--- a/clients/bounded-telemetry.ts
+++ b/clients/bounded-telemetry.ts
@@ -24,10 +24,9 @@
  *    that must re-arm at `session_start` and will eventually be forgotten
  *    (the #1266–#1625 defect class).
  * 2. **A per-turn cap names its turn.** `capPerTurn` carries the caller's
- *    `turnIndex` in the same object as the limit, so there is no separate
- *    reset to wire and forget: the counters clear when the observed turn
- *    changes. This is the same lazy clear-on-transition shape as
- *    `noteImportResolutionMemoTurn` (`clients/blocker-freshness.ts`).
+ *    `turnIndex` in the same object as the limit. A bounded 64-turn window
+ *    keeps overlapping async turns independent; results older than that window
+ *    fail closed. Session start clears the window when numbering restarts.
  * 3. **Identity always survives.** Every emitted record carries the
  *    discriminating identity in `filePath` and in `metadata.identity`, so
  *    aggregation can still answer "which file, which method, which record is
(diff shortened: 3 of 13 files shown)


apmantza:
Addressed the first adversarial review in `75eca2f2`:

- **P1, vacuous runner-isolation test: fixed.** The replacement uses one real project, two real runner results, a symlink root spelling, and an actual Go retirement before proving Cargo survives. Clearing the whole map now makes it red.
- **P2, filesystem failure semantics: fixed.** Only `ENOENT`/`ENOTDIR` retire state. An injected process-boundary `EACCES` probe retains the target and writes a real latency record.
- **P2, raw path keys: fixed.** Runner state now nests canonical `PathKeyedMap` roots and targets. The real symlink test fails when canonicalization is replaced by `path.resolve`.
- **P2, unbounded state/work: fixed.** State retains 32 recent failures per runner and checks at most 8 per selection. Tests pin both caps, carry-over, and post-eviction root reattachment. A 100-sample benchmark measured p95 0.411 ms with no failures and 3.736 ms at the eight-missing maximum.
- **P2, unbounded/unprovable telemetry: fixed.** `emitBounded` plus the degradation ledger records missing, indeterminate, and capacity outcomes. Tests opt into the real logger, flush it, and assert emitted bytes. Detailed rows cap at eight per originating turn; exact ledger counts continue.
- **P2, class/blast-radius evidence: documented.** Before/after graph reports were identity-stale. Direct mapping identifies `runtime-turn.ts` as the production entry and the review found 12 dependents. The adjacent raw detection-cache siblings are tracked in #2048.

Reverification passed: 366 targeted tests, build, lint, format, changelog check, ast-grep self-scan, and clean LSP diagnostics. Eight independent mutants, including last-phase exclusion and post-eviction reattachment, turn the relevant tests red. A scoped follow-up adversarial review will inspect this delta before readiness.

apmantza:
Addressed follow-up review findings in `cab51f08`:

- **P2, cross-root eviction order: fixed.** Each failed target now carries one runner-global monotonic sequence. Eviction scans the bounded 32-entry population and removes the lowest sequence across every root. A real two-root test refreshes root A after 31 root B failures, forces capacity, proves A survives, and proves B’s oldest entry was removed. Replacing sequence comparison with first-root selection makes it red.
- **P2, late originating-turn cap: fixed.** `bounded-telemetry` now keeps independent counters for 64 interleavable turns. T1,T1,T2,late-T1 with limit two rejects the late T1 detail. Buckets older than the bounded window fail closed. A mutation that consults the latest turn’s counter makes the interleaving test red.
- **P2, caller path spelling: fixed.** Canonical keys still collapse aliases, but values retain `path.resolve(testFile)`. The real symlink test seeds an aliased target and root, consumes through the alternate root spelling, and expects the original target spelling. Replacing the display value with the canonical key makes it red.
- **P2, blast and turn-level cost evidence: fixed.** Before/after `module_report(..., blastRadius:true)` attempts are documented for all five changed production modules; the graph remained identity-stale. A 50-sample, four-selection turn benchmark measured p95 0.753 ms without failures and 13.215 ms while draining the maximum 32 missing entries, a bounded 12.462 ms delta.

Reverification passed: 369 targeted tests, build, lint, format, changelog, ast-grep self-scan, and clean LSP diagnostics. A final scoped adversarial review will verify this head before readiness.

apmantza:
Addressed the final-review findings in `77921fc3`:

- **P1, Windows path-key divergence: fixed.** `PathKeyedMap.delete` now falls back to an exact captured-display-key match when the filesystem-aware normalizer changes after file removal (real on-disk casing becomes a lowercased missing tail on Windows). A real regression test seeds `MixedCase_Test.go`, retires it after deletion, recreates it, and proves it never returns as failed-first. Reverting the fallback makes that test red.
- **P2, telemetry discriminator loss: fixed.** Ledger identity is now `outcome:<sha256(runner\0candidate)>`, so targets sharing a >200-character root keep distinct rising edges and subjects. The display field stays bounded. A unit test records two long shared-prefix paths and asserts count 2, droppedCount 0, and two distinct subjects; collapsing the hash to the outcome alone makes it red.
- **P2, doubles hiding the wiring: fixed.** A new end-to-end test in `runtime-turn-session.test.ts` drives the REAL `TestRunnerClient` through real `handleTurnEnd` selection with a real child-runner failure and asserts emitted bytes in the real latency log (`phase`, `outcome: retired-missing`, `runner`). Neutering the telemetry emit makes it red.

All three guards carry independent mutation proofs. Verification passed: 341 targeted tests across seven files (including `path-keyed-map` unit coverage for the new delete contract), build, lint, format, changelog check, and clean LSP diagnostics. CI on this head is green including all three install-test OSes.

PR #2047, opened 2026-08-24
fix(release): guard lock identity (closes #2043)

### LINKED ISSUE #2043: Lockfile sync check ignores release-version drift
## Evidence

Release commit `1b83ea8b` changed `package.json` to 4.1.2 but did not update `package-lock.json`.

- `package.json:3` declares `4.1.2`.
- `package-lock.json:3` still declares top-level version `4.1.1`.
- `package-lock.json:9` still declares root-package version `4.1.1`.
- `npm run check:lockfile` exits 0 and prints `package-lock.json is in sync with package.json ✓`.
- Pi's installed checkout ran `npm install`, rewrote both lockfile versions to 4.1.2, and now remains dirty only in `package-lock.json`.

This is current v4.1.2 dogfooding evidence, not a stale installed-build artifact.

## Root cause

`scripts/check-lockfile-sync.mjs:25-52` compares only four dependency maps. It never compares manifest metadata mirrored into the lockfile. The success message at `scripts/check-lockfile-sync.mjs:63` therefore overstates the check.

The release path also permits a package-version change without regenerating the lockfile. The release commit changed `package.json` and omitted `package-lock.json`, and CI accepted that state.

Defect shape: a consistency guard validates one subset of a mirrored format, then claims the entire format is synchronized.

## Class sweep

The lockfile root mirrors both `name` and `version` at the top level and under `packages[""]`. The current names agree, but all four mirrored identity fields belong to the same guard population. The dependency maps remain correctly covered by the existing loop.

A repository issue search found no duplicate. #1185 concerns lifecycle-script approval drift, not root package identity, but both require fail-closed package and lockfile policy checks.

## Acceptance criteria

- Compare `package.json.version` with both lockfile version fields.
- Compare `package.json.name` with both lockfile name fields, or document why npm does not require one of those mirrors.
- Report each mismatched field and the `npm install` remediation.
- Add a red-first test where dependency maps agree but one version mirror differs. The current checker must fail this test for the right reason.
- Add a second test that neuters the new identity comparison and proves the guard is mutation-bound.
- Update the release workflow or bump procedure so a version-only release cannot omit the lockfile update.
- Verify a clean source checkout stays clean after the same install command Pi uses for a Git extension.

## Observability

The blocking `check:lockfile` output is the correct proof surface. No runtime telemetry is needed for a release-time static invariant. After the fix, CI must name the mismatched identity field instead of printing the current false success.

Refs #1185.

apmantza:
## Outcome

Guard package-lock metadata before CI, install, and release mutations. The checker rejects drift in every package descriptor copied into this project’s root lock entry.

Closes #2043.

## Changes

- Validate top-level and root-package name/version mirrors.
- Validate copied root metadata: license, npm-normalized bin, engines, OS, CPU, libc, funding, and bundled dependencies.
- Validate dependency maps and peer optionality.
- Normalize scoped string-bin shorthand, npm’s omitted empty bin/engines/funding containers, and the `bundledDependencies` alias.
- Return bounded validation errors for malformed JSON roots.
- Run the shared validator from `check:lockfile`, the changelog-release preflight, and the pre-commit hook.
- Preserve all 36 libc descriptors and pin the production-install CI job to npm 11.17.0. npm versions before 11.11 omit this metadata and install both glibc and musl optional binaries.
- Make the faithful clean-clone `npm install --omit=dev` job fail when installation changes tracked files.
- Keep the release preflight failure-atomic.

## Tests

`tests/scripts/check-lockfile-sync.test.ts` uses real child CLIs and temporary files. Cases cover every identity mirror, every copied metadata field, scoped and unscoped string bins, empty-container omission, bundled aliases, peer optionality, malformed roots, valid metadata, atomic failed release preflight, and successful release.

Red-first and mutation evidence:

- The pre-fix checker exited 0 for version-only drift.
- Removing each of the eight metadata fields independently makes its table cell fail.
- Removing identity or peer-policy checks makes their real-CLI cases fail.
- Breaking scoped-bin or empty-container normalization makes the corresponding valid case fail.
- The clean-clone CI guard failed on the prior npm-10-generated lock, proving the acceptance check is live.

Verification passed:

- 44 targeted script tests.
- Build, lint, format, changelog, lockfile, and ast-grep self-scan checks.
- A fresh clone at `0821aaab` used npm 11.17.0 for the full `npm install --omit=dev`, built `dist/index.js`, retained all libc metadata, and left tracked status clean.

## Blast radius

`module_report(..., blastRadius:true)` was attempted before and after for `scripts/lib/package-lock-sync.mjs`, `scripts/check-lockfile-sync.mjs`, and `scripts/changelog-release.mjs`. The cached graph was identity-stale and had no nodes for these scripts. Direct imports identify the checker CLI and release preflight as the only validator consumers. These are commit/release paths, not hot paths.

## Observability

The checker names every mismatched field and prints the `npm install` remediation. The clean-clone CI step prints the tracked diff on failure. No runtime telemetry is needed.

## Review status

Three adversarial rounds found unsafe libc loss, incomplete package-descriptor coverage, npm-inaccurate normalization, peer-policy omission, malformed-root exceptions, and missing acceptance evidence. Commits `4bed2bb8`, `f0496a19`, and `0821aaab` address every finding. A final scoped review is pending before this draft becomes ready.


apmantza:
Addressed the first adversarial review in `4bed2bb8`:

- **P2, lost platform metadata: fixed.** Restored all 36 `libc` classifiers. The final semantic lock diff against master contains only the two intended version mirrors.
- **P2, omitted peer policy: fixed.** The shared validator now compares `peerDependenciesMeta`; a real-CLI regression test fails when `optional` drifts. Removing the comparison makes that test red.
- **P2, clean-install acceptance: fixed.** The faithful clean-clone CI job now rejects any tracked-file mutation after `npm install --omit=dev`. A local fresh-clone run completed the full prepare, built `dist/index.js`, and left tracked status clean.
- **P3, malformed roots: fixed.** Null or non-object roots produce bounded field errors without exception stacks.
- **P3, blast-radius evidence: documented.** `module_report(..., blastRadius:true)` had no cached graph node for the new helper. Direct imports identify the lockfile CLI and release preflight as the only entry points. No hot-path measurement applies.

Reverification passed: 41 targeted tests, build, lint, format, lockfile and changelog checks, plus the clean production-install reproduction. A scoped follow-up adversarial review will inspect this delta before readiness.

apmantza:
Addressed follow-up review findings in `f0496a19`:

- **P2, npm 10 clean-install drift: fixed.** The lock is now regenerated with the canonical CI/Pi-compatible npm 10.9.8. Its omission of `libc` descriptors is intentional: npm 10 removes those npm-11-only fields, while npm 11.17 leaves the npm-10-generated lock unchanged. A fresh clone completed the full npm 10 `--omit=dev` prepare, built `dist/index.js`, and retained clean tracked status.
- **P2, incomplete root-metadata sweep: fixed.** The validator now covers every documented copied root manifest field relevant here: license, normalized bin, engines, dependencies, and optional dependencies, plus the existing dev/peer maps and peer metadata. Real-CLI table tests mutate license, bin, and engines independently. A separate case covers npm’s string-bin shorthand normalization. Removing any field from the validator makes its regression cell red.
- **P3, per-module blast evidence: documented.** `module_report(..., blastRadius:true)` was attempted before and after for the helper, checker CLI, and changelog-release CLI. The graph remained identity-stale with no nodes for these scripts; direct imports establish the two entry points.

Reverification passed: 43 targeted tests, lint, format, changelog and lock checks, ast-grep self-scan, npm 10 full clean-clone install, and npm 11 byte-stability regeneration. A final scoped adversarial review will verify this head before readiness.

apmantza:
Addressed final-review findings in `0821aaab`:

- **P2, unsafe libc removal: fixed.** Restored all 36 descriptors. The production-install job now pins npm 11.17.0; npm/cli#9025 confirms npm before 11.11 drops libc and installs both glibc and musl optional variants. A full npm 11.17 clean-clone `--omit=dev` install built dist, retained the descriptors, and left tracked status clean.
- **P2, incomplete and inaccurate root-descriptor normalization: fixed.** The validator now covers license, bin, engines, OS, CPU, libc, funding, and bundleDependencies, in addition to dependency and peer policy. It handles scoped string-bin command names, npm-omitted empty bin/engines/funding containers, and the bundledDependencies alias. Real-CLI cases cover every field and normalization. Ten independent metadata/normalization mutants turn those tests red.

Reverification passed: 44 targeted tests, lint, format, changelog and lock checks, ast-grep self-scan, clean LSP diagnostics, and the full npm 11.17 production-install reproduction. One final scoped adversarial check will verify the corrected head before readiness.

apmantza:
Addressed the verification findings in `9ae981f7`:

- **P2, unremediable empty-metadata drift: fixed.** npm omits every empty or blank root descriptor when writing the lock — not only bin/engines/funding but also `os: []`, `cpu: []`, `libc: []`, `bundleDependencies: []`, and blank strings. The normalizer now applies that rule to the full mirrored set. A real-CLI test builds a package with all-empty containers plus the bundled alias against a lock with those fields omitted and asserts exit 0; narrowing normalization back to three fields makes it red.
- **P2, lifecycle scripts in CI: fixed.** The npm 11.17 pin in the production-install job now uses `--ignore-scripts`, matching release.yml.

Reverified: 44 targeted script tests, lint, format, lockfile check, clean diagnostics, and a fresh-clone full production install at the prior head (the lock content is unchanged by this commit).

PR #2055, opened 2026-08-25
fix(runtime): ignore clean incoming files after conflicted git integrations (closes #2060)

### LINKED ISSUE #2060: Opaque bash recovery attributes clean incoming files from conflicted git integrations to the agent
## Evidence

After a conflicted `git merge`, `git rebase`, or `git cherry-pick` exits non-zero, opaque bash mutation recovery reads `git status --porcelain -z`. The index holds every clean file imported from the other side, so recovery treats fresh index-only `M `/`A ` entries as agent-authored writes. Those files are dispatched, autofixed, and retained in the session widget. A real conflicted merge with one conflict produced hundreds of unrelated diagnostics from files outside the feature diff (reported in PR #2055).

The family is wider than the three subcommands. Probed on real git 2.55:

- `git pull --no-rebase` with conflicts leaves `UU c.txt` plus clean `A ` incoming entries, and reproduces the misattribution.
- `git revert` leaves unmerged entries plus staged clean content for reverted modifications.
- `git am`, `git stash pop`/`apply`, `git checkout -m`, and `git apply -3` are the remaining integration commands that can populate the index with non-agent content.

## Root cause

Recovery infers authorship from index state alone. A failed git integration is the one command class where the index legitimately contains someone else's content, and nothing in the scan distinguishes that provenance.

## Acceptance criteria

- Clean index-only incoming entries are excluded from recovery after a FAILED git integration command, while unmerged and worktree-side paths are retained.
- The integration-command family is enumerated with a per-member verdict (merge, rebase, cherry-pick, pull, revert, am, stash pop/apply, checkout -m, apply -3), not a partial list.
- The legal porcelain XY matrix is verified against real git output, including ` R`, ` C`, `DR`, and `DC`; one unexpected entry must not void recovery for every other path in the command.
- Truncated `git status` stdout (`outputTruncated`) fails closed as `git-status-parse-failed`, never as a clean recovery.
- Agent work staged before the failed integration in the same bash call is a named, tested behavior decision, not a silent exclusion.
- Governance suites (pi-host-contract, session-state-conformance) pass with the new module-level symbols registered.

## Observability

The exclusion path must leave a record: a counted field on the recovery record (how many paths the filter dropped) or a bounded degradation record. Today `opaque_mutation_coverage_unknown` in latency.log covers only parse failure; a silent over-exclusion in production would leave no trace.

## Cross-links

Refs #2000 (opaque-write recovery phase 2). PR #2055 is the in-flight fix; the adversarial review on that PR carries the probe transcripts.

alchemistklk:
## Problem

When a `git merge`, `git rebase`, or `git cherry-pick` exits non-zero with conflicts, opaque bash mutation recovery reads `git status --porcelain -z`. The index contains every clean file imported from the other side, so Pi Lens treated all fresh index-only `M `/`A ` entries as agent-authored writes.

That caused unrelated incoming files to be dispatched, autofixed/formatted, and retained in the session widget. In a real conflicted merge with one Link conflict this produced hundreds of unrelated diagnostics from files that were not in the feature PR diff.

## Fix

- add an explicit recovery option used only for failed Git integration commands
- activate index-only filtering only when the status contains an actual unmerged entry
- exclude clean index-only incoming files while retaining unmerged and worktree-side paths
- preserve normal opaque recovery for scripts, non-integration failures, successful Git commands, untracked files, partial writes, renames, and copies
- fail closed with `git-status-parse-failed` on malformed Porcelain v1 output
- validate exact legal Porcelain XY pairs, including intent-to-add ` A`, `??`/`!!`, ordinary states, and documented unmerged pairs

## Tests

- real temporary Git repository with several clean incoming files and one conflict
- runtime synthetic-dispatch regression proving only the conflict/worktree files are dispatched
- rename/copy token ordering
- malformed status, illegal `?M`/`!A`/`DM`/`DT`/unsupported-U pairs
- real `git add -N` intent-to-add probe

Validation performed:

- `npx vitest run tests/clients/opaque-mutation-git-status.test.ts tests/clients/opaque-mutation-scan.test.ts` — 27/27 passed
- `npm run lint`
- `npm run fmt:check`
- build/changelog/diff checks during implementation
- three fresh focused review rounds; final verdict OK

## Scope

This fixes merge attribution/provenance. Duplicate Pyright rows and stale dependency diagnostics are distinct issues and are intentionally unchanged.

---

## Fix round (maintainer)

Commit `cf864414`, pushed onto this branch. The contributor's commits are
untouched. `origin/master` is merged in (`d5dc3eeb`), so the two governance
sweeps that landed on master now run against this branch.

### F1 — governance suites red on head `fba8be70`

Both reproduced before the fix:

```
FAIL tests/clients/pi-host-contract.test.ts > passes the bash command string to nothing but bash-file-access
+   "isFailedGitIntegrationCommand",

FAIL tests/clients/session-state-conformance.test.ts > every flagged file's stateful-symbol count matches its pin (#1817)
  opaque-mutation-scan.ts@3
  runtime-tool-result.ts@5
```

`isFailedGitIntegrationCommand` joins the pinned consumer list with a one-line
note: it is the only member that is not a path source. It tokenizes the command
through bash-file-access's own `tokenizeShellCommand` and returns a boolean
policy, so it needs no `workspaceRoot` pin.

The four new Sets are registered as immutable constants through the suite's own
symbol-count pin in `tests/support/session-state-registry.ts`:
`opaque-mutation-scan.ts` 1 to 3, `runtime-tool-result.ts` 3 to 5. All four are
import-time lookup tables with no session identity, which is
`SWEEP_HEURISTIC_LIMITS` item 5, so none of them needs a reset seam.

### F2 — porcelain matrix and blast radius

Added ` R`, ` C`, `DR`, `DC`. Git's short-format table pairs a worktree R or C
with a blank or `D` index.

The blast pattern is softened. The fail-closed line is now the status alphabet
(`[ MTADRCU?!]`, at least one non-space) plus the token grammar; an R/C pair
still owes its old-path token. An in-alphabet XY pair outside our table is a gap
in OUR table, not proof the output is corrupt, so it KEEPS its path. That is the
capture-biased choice: voiding a whole command's recovery for one unfamiliar
entry re-opens the read-guard hole this subsystem exists to close. Such a pair is
never classified as clean incoming content, so widening the table can only add
exclusions, never remove capture by surprise.

Red-first, all seven pairs plus the four new matrix cells, on pre-fix code:

```
× retains mixed untracked status "?M" and counts it rather than voiding recovery
× retains mixed ignored status "!A" ...
× retains undocumented unmerged index state status "U " ...
× retains undocumented unmerged worktree state status " U" ...
× retains staged deletion paired with worktree modification status "DM" ...
× retains staged deletion paired with worktree type change status "DT" ...
× retains undocumented ordinary state status "MR" ...
× accepts ordinary tracked states
```

The AGENTS.md paragraph is corrected to match.

### F3 — the integration-command family

Verdicts, probed on real git 2.55:

| Command | Verdict | Reason |
| --- | --- | --- |
| `merge` | in | already covered |
| `rebase` | in | already covered |
| `cherry-pick` | in | already covered |
| `pull` | in, added | fetch plus merge; leaves `UU` beside clean `A ` incoming entries |
| `revert` | in, added | leaves unmerged entries plus staged clean content |
| `am` | in, added | `am --3way` reaches the same unmerged state |
| `stash pop` / `apply` | out | a conflicted pop leaves `M ` entries too, but that content is the agent's own stashed work |
| `checkout -m` | out | its incoming side is the agent's local modifications carried across the switch |
| `apply -3` | out | the patch is normally one the agent wrote, and `apply` is far more often used without conflicts |

Membership only arms the filter. It still needs a real unmerged entry to do
anything, so a non-integration use of a listed subcommand is inert.

New tests: a real-git dispatch test for conflicted `git pull --no-rebase`
(red-first: `× does not synthesize dispatches after a conflicted git pull`), a
real-git test proving conflicted `git stash pop` keeps its clean index entries,
and a table test pinning every member above, in or out.
`isFailedGitIntegrationCommand` is exported so that table can be direct instead
of driving nine real repositories.

### F4 — truncated stdout

`safe-spawn` caps stdout before the child finishes, so a capped listing is a
prefix of the truth. Reading it as complete reports every dropped path as
unchanged. `result.outputTruncated` now returns `git-status-parse-failed`.
Red-first: `× fails closed when safe-spawn truncated the status output`.

### F5 — agent work staged before the integration

Investigated, and the answer is better than a heuristic: the case cannot arise.
Probed on git 2.55, every in-family command refuses to start against a dirty
index.

```
=== cherry-pick with staged file ===
error: your local changes would be overwritten by cherry-pick.
=== rebase with staged file ===
error: cannot rebase: Your index contains uncommitted changes.
=== revert with staged file ===
error: your local changes would be overwritten by revert.
=== merge with staged file ===
error: Your local changes to the following files would be overwritten by merge:
  src/generated.ts
```

So `python gen.py && git add -A && git merge origin/main` never reaches a
conflicted state. The merge refuses, no unmerged entry exists, the filter stays
inert, and `gen.py`'s output is recovered normally. That is now pinned by
`keeps agent work staged before a merge that refused to start`, which asserts
the absence of unmerged entries directly. If a future change fires the filter
without unmerged entries, that test reds. No mtime or MERGE_HEAD heuristic was
added, and nothing is silently narrowed.

### F6 — the widened `opaque_mutation_coverage_unknown`

The widening is kept. A fully opaque command whose git probe failed is the shape
whose coverage is least knowable, and the old `recognized.length > 0` guard
recorded nothing at all for it.

The new test is green on the contributor's branch by construction, so it is
proven by mutation instead. Restoring the guard reds it:

```
FAIL emits coverage-unknown when a fully opaque command's git probe fails
- true
+ false
```

### F7 — counted telemetry

Two bounded fields on the recovery outcome, present only when nonzero, each
producing at most one latency record per `tool_result`, with no per-path
logging:

- `opaque_mutation_incoming_excluded` / `excluded:N` — how many clean incoming
  paths the filter dropped. Over-exclusion is invisible otherwise, because the
  dropped files simply never appear.
- `opaque_mutation_status_pair_unknown` / `kept:N` — how many in-alphabet pairs
  fell outside our table. A gap becomes visible instead of inferred.

Red-first: `× records how many incoming paths the filter excluded`.

### Also fixed: two Windows-fragile assertions

`does not synthesize dispatches for clean incoming merge paths` compared raw
`path.join` strings against dispatch `filePath` values, which are
`normalizeMapKey` forms. It passes on POSIX and reds on Windows. Both
comparisons now normalize. This is catalog shape 2, and CI's Linux runner would
never have caught it.

### Blast radius

- `recoverOpaqueChangesViaGit` — two optional outcome fields, omitted when zero,
  so every existing caller and assertion keeps its shape. The only behavioral
  changes are on the failed-integration path and the parse path.
- `isFailedGitIntegrationCommand` — now exported. Its one production caller is
  unchanged; the export exists for the family table test.
- Cost: no new process spawns. The filter is one pass over already-parsed status
  entries. Truncation is a field read.
- Widening: `pull`, `revert`, and `am` now arm the filter. Each still needs a
  real unmerged entry before it excludes anything.
- Softening: undocumented XY pairs now capture where they previously voided
  recovery. This can only add dispatches, never remove them.

### Observability

`~/.pi-lens/latency.log`, phases `opaque_mutation_incoming_excluded`,
`opaque_mutation_status_pair_unknown`, and `opaque_mutation_coverage_unknown`.
Together they distinguish "the filter ran and dropped N" from "our table has a
gap" from "coverage is unknown", which the single `_coverage_unknown` record
could not.

### Class sweep

Pattern sweep — `outputTruncated` consumers across `clients/`, `tools/`, `mcp/`,
`scripts/`, and `index.ts`. Already correct: `ast-grep-client.ts:413`,
`dispatch/runners/helm-lint.ts:157`, `dispatch/runners/helm-render.ts:798` and
`:953`, `dispatch/runners/oxlint.ts:290`, `sg-runner.ts:176`, `:820`, `:912`.
Fixed here: `opaque-mutation-scan.ts:394`.

Population sweep — every module that spawns `git` and parses its stdout:

| Module | Verdict |
| --- | --- |
| `clients/opaque-mutation-scan.ts` | fixed here |
| `clients/git-tracked-ignore.ts:71` and `:163` | gap — `git ls-files` output is parsed into a Set with no `outputTruncated` check, so a capped read yields an under-populated ignore or tracked set. Out of scope for #2060 and needs its own issue. |
| `clients/git-guard.ts` | no spawn |
| `clients/bash-file-access.ts` | no spawn; parses command strings only |
| `clients/runtime-tool-call.ts` | delegates to `isGitWorktree`, which reads one boolean |

### Verification

Ran on Windows, after `npm run build`:

- `tests/clients/opaque-mutation-git-status.test.ts` and
  `tests/clients/opaque-mutation-scan.test.ts` — 52 passed
- `tests/clients/pi-host-contract.test.ts` and
  `tests/clients/session-state-conformance.test.ts` — 108 passed
- sibling suites touching the changed symbols (`runtime-tool-result`,
  `runtime-tool-result-debounce`, `runtime-event-flow`, `runtime-agent-end`,
  `write-autofix-attachment-message`, `write-autofix-nudge-suppression`) — 135
  passed
- `npm run lint` and `npm run fmt:check` — clean

Not run: the full suite. CI is authoritative. Several real-git tests needed
explicit 20 to 30 second timeouts, because the default 5 seconds is not enough
when the machine runs several suites at once.


apmantza (changes requested):
Thank you for this contribution. The problem is real, the direction is right, and the test investment (real temporary repos, porcelain edge cases) is exactly what this codebase asks for. The activation gate and the rename-token handling both held up under probing.

This review is AI-generated, with the maintainer supervising the process.

I filed #2060 for this defect; please add `Closes #2060` to the PR title (the `PR title` check requires an issue reference).

Verdict: needs changes. Unit tests fail on the head SHA, and one real-git regression got past the PR's validation. Findings ranked by severity, all reproduced by execution:

**F1 (blocker): Unit tests fail on head `fba8be70`.** Two governance sweeps, not your feature tests:
- `tests/clients/pi-host-contract.test.ts:910` pins the consumers of `bash-file-access.js`; the new `isFailedGitIntegrationCommand` import (`clients/runtime-tool-result.ts:83`) must be added to the pinned list.
- `tests/clients/session-state-conformance.test.ts:352` flags the new module-level `Set` constants in both files; register or exempt them.
Targeted files green is not enough here; these sweeps run in the full suite only.

**F2 (high): the porcelain matrix rejects a legal real-git status and voids all recovery.** `LEGAL_ORDINARY_PORCELAIN_STATUSES` (`clients/opaque-mutation-scan.ts:288`) omits ` R`, ` C`, `DR`, `DC`. Reproduced on git 2.55: a worktree rename alongside an ordinary edit returns `verdict: "unknown"`, so `other.ts`, a genuine agent write, is dropped. On master it was recovered. One unexpected entry must not discard every other path in the command. The AGENTS.md paragraph encodes the same incomplete matrix and needs the same correction.

**F3 (high): `git pull` reproduces the reported bug and is not covered.** `GIT_INTEGRATION_SUBCOMMANDS` (`clients/runtime-tool-result.ts:67`) lists merge, rebase, cherry-pick. A conflicted `git pull --no-rebase` leaves `UU` plus clean `A ` incoming entries and dispatches them today. `git revert` also leaves unmerged states. Please enumerate the family (pull, revert, am, stash pop/apply, checkout -m, apply -3) with a per-member verdict in the PR body, and cover at least pull.

**F4 (medium): truncated stdout is fail-open.** The parser never reads `outputTruncated` (`clients/safe-spawn.ts:150`). A capped status that happens to end on `\0` parses as a successful, smaller recovery. Return `git-status-parse-failed` when truncated.

**F5 (medium): agent-authored staged files are silently excluded.** In one call like `python gen.py && git add -A && git merge origin/main`, the generated output is index-only clean and gets dropped as incoming. If that trade-off is intended, name it in the PR body and cover it with a test; today the body claims partial writes are preserved.

**F6 (medium): unrelated widening.** `clients/runtime-tool-result.ts:653` drops the `recognized.length > 0` guard on the coverage-unknown record. Defensible, but it is a second behavior change with no test and no mention; either revert it or state and test it.

**F7 (low): no record when the filter activates.** Nothing logs how many incoming paths were excluded, so over-exclusion in production would be invisible. A counted field on the recovery record closes it.

**F8 (low): resolved by #2060 above.**

Red-first verification: I reverted the two client files to master, rebuilt, and ran your two test files: 11 failed / 16 passed, with the behavior tests failing for the right reasons. The regression coverage is honest; F2 shows the `accepts ordinary tracked states` matrix needs the missing pairs so it would have caught its own gap.

Also cleared under attack, for your confidence: rename `-z` token ordering, trailing-NUL handling, the unmerged-pair set, path canonicalization, the `-C`/`-c` global-option skipper, and the changelog fragment.

The branch is BEHIND master; update it when you push the fix round and re-run the two governance suites listed in F1. Happy to help with any of the above if maintainer edits are enabled.

apmantza:
Merged — thank you again @alchemistklk for a genuinely valuable contribution and for the real-git test investment. Summary of the maintainer-side additions on top of your work, all preserving your authorship: the porcelain matrix gained the missing legal pairs with a softened blast pattern, the integration family widened to pull/revert/am with probed exclusions for stash-pop, truncated-stdout now fails closed, two bounded observability records landed, and the governance pins were updated. Residual polish items are tracked in #2081. This review-and-merge process was AI-assisted, with the maintainer supervising.

Issue #2052, opened 2026-08-25
fix(lsp): decline roots outside the session cwd instead of serving them


apmantza:
## Evidence

2026-08-25 session (refs #2050): a coordinator pi session at `C:/Users/apman/OneDrive/Desktop/pi-lens` ran `lsp_diagnostics` against `.ts` files in sibling git worktrees under `C:/Users/apman/AppData/Local/Temp/pi-agent-*/`.

Result: two files were flagged with **49 blocking type errors** — `tool-discovery.test.ts` (41) and `managed-tool-refresh.ts` (8) — every one of the form `Cannot find name 'process'`, `Cannot find module 'vitest'`, `Cannot find module 'node:path'`, `Cannot find namespace 'NodeJS'`, `Cannot find name 'setImmediate'`.

Those verdicts are provably false: the same worktree's own `npm run build` and `npm run lint` passed, and CI's required `Lint & type-check` + `Unit tests` passed on Linux, macOS, and Windows for those exact commits. The session summary then presented the false errors as `🔴 49 blocking errors`, dominating the diagnostics report.

## Root cause (mechanism documented, gap labeled a finding)

`resolveLanguageRootForFile` clamps a found marker root back to the workspace root via its stricter workspace-relative check (#807 second wave). For a file whose nearest `tsconfig.json` lies OUTSIDE the session cwd (a sibling worktree), the request is then answered under a project context whose `@types`/dependency tree does not apply. Nothing at the touch or decline boundary distinguishes "file belongs to a foreign root" from a normal answer, so the server produces systematic false type errors and delivery surfaces count them as blockers.

This is adjacent to defect shape 10: a wrong-context answer looks identical to a real verdict downstream.

## Acceptance criteria

1. When the nearest marker/config root for a requested file lies outside the hard session-cwd ceiling, the LSP paths return an EXPLICIT outcome (e.g. `outside-project-root`, carried as an unconfirmed/skip reason) instead of type-error results. Paths that must honor it: `touchFile`'s diagnostic collection, `lsp_diagnostics` (single, batch, and directory walk), `lens_diagnostics` full-sweep file enumeration.
2. Blocker tallies exclude outside-root declines. Widget footer and `mode=all`/`delta` summaries either omit declined files or surface them in the existing skipped/unconfirmed lane — never counted as blocking errors.
[...]

Issue #2089, opened 2026-08-25
Two tests report PASSED on a bare return: one never runs on Linux CI, one passes on the failure it guards


apmantza:
Two tests exit early with a bare `return` before any assertion. Vitest reports both as PASSED, not skipped, so the suite claims coverage it does not have.

## Evidence

`tests/monorepo/pnpm-symlink.test.ts:128` returns early when `process.platform !== "win32"`. The only "Unit tests" job runs on `ubuntu-latest` (`.github/workflows/ci.yml:91`). This test therefore asserts nothing in CI, on every run, and reports green.

`tests/clients/tree-sitter-client-init.test.ts:58` wraps `_require.resolve("tree-sitter-wasms/package.json")` in a try/catch whose catch body is a bare `return`. The test's stated purpose is that `findGrammarsDir` resolves external packages through `require.resolve`. The exact failure it guards — the package not resolving — is what makes it pass.

Census: 11 bare-return sites inside `it()` bodies across 777 test files. The other nine call `ctx.skip()` or `skip(reason)` first, which is the correct pattern; `tests/clients/dispatch/runners/helm-render.test.ts:1030` carries a comment stating exactly why ("a silent pass would claim coverage this host never had"). These two predate that discipline.

## Acceptance criteria

- Both sites call `ctx.skip(reason)` before returning, so the run reports a skip.
- `pnpm-symlink.test.ts`'s POSIX variant is either written (a real symlink does report `isDirectory() === true`, per the site's own comment) or the gap is recorded as a named skip reason.
- A mechanical guard: a lint rule or meta-test that fails on a bare `return` inside an `it()`/`test()` body that precedes every assertion in that body, unless the preceding statement is a `skip` call.

## Observability

CI output is the record: a skipped test prints as skipped with its reason. Today these print as passed.

## Cross-links

Refs #1842, #1506 (Linux-only CI cannot see the win32 branch — the same blind spot, here reported as a pass rather than a gap). Companion to the governance-sweep issue.

*This issue is AI-generated, with the maintainer supervising the process.*

PR #2098, opened 2026-08-25
fix(lsp): release synced text on document close (closes #2065)

### LINKED ISSUE #2065: documentContentHashes retains full document text unbounded under Incremental sync; closeDocument misses it (36.9 MB at field scale)
Under Incremental sync — which the production TypeScript server negotiates — `documentContentHashes` retains the full text of every document ever sent, for the lifetime of the client. Nothing evicts it. `closeDocument` deletes six sibling per-path maps and misses this one.

### The retention is live on the dominant server

`bench-spawn.mjs` against the production-managed server (`~/.pi-lens/tools`: typescript-language-server 5.1.3, TypeScript 5.9.3) reports:

```
negotiated textDocumentSync = 2
```

2 is `Incremental`. `clients/lsp/client.ts:1712` retains the full text on exactly that branch:

```js
...(state.syncKind === TEXT_DOCUMENT_SYNC_KIND_INCREMENTAL && {
    text: content,
}),
```

913 of 941 `lsp_client_selected` records in the field window carry `serverId: "typescript"`, so this branch is the common case, not the edge case.

### Measured retention

Harness `bench-lsp-sync.mjs` drives the compiled `clients/lsp/client.js` `handleNotifyOpen` / `handleNotifyChange` over 200 real repo files (mean 58 KB), 3 edits each, with a count-instrumented connection:

| sync kind | entries | entries carrying full text | retained text |
|---|---|---|---|
| Full (1) | 200 | 0 | 0.00 MB |
| Incremental (2) | 200 | 200 | **11.30 MB** |

Same workload, same wire bytes (62.5 vs 62.6 KB per edit). The 11.30 MB is pure retention.

### Field scale

`~/.pi-lens/latency.log`, 2026-08-24T21:31:20Z to 2026-08-25T19:18:35Z (21.8 h), 1083 `lsp_document_send` records:

- 434 distinct paths sent.
- Sum of the latest `contentLength` per path: **18.44 MB of characters**.
- As UTF-16 heap: **about 36.9 MB** held by `documentContentHashes.text` alone, per client, never released.

### `closeDocument` misses it

`clients/lsp/client.ts:3674` onward deletes the per-path entry from `openDocuments`, `openDocumentUris`, `documentVersions`, `documentOpenedAt`, `diagnosticPublicationCounts`, and `projectIdentityProbedFiles`. It does not touch `documentContentHashes`. Closing a document releases every sibling record except the one holding the bytes.

`clearDiagnosticsForPath` (`client.ts:1584`) skips it deliberately, and correctly — the next publish still needs the version and hash. But that reasoning covers the `{version, hash}` pair, not the retained `text`.

## Root cause

`recordSentContent` (`clients/lsp/client.ts:1700`) writes into a plain `Map` with no eviction policy and no owner responsible for dropping the `text` field. The single reader is `buildContentChanges`, which needs only the *previous* text for a path that is about to receive another change. A path that is closed, or never edited again, holds its full text until the client dies.

The defect shape is the per-path teardown that misses one member of its family: six of seven per-path maps are cleared on close.

## Acceptance criteria

Assert on counts, not wall time.

- [ ] `closeDocument` drops the `documentContentHashes` entry for the closed path, or at minimum its `text` field. Removing that line must red a test.
- [ ] A test opens N documents under `syncKind = 2`, closes them all, and asserts the count of entries carrying a `text` field is **0**. The same test under `syncKind = 1` asserts 0 before and after, proving the Full path never regressed.
- [ ] A retention bound exists and is named. Either a cap on how many paths retain `text` (evicting least-recently-sent), or documented reasoning for why unbounded growth keyed on distinct paths touched is acceptable, with the ceiling stated.
- [ ] A test asserts entries-carrying-text never exceeds the declared bound after sending 2x the bound in distinct paths.

## Observability record

No record today reports how many paths a client retains text for. The gap: `lsp_document_send` in `~/.pi-lens/latency.log` carries `contentLength` for the send but nothing reports `documentContentHashes.size` or the count of entries carrying `text`. Add a bounded periodic record, or fold the count into the existing `memory_sample` phase, so the ceiling is provable from production rather than from a bench.

## Cross-links

- Refs #1999 — long-session parent heap reaches 1.3 GiB with the growth unattributed. This is an attributable LSP-side contributor: about 36.9 MB per client at field scale, never released. Not the whole story, but it is measurable and it is ours.
- Refs #1669 — introduced the Incremental retention.
- Refs #1095 — owns the `{version, hash}` pair in the same Map, whose retention across a clear is deliberate and unaffected by this fix.
- Refs #1332, #1126 — prior LSP-side memory work.

apmantza:
## Summary

Bound full-text retention for Incremental LSP synchronization and release the retained text when a document closes. The client now keeps at most 128 least-recently-sent text paths and 64 MiB of UTF-16 text per client, while preserving the `{version, hash}` binding needed for diagnostics. Evicted open paths safely fall back to a full-document change until their next confirmed send.

`activeLspClients` (the telemetry Set backing this) is a projection registry, not a second source of truth: considered projecting straight off `LSPService.state.clients` (`clients/lsp/index.ts`) instead, and rejected it because that file imports `client.ts`, not the reverse, and because that map is itself only lazily reconciled against real client death, so projecting off it would buy no more accuracy. See the doc comment above the Set's declaration for the full reasoning.

## Tests

- `tests/clients/lsp/refresh-and-sync-kind.test.ts` adds the #2065 regression matrix: 256 distinct Incremental paths stay within the 128-entry and 64 MiB bounds, closing all paths leaves zero text entries and zero retained bytes, Full sync retains zero text before and after close, and the real `handleNotifyChange`/`buildContentChanges` path proves a retained incremental edit round trip.
- `tests/clients/memory-sampler.test.ts` verifies the new LSP retained-entry and retained-byte fields are non-negative.
- Red-first: the pre-fix production files failed with 256 retained entries versus the 128 bound. Mutation proof: removing only the close eviction failed with 128 entries remaining after close.
- `npm run build`
- Standalone targeted LSP suite: 29 passed.
- `npm run fmt:check`
- `npm run astgrep:self-scan`: 0 findings.
- TypeScript lint and changelog validation passed. A combined targeted run also passed the memory-sampler suite; two unrelated workspace-refresh tests hit their existing 5-second timeout under parallel load, while the standalone LSP suite passed.

## Blast radius

The retained-text map has three production readers or writers:

- `recordSentContent` writes the hash, version, and optional Incremental text after a confirmed didOpen/didChange send.
- `buildContentChanges` reads only the previous text to calculate the ranged full-document replacement; a missing text entry falls back safely to a full-document change.
- Diagnostic binding reads only the hash and version, which remain available after text eviction.

`closeDocument` removes the closed path's complete content-binding entry after didClose. Client shutdown removes the client from the bounded telemetry registry. No `clients/lsp/index.ts` changes were required, so the concurrent #2052 root-selection work is unaffected.

## Class sweep

The closeDocument per-path family is complete:

| State member | Close verdict |
| --- | --- |
| `openDocuments` | deleted directly |
| `openDocumentUris` | deleted directly |
| `documentVersions` | deleted directly |
| `documentOpenedAt` | deleted directly |
| `diagnosticPublicationCounts` | deleted directly |
| `projectIdentityProbedFiles` | deleted directly |
| `documentContentHashes` | deleted directly; this fixes #2065 |
| `pullGenerations` | deliberately NOT cleared — fix round 1 briefly added a delete here (F4), fix round 2 reverted it: the same reasoning `clearDiagnosticsForPath` documents for `pullRequestSequences` applies to this counter, and a delete-on-close lets a stale in-flight pull's captured generation match the counter again after a close/reopen. Re-homed to the #2065 issue as a census remainder rather than closed here. |

The broader per-path diagnostic maps (`pushDiagnostics`, pull diagnostics, timestamps, bindings, publication versions, pull result IDs, workspace pull cache, and legacy diagnostic maps) continue through `clearDiagnosticsForPath`, which `closeDocument` calls. A tree-wide production grep found no other negotiated-mode full-text retention; only `recordSentContent` stores Incremental text and only `buildContentChanges` consumes it.

## Observability

Periodic `memory_sample` records now include aggregate active-client counts for Incremental retained text paths and UTF-16 bytes. The cap is therefore visible in production logs and can be compared with process heap growth without scanning document contents.

## Fix round 1

The review verified the eviction design (protocol convergence, re-prime, no thrash, exact accounting) and found a real crash-path leak plus two untested guards. All six findings are fixed on this branch, each red-proven on pre-fix (or pre-round) code before the fix landed.

| Finding | Disposition |
| --- | --- |
| F1 (blocker): `activeLspClients` never deregistered on a crash, only on graceful shutdown, so a crashed client's retained text stayed counted for the rest of the process lifetime | Fixed. Deregistration now lives in `disposeClientConnection`, the one place every death path (`connection.onError`, `connection.onClose`, the process `exit` handler, and `clientShutdownOnce`'s `finally`) already converges on, guarded by the same `connectionDisposed` idempotency flag. Registered `activeLspClients` in `tests/support/session-state-registry.ts` (lsp/client.ts pin: 2 → 3) with an explicit reset story: it deliberately spans sessions, matching the LSP clients it mirrors, which are themselves kept warm across session boundaries. |
| F2: the 64 MiB byte cap was untested at meaningful scale (neutering it stayed 29/29 green) | Fixed. New test sends two ~40 MiB documents (entry count 2, far below the 128-entry cap) and proves the byte clause alone evicts the older one. |
| F3: the LRU recency reinsert was untested (removing it stays green — silent FIFO degradation) | Fixed. New test fills to the 128-entry cap, re-touches the oldest path, adds one more, and proves the re-touched path survives while an untouched-but-newer path is evicted instead. |
| F4: `pullGenerations` is an eighth per-path map that survived `closeDocument` with no eviction | Reverted in fix round 2 (see below) — the fix was itself a regression. |
| F5: the eviction scan spread and scanned the ENTIRE content-binding map (including already-stripped paths) on every `didChange` past the cap | Fixed. Added `incrementalTextBearingPaths`, a Set tracking only text-bearing paths in the same recency order, kept in lockstep with `documentContentHashes`'s `text` field. Eviction now reads the first entry of that Set instead of spreading and filtering the full map — cost is bounded by the 128-entry cap, not by open-document count. At field scale (hundreds of open documents), this removes a per-send allocation and scan that used to grow with total open documents, not just the retained subset. |
| F6: the three `memory_sample` assertions were structurally always-0 (mock states never enter `activeLspClients`) | Fixed. Added a test using a real `createLSPClient` + fake-server client with known retained bytes, asserting `subsystems.lsp.clients/incrementalTextEntries/incrementalTextBytes` are all positive and match the known content size. |

Verification this round: `npm run build` before every run; targeted suites (`refresh-and-sync-kind.test.ts`, `memory-sampler.test.ts`, `session-state-conformance.test.ts`, `index-memory-sample-wiring.test.ts`) all green; sibling files referencing the touched symbols (`client-crash-logging`, `client-internals`, `rename-notify-wedged`, `service-rename-file`) also green in isolation — `service-rename-file` hit its existing 5-second timeout only under parallel load, reproduced as flake-not-regression by rerunning it alone. `npm run fmt:check`, `npm run lint`, `npm run astgrep:self-scan` (0 findings), and `npm run changelog:check` all pass.

## Fix round 2

The round 1 review confirmed F1/F2/F3/F5/F6 fixed and endorsed the `activeLspClients` layering rationale after probing the alternative directly. It also found that round 1 introduced two regressions of its own.

| Finding | Disposition |
| --- | --- |
| N1 (high): `closeDocument` never deleted a path from `incrementalTextBearingPaths`, the aux set fix round 1 added for F5. Only `recordSentContent` and the eviction loop deleted from it. Below the retention cap, eviction never runs, so nothing ever drained a closed path's membership — an unbounded per-path store, the exact class #2065 exists to close, now living in the aux index instead of the byte count. The reviewer's audit probe (4 rounds of 100 open/edit/close) measured real=0, aux=400, stale=400. | Fixed: one `state.incrementalTextBearingPaths?.delete(normalizedPath)` added to `closeDocument`. Added an assertion to the existing "#2065 bounds and tears down" test proving the aux set is empty after closing every path; red-proven by disabling the new delete first — see the red-run evidence below. |
| N2 (high): fix round 1's F4 (`state.pullGenerations?.delete(normalizedPath)` in `closeDocument`) was itself a bug. `pullGenerationFor` defaults a missing entry to `0`, so deleting it on close resets the counter; a pull that captured its generation before the close, then lands after a close/reopen, would find the counter back at the same value it captured and pass the staleness check it should fail. The file's own comment already explains why `pullRequestSequences` — the sibling counter — is deliberately never cleared for the same reason. | Reverted. Removed the F4 test (it pinned the wrong behavior) and rewrote the comment in `closeDocument` to state plainly why `pullGenerations` is retained, cross-referencing the `pullRequestSequences` comment it should have matched the first time. The class-sweep table above reflects the reverted verdict. Re-homing the `pullGenerations` census remainder to the #2065 issue as a comment, not closing it here — the coordinator is doing that directly. |

Red-run evidence (N1): before adding the `closeDocument` delete, the new assertion failed with `AssertionError: expected 128 to be +0` (128 stale aux-set entries survived closing all 256 test paths — matching the reviewer's audit shape). After the fix, the same test passes.

Verification this round: `npm run build`; `refresh-and-sync-kind.test.ts` (32/32), `memory-sampler.test.ts` (22/22), and `session-state-conformance.test.ts` (76/76) all green in isolation — one workspace-refresh timing test flaked under combined parallel load and passed cleanly alone, matching the same pre-existing flake noted in round 1, not a regression. `npm run fmt:check`, `npm run lint`, `npm run astgrep:self-scan` (0 findings), and `npm run changelog:check` all pass.

Refs #2065


apmantza:
Post-merge note from the round-3 review: the N2 revert (pullGenerations deliberately uncleared on close) is documented in a comment but untested. The reviewer's P4 probe is an eight-line regression test that reds if anyone re-adds `state.pullGenerations?.delete(...)` in `closeDocument` — probe shape: capture a pull generation, close and reopen the path, assert the reopened generation differs from the capture (post-fix: captured=1, afterCloseReopen=3). Non-blocking per the mutation-proof rule (N2 removes code, adds none). Fold into the next test-hygiene pass (#2089/#2090).

Issue #1842, opened 2026-08-20
Umbrella: test-infrastructure discipline


apmantza:
Tracking issue for the test-infrastructure design umbrellas opened across the 2026-08 retrospectives. This issue coordinates; it does not replace any member. Each linked issue stays open as its own work item, with its own acceptance criteria and sequencing.

## Members

- [ ] #1605 — Smoke-test umbrella: six real-host smoke lanes (classification, availability-lifecycle latch, staleness round-trip, exclusion-scope, language-coverage, log-contract) for defect classes already observed this arc.
- [ ] #1609 — Prospective test classes: six classes for defects the suite structurally cannot see yet (latch property tests, hostile-LSP-server fixture, crash-atomicity, filesystem hostility, perf budgets, tool-version golden corpus).
- [ ] #1635 — State-evolution and truth-vs-render: turn-simulation harness, session-state lifecycle registry, log-contract tests, truth-vs-render metamorphic tests, renderer round-trips — the multi-turn and rendered-claim families dogfooding still catches exclusively.
- [ ] #1661 — Verification discipline: perf hypothesis ledger, committed micro-benches for safety-gate hot paths, log-calibrated fixtures, an anti-flake contract, an issue-numbered regression convention, and a named test-admission filter.
- [ ] #1741 — State charter: extends the session-state registry to a four-lifetime charter (turn, session, process, persisted), each store declaring scope, invalidation, and re-arm.
- [ ] #1742 — Seams over mocks: three injectable seams (spawner, wire-speaking fake LSP server, telemetry sink) sized to eliminate the largest module-mock clusters, replacing a ratchet with real replacements.
- [ ] #1829 — Tool-contract smoke suite: a declared exit/output contract per wrapped binary, a gated real-binary test per tool, and a CI lane that installs a rotating subset so the gated tests execute somewhere.
- [ ] #1838 — Fault-injection kit: shared wedge/delay/starve/reset primitives (wedgedChild, delayInside, starveBudget, fireResetAt), each with its own fidelity test.
- [ ] #1839 — Runner-parser garbage harness: one adversarial-shapes battery driven through every dispatch runner's parser, asserting classification invariants never go clean-on-garbage.
- [ ] #1840 — Reset-interleaving explorer: instrument every await point of a path under test and fire a session reset at each one, asserting no cross-session write or latch survival.
[...]

PR #2109, opened 2026-08-26
fix(tests): red-first sweep for test bodies that return before asserting (closes #2089)

### LINKED ISSUE #2089: Two tests report PASSED on a bare return: one never runs on Linux CI, one passes on the failure it guards

AngriestBird:
## Summary

Thirteen test bodies exited early with a bare `return` before any assertion and reported PASSED (the issue's census named two; the grep missed the inline `if (…) return;` forms). All 13 are converted, and a mechanical sweep now prevents the class from returning. The pnpm-symlink win32 gate is deleted outright: its comment claimed POSIX symlinks report `isDirectory() === true`, which is false on every platform (`fs` Dirents report `isSymbolicLink() === true` / `isDirectory() === false`, junctions included), so the test now runs and asserts on Linux with a new positive control. Five win32 gates became `it.skipIf`, six fixture/tag-dependent gates became `ctx.skip(condition, reason)`, and the tree-sitter resolve case is deleted as a parallel-path test that never exercised `findGrammarsDir` (follow-up for that coverage gap to be filed separately).

Closes #2089. All three acceptance criteria are met: both named sites (and eleven more) now skip honestly or assert, the POSIX variant question resolved into the gate deletion (the POSIX behavior is identical, so the test simply runs), and the mechanical guard ships.

## Type of change

- [x] Bug fix

## Area

- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [ ] `npm run build:dist` — not applicable, no code under `clients/`, `commands/`, `tools/`, or `index.ts` changed
- [x] `AGENTS.md` is updated (screen 2's Detect clause now points at the sweep)
- [x] `.changelog/fix-2089-vacuous-skip-guard.md` has one valid entry

## Tests

New:
- `tests/support/vacuous-skip-scan.ts` — the detector: strips comments/strings via sweep-kit's `stripSource`, walks `it`/`test` bodies (including `.each`, `.skipIf(...)(`, `.runIf(...)(`), tracks function braces only, flags a return at body depth that precedes the first assertion with no earlier skip call. Known blind spots declared in the module head, all false-negative direction.
- `tests/config/vacuous-skip-coverage.test.ts` — the sweep (invisible-skip screen): `assertNonEmptyScan(786, 390)` + `auditRegistry({minScanned: 390, minFlagged: 0, minReasonLength: 20})`, empty exemption map. Liveness proven four ways: 5 must-catch fixtures (including skipIf/each evasions), 8 must-not-flag fixtures, a real-source probe that strips `ctx.skip()` out of `helm-render.test.ts` and demands ≥4 new flags, and a real-source probe that the C fixture string in `tree-sitter-c-rules.test.ts` stays clean.

Edited (the 13 conversions): `tests/monorepo/pnpm-symlink.test.ts` (gate deleted, positive control added), `tests/clients/tree-sitter-client-init.test.ts` (parallel-path case deleted), `tests/clients/path-utils.test.ts`, `tests/clients/pipeline.test.ts`, `tests/clients/read-guard-path-normalization.test.ts`, `tests/clients/dispatch/runners/elixir-parser.test.ts`, `tests/clients/review-graph.service.test.ts` (all `it.skipIf`), `tests/clients/actionable-warnings.test.ts` (2 sites), `tests/clients/code-quality-warnings.test.ts`, `tests/clients/diagnostic-dispositions.test.ts`, `tests/clients/finding-identity.test.ts`, `tests/scripts/changelog.test.ts` (all `ctx.skip(condition, reason)`).

Red proof, sweep run against pre-fix code (flags exactly the 13 converted sites):

```
× no test body returns before it asserts or skips
AssertionError: … expect(audit.problems).toEqual([])  — 13 unaccounted items listed by file:testName
```

Mutation proof: re-inserting `if (process.platform !== "win32") return;` into a test body reds the sweep (`expected [ Array(1) ] to deeply equal []`); removing it greens. Independent reviewer mutation on `clients/source-filter.ts:653` (`|| entry.isSymbolicLink()`) reds the rewritten pnpm-symlink assertion, proving the Linux walk reaches the symlink and the new assertions bite.

Totals: 314 passed, 10 skipped, 0 failed across the 20 targeted files; `tests/config/` + `tests/support/` 145 passed. Every skip now prints as a skip; the `ctx.skip` sites print their reason.

## Blast radius

Test-tree and AGENTS.md only; zero production modules touched, so `module_report` blast radius is not applicable. The sweep adds one file walk over `tests/` per suite run (same shape as the existing `sweep-floor-coverage` and `module-instance-coverage` sweeps). Open PRs #2106/#2107 touch AGENTS.md hundreds of lines away and add no bare-return test bodies, so this sweep stays green in any merge order.

## Observability

CI test output is the record: the 11 previously silent sites now print as `↓ skipped` (with reasons at the `ctx.skip` sites) instead of PASSED. Residual, pre-existing and now visible instead of hidden: no CI lane executes the win32-gated tests (`install-test` is the only windows job and never runs vitest) — #1842/#1506.

## Class sweep

Defect shape: AGENTS.md test-authoring screen 2, "a skip that reports as a pass" (invisible skip). Population: all 786 `.test.ts` files under `tests/` (the sweep's own census; scripts/tools/clients carry no vitest bodies). Per-member verdict: 13 offenders converted (listed above), 7 compliant skip-first sites left untouched (`helm-render.test.ts` ×4, `test-runner-cache-roots`, `service-rename-file`, `edits`), ~40 declarative `skipIf`/`runIf` gates already correct, everything else asserts before returning. Consolidation verdict: folded onto one seam — the `vacuous-skip-coverage` sweep now enforces the screen mechanically, and AGENTS.md's Detect clause points at it.


apmantza:
Thank you for taking this on — the bare-return vacuity class is exactly the kind of sweep #2089 needed, and an external pair of eyes on the test corpus is very welcome. A detailed review is underway and will follow shortly; heads up that it is AI-generated with a maintainer supervising the process. One early note: the red "PR body (advisory)" check means the body is missing some of the repo's template sections (Tests with red-first outputs, Class sweep, Observability) — no need to act yet, the review will spell out exactly what would help.

apmantza (changes requested):
Thank you for this contribution — this is an unusually well-built sweep. The detector is documented, probed in both directions, and your red-run claim held exactly: reverting the twelve test files against your detector flags precisely the 13 sites you list, including both sites #2089 named. The eleven `ctx.skip` conversions preserve their assertions verbatim, and the `pnpm-symlink` positive control is a nice touch.

Two asks before merge, both small:

**1. The detector can manufacture a false positive on a return-type-annotated nested function (medium-high).** In `tests/support/vacuous-skip-scan.ts`, `isFunctionBrace` classifies the brace from the token immediately before `{`. For `async run(): Promise<RunnerResult> {` that token is `>`, so the method body is classified as a block, and a healthy nested guard like `if (aborted) return;` inside it is counted at the test callback's own level. Reproduced on real source: inserting such a guard into the `async run(): Promise<RunnerResult>` at `tests/clients/dispatch/runner-timeout.test.ts:83` reds the sweep with a bogus finding. This contradicts the module's own invariant ("the sweep can miss a defect, but it cannot manufacture one"), and `tests/` already has 116 sites spelling `): Promise<…> {`, so the first future guard in one of them reds a repo-wide gate. Suggested fix: when the preceding token is neither `)` nor `=>`, walk back across a `:`-introduced type annotation to the parameter list's `)` before deciding, and add the `): Promise<T> {` method shape to the `CLEARED` fixtures so the false-positive direction stays probed.

**2. Deferral hygiene for the deleted `findGrammarsDir` case.** The deletion itself is right — `tree-sitter-wasms` was never installed, so the case was vacuous since inception. But the code comment defers the real coverage gap at `clients/tree-sitter-client.ts:896` to a follow-up "to be filed separately", and `closes #2089` requires that remainder to have a home. We filed it for you as #2112 — just cite that number in the code comment and you are done; `closes #2089` then stands.

Notes needing no action from you: the red "PR body (advisory)" check is stale — it ran six seconds after PR creation against the creation-time body, and your current body has all the required sections. And your branch composes cleanly with current master (we merged and re-ran the sweep: green).

Verified on our side: Unit tests and Lint & type-check executed and passed on head b6dbd8eb; mutation probes (re-inserting a bare win32 gate reds the sweep, forced `ctx.skip` genuinely skips and prints); the valued-return blind spot you declare costs nothing today (32 extra sites under a relaxed matcher, all inspected, none vacuous).

apmantza:
Status update from the maintainer side: master has moved quite a bit today and this branch is now merge-conflicted (a conflicted PR silently skips the Unit tests and Lint checks, so current green marks are not meaningful until it's resolved — resolving and pushing re-triggers them). Two changes to the earlier review asks that work in your favor: (1) issue #2112 — the findGrammarsDir coverage gap your deletion deferred — has since been fixed and closed independently (PR #2136 landed injectable resolver seams with all branches covered), so your code comment can simply cite #2112 as resolved; the deferral ask is fully discharged. (2) That leaves only the detector false-positive fix (the `): Promise<T> {` method-body classification and the CLEARED fixture for it) between this PR and merge-ready. If you'd like, we're happy to resolve the merge conflict on our side via maintainer edits, preserving your authorship — say the word, or push your own resolution, whichever you prefer.

apmantza:
Maintainer update: I resolved the merge conflict directly on your branch (maintainer edits were enabled — thank you for that), preserving your authorship of the sweep. The one substantive conflict was in `tests/clients/tree-sitter-client-init.test.ts`: since your branch was cut, #2112-era work on master replaced the vacuous `findGrammarsDir` case your sweep had deleted with a faithful injected-resolver version, so the resolution keeps master's new tests — they supersede both the old test and the deletion. Both touched suites pass locally on the merged tree (35 tests).

CI on the new head needs maintainer approval to run (standard for fork PRs); I am approving the runs now. This review process is AI-generated, with a maintainer supervising.

apmantza:
Merged — thank you for the contribution! The sweep landed with your authorship preserved; the one conflicting test resolved to master's newer faithful version as described above. This review process was AI-generated with a maintainer supervising.

Issue #2066, opened 2026-08-25
Per-send document sync walks the whole document three times where one pass suffices (35% of edit-path self time)


apmantza:
Every `didOpen` and `didChange` walks the whole document twice more than it needs to. One pass counts newlines for a telemetry field. Another splits the previous document into every line to read one of them. Together they are 35% of self time in a CPU profile of the edit path, and neither pass produces anything the server sees.

## Evidence

CPU profile of the compiled runtime (`node --cpu-prof`, `bench-lsp-sync.mjs` driving `clients/lsp/client.js` `handleNotifyOpen` / `handleNotifyChange` over 200 real repo files, 1400 notifications, 4532 ms wall, 2958 samples). Self time:

```
21.4%  971ms  sendNotification @ bench-lsp-sync.mjs:40   <- harness JSON.stringify, models the transport
17.4%  789ms  recordSentContent @ client.js:824
15.4%  697ms  realpath @ :-1
12.1%  549ms  update @ hash:144                          <- sha256, load-bearing for #1095
 2.6%  116ms  RegExp: \r\n|\r|\n @ :-1
 1.7%   77ms  buildContentChanges @ client.js:888
```

`recordSentContent` at 17.4% is the largest pi-lens-owned self cost in the profile. Isolating each full-document pass over the same 200 documents (11.3 MB of text), `bench-counts.mjs`:

| pass | cost | per send | what it produces |
|---|---|---|---|
| `recordSentContent` newline count | 36 ms | 0.179 ms | `contentLineCount`, a log field |
| `hashDiagnosticContent` sha256 | 40 ms | 0.201 ms | #1095 content binding (load-bearing) |
| `buildContentChanges` regex split | 32 ms | 0.158 ms | one line/character position |

The split allocates **306,834 substrings per pass to read one of them**.

## Member 1: a full-document scan for a log field

`clients/lsp/client.ts:1730-1733`:

```js
let newlineCount = 0;
for (let i = 0; i < content.length; i++) {
    if (content.charCodeAt(i) === 10) newlineCount++;
}
```

[...]

Issue #2016, opened 2026-08-23
Hot-path tax class: repeated realpath-canonicalizing path normalization on the per-edit pipeline (Windows-only, invisible on CI)


apmantza:
## Evidence (measured, Windows dev box / OneDrive)

`normalizeMapKey` = `normalizeFilePath` = `realpathSync.native()` on Windows (`clients/path-utils.ts:95-107`) — measured **~200µs/call** this machine. On POSIX the same function short-circuits without fs access for clean slash paths, so **Linux CI cannot see this cost**. Any hot-path operation that normalizes the same stable value twice pays a pure syscall tax.

Found live in #2014: `recordProjectMutation` paid it twice per mutation → **346µs/mutation**; fixed to 1.2µs by reusing `bumpFileSeq`'s returned key.

## Class members swept (grep + read of the per-edit chain)

| Site | Redundancy | Est. tax/edit |
|---|---|---|
| `dispatcher.ts:277-283` \`createDispatchContext\` | normalizes \`projectRoot ?? cwd\` AND language-root EVERY dispatch; both are session-stable values | ~400µs |
| `dispatcher.ts:545/602/912-913` once-keys + baseline keys | re-normalize \`ctx.filePath\`, already normalized at :283; :913 normalizes a second relative form | ~200-600µs |
| `lsp-mutation.ts:163+183` | same file normalized as map key then again for lookup | ~200µs/batch |
| `cache-manager.ts:114-115` \`addModifiedRange\` | normalizes cwd+file per call; cwd is session-stable | ~400µs |
| fixed in #2014: \`recordProjectMutation\` receipt key | re-derived bump's key | 0 now |

Read-guard/widget-state single-normalization sites are correct-by-design (one key per op). Rough total: **~1-2ms of redundant realpath syscalls per edited file on Windows**, spread across dispatch — invisible in CI timings, real on Windows/OneDrive setups where realpath is slow.

## Defect shape

Sub-member of the #1673 family (silent per-call taxes), with a twist worth naming: **the tax only manifests on Windows because POSIX short-circuits** — so CI timing gates structurally cannot catch it. Detection must be by code inspection (this sweep) or Windows-local measurement.

## Remediation directions

1. Normalize session-stable values ONCE per session (cwd/projectRoot) and thread the keys through — mirrors #2014's fix pattern.
2. Reuse an already-normalized value instead of re-deriving (dispatcher ctx carries \`normalizedFilePath\`; downstream consumers should use it).
[...]

Issue #1740, opened 2026-08-19
Offload CPU-bound scans from the extension event loop — the reader absence behind the wedge and timeout classes


apmantza:
## Evidence

Dogfood 2026-08-19: `loop_block 18270ms` (`windowCpuMs` 21844) during a full scan — the ast-grep-napi scan (12.8s) and tree-sitter scan (5.4s) run CPU-synchronous on the extension's event loop. While blocked, pi-lens cannot drain LSP pipes or read pull answers. Downstream costs currently handled symptomatically:

- ast-grep LSP wedge under sweep volume (#1714, mitigated by #1722's notify bounds).
- Scan-window `lsp_diagnostics_timeout` records plausibly being answers-on-the-wire, unread (#1713's instrument, #1549's budget question).
- The #1723 attribution gap exists because silent CPU work logs nothing.

Throttles and budgets bound the damage; removing the reader absence removes the class.

## Design

Two stages, cheap first:

1. **Chunked yielding** in the napi and tree-sitter scan loops (yield to the event loop every N files). Small diff, keeps everything in-process, lets pipes drain during sweeps. Measure the pipe-drain latency and the sweep-duration delta.
2. **Worker-thread pool** for the scans, only if yielding is insufficient. Real isolation, real cost: WASM and napi instances per worker, with the #417/#418 heap discipline applying per worker. Gate stage 2 on stage 1's measurements.

## Acceptance criteria

- Stage 1 lands with measured numbers: max event-loop block during a full scan before and after (target: no block above the LSP notify budget), sweep-duration delta, and #1722's wedge reproduction not reached even without notify throttling.
- #1723's telemetry confirms sub-budget blocks in a live session post-fix.
- Stage 2 is a separate decision, recorded with stage 1's data.

Refs #1714, #1722, #1713, #1549, #1723, #417.

apmantza:
Scope addition (maintainer decision, 2026-08-20): stage 1 also owns throttling the full diagnostics sweep's write fan-out, folded in from #1714.

[...]

PR #2110, opened 2026-08-26
perf(lsp): fold the two didChange document walks into one (refs #2066)

AngriestBird:
## Summary

Per-send LSP document sync made two full-document JavaScript walks that one pass can serve: `recordSentContent`'s LF newline count (a telemetry field) and `buildContentChanges`'s `/\r\n|\r|\n/` split allocated per line to read only the last one. A new `scanSentContent()` makes one `charCodeAt` walk producing both derived values, carried on the existing `documentContentHashes` entry beside `text`; `buildContentChanges` uses the carried last-line index+start plus a slice, with a recompute path through the same scanner for entries that carry text without a position. Eviction drops the position with the text it describes.

Non-obvious constraints kept: `contentLineCount` stays LF-only (it pairs with `diagnostic_past_eof`'s LF-byte gate), while the change range counts all three terminators; the two conventions diverge by design on lone-CR documents and a test now pins that. The `text` ⟺ `lastLine` invariant is enforced by the eviction test rather than a discriminated union: making it structural forces `lastLine` into the hand-seeded test entries that exist precisely to prove the recompute fallback matches the carried path.

Refs #2066 — AC1, AC3, and AC4 are met. The remainder is AC2's literal "one pass total": the #1095 sha256 content binding is a second, native full-document pass and still runs on every send; folding the scan into it means touching the content-binding contract, deliberately out of scope. Issue comment naming the remainder to follow.

## Type of change

## Area

- [x] area:lsp
- [x] area:perf

## Checklist

- [x] `npm run lint` passes
- [ ] `AGENTS.md` — no documented behavior, command, convention, or invariant changes
- [x] `.changelog/fix-2066-single-pass-document-sync.md` has one valid entry
- [x] Commit subject includes the issue number: `(refs #2066)`

## Tests

New, in `tests/clients/lsp/refresh-and-sync-kind.test.ts`:
- 9-case range-parity table (`\n`, `\r\n`, lone `\r`, mixed, single-line, empty, utf-8 2-byte, utf-8 astral after CRLF, utf-32 astral after lone CR), each asserted through the carried path AND the recompute path, pinning that both agree with the old split's ranges including the encoding-converted end character. Deliberately green pre-fix (AC1: reverting to `.split` must keep it green).
- Pass-count test (loose-bound screen, bounded tight): instruments `String.prototype.charCodeAt` visits and asserts < 1.5 full-document JS passes per didChange plus zero per-line arrays. Red pre-fix:

```
AssertionError: expected 1.9999183740102848 to be less than 1.5
AssertionError: expected 4001 to be less than or equal to 1
```

- Carried-field lifetime test (eviction strips `lastLine` with `text`; next send re-arms). Red pre-fix: `expected undefined to deeply equal { index: 2, start: 5 }`.

Edited, `tests/clients/lsp/client-internals.test.ts`: adds the previously missing CRLF `contentLineCount` case and a lone-CR pin of the LF-count vs range-line-count divergence.

Mutation proofs (all red): revert to `.split`; reintroduce a separate LF loop; drop the CRLF lookahead (4 tests); make eviction retain `lastLine`; neuter the recompute fallback (9 tests). Independent review additionally brute-forced all 364 strings over `{a, \r, \n}` up to length 5 plus 120 random mixed/astral docs through both paths against the pre-fix algorithm: zero divergence.

Totals: `tests/clients/lsp/` whole directory 1124 passed / 10 skipped (110 files), integration suite with real servers 37 passed, `tests/config` 30 passed.

## Blast radius

One production module: `clients/lsp/client.ts` (`recordSentContent`, `buildContentChanges`, the `documentContentHashes` entry shape). Consumers are the didOpen/didChange notify paths only; the #1095 hash pass and its consumers (`read-guard.ts`, `tools/lens-diagnostics.ts`, `tools/lsp-diagnostics.ts`) are untouched. Hot-path cost, measured through the compiled runtime: JS full-document passes per didChange 1.9999 → 1.0000; per-send median on a 400 KiB / ~8,500-line document −9% (0.751 → 0.686 ms), 200 KiB −7%, CRLF 400 KiB −2%; substring allocations 8,501 → 0 per didChange. didOpen never built ranged changes: 1 pass before and after, wall time within run-to-run spread (0.776-0.781 vs 0.790-0.825 ms min-of-5). Measurement note: a single-process multi-size benchmark shows a phantom 22% regression from heap/JIT contamination; per-process runs and `--cpu-prof` both dispel it.

## Observability

`lsp_document_send` in `~/.pi-lens/latency.log` carries `contentLength` and `contentLineCount` and proves the field stays correct in production; `loop_block` in the same file catches any regression large enough to stall the loop. No per-send CPU record exists or is added; the pass-count assertions are the durable guard (per the issue's own observability section).

## Class sweep

Defect class: O(document) work per edit where O(delta) or a single pass suffices (#1740 / #2016 family; this issue is the #2066 member). Tree-wide sweep of full-document `.split(/\r\n|\r|\n/)` and per-send scan sites across `clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts`: the two members fixed here were the only ones on the per-send path. One remaining family member found and left deliberately: `lineTextAt` in `clients/lsp/position-encoding.ts:64-70` full-splits to read one line, on diagnostic-conversion call sites, not the per-send path; needs its own issue rather than a ride-along. Consolidation verdict: terminator semantics now live in exactly one function (`scanSentContent`), which is the seam the family folds onto for this path.

### DIFF
--- a/.changelog/fix-2066-single-pass-document-sync.md
+++ b/.changelog/fix-2066-single-pass-document-sync.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Per-send LSP document sync makes one JavaScript pass over the previous document instead of two, on `didChange` (refs #2066)** — the newline count behind `lsp_document_send`'s `contentLineCount` and the last-line position behind an Incremental `didChange` range now come from a single scan whose result rides on the existing per-path content-binding entry, so a change no longer splits the previous document into one substring per line to read one of them. Line-count semantics are unchanged: `contentLineCount` still counts `\n` only, while the range still treats `\r\n`, lone `\r`, and `\n` as terminators. The remainder of #2066 is its "one pass total" target: the #1095 sha256 content binding is a second, native full-document pass and still runs on every send.
--- a/clients/lsp/client.ts
+++ b/clients/lsp/client.ts
@@ -912,10 +912,21 @@ export interface LSPClientState {
 	 *  ONLY when `syncKind` is Incremental — the one case that needs it, to
 	 *  compute the NEXT change's full-range replace against the document as the
 	 *  server last saw it (see `buildContentChanges`). Full/None servers never
-	 *  populate it, so the common case pays no extra retained memory. */
+	 *  populate it, so the common case pays no extra retained memory.
+	 *
+	 *  #2066: `lastLine` is where that same send's `scanSentContent` pass left
+	 *  the end of `text`, so the next change reads two numbers instead of
+	 *  re-walking (and re-splitting) the whole document for them. It travels
+	 *  with `text` and only with `text`: the eviction rewrite below drops both
+	 *  together, and a position describing text that is gone would be a lie. */
 	readonly documentContentHashes: Map<
 		string,
-		{ version: number; hash: string; text?: string }
+		{
+			version: number;
+			hash: string;
+			text?: string;
+			lastLine?: LastLinePosition;
+		}
 	>;
 	/** #2065: full Incremental-sync text is an LRU-like bounded subset of the
 	 * per-path content-binding map. The binding remains after text eviction. */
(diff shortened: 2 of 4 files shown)


AngriestBird:
SonarCloud flags two MINOR S7758 findings (prefer `codePointAt` over `charCodeAt`) on the new scanner loop. Deliberately not taken: both reads compare against 10/13, which no surrogate unit or astral code point can equal, so the two are interchangeable here — and `codePointAt` measured 8% slower on a 400 KiB LF document (0.716 vs 0.663 ms/scan, per-process medians), on exactly the pass this PR exists to make cheaper. Equivalence was verified over a probe corpus including lone surrogates and astral pairs before rejecting the swap. A comment at the call site records the measurement; no suppression is used, so the findings stay visible for a maintainer Won't-Fix if desired.

This comment is AI-generated, with the maintainer supervising the process.

apmantza:
Thanks for picking this up — #2066's didChange walk fold is a real win if the semantics hold, and it is great to see it tackled. A detailed review is underway and will follow shortly; it is AI-generated with a maintainer supervising the process. Early note: the red "PR body (advisory)" check flags missing template sections (Tests with red-first outputs, Blast radius, Class sweep, Observability) — the review will list exactly what would help, so no need to act on it yet.

apmantza (approved):
Thank you for this one — it is a genuinely strong perf PR. The fold is semantically exact under our independent probe (exhaustive over `{a, \r, \n}` to length 6 plus 200,000 random strings including astral characters, U+2028/U+2029, and lone surrogates: 201,093 checked, 0 divergences against both pre-fix algorithms), your red run reproduces character for character, and every claimed mutation proof holds. The #2098 retention seam is untouched and its suites pass.

This review is AI-generated, with a maintainer supervising the process. Approving; we will land it shortly.

Four optional notes, none blocking — feel free to address in a follow-up or leave them to us:

1. The "-9% per-send median" absolute did not reproduce in our per-process measurement (the direction did: JS passes over the document drop from 2.0 to 1.0, and the LF-document scan cost roughly halves). The pass count is the durable evidence; consider leading with it.
2. `client.ts:2009`'s `?? scanSentContent(previousText).lastLine` fallback is reachable only from hand-seeded test doubles today — a one-line comment saying so would keep it from silently reviving the extra pass if a future writer forgets `lastLine`.
3. `measureDocumentScans` patches `String.prototype` across an `await`; safe under sequential execution, worth a comment so the file never adopts `it.concurrent`.
4. The pre-existing overlapping-`didChange` window your change sits next to (payload built synchronously, `recordSentContent` after the await) is now filed as #2113 — no action needed from you; noting it so the adjacency is on record.

One body nit: the checklist ticks "proven RED on pre-fix code" for all new tests, but the 9 range-parity cases are green pre-fix by design (that is what equivalence requires); your prose already says this honestly — the checkbox just reads stronger than the truth.

apmantza:
Marking this ready and landing it — CI is green on the head, the review is approving, and the equivalence probe cleared it. If you had more polish queued, follow-ups are very welcome (the four optional notes in the review are a ready-made list).

Issue #1487, opened 2026-08-16
A runner that fails to start is reported to the agent as a failing test


apmantza:
`parseGenericRunnerOutput` initializes `failed = exitCode === 0 ? 0 : 1`. When a runner exits non-zero without running anything — a config error, a load failure, a missing dependency — no count parser matches, so that initial 1 survives.

The agent is then told:

```
[Tests] ✗ 1/1 failed, 0 passed — /tmp/proj
  ✗ /tmp/proj failure
    error: could not load configuration file
  → Fix failing tests before proceeding
```

There is no failing test. The directory is named as the failing case, and the instruction is to fix tests that never ran.

## Why the existing guard does not catch it

`clients/test-runner-client.ts` already builds an `error` field for exactly this situation, but its condition requires `failed === 0`. Since `failed` was initialized to 1 by the non-zero exit, the condition never holds on this path, and `formatResult`'s runner-error branch — which requires `passed === 0 && failed === 0` — is unreachable with it. The two halves disagree about what a non-zero exit means.

Pre-existing, and confirmed present on master. Found while reviewing #1485, which changes nearby code but neither introduces nor worsens this.

## Direction

Distinguish "the runner rejected the run" from "a test failed". A non-zero exit with no parsed counts and error text is the first; a non-zero exit with parsed counts is the second. The `error` field and `formatResult`'s runner-error branch already exist to express the first — they just cannot currently be reached.

Take care not to weaken the guard #1485 adds: a non-zero exit with real counts must still never render as PASS.

## Observability

The turn-end line is the record, and today it actively misleads: `✗ 1/1 failed` for a run that never started is indistinguishable from one real failing test. After the fix a reader can tell the two apart without opening the runner output.

## Acceptance criteria

- A non-zero exit with no counts and error text reports as a runner error, not as a failing test, and does not invent a failure name.
- A non-zero exit with parsed counts still reports those counts and never PASS.
- A regression test for each, proven to fail on current code.

Refs #1485.

PR #2111, opened 2026-08-26
fix(dispatch): make the outputTruncated guards fire on real cap kills (closes #2100)

### LINKED ISSUE #2100: Four dormant truncation guards: spawn sites pass no output cap, so outputTruncated can never fire
## Evidence

Found by the PR #2096 review (probe-verified 2026-08-26): four parse sites carry outputTruncated guards that are textually correct but UNREACHABLE, because their safeSpawnAsync call sites pass no maxOutputBytes — and safeSpawnAsync only sets outputTruncated when a cap is configured (clients/safe-spawn.ts appendOutput returns early with no cap; there is no default).

Dormant sites:
- clients/opaque-mutation-scan.ts:394 (the #2055 F4 guard)
- clients/dispatch/runners/oxlint.ts:290
- clients/dispatch/runners/helm-lint.ts:157
- clients/dispatch/runners/helm-render.ts:953

Live reference implementations: clients/git-tracked-ignore.ts:92/:190 (16 MiB, PR #2096), clients/sg-runner.ts (4 caps), clients/helm-render.ts:776→:798.

The uncapped sites also retain unbounded stdout (a probe held 20 MB in full), which is the same wrong-axis exposure #2065 closed for LSP document text.

## Acceptance criteria

- Each dormant site either gains a rationale-cited maxOutputBytes (making its guard live, with the cap value justified against the real output scale of that tool) or the guard is removed with a comment naming why unbounded is acceptable there.
- Cap values are pinned by tests (the #2096 pattern: expect.objectContaining on the spawn options).
- The truncation path per site is red-first proven with the REAL cap-kill shape (status 1 + outputTruncated true), not the fictional status-0 pairing.
- Class sweep: re-run the #2096 census after the change; no site may claim a guard its spawn cannot trigger.

## Cross-links

Refs #2075 (the two sites already fixed), PR #2096 (census and live-cap pattern), #2055 (where the first dormant guard shipped). Shape 7's production variant: a guard that cannot fire.

AngriestBird:
## Summary

`safeSpawnAsync` now records `killedForOutputCap` when `stopForOutputLimit` starts ending a child. POSIX commonly reports that action as a SIGTERM. Windows reports status 1 with no signal or failure. Helm render now uses the ownership field instead of inferring intent from either exit shape, so a Windows cap kill yields an UNCHECKED coverage gap instead of a chart failure.

`truncatedByOutputCap` still owns incomplete-output classification. Timeout and abort retain precedence. Oxlint and `git-tracked-ignore` now use that predicate, so a capped timeout or abort is not mislabeled as truncation telemetry. The fixtures model POSIX, Windows, timeout, abort, and fast-exit cap outcomes.

Closes #2100.

## Tests

- `safe-spawn-ambient-signal.test.ts` pins live portable cap-kill invariants and timeout/abort precedence.
- `helm-render.test.ts` exercises both POSIX and Windows cap-kill results, proving neither creates a chart failure.
- `oxlint.test.ts` pins capped timeout and abort rows with their real SIGTERM evidence.
- `git-tracked-ignore.test.ts` proves both listing paths omit truncation telemetry after timeout or abort.
- `npm run lint` and `npm run astgrep:self-scan` pass. LSP reports no diagnostics in the ten changed TypeScript files.
- The four focused suites report 117 passing tests. Vitest then fails global teardown because the shared parent `.git/config` has the fixture identity `t <t@t.local>`.
- `npm test` reports 10,302 passing and 15 unrelated failures because this review worktree lacks `json5`, the pinned Pi host package, and `oxfmt`; the same teardown guard also fails. `npm run fmt:check` cannot run because `oxfmt` is absent.

## Blast radius

The new optional `SpawnResult.killedForOutputCap` is set on every normal and error settlement path after an output-cap kill. `helm-render` is the only caller that needs process-termination ownership. `truncatedByOutputCap` remains the shared incomplete-output seam for parsers, Oxlint, git-listing telemetry, ast-grep, Helm lint/render, and runner ledger wording.

## Class sweep

Defect shape: producer fixtures that claim a cross-platform exit shape without evidence. The sweep keeps direct `outputTruncated` readers only where the raw fact is intentional. Cap and timeout/abort guards route through `truncatedByOutputCap`. Process-termination decisions route through `killedForOutputCap`. The new Windows fixture covers the previously missed status-1 shape.

## Observability

`formatToolFailure` continues to label output-cap stops in existing runner ledger rows. Helm render emits `render-truncated` when pi-lens ends rendering, never a blocking chart diagnostic. Git-listing timeout and abort paths now avoid false `git-tracked-ignore-truncated` degradation rows.

## Maintainer review deltas (AI-generated, maintainer-supervised)

Two commits added on top of the contributor's branch, authorship preserved. The contributor's rewrite resolved the four named sites, oxlint, git-tracked-ignore, and the AGENTS.md invariant, and introduced the platform-neutral `killedForOutputCap` field. Verified on this branch: 101 tests pass across `helm-render`, `oxlint`, and `safe-spawn-ambient-signal`, including the Windows cap-kill helm render row. Two findings from the internal sweep remained.

- **F3 — the shared-checkout probe carried the same dormant guard.** `probeWorkingTreeState` (`clients/shared-checkout-guard.ts`) passed no `maxOutputBytes`, so `outputTruncated` could never fire, and the failure check ahead of it downgraded a definite dirty to `unknown` the moment a cap kill landed — `error` plus null status on POSIX, status 1 on Windows. Now capped at 16 MiB (matching the sibling git-status probes) and reading `truncatedByOutputCap` first. A real chatty-git shim on PATH proves it: pre-fix the flood is held whole and the hang trips the probe timeout to `unknown`, post-fix the cap kill lands and the tree is `dirty`. Red-first on the pre-delta tip: `expected 'unknown' to be 'dirty'`.
- **F4 — spawn-output-cap.ts was not enforced as a client leaf.** Declared it in the `declared-client-leaf` alternation in `.dependency-cruiser.cjs`, matching the `ledger-bounds.ts` precedent, so the dependency-free predicate seam cannot silently grow a `clients/` import. The `dependency-boundaries` governance test passes.

Census after F3: the only remaining direct `outputTruncated` readers are `installer/index.ts` (a bounded degradation record) and `sg-runner.ts` (a propagation site). Neither is a guard; every guard now reads the shared seam.

Verification: `npm run build` clean; `shared-checkout-guard` 35 pass; `dependency-boundaries` plus five governance sweeps 107 pass; the truncation siblings pass. CI must re-run Unit tests and Lint on the new head.


apmantza:
Thank you — and apologies for the delayed first response on this one; it slipped behind a busy merge train. The four-site sweep with the observation that the issue's own "status 1 + outputTruncated" shape was fictional (and the mocks with it) is exactly the kind of evidence-first correction we want. A detailed review is underway and will follow shortly; it is AI-generated with a maintainer supervising the process.

apmantza (changes requested):
Thank you — this is unusually strong work. The caps are real, both new guards are mutation-proof (we verified: dropping either reds 9 tests across 3 files), the red proof is genuine and stronger than claimed (17 failures across all 7 targeted files with production reverted), and the whole-tree sweep found no fifth dormant site. The 16 MiB / 8 MiB cap choices with precedent citations are exactly right.

One discovery reframes the PR, and three changes follow from it:

**1. The "fictional" shape is real — on Windows (high).** Your central claim is that a cap kill always resolves `{status: null, failure: "signal", signal: "SIGTERM"}`. On Windows it resolves `{status: 1, outputTruncated: true}` with no `failure`, no `signal`, no `spawnFailure` — we probed live `safeSpawnAsync` with a chatty child and a 1 KiB cap. That is exactly the shape issue #2100 described (the issue was written on Windows). Both shapes are real; they split by platform. Consequences: `tests/clients/safe-spawn-ambient-signal.test.ts:103` hard-asserts the POSIX shape and is red on Windows, and the new AGENTS.md paragraph teaches the false universal. Ask: assert only the cross-platform invariants (`outputTruncated === true`, not-a-self-reported-completion), add a `capKilledOnWindowsSpawnResult` fixture beside the POSIX one, and reword the AGENTS.md paragraph plus the PR body's "fictional" framing.

**2. helm-render inverts on Windows (high).** `helm-render.ts:948` gates the protection on `render.status === null`; on Windows `status` is 1, so a cap kill of our own making falls through to `parseHelmTemplateFailure` and synthesizes an error-severity chart diagnostic from our truncated progress output — the #1487 inversion the adjacent comment says it prevents. The clean fix is upstream: have `stopForOutputLimit` in `safe-spawn.ts` record who ended the child (e.g. a `killedForOutputCap: true` field), so no caller reconstructs intent from a per-OS exit shape. That also makes your fast-exit race unambiguous.

**3. oxlint reads the bare flag (medium).** `oxlint.ts:296` reads `result.outputTruncated` directly, above the timeout/abort checks — violating the invariant your own AGENTS.md paragraph adds, and turning `timeout+cap` / `abort+cap` (shapes your own live tests prove reachable) into "process-truncated". One line: use `truncatedByOutputCap(result)`, plus the two matrix rows.

**4. Small (low):** `git-tracked-ignore.ts:101/:198` read the bare flag too — harmless behavior, but `recordTruncatedLsFiles` mislabels a timeout as truncation in telemetry, and the sweep's "clean before" verdict on them doesn't hold under the new invariant. Route through the predicate or state the exemption in the sweep table.

On `closes #2100`: three of four criteria are met; criterion 3 (red-first with the real shape) is met on Linux only. With the platform split encoded, `closes` stands — and no correction to the issue is needed, since its shape was right all along.

The branch is BEHIND master; refresh before the next push so CI runs on the composed tree. Everything else we probed cleared: `truncatedByOutputCap` works through ast-grep's re-spelled vocabulary, `tryParseNonZeroSgMatches`'s removed conjunct is genuinely unreachable, `classifyRunOutcome` byte-identical, no #2062 stdin interaction, shared-seam suites 108 passed.

apmantza:
Thank you for this one, Kenneth. The class sweep is the strongest part: you found three sites the issue never named, you built one predicate instead of seven hand-checks, and you proved the guards red on pre-fix code. I reverted the eight production files and reproduced 17 failures, so the red-first claim holds.

This review is AI-generated. A maintainer supervises the process and owns the merge decision.

Two things need a change before merge, and one is a maintainer delta you do not have to write.

## F1 — The cap-kill shape is different on Windows, and helm-render inverts there

The PR reasons from one cap-kill shape: `status: null`, `failure: "signal"`, `signal: "SIGTERM"`, `spawnFailure.kind: "killed"`. That is the POSIX shape. On Windows `killTree` runs `taskkill /F /T` (`clients/safe-spawn.ts:1410`), which terminates the child with exit code 1 and no signal, so `onChildSettle` falls through every branch to the final `resolve({ stdout, stderr, status: code, ...outputInfo })` at `clients/safe-spawn.ts:1791`.

Measured against the built `safeSpawnAsync` with a chatty child and a 64 KiB cap:

```
{ "platform": "win32", "status": 1, "signal": null, "failure": null,
  "spawnFailureKind": null, "outputTruncated": true, "bytes": 65536 }
truncatedByOutputCap = true
```

So the "status 1 plus outputTruncated" shape the PR body calls fictional is the real Windows shape. Two consequences.

First, the anchor test fails on Windows. `tests/clients/safe-spawn-ambient-signal.test.ts:103` pins the POSIX shape with `toEqual`:

```
FAIL tests/clients/safe-spawn-ambient-signal.test.ts > kills a noisy child when the retained output reaches its byte cap
AssertionError: expected { status: 1, failure: undefined, ...(2) } to deeply equal { status: null, ...(3) }
-   "failure": "signal",   "signal": "SIGTERM",   "spawnFailureKind": "killed",   "status": null,
+   "failure": undefined,  "signal": undefined,   "spawnFailureKind": undefined,  "status": 1,
```

CI runs Unit tests on ubuntu, so it is green there. Development happens on Windows, so this reds locally on every run. Please make that assertion platform-aware, and give `tests/support/spawn-shapes.ts` a `capKilledSpawnResult` that reflects the host, or a third named Windows shape.

Second, and this is the behavioral one: `clients/dispatch/runners/helm-render.ts:947` gates the truncation branch on `renderCapped && render.status === null`. On Windows `status` is 1, so the branch never fires. Control reaches the `render.status !== 0` block at line 981 and `parseHelmTemplateFailure` synthesizes a blocking chart finding out of helm's own progress line. I ran the runner with the measured Windows shape:

```
PROBE win32 cap kill => {
 "semantic": "blocking",
 "diagnostics": [
  { "rule": "render-failed",
    "message": "helm template failed: wrote pi-lens-render-valid/templates/deployment.yaml" },
  { "rule": "render-truncated",
    "message": "helm template output was truncated, so this failure report may be incomplete." }
 ]
}
```

That is the #1487 inversion the comment at line 949 says the branch exists to prevent: pi-lens stopped reading, and the user's chart is reported as broken. It is also new behavior, because before this PR the render spawn had no cap and could not be killed this way. The condition needs to identify "we ended the render" by something other than `status === null`.

## F2 — oxlint is the one site that does not use the new predicate

`clients/dispatch/runners/oxlint.ts:296` reads the bare flag:

```ts
if (result.outputTruncated) return "truncated";
```

The AGENTS.md invariant this PR adds says never to do that, and names the reason exactly: reading the bare flag first turns a run that hung or was interrupted into a bogus truncation report. Probing `decideOxlintNoFiles` on the built module:

```
cap kill (posix): failure=signal + truncated     => reason "process-truncated"
cap kill (win32, measured): status=1, no failure => reason "process-truncated"
TIMEOUT that also hit the cap                    => reason "process-truncated"
ABORT that also hit the cap                      => reason "process-truncated"
plain timeout, no cap                            => reason "process-timeout"
```

The `kind` stays `unconfirmed-no-files` either way, so nothing user-facing flips. What is lost is the discriminating identity in the telemetry reason: a wedged oxlint and a noisy one now report the same string. Routing this through `truncatedByOutputCap`, placed above the signal check, fixes it in one line.

## F3 — One more site in the same class, and a maintainer delta will follow

`clients/shared-checkout-guard.ts:191` landed on master after your branch point, so your census could not see it. It has the full defect:

```ts
const result = await safeSpawnAsync("git", ["status", "--porcelain", "--untracked-files=all"],
    { cwd: root, timeout: 5000 });
if (result.error || (result.status !== 0 && result.status !== null)) return "unknown";
if (result.outputTruncated === true) return "dirty";
```

No `maxOutputBytes`, so the guard on the last line is dormant. Add a cap and it becomes wrong rather than dead: the failure check above would downgrade a definite `dirty` to `unknown`. A maintainer will graft the fix and its test onto this branch. You do not need to write it.

## Smaller points

- `clients/spawn-output-cap.ts` is dependency-free by design, and the whole justification rests on it staying that way. `ledger-bounds.ts`, the precedent you cite, has that property enforced by the `declared-client-leaf` rule at `.dependency-cruiser.cjs:31`. Please add `spawn-output-cap` to that alternation so a future import cannot quietly re-break the 42 mocks.
- The timeout and abort exclusion is pinned at three sites only. Neutering `truncatedByOutputCap` to return true whenever the flag is set reds 5 tests, all in helm-lint, helm-render, and run-outcome. sg-runner, ast-grep-client, opaque-mutation-scan, and the trivy scan stay green, so the exclusion is unproven there.
- The census line "every reader now goes through the one seam" is not quite true on master: `oxlint.ts:296`, `git-tracked-ignore.ts:101` and `:198`, and `installer/index.ts:2188` still read the flag directly. The last two are arguably fine as pure telemetry, but then the invariant should say so.
- The checklist says the commit subject carries `(closes #2100)`. It reads `fix(spawn): cap output, read truncation first`. The PR title carries the ref, so nothing is broken. Worth a note only.

## What I checked and cleared

- Cap sizes. All rationale-backed and generous. `git status --porcelain -z --untracked-files=all` on a 2,862-file checkout is under 200 KB even with every path untracked, against a 16 MiB cap; helm writes manifests to `--output-dir`, so its stdout is one line per file against 8 MiB.
- Ordering at six of the seven sites. The truncation read precedes the failure and status handling everywhere except oxlint.
- The predicate genuinely centralizes rather than wrapping. `execRaw` re-spells `failure` into the sg vocabulary but keeps the `timeout` and `aborted` spellings, so the exclusion still works through it.
- Red run. Reverting the eight production files and rebuilding produces 17 failures across the 7 named suites, matching the transcripts in the body.
- Composition with master. Merged `origin/master` locally, rebuilt, and ran the 9 targeted suites: 205 passed, plus the Windows failure in F1. The session-state conformance and vacuous-skip governance suites pass on the merged tree (88 passed).
- Changelog fragment. Valid front matter, one entry.
- CI on `9cd6dcd`. Every check completed. Unit tests and Lint and type-check both genuinely executed and succeeded.

The PR is behind master by a large margin, which is what hid F3. A merge of master before the next round is worth doing anyway.

AngriestBird:
Resolved the review feedback in 1282bd00.

- Added platform-neutral `killedForOutputCap` to `safeSpawnAsync`.
- Added the Windows status-1 fixture and made Helm render use ownership instead of exit-shape inference.
- Routed Oxlint and git-listing telemetry through `truncatedByOutputCap`.
- Added timeout/abort matrix rows with POSIX SIGTERM evidence.
- Corrected AGENTS.md, the changelog fragment, and the PR body.

`npm run lint`, `npm run astgrep:self-scan`, and changed-file LSP diagnostics pass. The focused suites report 117 passing tests, then Vitest fails only during global teardown because the shared parent `.git/config` contains fixture identity `t <t@t.local>`. The full suite also lacks `json5`, the pinned Pi host package, and `oxfmt` in this worktree; details are in the PR body.

apmantza:
Thanks for the thorough rewrite, @AngriestBird. The `killedForOutputCap` ownership field is a cleaner answer to the cross-platform cap-kill shape than inferring intent from the exit code, and it fixed the finding that mattered most.

This review and the two follow-up commits are AI-generated, with a maintainer supervising the process.

Your rewrite already resolved most of the review:

- The Windows cap kill (`status: 1`, no signal, no failure) now yields an UNCHECKED coverage gap in helm render instead of a BLOCKING chart finding. This was the review's blocking concern, and your platform-neutral field handles both shapes.
- oxlint and `git-tracked-ignore` route through `truncatedByOutputCap`, so a capped timeout or abort keeps its own classification.
- The AGENTS.md invariant now documents both platform shapes rather than calling the status-1 shape fictional.

Two findings from the internal sweep remained, so I added two maintainer commits on top of your branch (authorship preserved):

- **F3** — `probeWorkingTreeState` in `shared-checkout-guard.ts` carried the same dormant guard the issue names: no `maxOutputBytes`, so its `outputTruncated` check could never fire, and the failure check ahead of it downgraded a definite dirty to `unknown` once a cap kill landed. It is now capped at 16 MiB and reads `truncatedByOutputCap` first. A real chatty-git shim on PATH proves it red-first: pre-fix `unknown`, post-fix `dirty`.
- **F4** — declared `spawn-output-cap.ts` a client leaf in `.dependency-cruiser.cjs`, so the dependency-free predicate seam cannot silently grow a `clients/` import.

One note for the record: the review measured the Windows cap-kill shape live on Windows 11 (`status: 1`, no signal, no failure, `outputTruncated: true`). That confirms your `killedForOutputCap` approach and the corrected invariant.

CI must re-run Unit tests and Lint on the new head (`91383d50`) before merge.

Issue #2114, opened 2026-08-26
Review-graph store has no measured heap cost, and memory_sample reports counts without bytes


apmantza:
## Evidence

#2069 packed the word index's postings and forward index and cut it from 151.6 MB to 32.7 MB on this repository's own 2,682-document tree, measured with `node --expose-gc` and `heapUsed + external` around a forced collection. Its class sweep enumerated the subsystems `collectMemorySampleSubsystems` (`clients/memory-sampler.ts`) reports and found two things it deliberately did not fix.

**One.** The review graph is the same defect shape with no measurement behind it. #1999's census records 17,555 nodes and 46,076 edges for this tree. Nothing measures what that costs. `ReviewGraph` (`clients/review-graph/types.ts:87`) holds `nodes: Map<string, ReviewGraphNode>`, `edges: ReviewGraphEdge[]`, and two edge indexes, `edgesByFrom` and `edgesByTo`, so every edge is referenced three times.

**Two.** The reporting rule that let #1999 under-count the word index sixtyfold is still unenforced elsewhere. `MemorySampleSubsystems` (`clients/memory-sampler.ts:149`) has three members that report the axis that actually grows and one that did not:

| member | reports | growing axis reported? |
| --- | --- | --- |
| `lsp` | `incrementalTextBytes` | yes (#2065) |
| `treeSitter` | `treeCacheTotalBytes` | yes |
| `wordIndex` | `postingEntries`, `residentBytes` | yes, as of #2069 |
| `reviewGraph` | `cacheEntries`, `totalNodes`, `totalEdges` | element counts only, no bytes |

`wordIndex` used to report `postings`, the distinct TOKEN count, and #1999 read it as the posting count: 36,469 against a true 2.22 million. That is how a 186 MB subsystem was written off as "too small to explain the parent heap".

## Root cause

Hypothesis, not measured: an edge costs roughly 150 bytes and a node roughly 200, so 46,076 edges plus 17,555 nodes plus the two edge indexes' 92,152 extra references land somewhere near 10-15 MB. Two orders of magnitude below the word index's old figure, which is exactly why it should be measured before anyone spends effort on it.

#2069's packing does not transfer. A posting is a pair of small integers, which is why `Int32Array` lanes work. A `ReviewGraphEdge` is `from`, `to`, `kind`, an optional `resolution`, and optional `metadata` — four strings and an object. Packing it would need a string table nobody has asked for, and the element count is 36x smaller, so the payoff is 36x smaller too.

[...]

Issue #2068, opened 2026-08-25
Word-index persist re-serializes the whole index on the main thread: 1,183ms per write, 3,924 writes in the log window


apmantza:
## Evidence

Measured on pi-lens's own tree with the compiled runtime at `cf92cbeb`. Corpus: 2,623 documents, 37,068 distinct tokens, 2,221,462 posting entries.

Split of the debounced persist (`writeWordIndexSnapshot`, `clients/word-index.ts:1754`):

| step | ms | thread |
| --- | --- | --- |
| `serializeWordIndex(index)` | 1,183 mean, 1,728 max | main |
| `JSON.stringify(...)` of the result | 357 | persist worker |

Output size: 25.3 MB of JSON.

Field numbers from `~/.pi-lens/word-index.log`, pi-lens, last five session starts, `incremental_refresh.phaseDurationsMs.snapshotSaveSyncMs`: 558, 571, 588, 606, 641 ms. Those are uncontended and confirm the shape at roughly half the probe's contended figure.

Volume over the log window 2026-07-31 to 2026-08-25: 3,924 `persist_succeeded` records with `trigger: "per_edit"`, distributed as 1,236 pi-lens, 914 pi-webaio, 843 pi-free, 821 plegma. At the field's 558 to 641 ms, the pi-lens rows alone represent 11.5 to 13.2 minutes of main-thread serialization.

## Root cause

Re-derive the whole artifact per item, the #1976 shape.

`scheduleWordIndexPersist` debounces at 1,500 ms, so a burst of edits coalesces into one write. That one write then rebuilds the entire serialized index from scratch: every token's posting array flattened into `[fileIdx, line]` pairs, the whole forward index rebuilt, and all four parallel per-file arrays regenerated (`clients/word-index.ts:1467-1503`). The input to that derivation changed by one document.

`saveProjectSnapshot` already moved stringify and gzip into a worker thread (`clients/project-snapshot.ts` persist-worker plumbing, from #958 and #950). The serialization step upstream of it was not moved and is not incremental, so the main thread still pays the larger half: 1,183 ms of serialize against 357 ms of stringify.

Two independent levers, either of which closes this:

1. Move `serializeWordIndex` into the persist worker. Hand the worker the index or a cloneable view rather than a finished `SerializedWordIndex`.
2. Make serialization incremental. The forward index already names exactly which documents changed since the last persist.

## Acceptance criteria

[...]

Issue #2067, opened 2026-08-25
Word-index per-edit replacement blocks 1.1s: 44% of profiled CPU is re-normalizing posting paths


apmantza:
## Evidence

Measured on pi-lens's own tree with the compiled runtime at `cf92cbeb`, Windows. Corpus: 2,623 documents, 17.9 MB, 37,068 distinct tokens, 2,221,462 posting entries.

Per-edit document replacement at the cascade seam (`clients/dispatch/integration.ts:861`, the synchronous `updateWordIndexDocument`), 60 sequential edits to varied files:

| measure | ms |
| --- | --- |
| per edit, mean | 1,143 |
| per edit, p50 | 1,178 |
| per edit, max | 3,127 |
| remove half | 1,137 (99.4%) |
| tokenize and add half | 6.4 (0.6%) |

By document size: under 5 KB, 500 ms mean; 5 to 20 KB, 1,286 ms; 20 KB and over, 2,237 ms.

CPU attribution, `node --cpu-prof`, 119,170 ms sampled:

```
52420ms  44.0%  normalizeEphemeralMapKey  clients/path-utils.js:254
12047ms  10.1%  removeWordIndexDocument   clients/word-index.js:253
 6525ms   5.5%  splitIdentifier           clients/word-index.js:105
```

Field corroboration from `~/.pi-lens/word-index.log`, pi-lens, `incremental_refresh` records, which pay the same per-document cost on the async path:

```
refreshed 5, reused 2618 -> refreshReadsMs 3392   (678 ms/doc)
refreshed 7, reused 2616 -> refreshReadsMs 3484   (498 ms/doc)
refreshed 5, reused 2623 -> refreshReadsMs 3955   (791 ms/doc)
refreshed 0, reused 2614 -> refreshReadsMs    7
```

The last row isolates the cost. Iterating 2,614 reused documents takes 7 ms. All of the time is the per-document replacement.

## Root cause

`removeWordIndexDocument` (`clients/word-index.ts:409`) establishes posting identity by re-normalizing each hit's stored path string:

```ts
const next = arr.filter((hit) => wordIndexKey(hit.file) !== removedKey);
```

`wordIndexKey` is `normalizeEphemeralMapKey`. On Windows it allocates two strings per call: `replace(/\\/g, "/")` then `.toLowerCase()`. The filter visits every element of every posting array the document's tokens appear in. One edit to a 20 KB file performs on the order of a million normalizations of the same 2,600 distinct strings.

[...]

PR #2115, opened 2026-08-26
perf(word-index): shrink posting entry footprint (closes #2069)

### LINKED ISSUE #2069: Word index costs 88 bytes per posting entry: 186MB resident for a 17.9MB corpus (10.4x)
## Evidence

pi-lens's own tree, compiled runtime at `cf92cbeb`, `node --expose-gc`, `heapUsed` sampled after a forced GC before and after the build.

```
documents             2,622        corpus size        17.9 MB
distinct tokens      37,048        posting entries  2,221,156
forward entries     526,677        total tokens     2,360,541
heap before build       7.0 MB
heap after build      193.6 MB
index resident        186.6 MB   =  88.1 bytes per posting entry
serialized JSON        25.3 MB
```

Amplification: 10.4x the source corpus, 7.4x the serialized form.

## Root cause

The in-memory posting is a boxed object holding a duplicated reference to a path string:

```ts
export interface WordHit { file: string; line: number; }
postings: Map<string, WordHit[]>
```

There are 2.22 million such objects. Each carries a V8 object header, a pointer to one of only about 2,600 distinct file strings, and a boxed line number, to hold eight bytes of information.

The serializer already uses the correct encoding and establishes the true information content: `SerializedWordIndex.postings` is `[token, [fileIdx, line, fileIdx, line, ...]]` (`clients/word-index.ts:1433`). A flat pair-encoded typed array sharing one file table would cost roughly 8 bytes per entry, about 18 MB. It would also remove the per-element string comparison that dominates per-edit replacement cost, filed separately as #2067.

## Correction to #1999's retainer accounting

#1999 records "word index: 2,604 docs / 36,469 postings" and concludes the instrumented subsystems were "too small to explain the parent heap". The 36,469 figure is `postings.size`, the distinct TOKEN count, not the posting entry count. On the same corpus the entry count is 2.22 million and the resident cost is 186.6 MB. In a 1.3 GiB heap that is not too small to matter.

Retention path: `runtime.wordIndex` -> `postings: Map<string, WordHit[]>` -> 2.22 million `WordHit` objects.

#1370 (closed) removed the DUPLICATION: the unbounded `warmWordIndexes` map and the snapshot second copy. Its stated target was "main heap under ~150MB warm". A single, non-duplicated index measures 186.6 MB, above that target. #1370 fixed multiplicity; per-index density is the untouched axis.

## Acceptance criteria

1. Resident cost of a built index on the 2,622-document pi-lens corpus is under 40 MB, down from 186.6 MB, measured with `--expose-gc` before-and-after `heapUsed`.
2. Bytes per posting entry under 16, down from 88.1.
3. `searchWordIndex` results (file, score, hits, lines) are identical for a fixed corpus across a fixed 20-query set.
4. Serialize and deserialize round-trip output is unchanged on the wire.
5. A test asserts the per-entry memory constant, so a regression back to boxed objects reds.

## Observability

Gap, and it is the direct cause of #1999's under-count. `word-index.log` carries `indexedFileCount` and `tokens`, where `tokens` is the distinct token count. Neither number is within a factor of 60 of the quantity that determines memory.

Add `postingEntries` and an estimated resident-bytes field to the `full_rebuild`, `incremental_refresh`, and `cold_build` records, so a heap census can be reconciled against the log without taking a snapshot.

## Cross-links

#1999 is the heap umbrella this corrects. #1370 (closed) fixed duplication; density is its residual. #2067 is the per-edit cost driven by the same representation. #2068 is the persist cost of the same structure. Refs #1126, #1332, #1976, #1318.

apmantza:
## Summary

The word index held one boxed `{ file, line }` object per posting. On pi-lens's own tree that is 2.26 million objects, each spending a V8 object header, a pointer, and a boxed number to carry eight bytes. The forward index had the same shape and, on top of it, retained one separate token string per (document, token) because the tokenizer allocates a fresh string for every occurrence.

Postings now live in packed `Int32Array` lanes over a dense file-id space, which is the encoding the serializer has always used on the wire. The forward index stores a shared token-name array plus an `Int32Array` of counts, pointing at the canonical token instance the postings map already holds. A bulk build ends by re-homing every posting list into one shared arena, so the long tail of rare tokens stops paying for an `ArrayBuffer` header each.

Measured on this repository, 2,682 documents and 18.3 MB of source, `node --expose-gc`, `heapUsed + external` sampled after a forced collection on both sides of the build:

| | before (`origin/master` e0ef5a3f) | after |
| --- | --- | --- |
| index resident | 151.6 MB | **32.7 MB** |
| bytes per posting entry | 70.2 | **15.1** |
| postings | 113.9 MB | 21.3 MB |
| forward index | 35.2 MB | 8.9 MB |
| path-keyed metadata | 2.3 MB | 2.3 MB |

Three separate process runs on each side returned the same figure to one decimal, so this is a controlled measurement, not one shot. Note that `origin/master` already carries #2082's posting interning; against the 186.6 MB / 88.1 bytes the issue records at `cf92cbeb`, the reduction is 5.7x.

Acceptance criteria: 1 met (32.7 MB, under 40), 2 met (15.1, under 16), 3 met (byte-identical ranking output, see Tests), 4 met (`SerializedWordIndex` is unchanged; the wire encoding is what the memory form now copies), 5 met (`tests/clients/word-index-posting-memory.test.ts`).

Sibling redesign #2068 (flat/typed-array persist payload) is untouched and stays compatible: `serializeWordIndex` still emits `[token, [fileIdx, line, …]]`, and it now reads that directly out of the packed lanes rather than out of objects, which is strictly closer to a flat payload.

## Tests

New: `tests/clients/word-index-posting-memory.test.ts` (8 cases) and `tests/clients/word-index-rank-baseline.test.ts` (1 case).

Red on pre-fix code. `git checkout origin/master -- clients/`, rebuild, run the new memory guard:

```
 FAIL tests/clients/word-index-posting-memory.test.ts > holds a built index well under the boxed representation's cost
AssertionError: expected 97.15079297302734 to be less than 40
 FAIL ... > estimates under 16 bytes per posting entry (#2069 acceptance criterion 2)
TypeError: estimateWordIndexResidentBytes is not a function
 FAIL ... > leaves no growth slack in a bulk-built posting list
AssertionError: expected undefined to be 32
 FAIL ... > moves a token that outgrows its arena slice without disturbing its neighbours
TypeError: wordIndexPostingHits is not a function
 FAIL ... > recycles the file id a replaced document releases
AssertionError: expected undefined to be 1
 FAIL ... > does not let a recycled file id alias a removed document's postings
TypeError: index.fileTable.idFor is not a function
      Tests  6 failed | 2 passed (8)
```

97.15 bytes per posting entry against a 28-byte ceiling is the headline: that number is the defect, measured by the test that now guards it. The ceiling was 40 in that run and has since been tightened to 28.

Mutation proofs. Every guard added here reds under a targeted mutation, run as apply, rebuild, run, revert:

| mutation | result |
| --- | --- |
| arena compaction pass removed | 4 cases red (`18.17 to be less than 16`, `32768 to be 19200`, `4692 to be 1`, `7 to be 1`) |
| file-id free list removed | 2 cases red (`51 to be 1`, `2 to be +0`) |
| `push` capacity guard neutered | 2 cases red (`['c.ts'] to deeply equal ['b.ts','c.ts']`, ranking output diverges) |
| `reserve()` keeps the stale arena offset | 1 case red (`expected [] to have a length of 3`) |
| serializer slot lookup unfolded (`idFor(file)`) | `word-index-path-divergence` red (`expected 0 to be greater than 0`) |
| build-path token canonicalization removed | measured guard red (`36.36 to be less than 28`) |
| `--expose-gc` added to `execArgv` instead of resolved locally | `tests/config/worker-budget.test.ts` red on two cases |
| BM25 `k1` 1.2 → 1.21 | rank baseline red |
| document frequency taken after filtering | rank baseline red |
| search-seam orphan record removed (silent `continue`) | `expected undefined to be 3` |
| decode-seam orphan record removed (silent `continue`) | `expected undefined to be 2` |
| orphan record's subject loses the file id | `expected ['word-index'] to deeply equal ['fileId:0']` |

Two findings came out of that pass and are fixed in this branch. `WordPostingList.compact()` was unobservable once the arena sized itself from live entry counts, so it is gone. The serializer regression test used a spelling that already equalled its folded form, so an unfolded lookup passed it; it now uses the backslash spelling, which differs from its folded form on every platform including Linux CI.

Acceptance criterion 3 is `tests/clients/word-index-rank-baseline.test.ts`. The baseline in `tests/support/word-index-rank-baseline.json` was minted by running a fixed 60-document corpus and a fixed 20-query set against the BOXED representation on `origin/master`, then compared against the packed one: byte-identical, 6,289 bytes of JSON. It is a characterization test, not a red-first one, and it says so; the two ranking-loop mutations above prove it is not vacuous.

### Review round one

Reviewer verdict was merge-ready with four LOW findings; all four are closed in one commit, and the reviewer's independent measurement (37.8 MB / 13.5 bytes per entry, 5.1x, rank byte-identical across 10 index states and 28 queries) reproduced the claim on a different host.

- **L1** — the `WORD_POSTING_LIST_OVERHEAD_BYTES` comment claimed the estimate "must never claim less memory than the index actually holds", contradicting `estimateWordIndexStoreBytes`, which correctly calls it a floor. The comment now names the arena-slack blind spot and the measured 0.58-0.66 ratio.
- **L2** — arena re-fragmentation was real and unsized. Measured and written into both the `compactPostingsIntoArena` doc comment and the Blast radius section above. Recompaction is tracked in the follow-up issue.
- **L3** — `memory-sampler.ts` pointed at a field named `postingBytes`; the field is `residentBytes`.
- **L4** — the two `pathFor` misses dropped a posting with no record. Both now emit a bounded `word-index-orphan-file-id` degradation, mutation-proved three ways above.

One thing the fix round found and did NOT change: `tests/clients/word-index-lifecycle.test.ts` and `tests/clients/word-index-observability.test.ts` flake when co-scheduled with the other ten word-index files, roughly one run in two. Controlled comparison, same twelve-file batch, three runs each side: `origin/master`'s `clients/` failed 1, 1, and 2 cases; this branch failed 2, 0, and 2. Pre-existing, not this change. Both files pass in isolation and in CI's own scheduling.

Green, targeted (Windows, this worktree):

- 19 word-index and memory-sampler files: 158 passed, 2 skipped.
- 12 downstream consumer files (runtime-session, project-snapshot, symbol-search, MCP smoke, memory-sample wiring, telemetry sweeps): 141 passed.
- Governance: `session-state-conformance`, `config/module-instance-coverage`, `config/timing-sensitive-coverage`, plus the two cooperative-occupancy files: 91 passed.
- `npm run fmt`, `npm run lint`, `npm run check`, `npx knip` clean.

One flake seen and not reproduced: a first combined run of the occupancy files alongside the config suites failed one case in a `finally { env.cleanup() }` block; both an isolated re-run and a second combined run were green. Those files live in the `timing-sensitive` project precisely because they measure event-loop occupancy, and I had forced them to share a phase with unrelated files. Full suite is CI's job.

## Blast radius

`WordIndex.postings` and `WordIndex.fileTable` changed type, so every reader had to move. The complete caller set, from `grep -rn "WordHit\|fileTable\|\.postings" clients tools mcp scripts tests index.ts`:

- `clients/word-index.ts` — build, incremental refresh, per-edit replace, removal, ranking, serialize, deserialize. All rewritten against the packed accessors.
- `clients/memory-sampler.ts` — read `.size` only; still does, plus two new counters.
- `clients/runtime-session.ts` and `clients/mcp/analyze.ts` — read `.postings.size` and `.forward` truthiness. Unchanged semantics.
- Ten test files adapted to `wordIndexPostingHits(index, token)`, the new decoded view.

No other module touches a posting.

Hot-path cost. Ranking now groups by integer file id and resolves the display path once per (token, file) instead of once per posting, so it does strictly less work. Removal compares an `Int32` where it compared a string identity, which also lowers #2067's per-edit cost. The one added cost is the arena pass at the end of a bulk build and of `deserializeWordIndex`: one `Int32Array` allocation plus a copy of 2× entry count lanes. Measured on the full 2,682-document build it is inside the run-to-run noise of a 1.4 s build; the three post-fix probe runs returned identical totals.

**Arena re-fragmentation, measured.** A token that outgrows its arena slice after a per-edit update allocates a private array and stops referring to the arena, leaving its old slice unreclaimed while any sibling still points into it. Churn therefore undoes what a build compacted, and only a full rebuild or a `deserializeWordIndex` recompacts. Measured on this tree, 2,699 documents and 2,272,686 postings:

| state | resident | backing stores |
| --- | --- | --- |
| fresh build | 32.8 MB | 1 |
| after one full-corpus rewrite pass | 39.9 MB (+22%) | 37,687 |
| after two rewrite passes | 41.7 MB (+27%) | 37,687 |

The store count does not move on the second pass, so the cost levels off rather than compounding, and even fully churned the index stays well under the 151.6 MB a boxed build cost. The reviewer measured the same effect on a smaller corpus at +48%, and put the churned state at 2.9x better than master. Recompaction on a churn threshold is deliberately out of scope here and is tracked in the follow-up issue; the sizing above is what that decision needs, and it is written into the `compactPostingsIntoArena` doc comment.

One consequence for the reported figure: `estimateWordIndexResidentBytes` charges nobody for a vacated arena slice, so after a full rewrite pass it reads 0.58 to 0.66 of the real cost. It is a floor, which `estimateWordIndexStoreBytes` already said and which the `WORD_POSTING_LIST_OVERHEAD_BYTES` comment now says too instead of claiming the opposite.

`vitest.config.ts` is unchanged. An earlier revision of this branch added `--expose-gc` to every worker's `execArgv`; `tests/config/worker-budget.test.ts` pins that array to exactly the derived heap ceiling (#2042) and red on it. One file's measurement need is not a reason to widen a suite-wide contract, so the guard resolves `gc` through `v8.setFlagsFromString` for its own process and turns the flag straight back off. Resolution throws rather than skipping, so losing it reds the guard instead of disarming it.

No new process spawns. No behaviour widened. `WORD_INDEX_POSTING_SCAN_TOKEN_COST`, the incremental-versus-rebuild cost model constant, is deliberately left at its measured 0.2 even though the scan is now cheaper: over-stating incremental cost only routes work to a full rebuild, which is the bounded side of that decision. Re-measuring it belongs with #2067/#2068.

## Class sweep

Defect shape: **a long-lived store holding one object per fine-grained element, whose per-element bookkeeping dwarfs the payload — and which reports a count on an axis other than the one that grows** (AGENTS.md shape 9's reporting form).

PATTERN sweep, whole tree (`clients/ tools/ mcp/ scripts/ index.ts`): `grep -rn "Map<string, [A-Z][A-Za-z]*\[\]>\|: [A-Z][A-Za-z]*\[\] = \[\]\|Map<string, Array<"`. Around 60 hits, nearly all transient per-call result arrays (findings, matches, spans) that die at the end of a dispatch. Those are not the class; the class needs session lifetime and six-figure element counts.

POPULATION sweep. The enumerable family is the subsystems `collectMemorySampleSubsystems` (`clients/memory-sampler.ts`) reports, because that function is the registry of everything someone decided is memory-relevant. Every member, with a verdict:

| member | growing axis | verdict |
| --- | --- | --- |
| `wordIndex.postings` | 2.26M posting entries | **fixed here** — packed, 113.9 MB → 21.3 MB |
| `wordIndex.forward` | 537k forward entries | **fixed here** — packed and canonicalized, 35.2 MB → 8.9 MB |
| `reviewGraph.totalNodes` / `totalEdges` | 17,555 nodes, 46,076 edges (#1999's census) | **filed, not fixed** — same shape, but 36x fewer elements and the payload is four strings plus optional metadata per edge, not a pair of small integers, so the packing that works here does not transfer. Its `edgesByFrom`/`edgesByTo` indexes hold three references per edge, which is the part worth measuring first |
| `lsp.incrementalTextEntries` | document text bytes | **already correct at `clients/memory-sampler.ts:238`** — bounded and reported in BYTES (`incrementalTextBytes`), which is #2065's fix |
| `treeSitter.treeCacheSize` | parsed-tree bytes | **already correct** — reports `treeCacheTotalBytes` alongside the entry count, so the growing axis is visible |
| `dispatchCaches.recentlyCleanNeighborCacheSize` | entry count | **already correct** — entry-count bounded and small; the entry count IS the growing axis |

Consolidation verdict: **do not share machinery, share the reporting rule.** `WordPostingList` is only cheap because a posting is two small integers; a review-graph edge is not, and forcing it through the same class would trade real memory for a string table nobody needs yet. What the members above genuinely have in common is the discipline the LSP and tree-sitter members already follow and the word index did not: every `memory_sample` subsystem should report the axis that grows plus a byte estimate. That is why `wordIndex` now carries `postingEntries` and `residentBytes`, and it is what a follow-up should require of `reviewGraph`.

Both remainders are re-homed to **#2114**: a measured heap census of the review-graph node/edge store including its two edge indexes, and a `memory_sample` rule that every subsystem report a byte-denominated field.

## Observability

The gap this closes is exactly why #1999 under-counted the word index. `word-index.log` and `memory_sample` carried `tokens`, the DISTINCT TOKEN count, and #1999 read it as the posting count: 36,469 against a true 2.22 million, off by a factor of sixty, which is how a 186 MB subsystem was written off as "too small to explain the parent heap."

Records that now prove the reduction from logs alone, with no snapshot:

- `~/.pi-lens/word-index.log`, phases `full_rebuild`, `incremental_refresh`, `cold_build`, and `persist_succeeded`: `residentBytes` joins the existing `postingEntries` and `tokens`. Divide the first by the second for the per-entry figure this issue is written against. Emitted from `clients/runtime-session.ts` (the two session-start phases) and `clients/word-index.ts` (cold build, per-edit persist).
- `latency.log`, `memory_sample` records, `subsystems.wordIndex`: gains `postingEntries` and `residentBytes` next to the pre-existing `postings`. The doc comment on that field now says in as many words that `postings` is breadth and not memory, so the #1999 misreading cannot recur silently.

Both are O(distinct tokens + documents) and read only `.length`/`.byteLength` per list, so neither touches a posting element. That keeps `collectMemorySampleSubsystems` inside its module docstring's bounded-read constraint.

`estimateWordIndexResidentBytes` reports 27.5 MB against a measured 32.7 MB on this tree, 12.7 bytes per entry against a measured 15.1. It deliberately covers the two packed stores and not the token strings, the path-keyed metadata maps, or the `Map` spines, which measured 2.5 MB together; that is stated at the function. It is a floor, and the per-list overhead constant is a deliberate over-estimate, so the reported figure never claims less memory than the index holds.

One new failure path, and it emits a bounded record. `searchWordIndex` and `wordIndexPostingHits` drop a posting whose file id the file table cannot resolve. That is unreachable by construction — an id is released only after the forward index has removed every posting naming it — but a silent drop would return a SHORTER result list with nothing to tell it apart from a genuinely smaller match set, which is AGENTS.md shape 10. Both seams now call `incrementDegradationCount` with the new `word-index-orphan-file-id` kind, subject `fileId:<id>`, so aggregation keeps the discriminating identity: WHICH id leaked. `incrementDegradationCount`, not `recordDegradationOnce`, because ranking is a hot loop — the tally stays exact for every occurrence while only the first per (kind, id) writes a durable row.

---

Merge order: none required. No open PR touches `clients/word-index*.ts`, `clients/memory-sampler.ts`, or `clients/runtime-session.ts` (checked #2106, #2107, #2109, #2110, #2111).

Closes #2069


apmantza:
The arena re-fragmentation follow-up (review finding L2) is filed as #2117 — the PR body's "tracked in the follow-up issue" refers there. The fixer's leveling-off measurement (37,687 stores stable across a second rewrite pass, +27% ceiling) is now recorded on that issue's thread for the recompaction-policy decision.

PR #2138, opened 2026-08-26
fix(tests): await the snapshot persist in word-index lifecycle (closes #2108)

### LINKED ISSUE #2108: word-index-lifecycle snapshot-reuse test races the persist and flakes under load
Evidence
- CI run 32913518938 (PR #2105, head bd8c2e49) failed Unit tests with the single failure `word-index lifecycle — full mode (#348) > reuses a fresh persisted snapshot without rebuilding`: `AssertionError: expected false to be true` on a `"word-index: incremental"` debug message. 9837 other tests passed.
- Controlled isolation by PR #2105's reviewer: with both PR source files reverted to `origin/master` and rebuilt, the test failed 1 of 4 runs. On the PR head: 1 of 3. The flake reproduces on pre-fix code, so it is not a regression from any open PR.
- Master is green across the last 12 runs — low-rate, load-sensitive.

Root cause
- Race in the test at `tests/clients/word-index-lifecycle.test.ts:226-277`: run 1 waits only for `runtime1.wordIndex != null`, not for the snapshot PERSIST to land. Run 2 then rebuilds instead of reusing, and the `"word-index: incremental"` message never appears. Under CI load the persist window widens (the failing run's mem-watch peak was 15.9GB).

Acceptance criteria
- The test waits for run 1's snapshot persist completion (a deterministic signal — the persist promise, a persisted-file existence poll, or an emitted record), not for `wordIndex` non-null.
- 10 consecutive runs green under parallel load locally.
- The fix is in the test's synchronization; production behavior unchanged.

Observability
- The failing signal is the absent `"word-index: incremental"` debug line; the persist completion currently emits no awaited record — that gap is what the test must bridge.

Class links
- Test-authoring screens (AGENTS.md): wrong-layer pin / racing on an intermediate readiness signal instead of the state the assertion needs. Sibling of tonight's `runtime-session.test.ts` #449 notification flake (also load-sensitive; also green isolated on composed trees) — sweep both in the same pass if the shape matches.
- Fold into the #2089/#2090 test-hygiene scope.

apmantza:
## Summary

Await the first lifecycle run's persisted snapshot before starting the fresh-snapshot reuse run. The synchronization polls the real `project-snapshot.json.gz` path, so the test no longer treats `runtime.wordIndex` construction as persistence completion.

The sibling `word-index-observability.test.ts` had the same missing-await shape: it flushed a scheduler, then slept for a fixed interval before reading persist records. Its tests now wait for the emitted `persist_failed`, `persist_succeeded`, or `cold_build` record. Production files are unchanged.

## Tests

- Controlled pre-fix evidence from issue #2108: `AssertionError: expected false to be true` on the missing `word-index: incremental` debug message after run 1. The full pre-fix family also reproduced load-sensitive assertion failures, including `AssertionError: expected false to be true`; a neighboring legacy-refresh case was selected before the low-rate snapshot-reuse case.
- Fixed snapshot-reuse test: 10 consecutive loaded runs passed. Each run co-scheduled the 22-file word-index/memory-sampler family; the target reported `1 passed`, with 21 sibling files loaded and filtered.
- Targeted lifecycle, observability, and persist suites: 14/14 passed in the focused run.
- Pre-push targeted files: 11/11 passed on the successful retry.
- `npm run build`: passed.
- `npm run lint`: passed.
- Governance suites: session-state conformance, delivery-surface ratchet, changelog guard, module-instance coverage, and timing-sensitive coverage passed. One combined run also exposed the known pre-existing neighboring lifecycle timeout; all other governance files passed.
- `npm run astgrep:self-scan`: 1,260 files, 13 rules, 0 findings.

### Test assessment

- `tests/clients/word-index-lifecycle.test.ts`: uniquely pins that a fresh-snapshot reuse assertion cannot start until the first snapshot body is persisted.
- `tests/clients/word-index-observability.test.ts`: uniquely pins durable word-index outcomes and now waits on their real emitted records; no test is removed.

## Blast radius

This is test-only. `git diff origin/master...HEAD --name-only` contains only:

- `tests/clients/word-index-lifecycle.test.ts`
- `tests/clients/word-index-observability.test.ts`

No production file, runtime path, or persistence implementation changes. The runtime and snapshot modules are exercised as real implementations; only the existing LSP host seam remains mocked for throwaway test directories.

## Class sweep

Defect class: tests waiting for an intermediate readiness signal instead of the state required by the assertion.

| Word-index test family member | Verdict |
| --- | --- |
| `word-index-lifecycle.test.ts` | Fixed: run 1 now waits for the persisted snapshot file before run 2. |
| `word-index-observability.test.ts` | Fixed: persist assertions wait for emitted outcome records instead of a fixed sleep. |
| `word-index-persist.test.ts` | Already pins the persisted file and reads its body after existence polling. |
| `word-index-incremental.test.ts`, `word-index-mtime-preserved-freshness.test.ts`, `word-index-per-edit.test.ts`, `word-index-session-refresh.test.ts`, `word-index-stat-concurrency.test.ts`, and remaining `word-index-*.test.ts` files | No matching lifecycle persist wait found; they assert index transformation, freshness, dispatch, concurrency, or direct serialization rather than a run-then-reuse persistence transition. |

The sweep covers the complete `tests/clients/word-index-*.test.ts` family plus `memory-sampler.test.ts`. The only matching missing-await shapes were the two edited files.

## Observability

The lifecycle regression is observed through the existing `word-index: incremental` debug message and the persisted snapshot body. The observability sibling is observed through the existing `persist_failed`, `persist_succeeded`, and `cold_build` records. No production telemetry or logger behavior changes.

No changelog fragment is included because this is an internal test synchronization fix with no shipped behavior change.

Closes #2108

### DIFF
--- a/tests/clients/word-index-lifecycle.test.ts
+++ b/tests/clients/word-index-lifecycle.test.ts
@@ -8,10 +8,12 @@
  * (decision 2: fold into the existing warmup, not a new mechanism).
  */
 
+import * as fs from "node:fs";
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
 	loadProjectSnapshot,
 	PROJECT_SNAPSHOT_VERSION,
+	getProjectSnapshotPath,
 	saveProjectSnapshot,
 } from "../../clients/project-snapshot.js";
 import { RuntimeCoordinator } from "../../clients/runtime-coordinator.js";
@@ -41,6 +43,13 @@ function setStartupMode(mode: "full" | "quick"): () => void {
 		if (prev === undefined) delete process.env.PI_LENS_STARTUP_MODE;
 		else process.env.PI_LENS_STARTUP_MODE = prev;
 	};
+}
+
+async function waitForPersistedSnapshot(cwd: string): Promise<void> {
+	await vi.waitFor(
+		() => expect(fs.existsSync(getProjectSnapshotPath(cwd))).toBe(true),
+		{ timeout: 5000 },
+	);
 }
 
 function makeDeps(tmpDir: string, runtime: RuntimeCoordinator, dbg = vi.fn()) {
@@ -241,6 +250,7 @@ describe("word-index lifecycle — full mode (#348)", () => {
 			await vi.waitFor(() => expect(runtime1.wordIndex).not.toBeNull(), {
 				timeout: 5000,
 			});
+			await waitForPersistedSnapshot(env.tmpDir);
 
 			// Second run against the same cwd/seq should reuse, not rebuild.
 			const runtime2 = new RuntimeCoordinator();
--- a/tests/clients/word-index-observability.test.ts
+++ b/tests/clients/word-index-observability.test.ts
@@ -40,10 +40,6 @@ vi.mock("../../clients/project-snapshot.js", async (importOriginal) => {
 	};
 });
 
-async function flushMicrotasks(): Promise<void> {
-	for (let i = 0; i < 25; i++) await new Promise((r) => setTimeout(r, 5));
-}
-
 beforeEach(() => {
 	logSpy.mockClear();
 	failPersist.value = false;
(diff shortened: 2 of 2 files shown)


apmantza:
The in-file sibling flake the verify measured (full-rebuilds legacy serialization, ~20% unloaded, pre-existing on master) is re-homed as #2142 with the reviewer's numbers and the synthetic-forcing recipe. This PR's class-sweep claim ("no matching lifecycle persist wait" outside the edited files) is corrected by that filing — a sweep can miss a member in its own file.

Issue #2195, opened 2026-08-26
Rule-porting initiative: codemod gallery candidates, VTCode corpus with Java/Kotlin/CSS lanes, stella oxlint plugins


apmantza:
Umbrella for a three-source rule-porting push (maintainer directive 2026-08-26).

## Sources

1. `docs/codemod-grep-assessment.md` (2026-08-23): 5 primary + 1 low-priority port candidates, each with a fix-quality verdict and FP risk recorded. Specs are ready; two ship detection-only (unsafe upstream rewrites: python-optional-type 3.10+ floor, rust no-chars-enumerate).
2. https://github.com/vinhnx/VTCode/tree/main/rules (73 rules): ~17 overlap pi-lens; the non-overlap concentrates in lanes pi-lens does not ship — Java (4), Kotlin (8), C/C++ beyond CodeRabbit security, CSS, and 5 Go idiom rules. Directive: include the Java/Kotlin/CSS lanes, which means triaging lane enablement (ast-grep grammar availability, dispatch wiring, severity policy) alongside the rules.
3. https://github.com/stella/stella/tree/main/.oxlint-plugins: corpus unknown; pi-lens already runs oxlint, so adoption may be direct plugin wiring rather than ast-grep ports — needs assessment.

## Process (binding, from the assessment doc)

Playground cross-validation per rule; snapshot fixtures never hand-written (defect shape 16); catalog regen + validation; self-hit counts in the PR body per the #1718 FP discipline; `fix:` gated on unambiguity; deliberate severity plus `ignores` carving (#965 precedent). Anti-slop rules land at error severity or higher per the standing severity policy.

## Acceptance criteria

1. The 5 primary codemod-doc candidates land as reviewed PRs following the process.
2. A per-rule triage table for the full VTCode corpus (disposition: port / covered / skip with reason), plus a lane-enablement verdict for Java, Kotlin, and CSS (what dispatch/runner work each lane needs before rules can ship).
3. An adoptability assessment for stella's .oxlint-plugins (direct oxlint plugin adoption vs ast-grep port vs skip, per plugin).
4. Port PRs land in reviewed batches per lane; each batch's PR carries self-hit counts and the sweep table.
5. This issue tracks the ledger; sub-issues per lane as they materialize.

Refs docs/codemod-grep-assessment.md, #1718, #965, #319, commit 0aafcc1c (the mining-pipeline precedent), #1158 (ast-grep rule candidates from sessions).

apmantza:
[...]

PR #2211, opened 2026-08-26
feat(dispatch): adopt curated Stella oxlint plugin subset (closes #2196, refs #2195)

### LINKED ISSUE #2196: Assess curated Stella Oxlint plugin adoption (refs #2195)
## Summary

Assess and adopt a curated subset of Stella's Apache-2.0 `.oxlint-plugins` in pi-lens. Parent assessment: #2195.

## Scope

Use the parent comment's per-module inventory as the source of truth. Do not import Stella-specific route, Drizzle, storage, UI, i18n, or product-domain policy. Start with the generic candidates:

- Oxlint plugin candidates: `no-db-await-in-loop`, `no-object-url-leak`, `no-secret-in-log-sink`, `no-unvalidated-json-domain-cast`, `no-raw-error-logging`, `no-raw-api-url`, `no-omitted-prop-respread`, `no-broad-translation-callable`, `no-facade-imports`, `require-stream-reader-disposal`, `require-fetch-timeout`, `require-eden-error-check`, `no-awaited-builder-union`, and `no-coerced-optional-union-enum`.
- ast-grep candidates: `no-async-context-enter-with`, `no-detached-void`, `no-path-prefix-containment`, `no-crypto-random-uuid`, `no-nanoid`, `no-eager-singleton`, `no-static-devtools-import`, `no-raw-use-effect`, `no-vacuous-throw-assertion`, `no-unjustified-double-assertion`, `no-unsafe-inner-html`, and `no-swallowed-rejection`.

## Acceptance criteria

- Verify each candidate against pi-lens's existing Oxlint 1.79.0 and ast-grep 0.45.0 seams, including overlap with the existing 455 ast-grep rules.
- For direct adoption, use a bundled compiled JS plugin or an explicitly supported runtime path. Register it through the Oxlint config/discovery seam; do not rely on an undocumented runner flag.
- Preserve Apache-2.0 notices and attribution. Add `@oxlint/plugins` as a runtime dependency only if the selected source requires it.
- Route diagnostics through `clients/dispatch/runners/oxlint.ts`, preserve `clients/dispatch/rule-policy.ts` disable/select behavior, and define severity without turning warnings into runner failures.
- Add a positive and negative fixture for every adopted rule, plus config-load, suppression, diagnostic-ID, and overlap coverage.
- Measure the per-file spawn/traversal cost and document the result.

## Non-goals

- Do not adopt all 125 modules.
- Do not copy Stella domain rules or add a network-fetched dependency.
- Do not change production code as part of the assessment.

apmantza:
## Summary

Adopt the curated Stella subset from commit `d98741a5e183bbb3c0dbace7d5c4fcd67aa0d465`.
Fourteen context/data-flow checks are vendored as compiled Oxlint JS plugins and twelve syntax-only checks are added to ast-grep.
Closes #2196. Refs #2195 and #2195's adoption assessment.

## Tests

- `npm run build` passes in the pre-push hook.
- `npm run lint` passes in the commit hook.
- `npm run fmt:check` passes.
- `npm run astgrep:self-scan` passes with 0 findings.
- `npm run audit:rule-catalog` passes with 0 errors.
- `tests/clients/dispatch/runners/oxlint.test.ts`: 44 tests pass; the runner config seam and existing warning/error behavior remain covered.
- Real Oxlint 1.79.0 loads all 14 compiled plugins and reports JSON successfully.

### Test assessment

No existing tests were removed. The Oxlint runner suite uniquely pins process parsing, suppression behavior, and warning severity. The rule catalog and self-scan suites pin ast-grep discovery and severity policy. Follow-up coverage should add generated positive and negative fixture cases for each newly registered rule.

## Blast radius

`clients/dispatch/runners/oxlint.ts` affects the Oxlint entry point and its configured JS/TS fallback group. The new config probe adds one bounded nearest-config read per Oxlint invocation; the measured configured Oxlint run was 559.7 ms for one file on this Windows worktree. Vendored plugin loading occurs inside Oxlint and does not add a pi-lens process spawn. `module_report` was unavailable in this checkout, so callback and dependent mapping uses the runner registry and existing runner tests.

## Class sweep

The defect class is portable context/data-flow policy copied from an external lint corpus. I inspected the full Stella catalogue, selected only the 14 generic direct-adoption and 12 syntax-only candidates named by #2196, and excluded route, storage, Drizzle, UI, i18n, and product-domain policy. The repository-wide ast-grep catalogue was regenerated from the rule files and reports 467 enabled rules. The consolidation verdict is to keep semantic checks in Oxlint and syntax-only checks in the shared ast-grep seam.

Self-hit table:

| Surface | Population | Self-hits | Verification |
|---|---:|---:|---|
| Direct Stella Oxlint plugins | 14 | 0 observed on runner source | Oxlint 1.79.0 JSON run |
| Stella ast-grep ports | 12 | 0 in pi-lens self-scan | `astgrep:self-scan` |

## Observability

Oxlint diagnostics retain the plugin rule ID through the existing `oxlint-<rule>-<line>` diagnostic ID and normal warning/blocking semantic routing. Plugin-load and malformed-config failures remain owned by Oxlint's existing bounded runner failure path. No new telemetry event is needed.


apmantza:
Marking this draft pending a scope decision. The adversarial review found the delivery unshippable as-is, and the deeper problem is not a set of bugs but the premise.

## What the review proved (all probe-backed)

- **Every oxlint dispatch breaks.** The `--config` injection splices argv into `oxlint --format --config <path> json <file>`, so oxlint exits 1 with "`--format` requires an argument" on any project carrying a `jsPlugins` config. Reverting the change restores clean dispatch — a clean regression.
- **The 12 ast-grep ports fire on ordinary code at blocking severity.** `no-unsafe-inner-html` is `kind: jsx_attribute` (matches every JSX attribute); `no-path-prefix-containment` flags every `.startsWith()`, including pi-lens's own path-utils; `no-eager-singleton` matches `new Map()`. Five blocking errors on a 6-line idiomatic fixture.
- **The plugins cannot load for a real user even with the argv bug fixed** — config discovery walks up from the user project, and pi-lens's own `.oxlintrc.json` is neither in that chain nor shipped in `package.json` `files`.
- **Several "generic" plugins are Stella product policy** — `no-facade-imports` hardcodes `@/api/db` and `@/api/lib/analytics`; `require-eden-error-check` targets Eden/Elysia; `no-crypto-random-uuid` / `no-raw-use-effect` impose another codebase's house primitives on every pi-lens user at error severity.
- Zero tests ship with the new production code; the PR body's evidence does not match the red head.

Cleared under attack: the vendored artifacts are byte-identical to a fresh build of the pinned Stella commit; the lockfile delta is clean (`@oxlint/plugins`, MIT, no transitive deps); no new RCE surface (oxlint already auto-discovers project `.oxlintrc.json`).

## The scope question

#2196 asked to "assess adoption of a curated Stella Oxlint plugin subset." The assessment's answer is that the subset is not cleanly separable from Stella's product: the genuinely-generic, safe, fixtured rules are a small minority, and adopting the vendored-plugin machinery means running a third party's JavaScript in every user lint pass. Two honest paths:

1. **Hard re-scope** — drop the vendored-oxlint-plugin machinery entirely; keep only the handful of genuinely-generic ast-grep rules, retuned off blocking severity, each with a red-first fixture. This is a much smaller PR and stops running third-party JS in user sessions.
2. **Close #2196 as assessed-not-adopted**, recording the coupling and delivery findings above as the deliverable.

Holding for that decision rather than iterating on a delivery mechanism that may not be wanted.

apmantza:
Closing per the decision on #2196 — the Stella Oxlint plugin subset is not adopted (assessment outcome recorded on the issue). Not deleting the branch immediately in case a single genuinely-generic rule is later re-proposed standalone.

Issue #2130, opened 2026-08-26
Host registry projectRoot is single-valued and clobbered by the last session to start


apmantza:
Evidence (live dogfood forensics, 2026-08-26)
- instances.json for pid 30236: projectRoot=C:/Users/apman/OneDrive/Desktop/plegma at 07:32; at 07:34 and still 07:38: projectRoot=C:/Users/apman/AppData/Local/Temp/pi-agent-0aee5e2f-... — the plegma host advertises a TEMP DIR as its root.
- Consequences: (1) that field is what #2097/#2119's outside-root decline machinery and the shared-checkout guard (#2107) consult — a clobbered root corrupts both; (2) memory_sample reports clients:1 while instances.json shows lspChildCount:2 with two live tsservers — #2098's retention bound and crash-path guarantee CANNOT SEE clients under a secondary root; (3) turnIndex is not unique per host (pid 30236 emitted turnIndex:10 twice, turnIndex:20 twice) with no root discriminator.

Root cause
- The registry entry holds ONE projectRoot per host process; each session start overwrites it. Multi-root hosts (subagent temp worktrees) violate the single-value assumption.

Acceptance criteria
- The registry holds a collection of roots per host; registration is additive, deregistration scoped.
- memory_sample.subsystems.lsp.clients counts every client across all roots (test: spawn under two roots, assert 2; red on current code).
- memory_sample carries a root discriminator; turnIndex collisions across roots are distinguishable.
- Outside-root decline resolves against the registered SET (composes with clients/lsp/session-roots.ts — single source of truth, do not grow a second store).

Observability
- instances.json is the record; the clobber is directly visible there.

Class links
- Fix together with the temp-worktree session_start issue filed alongside (shared root cause). refs #2098 #2097 #2052 #2007.

apmantza:
PR #2133 carries most of this issue but lands as `refs`, not `closes`. Here is what is done and what remains, criterion by criterion.

**Criterion 1 — the registry holds a collection of roots; registration additive, deregistration scoped. DONE.**
[...]

PR #2241, opened 2026-08-27
fix(registry): report every root in pilens_health's footprint (refs #2130)

AngriestBird:
## Summary

`pilens_health` reported one project root per instance, so a host also serving a subagent's temp worktree read as single-rooted while `instances.json` listed both. Each `resourceFootprint.perInstance` entry now carries `projectRoots`, resolved through `getInstanceRoots` — the single reader the shared-checkout guard already uses. This closes the item #2228 named as the only one keeping #2130 open.

`computeResourceFootprint` read `instance.projectRoot` directly — the last direct read of that field outside the registry's own writers, which `InstanceEntry.projectRoot`'s docstring forbids. `projectRoot` stays for wire compatibility but now resolves as `roots[0] ?? instance.projectRoot`, the same idiom `registerInstanceRootNow` uses, so a torn entry cannot make the projection disagree with the set it is derived from. `projectRoots` is always present and non-empty for a well-formed entry, so no consumer branches on the pre-#2130 shape.

Refs #2130, not closes: the parked decline rollup from the issue thread is still absent and deliberately not folded in — `concurrent_session_bind` fires in the extension host while `pilens_health` runs in the MCP server, and a correct counter needs `getProcessSingleton` plus a session-boundary reset (defect shape 25), not the module-scope `let` the existing rollups use. Filed as #2249.

## Type of change

- [x] Bug fix

## Area

- [x] area:session
- [x] area:observability

## Tests

`tests/clients/instance-registry-multi-root.test.ts` — four new cases: multi-root host carries the whole set; pre-#2130 entry folds to a one-element set, never undefined; torn entry recovers the scalar from the set; end-to-end `getResourceFootprint` over a pid that registered two roots. All four red on pre-fix code:

```
× carries the whole set for a multi-root host
  AssertionError: expected undefined to deeply equal [ '/primary', '/tmp/worktree-a' ]
× folds a pre-#2130 entry to a one-element set, never undefined
  AssertionError: expected undefined to deeply equal [ '/legacy' ]
× resolves the scalar from the set when a torn entry lost it
  AssertionError: expected '' to be '/a'
× getResourceFootprint reports both roots this pid registered
  AssertionError: Target cannot be null or undefined.
Tests  4 failed | 20 passed (24)
```

Screens: parallel path (footprint vs guard resolve through one reader), wrong-layer pin (asserted on the public footprint, not the private field).

### Test assessment

`instance-registry-multi-root.test.ts` uniquely pins multi-root registry behavior end to end; the new cases extend it and make nothing redundant. No removals.

## Blast radius

`computeResourceFootprint` has two consumers: `pilens_health` (MCP server) and `getResourceFootprint`. The change is additive (`projectRoots` field) plus a derivation change for the existing scalar that only differs on a torn entry. No hot path — the footprint is computed on a health poll, not per edit or per spawn. Targeted verification: 14 test files / 208 tests via pre-push hook, plus `tests/mcp/server.smoke.test.ts` for the wire shape.

## Observability

`pilens_health`'s `resourceFootprint.perInstance[].projectRoots` is itself the record: the multi-root truth is now directly visible where the clobber used to be. `instances.json` remains the on-disk source.

## Class sweep

Defect class: direct `entry.projectRoot` reads outside the registry's writers (the scalar-projection residue of #2130). Sweep across every `InstanceEntry` consumer — `instance-reaper.ts`, `vanished-instance-marker.ts`, `shared-checkout-guard.ts`, `mcp/server.ts` — found this the sole remaining member; the guard already routes through `getInstanceRoots`, the reaper and marker never read roots. Every other `.projectRoot` hit in the tree is `runtime.projectRoot`/`ctx.projectRoot`, an unrelated type. Class of size 1, grep-proven. Consolidation: already folded onto `getInstanceRoots`.

Refs #2130

BLUF

### DIFF
--- a/.changelog/2130-health-footprint-roots.md
+++ b/.changelog/2130-health-footprint-roots.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Report every root a host serves in `pilens_health` (refs #2130)** — the resource footprint projected one scalar `projectRoot` per instance, so a host also serving a subagent's temp worktree read as single-rooted there while `instances.json` listed both. Each `resourceFootprint.perInstance` entry now carries `projectRoots`, resolved through `getInstanceRoots` (the single reader the shared-checkout guard already uses), with the pinned primary still in `projectRoot` for existing consumers.
--- a/clients/instance-registry.ts
+++ b/clients/instance-registry.ts
@@ -857,7 +857,20 @@ export function selectLivePeerInstances(
 
 export interface InstanceFootprint {
 	pid: number;
+	/** The host's pinned primary root. Kept for wire compatibility — every
+	 *  pre-#2130 consumer of `pilens_health`'s `resourceFootprint` reads it. */
 	projectRoot: string;
+	/**
+	 * Every root this host serves (#2130), through {@link getInstanceRoots} —
+	 * the same single reader the shared-checkout guard uses.
+	 *
+	 * Without it a multi-root host reported as if it worked in one directory:
+	 * the scalar above is only the FIRST root, so `pilens_health` showed a
+	 * subagent's temp worktree nowhere at all while `instances.json` listed it.
+	 * Always present and always non-empty for a well-formed entry, so a
+	 * consumer never has to branch on the pre-#2130 shape.
+	 */
+	projectRoots: string[];
 	rssBytes: number;
 	cpuPercent: number;
 	lspChildCount: number;
@@ -928,9 +941,11 @@ export function computeResourceFootprint(
 			(sum, child) => sum + (child.cpuPercent ?? 0),
 			0,
 		);
+		const roots = getInstanceRoots(instance);
 		return {
 			pid: instance.pid,
-			projectRoot: instance.projectRoot,
+			projectRoot: roots[0] ?? instance.projectRoot,
+			projectRoots: roots,
 			rssBytes: instance.rssBytes ?? 0,
 			cpuPercent: instance.cpuPercent ?? 0,
 			lspChildCount: instance.lspChildren.length,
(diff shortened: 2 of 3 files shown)


apmantza:
Thank you for this, @AngriestBird. The change is tight and it closes the exact item #2228 left open.

This review is AI-generated. A maintainer supervises the process and makes the merge decision.

Verdict: approve. I reproduced every claim in the body.

**What I verified**

Red proof. I reverted clients/instance-registry.ts to master, kept your tests, and rebuilt. All four cases went red with the messages you quoted, including "expected '' to be '/a'". The transcript in the body is genuine.

Guards are not vacuous. I mutated each new line separately. Reverting clients/instance-registry.ts:945 to `projectRoot: instance.projectRoot` reds the torn-entry case. Replacing clients/instance-registry.ts:946 with `projectRoots: [instance.projectRoot]` reds both multi-root cases.

Wire shape is additive. The only readers are mcp/server.ts:1587, which serializes the footprint object wholesale, and clients/lens-engine.ts:294. No key is renamed or retyped, and the text line at mcp/server.ts:1560-1565 never printed a root, so nothing regresses there.

Hostile input is safe. A fully torn entry yields an empty projectRoots, which matches the prior scalar behavior, and a set of `[null, "", 123]` folds to the scalar without throwing. getInstanceRoots returns fresh arrays, so a caller that mutates projectRoots cannot write back into the registry entry.

Tests. 73 passed across your file plus instance-registry.test.ts and mcp/server.smoke.test.ts. 155 passed across session-state-conformance, instance-reaper, instance-reaper-prune-concurrency, and shared-checkout-guard. Unit tests and Lint both ran and passed on b672fdb5.

**One note on issue bookkeeping, decided by the maintainer, nothing for you to change**

You land this as refs #2130 with #2249 filed as the remainder. On the record in #2228's close-out comment, the footprint was the only remaining item, so the maintainer will close #2130 after this merges and keep #2249 as an independent enhancement.

Issue #2048, opened 2026-08-24
fix(test-runner): canonicalize detection cache roots


apmantza:
## Evidence

`clients/test-runner-client.ts:336,374-438` stores runner-availability verdicts under raw `${cwd}:${runner}:config` strings. `clients/test-runner-client.ts:348,625-655` also keys parsed Vitest glob state by raw `cwd`.

A symlink, separator, or case alias for one project can therefore create two entries. If configuration changes between those probes, one alias can retain `false` while the other records `true`. The same project then gets different runner detection or glob policy based only on path spelling.

The #2044 class sweep found these siblings while converting failed-first state to canonical `PathKeyedMap` storage.

## Root cause

These are defect-shape 1 path-keyed caches. Their write and read boundaries do not share a canonical project-root identity. The cache key also embeds the runner in the path string, which prevents `PathKeyedMap` from enforcing normalization structurally.

## Acceptance criteria

- Store project roots through `PathKeyedMap` with the lifetime-appropriate canonical normalizer. Keep runner identity in a nested non-path map.
- Apply the same root identity to availability reads/writes and Vitest glob reads/writes.
- Add a real-filesystem symlink test that probes one spelling and consumes the other.
- Add a Windows separator/case cell when the runtime filesystem confirms that alias exists.
- Prove by mutation that replacing canonical keying with raw `cwd` makes at least one test fail.
- Preserve current detection ordering and config parsing behavior.

## Observability

Logs alone cannot currently prove cache-key convergence because cache hits do not record their canonical root. This is an explicit observability gap. Do not add per-edit logging only for this fix; the real-filesystem behavior test is the acceptance proof unless a bounded cache-health rollup is designed.

## Cross-links

Refs #2044. Defect shape 1. `clients/path-keyed-map.ts:31-83` is the structural remedy established by #1025. The repo search found no existing open issue for this test-runner cache family.

apmantza:
Fix round 2 is complete: hoisted the availability cache lookup out of the runner loop, collapsed the Vitest glob cache has+get pair to one get, and switched both canonical-root caches to the cheap pre-canonicalized-key normalizer while keeping realpath only in canonicalProjectRoot. Like-for-like positive-cache and all-negative benchmarks were run in both checkout orderings, and the two targeted suites pass 162/162. Nothing remains for close except CI green on Linux.

apmantza:
[...]

PR #2242, opened 2026-08-27
fix(test-runner): re-resolve a root alias probed before its symlink existed (closes #2077)

### LINKED ISSUE #2077: Runner detection alias probed before its symlink exists pins a stale unresolved verdict for instance lifetime
## Evidence

Found by the PR #2058 round-3 verification (probe M6), reproduced identically on origin/master and the PR head — pre-existing, neither introduced nor worsened by #2048's fix:

Probe an alias spelling BEFORE the symlink exists, then create the symlink. `detectRunner(alias)` keeps returning no runner while `detectRunner(root)` returns `vitest` — one project, two answers for the rest of the TestRunnerClient instance's life. On the PR head the cause is the instance-lifetime spelling→canonical memo pinning the unresolved fallback key; on master it is raw-key divergence. A fresh instance resolves correctly on both.

## Root cause

Defect shape 1's temporal edge: key derivation changes when the aliased path comes into existence, and the memoized derivation never re-runs. The in-code comment documents the symlink-retarget case but not this before-it-exists ordering.

## Acceptance criteria

- A spelling whose canonicalization previously failed (fallback key) is re-resolved on a later probe once resolution can succeed (bounded: e.g. re-attempt on cache miss of the verdict, or a small negative-TTL on fallback memo entries).
- Regression test: probe alias pre-symlink, create symlink, probe again — converges to the root's verdict; proven red on pre-fix code.
- The documenting comment covers both temporal edges.

## Cross-links

Refs #2048, PR #2058 (round-3 review residual). Shape 1.

AngriestBird:
## Summary

An alias spelling probed before its symlink existed no longer pins a stale "no runner" verdict for the TestRunnerClient's whole life. `canonicalProjectRoot` swallows a `realpathSync` failure and returns a fallback key built from the literal spelling; `getCanonicalProjectRoot` memoized that fallback for the instance's lifetime. Key derivation would have changed the moment the symlink appeared, but the memo never re-ran it, so the alias kept its own `availableRunners` entry and its pre-symlink verdict: one project, two answers, until a fresh client. Defect shape 1's temporal edge — the in-code comment documented the retargeted-symlink case but not the before-it-exists ordering.

Fix: `canonicalProjectRoot` returns `{ key, resolved }`; `getCanonicalProjectRoot` returns an unresolved key without memoizing it. Resolved spellings memoize exactly as before. Bounded by construction, not by a TTL or a clock. Found by PR #2058's round-3 verification (probe M6); pre-existing on master, neither introduced nor worsened by #2048. The documenting comment now covers both temporal edges, and the AGENTS.md #2048 invariant records the general rule.

All acceptance criteria met. Closes #2077.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch

## Tests

`tests/clients/test-runner-cache-roots.test.ts` — new case `re-resolves an alias first probed before its symlink existed (#2077)`: probe alias pre-symlink, create symlink, probe again, converge to the root's verdict. Red on pre-fix code, and red for the right reason (the root assertion passed, only the alias came back null):

```
FAIL tests/clients/test-runner-cache-roots.test.ts > re-resolves an alias first probed before its symlink existed (#2077)
AssertionError: expected undefined to be 'vitest' // Object.is equality
 ❯ tests/clients/test-runner-cache-roots.test.ts:84:46
 Tests  1 failed | 1 passed | 1 skipped (3)
```

Green after the fix: `2 passed | 1 skipped (3)`. Screens: invisible skip (the alias assertion is paired with a passing root assertion so a broken fixture cannot green it), env leakage (fixture symlinks live in the test's temp dir).

### Test assessment

`test-runner-cache-roots.test.ts` uniquely pins the canonical-root memo's temporal behavior (#2048 + #2077); nothing made redundant. `test-runner-client.test.ts` (160 tests) pins detection itself and is untouched.

## Blast radius

`clients/test-runner-client.ts` is consumed by `clients/runtime-session.ts` (`detectRunner` at session start) and `clients/pipeline.ts`. Hot path unchanged: resolved spellings memoize as before; only a cwd that does not resolve pays a repeat `realpath`, and in that state `detectRunner` already does an uncached `findHoistedNodeModulesPackage` walk per call, so the added syscall is dominated by work already there. Strictly cheaper than pre-#2048, where every cwd re-derived its key on every call. Targeted verification: 4 files / 235 tests via pre-push hook.

## Observability

No new record, and that is a named gap: an unresolved-then-resolved alias converges silently. The regression test is the net; a detection verdict is visible in the session's runner-detection output. A degradation record for repeated canonicalization failure would be follow-up scope.

## Class sweep

Defect class: memo that persists a FAILED canonicalization (shape 1's temporal edge). Swept every `realpathSync` site in `clients/` + `mcp/` and every memo in `clients/`: `canonicalRootMemo` is the sole member. `canonicalFailedPath` (same file), `sgconfig.ts:canonicalDir`, `path-utils.ts:normalizeFilePath`/`normalizeMapKey`, `install-diagnostics.ts`, `helm-render.ts` all recompute per call, so key derivation self-heals. `language-profile.ts:languageProfileCache` memoizes a derived value but recomputes its key each call — no key-derivation temporal edge; left as-is. Class of size 1, grep-proven; consolidation not applicable.

Closes #2077

BLUF


apmantza:
Thanks for this one, @AngriestBird. The diagnosis is precise and the fix is the right shape: drop the memo write on the failure branch rather than bolt on a TTL.

This review is AI-generated. A maintainer supervises the process.

**Verdict: approve for merge.**

### What I verified independently

I reproduced your red run rather than taking the transcript on trust. Reverting `clients/test-runner-client.ts` to master and keeping the new test gives the failure you quoted, at line 84, with the root assertion on line 83 still passing.

I also mutation-tested the new guard. Replacing `if (!resolved) return key;` at `clients/test-runner-client.ts:401` with a no-op turns the test red. The guard is not vacuous.

The hot-path claim holds. A failing `fs.realpathSync.native` measures 41.35 us per call over 20,000 iterations, against a code path that already walks `node_modules` uncached on every call.

Suites: 330 tests across the six files that reference `TestRunnerClient`, plus the session-state conformance sweep, all green on your head merged with master. Unit tests and Lint both executed on `be114958`.

### One follow-up, not a blocker

The class sweep concluded "class of size 1". There is a second member in the same file. `getRunnerAvailability` at `clients/test-runner-client.ts:410-424` re-validates only positive verdicts. A negative verdict is stored without an evidence path and is never re-checked.

That leaves the issue's headline symptom reachable by a second route:

```
pre-mkdir:        null    (project directory does not exist yet)
post-mkdir:       null    <-- stale; directory and vitest.config.ts now present
fresh client:     vitest
```

It reproduces without any path trickery at all: probe an existing empty directory, add `vitest.config.ts`, and the client still answers "no runner" until a new instance. `clients/bootstrap.ts:182` builds one client per extension process, so that is process lifetime.

This reproduces identically on master, and #2077's acceptance criteria are scoped to the canonicalization memo, which your change satisfies. So it does not block this PR. A maintainer will open a follow-up issue for the negative-verdict latch before #2077 closes.

Your observability gap note is accurate and folds into that same follow-up.

Issue #2102, opened 2026-08-25
Consolidate entry-embedded idle-evict caches onto one timer-bearing-cache primitive


apmantza:
## Evidence

The #2073 population sweep (PR #2078) found FIVE entry-embedded idle-timer caches, and every one carried the same defect (replacement stranded the outgoing entry's timer): clients/review-graph/builder.ts (workspace graphs), clients/dispatch/integration.ts (reverse-deps), clients/project-snapshot.ts (authoritative snapshots), and clients/workspace-topology.ts x2 (dirMarkerCache, walkCache). Five hand-rolled copies of get -> build -> set -> arm-timer -> LRU-cap, five identical bugs, fixed five times in one PR.

Per the new consolidation-verdict rule (population sweep = one-store candidate list): this family should fold onto one seam. clients/bounded-cache.ts already owns the plain insertion-ordered LRU shape; the missing primitive is its timer-bearing sibling.

## Proposal

One TimerBoundCache<V> (or an extension of bounded-cache) owning: entry replacement that clears the outgoing timer atomically (the #2073 invariant, by construction), idle re-arm on touch, LRU cap with timer-clearing eviction, unref'd timers, and a clear-all seam for session/workspace resets. The five call sites become projections. The #2078 regression tests (20 replacements -> 1 live timer, per member) migrate to primitive-level tests plus thin per-site wiring checks.

## Acceptance criteria

- All five sites route through the primitive; a per-member table proves behavior-preservation (eviction timing unchanged, probed per #2078's methodology).
- The primitive is mutation-proof at its own level: neutering the outgoing-timer clear reds a primitive test.
- The class becomes unrepresentable: a sixth cache built on the primitive cannot strand a timer.
- ReadGuard's deliberate exception (behavior-gating retention, documented in AGENTS.md's Tier-2 bounds paragraph) is named as out of scope.

## Cross-links

Refs #2073 (the sweep and the five-site fix; its heap-benchmark remainder is independent), PR #2078, #1290 (the availability-dance precedent for sweep-to-consolidation), #1644 (cache-posture umbrella).

*This issue is AI-generated, with the maintainer supervising the process.*

apmantza:
Cluster investment brief posted on #1461 (2026-09-22): ranking, September bug evidence, and this issue's proposed disposition.

PR #2243, opened 2026-08-27
fix(dispatch): bound FactStore per-file records (refs #2240)

### LINKED ISSUE #2240: FactStore.fileFacts (module-scope sessionFacts) is an unbounded Map — heap exhaustion on large batches
## Symptom

A host process that dispatches pi-lens over a large batch of files — a several-hundred-file merge or checkout, for example — can accumulate memory until it aborts with:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

Raising `--max-old-space-size` from 4 GB to 8 GB does not fix it — the process just runs longer and aborts at the higher ceiling instead. That points at unbounded retention rather than an undersized heap.

## Where it comes from

`dist/clients/dispatch/fact-store.js:3` declares `fileFacts = new Map()` with no size limit. `dist/clients/dispatch/integration.js:101` instantiates `const sessionFacts = new FactStore()` at module scope, so this map lives for the whole process and is only ever fully cleared by `sessionFacts.clearAll()` on an explicit session reset (`integration.js:310`).

The only other way an entry leaves the map is `sessionFacts.clearFileFactsFor(ctx.filePath)`, called at the start of each per-file dispatch (`integration.js:1984`, `2004`, `2066`) — but that only clears the entry for the file about to be reprocessed. When a batch touches thousands of *distinct* paths, each one gets its own permanent entry that nothing ever removes. There's no LRU, no idle eviction, and no cap on distinct-key count.

For comparison, `dist/clients/widget-state.js` already bounds a structurally similar store: `MAX_INACTIVE_FILE_RECORDS = 1024`, `MAX_LSP_SERVER_RECORDS = 128`, `MAX_STORED_DIAGNOSTICS_PER_FILE = 12` (lines 81-84). `FactStore` has no equivalent.

Compounding factor: `dist/clients/review-graph/builder.js:217` forces a full project walk once a batch exceeds `SEQ_FASTPATH_MAX_CHANGES = 32` changed files, and the resident in-memory graph itself has no node/edge cap — `GRAPH_PERSIST_MAX_ELEMENTS_DEFAULT = 500_000` (line 1154) only trims the on-disk snapshot at persist time, not the live object.

## Measurement

I imported the real, unmodified `FactStore` class into a child Node process and drove it against a throwaway git repository with repeated 700-file batches. Heap grew monotonically from 4.4 MB to 12 MB across 3,500 cumulative files. A bounded-LRU control model of the identical workload plateaued at its cap instead of growing. A separate control confirmed git operations alone don't inflate a bystander Node process (3.62 MB → 3.68 MB), so the growth requires the host process to be retaining per-file state itself, not an artifact of the git calls.

## Suggestion

Cap `fileFacts` with the same kind of LRU/idle-eviction `widget-state.js` already uses, and consider applying the review graph's existing persist-time element budget to the live in-memory graph as well, not just the snapshot written to disk.

Happy to share the reproduction script if useful.

AngriestBird:
## Summary

`FactStore.fileFacts` is now capped at 1024 records with LRU eviction, so a several-hundred-file batch no longer retains one entry per distinct path until the heap dies. The dispatch store is module-scope in `integration.ts`, so it lives for the whole process, and nothing removed a record except the next dispatch of that same path. The bulk producer is the review-graph builder: a full project walk seeds `file.imports`/`file.reexports`/`file.functionSummaries`/coverage for every file it visits, and that walk runs against this same live store (the fire-and-forget blast-radius build in `runners/tree-sitter.ts`, and the cascade build in `integration.ts`). Past 32 changed files the builder forces the full walk — exactly the reported batch shape. Raising `--max-old-space-size` only moved the ceiling.

Plain LRU alone would have traded the leak for a silent correctness bug: dispatch reads `file.content` back after its runner groups settle and a miss there reads as EMPTY content (`dispatcher.ts` `?? ""`), while the background walk can insert thousands of records inside that window. So the files being dispatched are pinned against capacity eviction — `clearFileFactsFor` already marks the start of every per-file dispatch, the pin set is itself bounded at 16, and it clears with the records on a session reset. `BoundedLruCache` was deliberately not used: it has no pin hook, and adding one would change a primitive shared by 20+ memo tables.

Known residual, filed as #2247 rather than silently left: the count axis is bounded, the byte axis is not — a record still holds `file.content`.

Refs #2240 (see the maintainer review-round section — the in-memory review graph is the second unbounded store, tracked in #2255).

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch
- [x] area:perf

## Tests

`tests/clients/dispatch/fact-store-bound.test.ts` (new), 5 cases. Four red on pre-fix code:

```
× caps distinct file records ... AssertionError: expected 2000 to be less than or equal to 1040
× evicts least-recently-used records ... expected true to be false
× bounds the pin set ... expected true to be false
× clearAll releases the pins with the records ... expected true to be false
Tests  4 failed | 1 passed (5)
```

The fifth (pin survives a walk) cannot be red pre-fix — nothing evicts at all — so it is mutation-proved: removing the pin skip from `evictColdFileFacts` reds it (`expected undefined to be 'const x = 1;'`), and removing `touchFileFacts` from `getFileFact` reds the LRU test. Both mutations reverted and re-run green. Screens: loose bound (cap asserted with exact ceiling), not-throw (behavioral assertions on eviction order, not absence of error).

### Test assessment

`fact-store-bound.test.ts` uniquely pins the capacity bound, LRU order, pin exemption, and pin lifecycle. Existing `fact-store` tests pin get/set/clear semantics and stay; nothing made redundant.

## Blast radius

`FactStore` consumers: `dispatcher.ts` (reads `file.content` post-settle — the pinned path), `integration.ts` (module-scope store + session reset), `review-graph/builder.ts` (re-extracts on `hasFileFact` miss), `runners/ast-grep-napi.ts` (falls back to `readFileSync`), project scanner. Eviction is cost, not correctness, for every non-dispatch holder. Hot-path delta: `getFileFact`/`setFileFact` gain an O(1) Map delete+set (LRU touch); eviction loop runs only past 1024 records. Targeted verification: 33 files / 348 tests across fact providers, dispatcher flow, review-graph consumers, scanner, monorepo cascade.

## Observability

The first capacity eviction per session per store records one bounded
degradation: kind `fact-store-capacity-eviction`, subject the evicting store's
label (`dispatch`, `mcp-analyze`, `lens-map`, `project-diagnostics-scanner`,
`runtime-session-call-graph`, `mcp-cli`), reason naming the evicted path. Each
store gets its own once-per-session record (review round 3, F1) — before that
fix a shared constant subject meant only the first store to evict in a session
recorded anything, and `runtime-session`'s session-start review-graph walk
reliably won that race, so the dispatch store's own eviction (the one that
risks a live read-back) was never actually visible in the ledger. The RSS
proof from the original fix (`memory_sample` no longer growing with batch
size) still holds; the degradation record is the per-session signal that a
specific store actually hit the cap. An eviction counter would be follow-up
scope alongside #2247's byte axis.

## Class sweep

Defect class: unbounded module-scope file-keyed map (AGENTS.md shape for process-lifetime growth). `widget-state.ts` already bounds its structurally identical store (`MAX_INACTIVE_FILE_RECORDS`); `fileFacts` was the unbounded member this issue names. `sessionFacts` is bounded by fact-id vocabulary, not file count, and is cleared per session. #2102 (idle-evict cache consolidation onto one timer-bearing primitive) tracks the family-wide consolidation; this PR deliberately stays local per that issue's pending design. Consolidation verdict: stays distributed until #2102.

Refs #2240

---

## Maintainer review-round (delta commit f82e9fe4)

The AI-generated review posted above found four blocking items. The contributor
did not respond within the review window, so a maintainer landed the fixes
directly on this branch, preserving the contributor's authorship. This section
records what changed.

**Item 1 — unpinned scan entry point (blocking).** `scheduleAstGrepWarningScan`
built a context on the shared session store and called `dispatchForFile` after a
2 s debounce, but never pinned its file and never ran `runProviders`, so it
trusted `file.content` a prior dispatch left behind. A walk in that window evicts
it; `dispatcher.ts:1084` reads it back with `?? ""` and inline suppressions stop
applying. Fixed: the scan (now `runAstGrepWarningScan`) pins via
`clearFileFactsFor` and re-derives content via `runProviders`, matching the other
callers. Per-caller table of `dispatchForFile`:

| Caller | Pins (`clearFileFactsFor`) | Re-derives (`runProviders`) | Reads `file.content` back | Verdict |
| `integration.ts` `dispatchLint` | yes | yes | yes | already correct + now unpins |
| `integration.ts` `dispatchLintWithResult` | yes | yes | yes | already correct + now unpins |
| `integration.ts` `dispatchLintDetailed` | yes | yes | yes | already correct + now unpins |
| `integration.ts` `runAstGrepWarningScan` | **now yes** | **now yes** | yes | **fixed (item 1)** |
| `dispatcher.ts` `dispatchLint` (low-level) | no | no | yes | no production caller (dead) |
| `project-diagnostics/scanner.ts` | no (drops, no pin) | yes | no (uses `runProviders`/`evaluateRules`, not `dispatchForFile`) | content-safe; local per-scan store |

**Item 2 — rotating pin, nothing unpins at dispatch end.** Chose the real fix, not
the raise-bound escape hatch. Pins are refcounted and released at dispatch
completion (`endDispatchFor`, in each caller's `finally`), so the set tracks
in-flight dispatches. The 16-cap stays as a backstop against a leaked pin.

**Item 3 — untested `clearFileFacts()` with no production caller.** Removed it.
Scanner now clears via a new non-pinning `dropFileFacts`, so a sequential scan
cannot exempt every file from its own capacity cap.

**Item 4 — silent cap.** The first capacity eviction per session records one
bounded degradation (`fact-store-capacity-eviction`) stamped with the evicted
path. Re-arming needs no latch or generation compare inside `fact-store.ts`
itself: `recordDegradationOnce`'s own dedupe set lives in the ledger and is
cleared by `resetDegradationLedger()` at session_start, so a fresh session
always gets a fresh record for free.

**Scope correction.** Title is now `refs #2240`. #2240 names two unbounded stores;
this PR bounds only `fileFacts` (about 2.2 KB per file, which alone cannot reach a
4-8 GB abort). The in-memory review graph is the likely bulk —
`GRAPH_PERSIST_MAX_ELEMENTS_DEFAULT` trims only the on-disk snapshot — filed as
#2255. #2247 covers the `fileFacts` byte axis.

**Verification.** `npm run build`; targeted suites green (fact-store, fact-store
bound, ast-grep warning scan, dispatch integration, scanner, degradation ledger,
bounded-telemetry sweep, deps-centralization, session-state conformance). Every
new test proven red first: on PR-head source the four new tests fail
(`endDispatchFor`/`dropFileFacts` absent, no degradation record), and each guard
is mutation-proofed (dropping `runProviders`, no-op `unpinFile`, or removing the
degradation call each reds its test). Full suite deferred to CI.

---

## Maintainer review round 3 (this commit)

A second AI-generated review found seven items — one high, five medium, one
low-pair. A maintainer applied fixes directly on this branch, preserving the
contributor's authorship.

**F1 (high) — eviction telemetry was one shared subject across six stores.**
`dispatch/integration.ts`, `mcp/analyze.ts`, `lens-map.ts`,
`project-diagnostics/scanner.ts`, `runtime-session.ts`'s session-start
call-graph task, and `mcp/cli.ts` each own a `FactStore`, but the reporter was
wired with a constant `subject: "session-fact-store"`. `recordDegradationOnce`
dedupes on `kind\0subject`, so whichever store hit the cap FIRST in a session
won the once-per-session slot for every store. `runtime-session`'s
session-start review-graph walk runs before any dispatch and can visit every
file in the project, so it reliably won — the dispatch store's own eviction
(the one that actually risks a live mid-dispatch read) was silently never
recorded. Fixed: `FactStore` now takes a `subject` in its constructor, and the
reporter forwards it through to `recordDegradationOnce`, so each store's
per-session record is independent. `tests/clients/dispatch/fact-store-bound.test.ts`
adds a case proving a differently-labeled store gets its own record even
after another store already fired one in the same session; red on pre-fix
code (mutating the reporter's `kind`/`subject` back to constants):
```
 ❯ ... > records one capacity-eviction degradation per session, naming the evicted path
 ❯ ... > gives a differently-labeled store its own record after another store already fired
```

**F2 (medium) — the pin half of item 1 was vacuous.** The existing
`ast-grep-warning-scan-eviction.test.ts` case only pre-evicted content before
calling the scan, so `runProviders`' re-derive masked the pin entirely —
swapping the pin for `dropFileFacts` at the scan's call site left it green.
Added a case that lets the eviction land WHILE the dispatch is in flight (a
`FactStore` subclass floods the store the instant `runProviders` sets this
file's content, simulating the concurrent walk), then asserts the content
`dispatcher.ts:1084` reads back afterward is still the real bytes, not `""`.
Red on pre-fix code (pin swapped for `dropFileFacts`):
```
AssertionError: expected undefined to be 'export const answer = 42; // pi-lens-…'
 ❯ ... > keeps file.content pinned when the project walk lands mid-dispatch
```

**F3 (medium) — production wiring of `endDispatchFor` was vacuous.** No test
drove the actual exported `dispatchLint`/`dispatchLintWithResult`/
`dispatchLintDetailed`/`runAstGrepWarningScan` and checked they release the
pin they took — every existing test either exercises `FactStore`'s own
pin/unpin logic directly, or mocks the runner layer without checking the
`finally` block around it. New `tests/clients/dispatch/dispatch-pin-unpin-wiring.test.ts`
spies on `FactStore.prototype.endDispatchFor` (a real instance method on the
module-scope singleton, observable without exporting it) and drives all four
production entry points once each. Red on pre-fix code (all four
`endDispatchFor` calls commented out — the "no-opping all four calls" shape
the review named):
```
AssertionError: expected "endDispatchFor" to be called with arguments: [ Array(1) ]
Number of calls: 0
 ❯ ... > dispatchLint calls endDispatchFor for the file it pinned
 ❯ ... > dispatchLintWithResult calls endDispatchFor for the file it pinned
 ❯ ... > dispatchLintDetailed calls endDispatchFor for the file it pinned
 ❯ ... > runAstGrepWarningScan calls endDispatchFor for the file it pinned
```

**F4 (medium) — the production reporter body was untested.**
`fact-store-bound.test.ts` hand-copied the `recordDegradationOnce({kind,
subject, reason})` wiring into a local test helper instead of exercising the
reporter `integration.ts` actually installs — a shared-seam double that could
not catch a mutation to that reporter's `kind` or `subject`. Fixed: the test
file now side-effect-imports `integration.ts` and captures the real installed
reporter via `getFactStoreEvictionReporter()` at module load, before any
`afterEach` can clear it, and installs that exact function for its assertions.
Same red-run evidence as F1 above (both fixes landed in the same test edit).

**F5 (medium, design) — item 1's pin did a needless full re-derive.**
`clearFileFactsFor` clears the file's facts before pinning, which is correct
for a fresh-edit dispatch but wasteful for the debounced ast-grep warning
scan: it fires ~2s after the SAME edit's inline dispatch already re-derived
every fact for this file, so the scan's clear+re-derive cost ~51ms of
redundant re-parsing across all five providers (a warm-cache read is ~1ms).
Adopted the reviewer's shape: a new `beginDispatchFor(path)` pins WITHOUT
clearing, paired with the same `endDispatchFor`. `runProviders` already skips
any fact still present (`fact-runner.ts`'s `allPresent` check), so a genuine
miss — the fact really was evicted meanwhile — still repopulates from disk.
The scan now calls `beginDispatchFor` instead of `clearFileFactsFor`; the
other three dispatch callers (fresh-edit paths) are unchanged. Blast radius:
this only touches the warning-scan path; ~51ms saved per debounced scan on a
jsts file whose facts are already warm.

**F6 (low) — fragment said "closes #2240".**
`.changelog/2240-fact-store-bound.md` still said `closes #2240`; changed to
`refs #2240` to match the PR title and the scope-correction note above (#2240
names two unbounded stores; this PR bounds one, `fileFacts`).

**F7 (low) — PR body drift.** The Observability section still said "No new
record", contradicting item 4's own record. Item 4's paragraph described a
"generation compare" fact-store never has — re-arming is the ledger's own
`resetDegradationLedger()` clearing `recordDegradationOnce`'s dedupe set, not
a latch inside `fact-store.ts`. Both fixed above.

**Verification (round 3).** `npm run build`; targeted files green:
`fact-store-bound.test.ts` (10 cases), `ast-grep-warning-scan-eviction.test.ts`
(2 cases), `dispatch-pin-unpin-wiring.test.ts` (4 cases, new),
`fact-store.test.ts` (12 cases), `fact-store-lifecycle.test.ts` (5 cases) — 33
tests total, all passing. A wider sweep (dispatch integration/groups, mcp
analyze + analyze-cli + build-graph-cli, project-diagnostics scanner, three
runtime-session suites, and every listed governance sweep — bounded-telemetry,
bus-producer-coverage, delivery-surface-ratchet, deps-centralization,
finding-delivery-gate, freshness-sweep, managed-tool-seam-coverage,
profiling-coverage, session-state-conformance, module-instance-coverage,
sweep-floor-coverage) also passed: 25 files, 297 tests. Every new/changed
assertion proven red first as quoted above, each mutation reverted and
re-verified green before commit. `npm run lint` (tsc) and `npx oxfmt --check`
pass on every changed file. `npm run changelog:check` passes (104 entries
validated). Full suite deferred to CI.


apmantza:
Thanks for this, @AngriestBird. The pin design is the right instinct. You saw the `?? ""` sink in `dispatcher.ts` and built for it rather than shipping a plain LRU, and the mutation proofs in the body all reproduce here.

This review is AI-generated. A maintainer supervises the process and makes the merge decision.

I verified your red run by reverting `clients/dispatch/fact-store.ts` to master and rebuilding. It reproduces exactly: 4 failed, 1 passed, with the four assertion messages you quoted. All four mutations behave as you describe. CI is green on `cef319e9`, with `Unit tests` and `Lint & type-check` both genuinely executed.

Four items before merge.

**1. One dispatch entry point is never pinned (blocking).** `clients/dispatch/integration.ts:238` (`scheduleAstGrepWarningScan`) builds a context on the module-scope `sessionFacts` and calls `dispatchForFile` after a 2000 ms debounce, but never calls `clearFileFactsFor`. It also never calls `runProviders`, so it relies on the `file.content` an earlier dispatch left behind. `clients/dispatch/dispatcher.ts:1084` reads that fact back and passes it to `applyInlineSuppressions`. A walk inside that window evicts the record, and the read returns `""`, so inline suppressions stop applying. Probe: seed content, run a 2000-file walk, read back yields `""`. Pre-fix this could not happen. Pin at every entry point into that read, or make the read re-derive from disk instead of defaulting to empty.

**2. The pin is a rotating last-16, not an in-flight set** (`clients/dispatch/fact-store.ts:92`). Nothing releases a pin when a dispatch completes, so the 17th `clearFileFactsFor` unpins a dispatch that is still awaiting its runner groups. Same probe, same `""` result. An explicit unpin at dispatch end replaces the estimate with a fact.

**3. The pin clear in `clearFileFacts()` is untested** (`clients/dispatch/fact-store.ts:80`). Removing the line keeps all 124 dispatch test files green. The method has no production caller either, only `tests/clients/dispatch/fact-store.test.ts:44`. Cover it or drop it.

**4. The cap is silent.** AGENTS.md asks every new failure path for a bounded record that keeps the discriminating identity, and evicting a fact a live dispatch needs is one. RSS is too coarse to serve as the proof here: the issue's own measurement is about 2.2 KB per file. PR #2236 has the pattern for read-guard eviction. One `recordDegradationOnce`-class record on the first capacity eviction per session, stamped with the evicted path, would surface items 1 and 2 in production.

One scope note. #2240 names two unbounded stores, and the live review graph is the second. Its measurement (4.4 MB to 12 MB over 3,500 files) does not account for a 4-8 GB abort. Change the title to `refs #2240` and leave a comment on the issue naming the graph cap as the remainder, so the OOM is not marked resolved on evidence that does not reach it. Filing #2247 for the byte axis was the right call.

apmantza:
Thanks again for the fix and for bearing with a second AI-generated review round — noting again, as before, that this review is AI-generated and a maintainer is supervising the process.

The second pass found seven items (one high, five medium, two low) and the fixes are pushed to this branch (`8f9260eb`), preserving your original commits.

- **F1 (high):** eviction telemetry was one shared subject across all six production `FactStore` instances, so `runtime-session`'s session-start review-graph walk reliably consumed the once-per-session ledger record before any dispatch ran, and the dispatch store's own eviction never got recorded. `FactStore` now takes a `subject` label per instance, so each store gets its own record.
- **F2 (medium):** the existing pin test only proved content is re-derived when it was already missing before the scan started, which stayed green even with the pin removed. Added a test where the eviction lands while the dispatch is in flight.
- **F3 (medium):** nothing proved the four production callers actually release the pin they take. Added a test that spies on the real method and drives all four entry points.
- **F4 (medium):** the reporter test hand-copied the production wiring instead of exercising the reporter `integration.ts` actually installs. Fixed to import the real one.
- **F5 (medium, design):** the warning-scan's pin did a needless full clear+re-derive (~51ms measured) of facts the same edit's inline dispatch had already computed seconds earlier. Split into a `beginDispatchFor` that pins without clearing.
- **F6/F7 (low):** a stale "closes #2240" in the changelog fragment and some drifted PR-body text, both corrected.

Full detail, including each red-run proof, is in the "Maintainer review round 3" section of the PR body above. Targeted suites (33 tests in the touched files, plus a 297-test wider sweep including dispatch integration, mcp/analyze, the scanner, runtime-session, and every governance sweep) are green; `npm run lint`, `oxfmt --check`, and the changelog check all pass. Full suite is CI's call from here.

AngriestBird:
Fixed item 2 in ee4dd752. Live FactStore pins now survive until their matching endDispatchFor call, and unpinning immediately restores the idle-record cap. The regression holds 1,025 concurrent dispatches open, proves every content value survives a flood, then proves the first record is evicted on release.\n\nLocal diff check passed. The worktree could not run Vitest or oxfmt because its dependency link is unavailable; CI will verify the pushed head.

apmantza:
Merged — thank you again, @AngriestBird. The verify round reproduced every fix with independent probes: the per-store eviction records discriminate correctly under the session-start walk, the pin/unpin wiring reds under mutation at all four entry points, and the warning-scan path no longer pays the redundant provider pass. Two low-severity residuals are noted in the review record (per-call-site label coverage, telemetry-test timing margin) and do not block. As before: this review process is AI-generated and maintainer-supervised. #2240 stays open for the remaining byte-axis work (#2247).

Issue #2217, opened 2026-08-26
scanner hard-codes kind jsts for every napi-evaluated file, mislabeling non-JS lanes


apmantza:
Evidence
- `clients/project-diagnostics/scanner.ts:388` hard-codes `kind: "jsts"` for every file handed to `evaluateAstGrepRules`, including `.css` and `.html` files.
- That kind drives `suppressLinterOverlap = kind === "jsts" && hasEslintConfig(cwd)` (clients/dispatch/runners/ast-grep-napi.ts:536), so non-JS files are evaluated under a JS/TS linter-overlap policy.
- Harmless today: no CSS rule is in `LINTER_OVERLAP`. But the argument is false, and the first CSS/HTML rule added to `LINTER_OVERLAP` inherits a wrong suppression decision silently.
- Found during PR #2206's adversarial review (probe of the linter-overlap seam), 2026-08-26.

Root cause
- The scanner predates non-JS napi lanes and never plumbed a per-file kind; it labels everything with the only kind that existed.

Acceptance criteria
- The scanner passes a kind derived from the file's actual language family (shared resolver / KIND_EXTENSIONS, not a hand list).
- A test asserts a `.css` file is not evaluated with `kind: "jsts"`.

Observability
- None; the wrong kind is invisible in logs. The test is the net.

Class links
- refs #2199 (CSS lane), refs #2215 (napi allowlist derivation — same seam family). Single-source-of-truth pattern (PR #883).

apmantza:
Update from the #2296 review (css/html now reach the napi runner via dispatch): the scanner's hard-coded jsts kind is now a live DIVERGENCE, not just a mislabel — the same css file is evaluated under kind "css" on the per-edit dispatch path and kind "jsts" on the scanner path. suppressLinterOverlap and any kind-gated semantics can disagree between the two paths for the same file. Raises the value of fixing this; the fix site is unchanged.

PR #2244, opened 2026-08-27
fix(dispatch): derive napi language matrix, cover module extensions (closes #2215)

### LINKED ISSUE #2215: napi fallback allowlist silently skips ~12 rule languages: derive it and pin coverage
Evidence
- `clients/dispatch/runners/ast-grep-napi.ts:638-644` gates the in-process fallback to `typescript`, `tsx`, `javascript`. PR #2206 adds `css` only.
- All 28 Go rules (23 shipped + 5 in PR #2207), plus every Rust, Python, Java, Ruby, and other non-JS-family rule, never run on the napi path; the ast-grep LSP/CLI is their only delivery route. Roughly 12 of the catalog's 16 languages are affected.
- PR #2206's own in-code comment documents the failure shape for CSS: "Before #2199, every `language: Css` rule silently never matched in this fallback engine". The identical hole remains for the rest.
- Found during PR #2207's adversarial review (cross-PR dependency probe), 2026-08-26.

Root cause
- The napi fallback's language allowlist is hand-maintained and was only ever extended reactively (tsx/js at birth, css in #2206, java/kotlin proposed in #2212). No coverage test asserts that every language with enabled catalog rules is either napi-served or deliberately excluded.

Acceptance criteria
- The allowlist is derived from a single source of truth (KIND_EXTENSIONS or the napi engine's actual language support matrix), or each excluded language carries a recorded reason.
- A coverage test reds when a catalog language with enabled rules is silently absent from the napi path.
- The population sweep across all 16 catalog languages is recorded on this issue: for each, napi-served / deliberately-excluded-with-reason / gap.

Observability
- None today: a rule that never runs in-process emits no record. The coverage test is the net; a startup record naming excluded-but-ruled languages would be the runtime net.

Class links
- Same shape as #2206's CSS finding (its diff fixes one member, not the class). refs #2195 (rules initiative), refs #2199, refs #2200. Sibling of the single-source-of-truth pattern (PR #883).

AngriestBird:
## Summary

The ast-grep in-process fallback ran zero rules on `.mjs`, `.cjs`, `.mts`, and `.cts` files, and skipped 247 of the catalog's 470 rules with nothing recording why. Four hand-maintained lists described the same thing — `SUPPORTED_EXTS`, `SUPPORTED_RULE_LANGUAGES`, `ruleLanguageForFile`, `getLang` — each extended reactively (tsx/js at birth, css in #2206, java/kotlin proposed in #2212), and they had drifted from `KIND_EXTENSIONS.jsts`: four module-flavored extensions the addon's own js/ts grammars parse were never admitted, so every shipped rule was silently skipped on them.

`NAPI_LANGUAGE_BINDINGS` is now the single matrix; the four lists derive from it. `getLang` resolves each binding's export against the module that actually loaded, so the effective set is the matrix ∩ the addon's real capabilities — and an admitted file the addon cannot parse records `ast-grep-napi-language-unavailable` once per language rather than returning an empty result no caller can tell from a clean one. The aggregated skip telemetry gains a per-language `route` (`ast-grep-lsp-cli`, or `unclassified` when nobody has decided).

Two claims were measured, not assumed: the addon's `Lang` export is an empty object at runtime (its enum is type-only), so the module's own grammar accessors are the capability surface; and all twelve unserved languages were run through the real `ast-grep run -l <lang>` with `cobol` as the negative control, so the "genuinely missing" bucket is empty.

The population sweep the issue asked for is recorded on the issue thread. Closes #2215.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch

## Tests

`tests/clients/dispatch/runners/ast-grep-napi-language-coverage.test.ts` (new), 15 cases. Pre-fix run (production reverted via `git apply -R`, rebuilt): 8 failed / 7 passed; six are genuine defect-reason reds:

```
expected [ '.cjs', '.cts', '.less', '.mjs', '.mts', '.sass', '.scss', '.svelte', '.vue' ] to deeply equal []
expected [ 'c', 'cpp', 'csharp', 'go', 'java', 'kotlin', 'php', 'python', 'ruby', 'rust', 'scala', 'swift' ] to deeply equal []
.mjs / .cjs: expected [] to include 'no-global-eval-js'
.mts / .cts: expected [] to include 'array-callback-return'
degradation record: expected undefined to be 1
```

The seventh red (`deliveryRouteForRuleLanguage is not a function`) is import-wiring, not a defect red — that helper is new API, covered by mutation instead. Mutation run (four at once, rebuilt): removing `.mjs` from the matrix, adding `.css` to the unserved map, adding `zig` and `css` to the LSP-only list each redded exactly the guard that binds it; `deliveryRouteForRuleLanguage("zig")` mutation flips `unclassified` → red. Post-fix 15/15 green. Screens: invisible skip (the exact shape this issue is about — the coverage test reds on a silently absent language), population sweep with per-member verdict.

### Test assessment

The new file uniquely pins: the addon's bundled grammar set against a real load; `canHandle` ⟺ `getLang` agreement across `KIND_EXTENSIONS`; routed-extension and catalog-language classification in both directions; `.mjs`/`.cjs`/`.mts`/`.cts` firing real catalog rules through the real runner; the admitted-but-unparseable degradation record. It makes no existing test redundant — `ast-grep-tsx-coverage.test.ts`'s `canHandle("App.java") === false` pin guards the complementary negative and stays.

## Blast radius

`ast-grep-napi.ts` consumers: the dispatch runner itself and `clients/project-diagnostics/scanner.ts`; both gate on `canHandle` first. Behavior change by design: `.mjs`/`.cjs`/`.mts`/`.cts` now run the full js/ts catalog in-process at the same per-file cost `.js`/`.ts` already pay — no new work on any previously served path. `clients/degradation-ledger.ts` gains one kind (additive union member; open PR #2236 appends kinds ~line 380, this one ~line 106 — no conflict either way). Targeted verification: 33 files / 576 tests across runner siblings, scanner, ledger, severity, meta-sweeps.

## Observability

`ast-grep-napi-language-unavailable` in the degradation ledger (once per language, subject = rule language) proves an addon/matrix disagreement in production; the `route` field on the aggregated skip record (`ast-grep-lsp-cli` / `unclassified`) proves the routing decision for every skipped language. `unclassified` appearing in a log is the signal this class regrew.

## Class sweep

Defect class: hand-maintained parallel lists describing one capability surface (#883 derive-don't-hand-maintain). Population: the four lists in this file — folded onto `NAPI_LANGUAGE_BINDINGS` (deletion test: removing the matrix re-distributes the ext→language mapping into four call sites, so the seam is deep). One member deliberately left distributed: `scripts/scan-ast-grep-rules.mjs:110` hand-rolls a narrower `getLang` for the dev-only FP census, whose recorded numbers are quoted in shipped rule notes — folding it risks changing recorded censuses. Adjacent same-class finding filed as #2248 (`appliesTo: ["jsts"]` keeps css/html from reaching this runner on the per-edit path). Consolidation verdict: folded, with the one recorded exception.

Closes #2215

BLUF


apmantza:
Thanks for this one, @AngriestBird. The derivation is honest about what it can and cannot prove, and the coverage test earns its place. That is the hard part of this issue and you got it right.

This review is AI-generated. A maintainer supervises the process and reads every finding before it is posted.

**Verdict: merge-ready.** One non-blocking finding, which we will close with a small test commit on this branch rather than a fix round.

### F1 - the `route` telemetry field is not covered by a test

`clients/dispatch/runners/ast-grep-napi.ts:698`, and `skipRouteFor` at `:364`. The PR body says the `route` field proves the routing decision for every skipped language. No test reads it off the record: deleting `route: skipRouteFor(language),` leaves 24/24 green, and collapsing `skipRouteFor` to drop the `mismatch:` branch also leaves 24/24 green. `ast-grep-utils-block.test.ts:98` already destructures `skippedByLanguage.python`, so one assertion on `python.route === "ast-grep-lsp-cli"` plus one on a `mismatch:` key closes it. A maintainer commit will add these to keep your merge unblocked.

### F2 - the quoted pre-fix output is from an earlier test iteration

Reverting the two production files and running your shipped test produces `[ '.cjs', '.cts', '.mjs', '.mts' ]` (9 failed / 6 passed), not the nine-entry list (8 / 7) the body quotes. The declared-unserved extensions cannot appear because line 169 filters them. The red is real and stronger than claimed, so this is transcript hygiene only; noting it here to keep the record straight.

### F3 - changelog prose

`.changelog/2215-napi-language-coverage.md:5` runs ~90 words with three em-dash connectives. Structure is valid; splitting into three sentences would fit house style. Not blocking.

### What I verified independently

The addon capability claim reproduces: on `@ast-grep/napi@0.45.1`, `Object.keys(sg.Lang).length` is 0, `sg.java`/`sg.python` are undefined, and `ts tsx js jsx css html` are live objects. Your matrix matches what napi actually loads.

The coverage test reds in both directions: a synthetic `language: Zig` catalog rule yields `expected [ 'zig' ] to deeply equal []`, and adding `.astro` to `KIND_EXTENSIONS.jsts` yields `expected [ '.astro' ] to deeply equal []`. That is the single-source-of-truth property the issue asked for.

`.mjs`/`.cjs`/`.mts`/`.cts` dispatch end to end through the real runner, not just in a list. The degradation record re-arms after a session reset, stays bounded at count 1 across 500 files, and keeps distinct subjects per missing grammar. Gate B still prevents double-reporting when the ast-grep LSP is alive.

All 18 checks pass on `4ef48b6f` including Unit tests and Lint, on a CLEAN merge ref. Leaving #2217 and #2248 as separate issues is the right call, and your population sweep on the issue covers the third acceptance criterion.

Issue #2117, opened 2026-08-26
Word-index arena never recompacts under incremental churn: 18,102 backing stores after one full-corpus rewrite


apmantza:
Evidence
- Measured during PR #2115's review (finding L2): one pass rewriting every document takes the posting store from 1 backing store to 18,102, and real resident memory from 8.3 MB to 12.3 MB (+48%). Further churn does not recover it.
- Only a full rebuild or a deserialize recompacts; `refreshWordIndexIncrementally` never calls the compaction path.
- Context: still 2.9x better than pre-#2115 master in the identical churned state (35.5 MB), so this is headroom, not a regression.

Root cause
- `compactPostingsIntoArena` runs at build/deserialize time only. Incremental document replacement allocates private per-token slices; orphaned arena ranges stay resident and unaccounted (the estimator's arena-slack blind spot, documented in the store).

Acceptance criteria
- A churn-triggered recompaction policy (threshold on backing-store count or slack ratio), bounded so it cannot stall an edit; or a deliberate, measured decision to keep rebuild-only compaction, recorded in the store docstring with the 48% figure.
- A test that drives full-corpus churn and asserts the chosen bound (stores or resident bytes) holds; red on current code.
- Estimator accounting for arena slack, or its floor semantics asserted in a test.

Observability
- `memory_sample`'s word-index member reports `residentBytes`; a recompaction event should emit a bounded record (before/after bytes, store count).

Class links
- refs #2069 (the arena landed there, PR #2115); sibling of #2114 (review-graph store cost) and #2068 (flat persist payload — a recompact could share its packing walk).

apmantza:
Additional measurement from PR #2115's fix round (this tree, 2,699 docs / 2.27M postings): fresh build 32.8 MB / 1 backing store; after one full-corpus rewrite pass 39.9 MB / 37,687 stores (+22%); after TWO passes 41.7 MB / 37,687 stores (+27%) — the store count does not move on the second pass, so fragmentation levels off rather than compounding. That bounds the worst case and weakens the urgency: the acceptance criteria stand, but rebuild-only compaction with the documented ceiling is a viable verdict.

PR #2245, opened 2026-08-27
perf(word-index): cut per-edit max sync block ~3x (refs #2067)

AngriestBird:
## Bottom line

The cascade and MCP-analyze per-edit seams now replace a word-index document
cooperatively instead of holding the event loop for the whole replacement. On
this repo's corpus (2,829 documents, 2.38M postings) the seam's longest
synchronous stretch drops from 76.7 ms to 32.4 ms with no change in per-edit
CPU, and search results are byte-identical.

## Summary

`updateWordIndexForCascade` called the synchronous replacement primitive, which
is what AC4 was left open on. Swapping in the existing async variant was not
enough on its own: `stageWordIndexDocumentRemoval` checked its cooperative
deadline per posting ELEMENT, so it paid one `performance.now()` per entry —
17.8% of profiled CPU, and 3.2x the sync path's per-edit cost. Staging now
filters each token with `list.withoutFile(removedId)`, the same packed
primitive the synchronous path uses, and checks the deadline between tokens.
The staged result is identical rather than merely equivalent.

The new `updateWordIndexDocumentForEdit` is the async replacement plus the
arena-recompaction gate the synchronous variant carried, so migrating the seam
does not silently drop #2117's fragmentation gate. Bulk refresh deliberately
keeps driving `updateWordIndexDocumentAsync` and recompacting once after its
loop: one refreshed document raises `postingStoreCount` by roughly its
distinct-token count, so a per-document schedule would recompact between
nearly every pair of documents.

`AGENTS.md` said the seam "deliberately uses the synchronous replacement
variant: unawaited concurrent cascades must not yield across a wholesale
posting snapshot." That is the invariant this supersedes, and the line is
rewritten rather than left contradicting the code. The hazard it named was
closed by the per-index operation queue: every async operation chains behind
the previous one in call order, so overlapping unawaited cascades are ordered,
not interleaved.

## Acceptance criteria

1. Met — 9.7 ms mean for a 20 KB document (target under 50 ms, was 2,237 ms).
2. Met — `normalizeEphemeralMapKey` is 0.000% of a 60-edit profile (target
   under 2%, was 44.0%). Top frames are now `splitIdentifier` 36.4%,
   `compactPostingsIntoArenaCooperatively` 21.0%, `indexWordLine` 12.2%.
3. Already met by the interning round; the guard test is unchanged.
4. The "migrates to the async variant deliberately" clause is met. The literal
   8 ms clause is NOT: the measured longest stretch is 32.4 ms. Two reasons,
   both named: the atomic publish cannot yield by construction, and #2117's
   arena recompaction fires on every edit (see Blast radius).
5. Met — see Tests.

## Measurements

`node scripts/bench-word-index-replacement.mjs`, corpus read from this repo's
own tree: 2,829 documents, 2,380,719 posting entries, 38,809 distinct tokens,
60 edits, event loop turned between edits so queued recompactions actually run.

| measure | synchronous | cooperative |
| --- | --- | --- |
| 20 KB document, latency mean | 9.6 ms | 9.7 ms |
| large document, latency mean | 19.6 ms | 21.8 ms |
| large document, block p95 | 52.3 ms | 31.1 ms |
| large document, block max | 76.7 ms | 32.4 ms |

Staging change in isolation (same fixture, old per-element staging vs new):
cooperative latency mean 68.9 ms to 21.8 ms, and `performance.now` self-time
17.8% to 0.9%. Without it this migration would have tripled per-edit CPU.

## Blast radius

Per-edit hot path, both seams that own it. Per-edit CPU is unchanged at 20 KB
(9.6 to 9.7 ms) and +11% on the largest documents (19.6 to 21.8 ms), bought for
a 2.4x cut in the longest synchronous stretch.

A per-edit update now queues behind at most ONE bulk-refresh document rather
than mutating mid-refresh, because the refresh enqueues each document
separately. That is strictly safer and bounded by one document's replacement.

Measured but deliberately not fixed here: `compactPostingsIntoArenaCooperatively`
is 19-22% of per-edit CPU and roughly 13 ms of the remaining 24 ms block.
`postingStoreCount` rises by about the edited document's distinct-token count,
so #2117's flat 64-store threshold is crossed by every edit and the whole arena
is rebuilt each time. That is #2117's guard and the CPU-for-memory trade needs
its own acceptance criteria; filed as a follow-up rather than folded in.

## Class sweep

Both call sites of the per-edit replacement primitive migrated:
`clients/dispatch/integration.ts` (pi cascade) and `clients/mcp/analyze.ts`
(MCP warm index). `rg` over `clients/` finds no third. The synchronous
`updateWordIndexDocument` / `removeWordIndexDocument` remain exported for tests
and are no longer called from any runtime seam.

## Observability

No new record. The bounded per-edit aggregate this issue asked for already
exists and is unchanged: `replacementCount`, `totalReplacementMs`, and
`maxReplacementMs` on `persist_succeeded` with `trigger: "per_edit"`, reset per
persist window. `recordWordIndexReplacement` is called from the shared commit
path, so it covers the cooperative primitive exactly as it covered the
synchronous one. `maxReplacementMs` is the field that proves this change in
production.

## Tests

New:

- `tests/clients/word-index-per-edit.test.ts` — occupancy guard on the seam
  itself. Red on pre-fix `clients/`: `expected 43.238690 to be less than 30`,
  `38.667686`, `37.583977`, all three retries. Green after: 9.3 to 15.7 ms.
- `tests/clients/word-index-rank-baseline.test.ts` — criterion 5. Pushes every
  document of the fixed 60-document corpus through the new primitive and
  reproduces the 20-query baseline minted on `origin/master` at `e0ef5a3f`,
  before any of this work.
- `tests/clients/word-index-observability.test.ts` — the new primitive drives
  the arena recompaction, so migrating the seam cannot silently drop #2117's
  gate.

Mutants run, both red as required: deleting `scheduleWordIndexRecompact` from
the new primitive, and neutering the staged removal to `const survivors = list`.

Targeted run: 17 files, 276 passed, 3 skipped. `npm run build`, `lint`,
`fmt:check`, `check`, `changelog:check`, and `depcruise:check` clean.

## Test assessment

The occupancy guard's 30 ms bound sits between a measured 37-43 ms pre-fix and
9-16 ms post-fix, with `retry: 2`, matching the sibling word-index occupancy
guards' shape. Its fixture is deliberately token-poor and character-rich so the
yieldable halves dominate and the non-yieldable atomic publish stays small; a
fixture with many distinct tokens would have measured the publish instead. The
criterion 5 test is a characterization test against a pre-change baseline, not
a red-first regression test, which is the claim it exists to make.

Refs #2067

BLUF

---

## Maintainer note (added after review)

The review found the caller-latency table does not reproduce off an idle
machine. Running this PR's own bench on the same corpus, five times on a
loaded box, gave cooperative caller-return latency 30% to 96% above the
synchronous path, not the 1% to 11% the table reports. That is expected
behavior, not an error in the numbers: a cooperative path yields to competing
work, so its caller-return time rises exactly when the loop is busy. No
re-measurement is asked of the contributor. Read the table as an idle-box
measurement, and expect absolute numbers to move with machine, OS, and load.

The block-time result is the headline and it holds. On the same loaded box the
large-document longest synchronous stretch went 126.1 ms to 38.5 ms and
136.1 ms to 44.6 ms, a larger cut than the 2.4x claimed above.

Three deltas are committed on this branch, preserving the contributor's
authorship of the change itself:

- A guard for the cooperative staging rewrite in
  `tests/clients/word-index-incremental.test.ts`. Review found the rewrite
  mutation-survivable. The guard counts `performance.now` calls across one
  staged removal rather than timing it, so it is invariant to machine speed
  and load: 16 to 18 reads now, 270,094 with the per-element form restored.
- The changelog fragment now states the ~3x cut in the longest synchronous
  stretch instead of "off the event loop", which the body itself notes is not
  the literal AC4 clause.
- `scripts/bench-word-index-replacement.mjs` gains the `--target-bytes`
  invocation that produces the large-document row.

*This note is AI-generated, with the maintainer supervising the process.*


apmantza:
Thanks, @AngriestBird. The concurrency work here is the strongest part: I could not break the ordering or atomicity of the migrated seam, and the block-time win on large documents is real and larger than you claimed.

Two things to fix before merge.

**1. The per-edit latency numbers do not reproduce.** I ran your own `scripts/bench-word-index-replacement.mjs` five times on the same corpus (2,830 documents, 2,380,811 postings, 38,809 tokens). Cooperative caller-return latency was 30% to 96% higher than synchronous in every run, not the 1% to 11% the body reports. At 20 KB: 14.8/23.5/17.0 ms synchronous against 29.0/28.7/24.8 ms cooperative. At 500 KB: 38.5/39.7 against 50.7/56.2.

My machine runs other work, and that is likely the explanation rather than an error on your part. A cooperative path yields to competing work, so its caller-return time inflates on a loaded loop, which is the production condition. Either restate the table with the measurement condition named, or re-measure under load.

The block claim reproduces and is the headline worth leading with. Large-document block max went 126.1 to 38.5 ms and 136.1 to 44.6 ms in my runs, better than your 2.4x. At 20 KB I saw no reliable block win.

**2. The staging rewrite has no guard.** `clients/word-index.ts:834` swaps the per-element deadline check for `list.withoutFile(removedId)` with a per-token check. The body credits this with the whole CPU outcome, 68.9 ms to 21.8 ms. I restored the per-element form, rebuilt, and ran `word-index-per-edit`, `word-index-rank-baseline`, `word-index-incremental`, and `word-index-cooperative-occupancy`: 27 tests, all green. Nothing reds. AGENTS.md requires a neutered branch to fail a test. A bound on per-edit CPU, or a count of `performance.now` calls per replacement, would close it.

Two smaller items. The body's "large document" row needs its invocation named: the script defaults to `--target-bytes 20480`, so a reader cannot reproduce that row. And the title and the changelog fragment both say "off the event loop" while the body correctly says AC4's 8 ms clause is not met at 32.4 ms. State what the change does: it cuts the longest synchronous stretch about 3x on large documents.

What I verified and cleared:

- Red run reproduced independently. Reverting `clients/` to master gave `expected 143.5987 to be less than 30` on all three retries.
- Both mutants you name go red for me too.
- Ordering: two unawaited edits to one file apply in call order. Atomicity: 400 samples of a query racing an in-flight replacement saw only old or new, never partial. An update racing a remove left consistent counters.
- Composition with the open #2227: merged locally, its pinned counts 1692, 395, and 543 stay green.
- No runtime caller of the synchronous primitives remains anywhere in the tree.
- CI on `478abbfc`: 18 checks green, with Unit tests and Lint and type-check both genuinely executed.

Issue #123, opened 2026-05-30
fix(lsp): TypeScript LSP silently fails for extensions in .pi/agent/extensions/


BingWuJ:
## Summary

When developing a pi extension inside `.pi/agent/extensions/` (with its own `package.json` + `tsconfig.json`), the TypeScript LSP server never starts. Running `lsp_diagnostics` returns "no LSP server configured", with no error or warning to indicate why.

## Root Cause

`clients/lsp/server.ts` ~line 1007, the `TypeScriptRoot` function:

```ts
const TypeScriptRoot: RootFunction = DenoExcludeRoot(async (file) => {
	const extensionRootKey = piAgentExtensionsRootKey(file);
	const projectRoot = await TypeScriptProjectRoot(file);
	if (extensionRootKey) {
		return projectRoot && isSameOrUnderSlashKey(projectRoot, extensionRootKey)
			? projectRoot
			: undefined;
	}
	if (projectRoot) return projectRoot;
	return FileDirRoot(file);
});
```

- `piAgentExtensionsRootKey(file)` returns the `.pi/agent/extensions` directory
- `TypeScriptProjectRoot(file)` walks up to find a `package.json`, returning the parent `.pi/agent`
- `isSameOrUnderSlashKey(child, parentKey)` checks if `child` is the same as or under `parentKey`

The call `isSameOrUnderSlashKey(projectRoot, extensionRootKey)` asks: "is `/path/.pi/agent` under `/path/.pi/agent/extensions`?" — this is never true, because the parent is not under the child. The condition always evaluates to `false`, so the function returns `undefined` and the TypeScript server never starts.

## Steps to Reproduce

1. Create `~/.pi/agent/extensions/` with `package.json` + `tsconfig.json` + any `.ts` file
2. Start pi from within the extensions directory (with pi-lens enabled)
3. Run `lsp_diagnostics` on the `.ts` file → "no LSP server configured"

## Expected Behavior

The TypeScript LSP should start with the project root set to `.pi/agent`, and provide diagnostics for `.ts` files inside `.pi/agent/extensions/`.

## Actual Behavior

The TypeScript LSP never starts. No diagnostics, no error message, silent failure.

## Impact

All users who develop pi extensions inside `.pi/agent/extensions/` with TypeScript cannot use LSP diagnostics, hover, go-to-definition, or any other LSP features.

## Suggested Fix

Change `clients/lsp/server.ts` lines 1009-1012 from:

```ts
[...]

PR #1, opened 2026-04-04
fix: harden runtime loading and startup scans

alexx-ftw:
## Summary
- fix runtime dependency declarations for published installs
- resolve bundled rules/assets from the installed package root
- skip eager startup warmups outside detected project roots or for oversized trees
- add regression tests for manifest/runtime deps, packaged assets, and startup scan gating

## Testing
- npm run build
- npm test
- npm pack

Issue #2103, opened 2026-08-25
CI failure classifier: label OOM/infra kills vs real assertion failures on the PR, auto-rerun infra once


apmantza:
## Evidence

The 2026-08-25/26 merge train hit at least ten Unit-tests failures needing manual log reads to classify: seven exit-137 runner OOM kills with zero assertion failures (#2042's class — heads across PRs #2058, #2059, #2062, #2080, #2084, #2094, #2096), one CodeQL tarball DNS error, and two real assertion failures (#2097's regression; master a6b9388c's contention flake). Every one cost an orchestrator or reviewer a log fetch and a judgment round-trip, and the standing rule (a red required check blocks until read and judged) makes the read mandatory.

## Proposal

A workflow step (or small follow-up job) that runs when the Unit tests job concludes failure:

1. Fetch the job log tail; classify: `Killed`/exit 137 with no FAIL lines = infra-oom; `getaddrinfo`/tarball-download = infra-net; otherwise real, extracting the first `FAIL`/AssertionError line.
2. Post one PR comment: `ci-classifier: infra-oom (0 assertions; auto-rerun triggered)` or `ci-classifier: real — first failure: <file> > <test>`.
3. For infra classifications only: `gh run rerun --failed` ONCE per head SHA (a marker in the comment prevents loops).

Detection-only judgment stays with the maintainer for merges; the classifier removes the log-fetch round-trip and the manual rerun.

## Acceptance criteria

- Real failures are never rerun automatically and never labeled infra.
- The once-per-SHA rerun guard is mutation-proof.
- The comment is one line, updated in place per SHA (no comment spam).
- Classifier heuristics live in a pure, unit-tested function (the check-pr-title.mjs pattern); PR #2093's [mem-file] lines, once merged, feed the infra-oom evidence.

## Cross-links

Refs #2042 (the OOM class and its attribution PR #2093), #1844 (mechanical-enforcement wave; the warden workflow is the sibling that should also comment on PRs that go DIRTY after a master push).

*This issue is AI-generated, with the maintainer supervising the process.*

apmantza:
[...]

Issue #2086, opened 2026-08-25
Close-keyword verification reads the stale closed-event body; a rerun after a body edit replays it


apmantza:
## Evidence

Found by the PR #2084 review's class sweep and independently confirmed: scripts/check-close-keywords.mjs:110 (verifyMergedPullRequest) reads pullRequest.body from the closed-event payload, and close-keyword-verification.yml triggers only on types: [closed] — no edited refresh, no live fetch. A rerun after a body edit always replays the stale body.

Impact is milder than the title-check sibling (#2083, fixed by PR #2084): a stale body produces a false POSITIVE (reports issues unresolved) post-merge, never a false pass. The script already has GH_TOKEN and shells out to gh api, so the fix is a one-line live-body fetch.

## Acceptance criteria

- The script fetches the live PR body (falling back to the payload with a logged ::warning:: on failure), mirroring PR #2084's resolveLivePrTitle shape.
- A rerun after a body edit verifies the edited body.
- Guard coverage per the #2084 round-1 pattern: non-2xx, malformed response, missing token — each mutation-proof.

## Cross-links

Refs #2083 (the shape), PR #2084 (the fix to mirror), #1844 (mechanical-enforcement wave).

*This issue is AI-generated, with the maintainer supervising the process.*

apmantza:
Record before this closes via PR #2315: criterion 1 as originally filed asked for a warn-and-fallback on live-fetch failure. #2267 deliberately superseded that with fail-closed for the verification gate (a warning that lets a stale body pass is the #2086 defect wearing a different hat), and #2315 extends the same fail-closed contract to the syntax lint, with the ::error did-not-run record. Criterion 3's guard set (missing token, non-2xx, malformed response) is now mutation-proof on both paths. The deviation is intentional and verified, not an omission.

PR #2267, opened 2026-08-27
fix(ci): read live PR body in close-keyword verification (refs #2086)

apmantza:
## Summary

`check-close-keywords.mjs --verify-merged` now reads the PR's live body instead of the closed-event payload's snapshot, and the fetch path actually runs in production. Refs #2086.

`verifyMergedPullRequest` read `pull_request.body` straight off the GitHub Actions closed-event payload — a snapshot frozen at close time. `close-keyword-verification.yml` only triggers on `types: [closed]`, with no `edited` refresh and no live fetch. A rerun of this check after the body was edited post-merge always relinted that stale snapshot: an edit that fixed a comma-separated close list never cleared the false-positive comment, and an edit that introduced one would go unseen.

## Fix round 1 (adversarial review)

Two blocking/high findings against the initial fix, both closed in this round.

**F1 (BLOCKING — the fix never fired in production).** `close-keyword-verification.yml` set `GH_TOKEN` for the `gh` CLI calls but not `GITHUB_TOKEN`, which `fetchLivePrBody`/`resolveLivePrBody` read specifically. Every real run threw `GITHUB_TOKEN is not set` before any fetch and silently used the stale payload — the exact bug #2086 exists to fix, wearing a different name. The sibling `lint.yml:83` workflow already carries `GITHUB_TOKEN` for the same seam; the initial fix reused the function without its env contract. Fixed by adding `GITHUB_TOKEN` alongside `GH_TOKEN` to the workflow step, plus a new test that parses the actual workflow YAML and asserts both env vars are present — pinning the WIRING, not just the function's env-var name.

**F2 (HIGH — verification failed open on any fetch failure).** A 503, `ECONNRESET`, or a malformed 200 all logged a `::warning::` and reported "OK" on the stale body — the original bug with a warning attached. That posture is right for the advisory PR body linter (`check-pr-body.mjs`), which `resolveLivePrBody` serves; it's wrong for a post-merge verification GATE, where silently passing on an unmeasured fetch failure is indistinguishable from a gate that actually checked and found nothing wrong. Exported `check-pr-body.mjs`'s previously module-private `fetchLivePrBody` (the strict, throwing version) and switched `verifyMergedPullRequest` to call it directly: any fetch problem now sets `exitCode=1` with an `::error::` annotation instead of falling back.

Also fixed: the changelog fragment claimed "it now fetches the live body," which was false until F1's wiring landed in the same rollup — reworded to describe the actual working behavior.

## Tests

- `tests/scripts/check-close-keywords.test.ts`: switched the "live PR body resolution" suite to exercise `fetchLivePrBody` directly (the function `check-close-keywords.mjs` actually calls now), added `verifyMergedPullRequest fails closed on a live-fetch problem (#2267 F2)` (missing token, non-2xx — both must set `exitCode=1`, log `::error::`, and never reach the success path or the `gh` comment side effect), and added `close-keyword-verification.yml carries GITHUB_TOKEN (#2267 F1)`, which parses the real workflow YAML and asserts the verify step's `env` carries both `GH_TOKEN` and `GITHUB_TOKEN`.
- `tests/scripts/check-pr-body.test.ts`: unchanged; `fetchLivePrBody`'s behavior was already covered there through `resolveLivePrBody`'s wrapping tests, and is now also directly reachable via the new export.

### Test assessment

- `check-close-keywords.test.ts`'s new workflow-YAML test is the only thing in this PR that can catch a REGRESSION of F1 specifically — a source-level assertion that the function reads `process.env.GITHUB_TOKEN` proves the function's contract, but nothing short of reading the deployed YAML proves the workflow actually satisfies it. This is exactly the class of bug that shipped in the first place: correct code, wrong wiring.
- The two `verifyMergedPullRequest` fail-closed tests replace nothing; they pin new behavior (F2) that had no prior test coverage.

Red-first, both rounds. Fix round 1 (#2086, `verifyMergedPullRequest` not exported):

```
TypeError: verifyMergedPullRequest is not a function
 Test Files  1 failed (1)
      Tests  1 failed | 15 passed (16)
```

Fix round 2 (#2267 F1/F2, reverted the workflow YAML and both `.mjs` source files, kept tests):

```
FAIL  ... live PR body resolution (#2086) > uses the live body ...  (fetchLivePrBody is not a function, x3)
FAIL  ... verifyMergedPullRequest fails closed ... GITHUB_TOKEN is unset ...
  expected "error" to be called with ::error::, got the STALE-BODY success comment instead
FAIL  ... verifyMergedPullRequest fails closed ... non-2xx response ... (crashed reaching the real `gh` CLI, because pre-fix code never short-circuits on a fetch failure)
FAIL  ... close-keyword-verification.yml carries GITHUB_TOKEN ... expected undefined to be truthy
 Test Files  1 failed (1)
      Tests  6 failed | 13 passed (19)
```

Restored both fixes, rebuilt, reran: `Test Files 2 passed (2)`, `Tests 104 passed (104)` (`check-close-keywords.test.ts` + `check-pr-body.test.ts` combined). Also ran `tests/clients/session-state-conformance.test.ts` (83 passed). Full suite deferred to CI.

## Blast radius

- `check-pr-body.mjs`'s `fetchLivePrBody` export is new surface; its only other caller remains `resolveLivePrBody` inside the same file, unchanged. No other script imports it.
- The workflow env change (`GITHUB_TOKEN` added to `close-keyword-verification.yml`) only affects that one step; it does not touch `lint.yml` or any other workflow.
- `verifyMergedPullRequest`'s new fail-closed behavior means a live-fetch outage now BLOCKS the post-merge check (exit 1) where it previously passed silently. That is the intended behavior change (F2), but it means a transient GitHub API blip on a merge now produces a visible failing check where none showed before — acceptable for a verification gate, and no automatic rerun is wired for this workflow (unlike #2103's classifier), so a real outage requires a manual rerun.

## Observability

The `::error::` annotation on a fetch failure is the record: it shows up in the Actions log and, because `console.error` writes to the job's own output, in `gh run view --log` for that job. There is no PR comment on a fetch-failure path (deliberately — the failure is about verification not running, not about an issue being unresolved, so the close-keyword-verifier marker/comment machinery is not the right channel for it). F2 is exactly the gap this PR's own missing Observability section let through the first time: writing this section for the initial fix would have surfaced the fail-open behavior before review did.

## Class sweep

Pattern grep across the whole tree (not just `clients/`) for `.pull_request` payload reads: `scripts/check-close-keywords.mjs` (this fix), `scripts/check-pr-body.mjs` (already fixed, #2085), `scripts/check-pr-title.mjs` (already fixed, #2083/PR #2084), `scripts/lib/stale-open-issues.mjs` (reads `issue`, not PR body — different shape), `.github/workflows/close-keyword-verification.yml` and `.github/workflows/osv-scan.yml` (workflow YAML, not script-level payload reads). No further siblings found. Filed #2268 for a related, out-of-scope finding in `fetchLivePrBody`'s null-body handling; corrected #2268's impact framing in a follow-up comment once F2 changed which caller is affected (`verifyMergedPullRequest` now fails closed on that same null-body shape, where it used to silently degrade through `resolveLivePrBody`).

## Scope not covered here

Out of scope, tracked separately: #2268 (`fetchLivePrBody`'s `body: null` handling — a legitimate empty PR description currently throws and, after this PR, fails the check closed rather than degrading silently).

Refs #2086.

### DIFF
--- a/.changelog/fix-2086-close-keyword-live-body.md
+++ b/.changelog/fix-2086-close-keyword-live-body.md
@@ -2,4 +2,4 @@
 section: Fixed
 ---
 
-- **Read the live PR body in close-keyword verification (refs #2086)** — a post-merge rerun of `--verify-merged` used to relint the closed-event payload's stale body; it now fetches the live body, so an edit that fixes or introduces a comma-separated close list is seen on rerun.
+- **Read the live PR body in close-keyword verification (refs #2086)** — a post-merge rerun of `--verify-merged` used to relint the closed-event payload's stale body and, in production, silently kept doing so because the workflow step never carried the `GITHUB_TOKEN` env var the live fetch needs. It now sets that var and fails the check loud on any fetch problem instead of falling back to stale data, so an edit that fixes or introduces a comma-separated close list is reliably seen on rerun.
--- a/.changelog/fix-2086-close-keyword-live-body.md
+++ b/.changelog/fix-2086-close-keyword-live-body.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Read the live PR body in close-keyword verification (refs #2086)** — a post-merge rerun of `--verify-merged` used to relint the closed-event payload's stale body; it now fetches the live body, so an edit that fixes or introduces a comma-separated close list is seen on rerun.
--- a/.github/workflows/close-keyword-verification.yml
+++ b/.github/workflows/close-keyword-verification.yml
@@ -21,5 +21,14 @@ jobs:
 
       - name: Verify referenced issues closed
         env:
+          # GH_TOKEN: the gh CLI calls in check-close-keywords.mjs (issue
+          # state lookups, the PR comment). GITHUB_TOKEN: the live-body
+          # fetch this step exists for (#2267 F1) -- check-pr-body.mjs's
+          # resolveLivePrBody/fetchLivePrBody read process.env.GITHUB_TOKEN
+          # specifically, not GH_TOKEN, so without this the live fetch
+          # always throws "GITHUB_TOKEN is not set" and every run silently
+          # falls back to the stale closed-event body -- the exact bug
+          # #2086 was filed to fix, still present under a different name.
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: node scripts/check-close-keywords.mjs --verify-merged
--- a/scripts/check-close-keywords.d.mts
+++ b/scripts/check-close-keywords.d.mts
@@ -11,3 +11,8 @@ export declare function lintCloseKeywords(body?: string): {
 	offendingLines: string[];
 	valid: boolean;
 };
+export declare function verifyMergedPullRequest(
+	fetchImpl?: typeof fetch,
+	event?: { pull_request?: { number: number; body?: string | null } },
+	getIssueState?: (repository: string, number: number) => string,
+): Promise<void>;
--- a/scripts/check-close-keywords.mjs
+++ b/scripts/check-close-keywords.mjs
@@ -1,7 +1,7 @@
 import { execFileSync } from "node:child_process";
 import { readFileSync } from "node:fs";
 import { fileURLToPath } from "node:url";
-import { resolveLivePrBody } from "./check-pr-body.mjs";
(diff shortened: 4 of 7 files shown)


apmantza:
Post-merge close verification found issue(s) that were not closed: #1 (open). GitHub only applies the first issue in a comma-separated close list; use one close keyword per issue (for example, "Closes #123. Closes #456.").

Issue #2255, opened 2026-08-27
Bound the in-memory review graph, not just its on-disk snapshot (refs #2240)


apmantza:
## Evidence

`#2240` names two unbounded dispatch stores behind the 4-8 GB OOM abort. PR
#2243 bounds the first (`FactStore.fileFacts`). At about 2.2 KB per file that
store alone cannot reach a multi-gigabyte heap, so it is not the bulk.

The second store is the in-memory review graph. Its only size guard is a
persist ceiling:

- `clients/review-graph/builder.ts:1698` — `GRAPH_PERSIST_MAX_ELEMENTS_DEFAULT = 500_000`.
- `clients/review-graph/builder.ts:1707` — `graphPersistMaxElements()` feeds
  that cap to the serialize step only.

The cap trims the centrality-ranked subgraph written to disk. It does not bound
the live `ReviewGraph` object that `buildOrUpdateGraph` holds in memory, which
retains every node and edge for the whole project. On a large repo that object,
not `fileFacts`, is the plausible bulk of the abort.

## Hypothesis

The OOM is dominated by the in-memory review graph, which grows with project
size and has no in-memory element bound — only an on-disk one. Bounding
`fileFacts` (#2243) and its byte axis (#2247) does not reach it.

## Acceptance criteria

- Measure the live `ReviewGraph` retained size across a large repo build and
  state the number.
- Bound the in-memory element/byte footprint, or document why the persist
  ceiling already bounds the live object (with evidence).
- Preserve the coverage-marker honesty the persist ceiling already provides.

## Observability

Name the record that proves the live graph stays bounded in production, or file
the gap. The degradation ledger is the existing seam
(`clients/degradation-ledger.ts`).

## Cross-links

- #2240 — parent OOM issue (names both stores).
- #2243 — bounds `fileFacts` (record count).
- #2247 — `fileFacts` byte axis.

apmantza:
## Cluster plan (owner: review-graph memory follow-ups)

### Shared seam

Four of the five issues live in `clients/review-graph/builder.ts`. One does not.

| issue | area | touches builder.ts | status |
| --- | --- | --- | --- |
| #2255 | bound the live graph | yes — `setWorkspaceGraph`, `capGraphForPersist` | this PR |
| #2114 | measured heap cost + `memory_sample` bytes | yes — `estimateReviewGraphStoreBytes` | reporting already merged (PR #2213); one census item left |
[...]

PR #2291, opened 2026-08-27
TEMP do-not-merge: Linux proof for #2256's mixed-path guard (refs #2255)

apmantza:
## Do not merge, do not review

Throwaway branch. It exists only to produce Linux CI evidence for one guard on #2256, then it gets deleted.

## What this is

`clients/review-graph/builder.ts` on this branch carries a deliberate mutation: the lazy-fold retry trigger is weakened from counting reachable NODES to counting covered FILES.

## What it proves

#2256's mixed-path guard, `does not under-select when only some node paths are pre-folded`, must go RED here.

The round-3 version of that fixture built its unfolded variant with `path.relative` on non-existent paths. That round-trips through `normalizeMapKey` on Windows but not on Linux, so unmutated code behaved like the mutant there and the guard was unprovable on CI. Round 4 rebuilt the fixture on real files with a redundant `/./` segment, which `path.resolve` collapses identically on every OS.

A Windows-only mutation run cannot settle whether that repair works on Linux. This branch asks CI directly.

Expected: Unit tests RED, on that test only.

Refs #2255


apmantza:
Purpose served. Linux CI confirmed the guard reds here as intended: `does not under-select when only some node paths are pre-folded`, `expected 0 to be greater than 0`, 1 failed of 835. The fix on #2256 is green on the same runner. Deleting the branch. Refs #2255

Issue #2292, opened 2026-08-27
No work-count gate on warmed getCachedRules: 8000x regression passed CI green (from #2287 review)


apmantza:
# Evidence

- PR #2287 (head `67e58ffb`) shipped a warmed-path regression in `getCachedRules`: 0.02 ms on master vs 161–186 ms on the PR head, measured on the bundled 286-rule catalog, same machine, same harness.
- `clients/project-diagnostics/scanner.ts:235` calls `evaluateAstGrepRules` per file, so a 500-file scan pays ~80 s where master pays ~0.01 s.
- CI on `67e58ffb` was fully green. No required check covers the cost of this seam.

# Root cause

The test suite asserts freshness behavior on the rules cache but not its work profile. There is no work-count assertion (stat calls, file reads, hash invocations per warmed call) and no perf gate, so an unbounded re-read of the bundled catalog on every call is invisible to CI.

# Acceptance criteria

- A load-invariant test asserts the warmed `getCachedRules` path performs zero content reads and at most one stat sweep per cadence window, using work-count assertions (the `countClockReads`-style pattern from the #2202/#2227/#2254 arc in `tests/support/perf-harness`), not wall-clock.
- The test reds when the content-confirm branch from #2287 head `67e58ffb` is reintroduced.
- The test is OS-independent (no timing thresholds).

# Observability record

The regression was caught only by an adversarial review probe (manual timing harness). No log line records rules-cache revalidation cost; `~/.pi-lens/latency.log` covers dispatch, not cache warm-up. Gap: none of the runtime logs would have shown this in dogfood use either — the cost hides inside per-file scan latency.

# Class sweep

Siblings with the same shape (hot cached path, no work-count gate):

- `clients/cache/rule-cache.ts` — content-confirms only the mutable subset by design; its exemption comment is the invariant, but no test asserts the bundled set skips content reads.
- `clients/file-utils.ts` project-ignore freshness (PR #2260) — has the cadence window; work-count coverage worth confirming when this lands.

Related: #2263 (rules-cache freshness follow-ups), PR #2287 (fix round in flight applies metadata-only + cadence).

PR #2297, opened 2026-08-27
fix(snapshot): validate authoritative body size (closes #2285)

### LINKED ISSUE #2285: Authoritative snapshot cache gates on mtime with no size axis, masking same-bucket external writes
## Evidence

Two cache tiers in `clients/project-snapshot.ts` guard the same snapshot body
against the same external writer, one branch apart, with different freshness
gates. The stricter one documents exactly why the weaker one is not enough.

`snapshotParseCache` requires mtime **and** size:

```
clients/project-snapshot.ts:831  const cached = snapshotParseCache.get(cacheKey);
clients/project-snapshot.ts:833  // Both mtime AND size must match (see the `size` field doc on
clients/project-snapshot.ts:834  // SnapshotParseCacheEntry for why: coarse FAT/exFAT mtime resolution can
clients/project-snapshot.ts:835  // otherwise alias a just-rewritten file onto a stale cache entry).
clients/project-snapshot.ts:836  if (cached && cached.mtimeMs === body.mtimeMs && cached.size === body.size) {
```

The rationale on the field itself (`:261-274`):

```
 * On-disk file size in bytes at cache time. FAT/exFAT round `mtime` to a 2s
 * ... `mtimeMs` even though the body changed — mtime alone would then serve a
 * stale cached parse for a file that was, in fact, just rewritten.
```

`authoritativeSnapshots` gates on mtime only. Both read sites:

```
clients/project-snapshot.ts:794  const diskMtime = body ? body.mtimeMs : Number.NEGATIVE_INFINITY;
clients/project-snapshot.ts:795  if (diskMtime <= authoritative.knownMtime) {

clients/project-snapshot.ts:817  const diskMtime = body ? body.mtimeMs : Number.NEGATIVE_INFINITY;
clients/project-snapshot.ts:818  if (diskMtime <= authoritative.knownMtime) {
```

`AuthoritativeSnapshotEntry` (`:320-334`) carries `knownMtime` and
`lastUsedAt`. There is no `knownSize` field to compare.

## Root cause

The #1664 mtime-plus-size class. The authoritative tier accepts the exact
aliasing the parse tier was hardened against, and it is the tier that wins:
both read sites consult `authoritativeSnapshots` **first** and return early.

The comparison makes it worse than a plain equality gate. It is `<=`, not `===`,
so an external write whose mtime lands in the same coarse FAT/exFAT bucket as
our own write satisfies `diskMtime <= knownMtime` and the in-memory snapshot is
served. The external write is not detected late; it is masked for as long as the
entry lives, which is the 20-minute idle window
(`PROJECT_SNAPSHOT_IDLE_EVICT_MS_DEFAULT`).

The comment at `:810-813` states the intended contract: "Authoritative
in-process write wins while our own (possibly still in-flight) write has not
been superseded on disk by a newer external mtime." On a coarse-timestamp
filesystem, "newer" is not decidable from mtime alone. That is the whole reason
the sibling tier reads size.

## Acceptance criteria

- `AuthoritativeSnapshotEntry` carries the on-disk size alongside `knownMtime`,
  and both read sites compare it, so a same-bucket external rewrite of a
  different length is detected.
- Decide and record what a same-bucket, same-size external rewrite should do.
  Size is a second axis, not a content hash, so it narrows the window rather
  than closing it. The sibling tier accepts that limit; say whether this tier
  does too.
- A regression test writes an external body with an mtime inside the same
  coarse bucket and a different size, then asserts the loader serves disk, not
  the in-memory snapshot. Proven red on pre-fix code.
- Mutation-proof: removing the size comparison reds that test.
- Note the `<=` versus `===` question explicitly. If `<=` is load-bearing for
  the in-flight-write case, keep it and gate the size check accordingly rather
  than tightening the operator by accident.

## Observability

No record proves this today. A masked external write surfaces as stale exports
and stale project rules, which read as a wrong answer rather than as an error,
so nothing in `~/.pi-lens/sessionstart.log` or `latency.log` distinguishes it
from a correct cache hit. Naming the gap.

## Cross-links

Found during the PR #2260 review (#2071's fix round) while sweeping the
"two caches over one source on two freshness disciplines" family. It was
recorded there as an unverified lead and has since been read in source. Refs
#1664 (the mtime-plus-size class), #1105 (the same second axis added to the
ignore-matcher caches), #2071.

apmantza:
## Summary

- Add the on-disk size to each authoritative project-snapshot stamp and validate it at both public load sites.
- Preserve the load-bearing `diskMtime <= knownMtime` rule for in-flight writes. A different-size body in the same or an older mtime bucket now invalidates the authoritative serve.
- Keep the tier metadata-only. A same-size, same-mtime external rewrite remains invisible by design, matching the issue's requested decision and avoiding #2292's unbounded hot-path hashing cost.
- Stop the unbounded mask on the next read. The full loader deletes the mismatched authoritative entry; the narrow loader serves disk without touching or re-arming the entry's idle timer.

Closes #2285.

## Tests

- Red proof on the pre-fix code: both new parameterized cases failed at `expect(load(cwd)?.seq).toBe(9)` with `expected 7 to be 9`; 2 failed, 53 skipped.
- Mutation proof: replacing both new size comparisons with `true` produced the same two assertion failures.
- `npm run build`
- `npm test -- --run tests/clients/project-snapshot.test.ts` (55 passed)
- `npm run lint`
- `npm run fmt:check`

## Class sweep

Searched `clients/`, `tools/`, `mcp/`, `scripts/`, and `index.ts` for `mtimeMs`, `.mtime`, `knownMtime`, `lastMtime`, and mtime comparisons.

- `clients/project-snapshot.ts` authoritative cache: defect confirmed at both load sites and fixed here.
- `clients/project-snapshot.ts` parse cache: already compares `mtimeMs` and size.
- `tools/lens-diagnostics.ts`, `clients/advisory-provenance.ts`, `clients/blocker-freshness.ts`, `clients/diagnostic-dispositions.ts`, `clients/diagnostic-line-freshness.ts`, `clients/file-utils.ts`, `clients/opaque-mutation-scan.ts`, `clients/pipeline.ts`, `clients/project-lens-config.ts`, `clients/dispatch/runners/yaml-rule-parser.ts`, and `clients/word-index.ts`: content-bearing metadata gates already include size or a stronger content confirmation.
- `clients/tree-sitter-cache.ts`: production callers provide content, so the content hash is authoritative; its mtime-only fallback is test-only by documented contract.
- `clients/review-graph/workspace-modules.ts`, `clients/workspace-topology.ts`, and project-config directory stamps: directory-membership invalidation, where file-size identity is not a valid axis; not members.
- `clients/freshness.ts`, `clients/read-guard.ts`, `clients/session-state-store.ts`, PID/lock age checks, and diagnostic scanned-at comparisons: event ordering or age gates, not content memos; not members.
- `mcp/build-staleness.ts`, `clients/installer/index.ts`, `clients/lsp/diagnostic-binding.ts`, `clients/lsp/workspace-diagnostics-cache.ts`, and `clients/recent-touches.ts`: mtime-only memo candidates, but separate cache contracts and call paths. This PR does not claim they are safe. The five leads are tracked in #2300 rather than expanding #2285's authoritative-snapshot seam.

No textual twin of the fixed authoritative gate remains.

## Blast radius

Caller grep shows the full loader feeds MCP analysis, reverse-dependency reads, runtime-session warmup/hydration, word-index persistence, and project-snapshot merge writers. The narrow loader feeds runtime-session export/rule hydration. Both now honor a different-size external body immediately.

The hot-path cost stays O(1): no new `stat` or file read was added. Both loaders reuse the existing resolved body `{mtimeMs, size}`. Post-promotion assigns two fields from its existing single stat. Each authoritative hit adds one numeric selection and one equality comparison.

## Test assessment

- `tests/clients/project-snapshot.test.ts`: the two table cases uniquely pin the full and narrow production loaders against a real different-size gzip rewrite in the saved file's same two-second mtime bucket. No existing test is redundant; the older external-writer test covers a strictly newer mtime instead.

## Observability

No production log or ledger record proves this cache decision. The masked-write failure returned a plausible stale snapshot rather than an error, and this fix adds no per-hit logging to the hot path. The regression test is the proof surface; the production observability gap remains explicit.

## AGENTS.md sections consulted

- Issue and PR design contract
- Contributing
- Recurring defect shapes, shape 6
- Test requirements and test-authoring screens
- Standing invariants: caches, durable stores, and path keys

## Fix round (review findings, commit 4aba08cc)

**F1 (blocking, fixed).** Both loaders read `body === null` (nothing on disk)
as a size mismatch, because `diskSize` fell back to `-Infinity` while
`knownSize` was a real stamped number after our own write. An external
deletion of the body — clearing the cache dir mid-session — therefore looked
like a different-size rewrite: both loaders dropped the carry-forward and
returned `null`, contradicting the documented `body === null` contract this
code already stated in its own comment. Rewrote both guards so `body ===
null` short-circuits to "not superseded" and keeps serving the authoritative
object, matching the pre-existing contract.

Kept the comparison in `<=`/`===` form (`body === null || (body.mtimeMs <=
knownMtime && body.size === knownSize)`) rather than the logically
equivalent inverted `>`/`!==` shape, because the inverted form trips
`tests/clients/freshness-sweep.test.ts`'s out-of-kernel mtime-comparison
ratchet, which matches on `>`/`>=`. The freshness kernel's own tolerance
(50ms) is not appropriate here — this gate deliberately uses an exact `<=`
bucket comparison, not a drift-tolerant reference check — so migrating to
`freshnessFromMtime` would change behavior, not just form.

Red proof on the pre-fix commit (fcd358d9), both new parameterized cases:

```
× full loader carries forward the authoritative in-process write when the body disappears from disk (#2296) 60ms
× exports-and-rules loader carries forward the authoritative in-process write when the body disappears from disk (#2296) 33ms
AssertionError: expected undefined to be 7 // Object.is equality
```

Green after the fix: `tests/clients/project-snapshot.test.ts` 57 passed.
Mutation proof: replacing `body === null` with `false` in both sites broke
the build (`'body' is possibly 'null'`) and, forced through, reproduced the
same red failures — the guard is load-bearing.

**F2 (addressed, documented).** The narrow loader still does not delete the
authoritative entry on a size mismatch. Judged the PR body's original
rationale (avoid touching the idle timer) sound rather than thin: the narrow
loader never calls `touchAuthoritativeSnapshot` even on a hit, so a
mismatched entry it leaves in place is not re-armed by it either — eviction
stays bounded by whatever idle window the full loader (or the original
write) last set. Added a comment at the narrow loader recording this instead
of aligning it with the full loader's delete.

**F3 (re-homed).** The five untracked class-sweep leads (`mcp/build-
staleness.ts`, `clients/installer/index.ts`, `clients/lsp/diagnostic-
binding.ts`, `clients/lsp/workspace-diagnostics-cache.ts`,
`clients/recent-touches.ts`) now live in #2300. Updated the class-sweep
section above to reference the issue instead of listing them as loose leads.

**F4 (documented).** Added one sentence to AGENTS.md's caches section:
`touchAuthoritativeSnapshot` re-arms the idle-eviction timer on every full-
loader hit, so a bucket the narrow loader alone keeps hitting rides an
indefinite mask, not one bounded by the 20-minute idle window.

**Verification this round:** `npm run build`; targeted suite
`tests/clients/project-snapshot.test.ts` (57 passed); the five other files
referencing the loaders (`reverse-deps`, `runtime-session-retro-hydrate-
warmup-race`, `word-index-lifecycle`, `word-index`, `symbol-search`, 97
passed, 1 skipped); the ten governance sweeps named in the fix-round brief
(`delivery-surface-ratchet`, `finding-delivery-gate`,
`session-state-conformance`, `bounded-telemetry-sweep`,
`bus-producer-coverage`, `deps-centralization`, `freshness-sweep`,
`managed-tool-seam-coverage`, `profiling-coverage`,
`module-instance-coverage`/`module-instance-binding`,
`sweep-floor-coverage`), all green (313 passed, 1 skipped across those plus
project-snapshot and freshness-sweep); `npm run lint`; `npm run fmt:check`;
`npm run changelog:check` (existing #2285 fragment still valid, no new
fragment needed — same bug, same PR, not yet merged). Full suite deferred to
CI.

`closes #2285` stays: every acceptance criterion from the issue is still
met, and this round only restores a contract this PR itself introduced.

### DIFF
--- a/.changelog/2285-authoritative-snapshot-size.md
+++ b/.changelog/2285-authoritative-snapshot-size.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Detect different-size project-snapshot rewrites inside one mtime bucket (closes #2285)** — The authoritative in-process cache now validates both body mtime and size before serving its snapshot, so a hot project root cannot mask a different-length external write indefinitely. Same-size, same-mtime rewrites remain outside this metadata-only hot-path guarantee.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -821,7 +821,10 @@ at save and after promotion. Both `loadProjectSnapshot` and
 disk mtime is `<=` the stamp and size still matches. Keep `<=` for the
 in-flight-write case; size detects different-length writes in a coarse mtime
 bucket. Same-size, same-mtime external rewrites remain invisible by design,
-and this hot read path does not content-hash. (#2285)
+and this hot read path does not content-hash. Only `loadProjectSnapshot`
+re-arms the idle-eviction timer on a cache hit (`touchAuthoritativeSnapshot`),
+so a bucket the narrow loader alone keeps hitting rides an indefinite mask,
+not one bounded by the 20-minute idle window. (#2285)
 
 Advisory caches must carry immutable capture provenance and validate it again
 at every delivery surface. A finding is current only when session/turn state
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -815,6 +815,14 @@ tier. (#2262)
 
 ### Caches, durable stores, and path keys
 
+The project-snapshot authoritative-write cache stamps both `mtimeMs` and size
+at save and after promotion. Both `loadProjectSnapshot` and
+`loadProjectSnapshotExportsAndRules` serve the in-process object only while
+disk mtime is `<=` the stamp and size still matches. Keep `<=` for the
+in-flight-write case; size detects different-length writes in a coarse mtime
+bucket. Same-size, same-mtime external rewrites remain invisible by design,
+and this hot read path does not content-hash. (#2285)
+
 Advisory caches must carry immutable capture provenance and validate it again
 at every delivery surface. A finding is current only when session/turn state
 matches and every affected file is SHA-256-confirmed (size+mtime is only the
--- a/clients/project-snapshot.ts
+++ b/clients/project-snapshot.ts
@@ -327,9 +327,12 @@ interface AuthoritativeSnapshotEntry {
 	 * pre-write file's mtime at save time (or -Infinity when no file exists
 	 * yet), then updated to the promoted file's mtime once the worker (or the
 	 * sync fallback) lands the write. Load prefers this entry while the on-disk
-	 * mtime is `<=` this value.
+	 * mtime is `<=` this value and its size still matches `knownSize`. The size
+	 * axis detects coarse-mtime collisions without hashing this hot read path;
+	 * a same-size, same-mtime external rewrite remains invisible by design.
 	 */
 	knownMtime: number;
+	knownSize: number;
 	lastUsedAt: number;
 	idleTimer?: ReturnType<typeof setTimeout>;
 }
@@ -792,7 +795,11 @@ export function loadProjectSnapshotExportsAndRules(
 	const authoritative = authoritativeSnapshots.get(key);
 	if (authoritative) {
 		const diskMtime = body ? body.mtimeMs : Number.NEGATIVE_INFINITY;
-		if (diskMtime <= authoritative.knownMtime) {
+		const diskSize = body ? body.size : Number.NEGATIVE_INFINITY;
+		if (
+			diskMtime <= authoritative.knownMtime &&
+			diskSize === authoritative.knownSize
+		) {
 			const { version, seq, cachedExports, projectRulesScan } =
 				authoritative.snapshot;
 			return { version, seq, cachedExports, projectRulesScan };
@@ -809,18 +816,23 @@ function loadProjectSnapshotInternal(
 	const key = normalizeMapKey(cwd);
 	const body = resolveSnapshotBodyPath(cwd);
 	// Authoritative in-process write wins while our own (possibly still
-	// in-flight) write has not been superseded on disk by a newer external
-	// mtime. `body === null` means nothing is on disk yet — our just-scheduled
-	// write is the only truth, so serve it.
+	// in-flight) write has not been superseded on disk by a newer external mtime
+	// or a different-size body in the same/coarser bucket. `body === null` means
+	// nothing is on disk yet — our just-scheduled write is the only truth, so
+	// serve it. Keep `<=`: the pre-promotion body can be older than our baseline.
 	const authoritative = authoritativeSnapshots.get(key);
 	if (authoritative) {
 		const diskMtime = body ? body.mtimeMs : Number.NEGATIVE_INFINITY;
-		if (diskMtime <= authoritative.knownMtime) {
+		const diskSize = body ? body.size : Number.NEGATIVE_INFINITY;
+		if (
+			diskMtime <= authoritative.knownMtime &&
+			diskSize === authoritative.knownSize
+		) {
 			touchAuthoritativeSnapshot(key, authoritative);
 			return authoritative.snapshot;
 		}
-		// An external writer moved past our write — honor disk and stop
-		// serving the now-stale in-memory object.
+		// An external writer changed the body beyond our metadata stamp — honor
+		// disk and stop serving the now-stale in-memory object.
 		deleteAuthoritativeSnapshot(key);
 	}
 	if (!body) {
@@ -1298,9 +1310,11 @@ function reconcileAuthoritativeAfterWrite(
 		return;
 	}
 	try {
-		entry.knownMtime = fs.statSync(pending.gzPath).mtimeMs;
+		const stat = fs.statSync(pending.gzPath);
+		entry.knownMtime = stat.mtimeMs;
+		entry.knownSize = stat.size;
 	} catch {
-		// If we can't stat our own write, leave knownMtime as-is; the worst case
+		// If we can't stat our own write, leave the metadata as-is; the worst case
 		// is one extra disk re-parse on the next load.
 	}
 }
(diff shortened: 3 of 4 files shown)


apmantza:
Post-push broader-suite result for `4ebf03fbbca64a18406d625383ef3200525cff14`:

- `npm test`: **10,503 passed, 81 failed, 59 skipped** across 839 test files (787 passed, 41 failed, 11 skipped).
- This broad local run is environment-invalid, not green. Failures span 41 unrelated files and are dominated by 5-second timeout contention; the serialized wall-clock check also exceeded its budget after the storm.
- The workspace Git guard refused fixture-local `git init`, directly invalidating 17 `opaque-mutation-scan` cases and related Git-fixture tests.
- The ast-grep self-scan fixture also failed to locate its temporary `sgconfig.yml` under this environment.
- No failure names `clients/project-snapshot.ts` or the touched regression file.

Valid scoped evidence on the same source tree remains:

- `npm run build`: pass.
- `npm test -- --run tests/clients/project-snapshot.test.ts`: 55/55 pass.
- `npm run lint`: pass.
- `npm run fmt:check`: pass.
- Pre-fix and size-guard mutation runs: both new loader cases fail at the intended `expected 7 to be 9` assertion.

Required GitHub checks provide the unguarded broad comparison for the pushed head.

apmantza:
Final-head verification after the commit-message-only amend:

- Head: `fcd358d9af129fe6a8e6496596d08d2d25a94dd0`.
- Tree is unchanged from the scoped and broad test runs.
- Required `Lint & type-check`: completed, success, on this SHA.
- Required `Unit tests`: completed, success, on this SHA.

The earlier post-push broad-run comment records the local environment-invalid totals. GitHub's required Unit tests job supplies the green unguarded broad result for the final head.

Issue #2208, opened 2026-08-26
playground-verify-rule harness returns zero matches for every rule, voiding the cross-validation step


apmantza:
Found during PR #2204's review (2026-08-26). The #2195 porting discipline's playground cross-validation step is currently non-evidence.

## Evidence

Control run against a long-shipped rule with a known-positive snippet:

```
scripts/playground-verify-rule.mjs jsx-boolean-short-circuit --code "const A = () => <div>{items.length && <b/>}</div>;"
{"ok":true,"rule_id":"jsx-boolean-short-circuit","language":"tsx","matches":0,"lines":[]}
```

`matches: 0` for a snippet the shipped rule demonstrably catches. The harness's own header (lines 6-12) says its purpose is match-count agreement; zero-for-everything means every "ok:true matches:0" table produced with it (including PR #2204's) verifies only that the rule loads.

## Acceptance criteria

1. Root-cause the zero-match behavior (payload shape drift against the playground API, language mapping, or response parsing) and fix it.
2. A self-test: the harness run against one shipped rule + known-positive snippet must report matches >= 1; red on the current code.
3. The #2195 process doc/AGENTS reference stays accurate about what the harness proves.

Refs #2195, PR #2204 (review F3).

apmantza:
The fix in PR #2234 uses a different JSX snippet than the one in this issue's evidence. Recording why, so the issue's own repro doesn't read as still-failing after merge.

The issue's snippet used a self-closing tag:

```
const A = () => <div>{items.length && <b/>}</div>;
```

`<b/>` parses to a `jsx_self_closing_element` node. `jsx-boolean-short-circuit.yml`'s rule requires `kind: jsx_element` on the `&&`'s right-hand side, so this exact snippet is a genuine 0-match result — confirmed against the local `ast-grep scan --rule` CLI (also 0 matches) and with `ast-grep run -p '... && <b/>' --debug-query ast`, which shows the `jsx_self_closing_element` node kind directly. This is a snippet-choice issue in the original report, not a remaining defect in the harness.

The root cause and fix are unaffected: swapping to a non-self-closing tag (`<b>hi</b>`, a `jsx_element` node) reproduces the same `matches: 0` before the fix and `matches: 1` after, with everything else identical. That's the fixture PR #2234's regression tests use.

PR #2311, opened 2026-08-27
fix(playground-verify): normalize tabs in sentinel match, fail fast on stable mismatch (#2306)

### LINKED ISSUE #2306: playground-verify sentinel defeated by tabs in the first source line; burns full 40s timeout (from #2304 review)
# Evidence

From PR #2304's review, proven on master AND the fix branch (pre-existing, unchanged by #2304): a source whose first line contains an internal tab (`const\tok\t= arr.indexOf(x) !== -1;`) makes playground-verify report false schema drift — Monaco does not render the tab as U+0009, so the sentinel never matches:

- With fix: 78 polls, sentinelFound:false, ok:false "likely upstream schema drift", 41530 ms.
- Master: 75 polls, same, 40639 ms.

The run burns the full 40 s timeout before concluding.

# Root cause

`firstNonEmptyLine` compares raw source text against Monaco's rendered textContent; the NBSP normalization added for spaces has no tab counterpart.

# Acceptance criteria

- Collapse tabs (and any other Monaco-rendered whitespace substitutions) in the sentinel comparison the same way NBSP is normalized.
- Regression fixture with an internal-tab first line: red pre-fix, green after.
- A sentinel mismatch should conclude faster than the full 40 s timeout when the first poll already renders the pane (fail-fast on stable-but-unmatched, or say why not).

# Observability record

The verifier's own "likely upstream schema drift" message is the (misleading) record; the fix should make the message name a sentinel-normalization mismatch when that is the cause.

Refs #2253 (parent seam), PR #2304 (where measured).

AngriestBird:
## Summary

A tab in the first source line made `playground-verify` report false schema drift and burn
the full 40s timeout: the sentinel comparison normalized NBSP but not tabs, so a sentinel
built from the caller's raw source never matched Monaco's rendering.

`buildScrapeExpr` now collapses every run of tab / NBSP / space to a single space on BOTH
sides of the comparison — the decoded sentinel as well as the rendered pane — instead of
hand-listing the substitutions Monaco happens to use. Scrape results carry `sourceLen`, and
a new exported pure state machine `trackStableUnmatched` ends the poll loop once a
non-empty `sourceLen` has repeated unmatched for 3 consecutive 250ms polls; any change
resets it, so a still-mounting pane never trips the early exit. Result JSON gains `polls`
and `concluded_early`.

Closes #2306

### Fix round 1 (review)

The early-exit failure message originally read "likely a sentinel-normalization mismatch in
this harness, **not upstream schema drift**". That claim is not supported by the evidence
the early exit has: upstream schema drift (a renamed/moved `state.source`) leaves the
playground rendering its **own default sample**, which is just as stable and non-empty as a
normalization mismatch. `concludedEarly` rules out "still mounting" and nothing else. The
message now names both remaining causes and lets the reader pick; the changelog fragment
was corrected to match. No behavior change.

## Tests

- EDIT `tests/scripts/playground-verify-rule-sentinel.test.ts` (+4 describe blocks, no
  existing case changed):
  - `buildScrapeExpr tab normalization (#2306)` — 3 cases. Two prove `sentinelFound: true`
    when Monaco renders each tab as a run of NBSP and as a run of regular spaces
    (regression proof for the reported failure). The third feeds the playground's own
    default sample and pins `sentinelFound: false`, so genuine drift still survives
    normalization — the mutation guard against a collapse that matches everything.
  - `trackStableUnmatched fail-fast (#2306)` — 4 cases: no early exit while `sourceLen`
    changes (still mounting), early exit on the 3rd consecutive identical non-empty
    reading, streak reset when `sentinelFound` flips true, and never counting a
    zero-length pane. That last one is the invisible-skip screen: without it, a pane that
    never renders would satisfy "stable" and exit early.
- Red-first, recorded by the authoring run: reverse-applying the source patch failed 6 of
  the 7 new assertions (both tab cases plus every `trackStableUnmatched` case).
- Re-run in this worktree after `npm run build`: 19/19 green. `npm run lint` and
  `npm run fmt:check` clean. Full suite deferred to CI per #1112.
- Screens: `trackStableUnmatched` is exported specifically so the poll decision is pinned
  at its own layer rather than through a live Chrome/CDP run (wrong-layer pin); the
  `buildScrapeExpr` cases evaluate the generated expression against a stubbed `document`,
  the real string the harness ships, not a reimplementation (ambient-inspection double).

## Test assessment

`playground-verify-rule-sentinel.test.ts` is the only suite that observes the generated
scrape expression as a string. The new tab cases sit alongside the existing #2208 F2 NBSP
cases and pin a distinct substitution class, so neither makes the other redundant — but the
NBSP cases are now partly subsumed: both go through the same `collapseWhitespace`, so a
regression in it reds all five. They are kept because they document the two separately
reported failures, and the cost is four assertions.

The `trackStableUnmatched` block is net-new coverage of a decision that previously had
none (it did not exist as a seam). Nothing was removed or loosened.

Not covered, stated plainly: Monaco's *actual* tab substitution was never verified against
the live site (no Chrome in this environment). The collapse-both-sides design is
substitution-agnostic by construction and the fixtures cover the NBSP-run and space-run
renderings, but if Monaco uses a fourth character the fix does not cover it and the new
early-exit message is what points the next reader at that possibility.

## Blast radius

`scripts/playground-verify-rule.mjs` is a developer verification harness, not shipped
runtime. `rg` finds no `import` of it outside `tests/`, and it is not reachable from
`index.ts`, `clients/`, `mcp/` or `tools/`. Its dependents are:

- `package.json:40` — the `audit:playground` npm script (CLI entry, argv contract
  unchanged)
- `tests/scripts/playground-verify-rule-sentinel.test.ts` — imports `buildScrapeExpr`,
  `firstNonEmptyLine`, and now the two new exports
- `docs/astplayground.md` and `AGENTS.md:2458` — prose, describing the CLI

Touched entry points: `buildScrapeExpr` (existing export, return shape extended with
`sourceLen`), `main()`'s poll loop, and two new exports (`initialPollStability`,
`trackStableUnmatched`) declared in the sibling `.d.mts`. Both new exports are pure and
allocate one small object per poll.

No hot path: the poll loop runs at 250ms intervals in a manual verification run. The cost
delta is the opposite sign from usual — a stable mismatch now concludes in roughly 750ms
instead of the full `--timeout` (40s in the reported case).

`module_report --blastRadius` is not applicable: the review graph indexes TypeScript
sources, and this is a `.mjs` script with a hand-written declaration file. Recorded as
unavailable-with-reason rather than empty.

## Observability

The record that proves this works is the harness's own stderr result JSON on the
`!scrape.sentinelFound` path, which now carries `polls` and `concluded_early` alongside
`matches_reported_by_page` and `engine_ms`. `concluded_early: true` with a low `polls` is
the positive signal that the fail-fast path fired; `engine_ms` shrinking from ~40000 to
under ~1000 on a mismatching rule is the same evidence in wall-clock form.

The success path is unchanged and still emits no new field.

## Class sweep

Defect shape: a normalization applied to only ONE side of a comparison, and a hand-listed
enumeration of the substitutions a renderer might use — a member of the recurring
"partial-normalization equality" family. Both halves are fixed here: `collapseWhitespace`
runs on both sides, and it collapses whitespace *runs* generically instead of naming
characters.

Whole-tree sweep for the shape, `clients/` plus `tools/`, `mcp/`, `scripts/`, `index.ts`:
`rg -n "fromCharCode\(160\)|\\\\u00a0|nbsp" -g '!tests/**' -g '!*.js' -g '!CHANGELOG.md'`
returns exactly two files.

| Site | Verdict |
| --- | --- |
| `scripts/playground-verify-rule.mjs` | The member. Fixed by this PR |
| `clients/path-utils.ts:664,723` | Not a member. `HostPathVariantResolution`'s `"narrow-nbsp"` is a candidate *generator* — it builds alternate path spellings and probes each for existence. There is no two-sided string comparison to normalize asymmetrically, so the shape cannot occur |

Class of size 1, proven by that grep.

Population sweep, family "renderer-text scrapers in `scripts/`": `playground-verify-rule.mjs`
is the only member — `playground-chrome.mjs` and `playground-cdp.mjs` are transport, they
never compare rendered text. Consolidation verdict: nothing to fold; a one-member family
does not earn a shared seam, and `collapseWhitespace` deliberately lives inside the
generated browser expression, where it must be self-contained.


apmantza:

apmantza:
Merged — thank you, @AngriestBird. The review re-ran the original tab repro live (78 polls burning 41.5 s → match on poll 1), verified the fail-fast (71 polls → 3 under simulated drift), and — since you flagged that Monaco's actual tab substitution was unverified in your environment — dumped it over CDP: tabs render as variable-length NBSP runs, which is exactly what your collapse-both-sides design handles and why the simpler one-sided mapping could never work. Your declared uncertainty resolved in your favor. All three #2306 criteria met; the fragment's refs was aligned to closes as a maintainer delta. As always: AI-generated review, maintainer-supervised.

Issue #2146, opened 2026-08-26
Module-scope singletons break when pi evaluates pi-lens more than once per process


apmantza:
Evidence (dogfood pass 3, 2026-08-26, new build d5fe7d3a installed 11:25Z)
- host_boot is emitted at module top level (index.ts:472): one record = one module evaluation. Three hosts multi-evaluated in one pid: pid 84428 host_boot=1/binds=3 (guard works); pid 71216 host_boot=4/binds=0; pid 30236 host_boot=9/binds=1. The #473/#2129 guard declines correctly ONLY in the single-evaluation process.
- One 90s burst in pid 71216 on the NEW build: three session_start_total mode=full (plegma + two pi-agent temp roots, 11:32:00-01), then THREE identical word-index full_rebuilds (79,437 / 80,064 / 81,309 ms, identical postingEntries=454143 and residentBytes=6207520) — 240.8s of CPU for one index.
- Concurrently, three instance-registry-corrupt degradation records (11:32:16-25): registryMutationTail (clients/instance-registry.ts:468, docstring claims one shared per-process tail) exists once PER MODULE INSTANCE; concurrent read-modify-write tore instances.json — both temp roots absent from projectRoots, pid 8232's entry lost. projectRoot pinning held (#2130's clobber IS fixed); the root set is not populated.

Root cause
- classifySessionStart reaches secondary-root only when hasPrior===true; hasPrior reads THAT MODULE INSTANCE's scope. A second evaluation starts empty → "primary" → full battery. Same shape for every module-scope mutable singleton.

Acceptance criteria
1. Per-process-unique state moves off module scope onto globalThis-keyed process singletons, following the existing precedent (clients/runtime-session.ts:1826 __piLensFirstSessionDone / __piLensWarmupScheduled).
2. A test evaluates the session-lifecycle module twice and proves the second evaluation sees the first's registered primary; deleting the sharing reds it.
3. A test drives concurrent registry mutations from two module instances: no lost update, no instance-registry-corrupt record.
4. Live proof: a pid with host_boot>1 emits concurrent_session_bind with classification secondary-root per temp root, and at most ONE full_rebuild per distinct content root.

Observability
- concurrent_session_bind.metadata.classification (latency.log), full_rebuild count per burst (word-index.log), absence of instance-registry-corrupt. Gap to close: add the evaluation ordinal to host_boot.metadata so multi-eval is greppable.

Class sweep (required in the fix PR)
[...]

PR #2312, opened 2026-08-27
feat(observability): session-end rollup of declined session binds (#2249)

### LINKED ISSUE #2249: No rollup for declined session starts: concurrent_session_bind is per-bind only
The remainder that keeps #2130 from closing after PR #2241; parked there because it is its own subsystem and process.

## Shape
- `logConcurrentSessionBind` (`clients/session-start-observability.ts`) is the only writer, called once per bind from `index.ts:1821`; nothing aggregates.
- The rollup belongs in the extension host (a `*_rollup` latency phase), not `pilens_health` (MCP server process).
- A correct counter needs `getProcessSingleton` with a family version and a session-boundary reset (defect shapes 25 and 17) — not the module-scope `let` the existing rollups use — plus phase-name registration in `tests/scripts/latency-log-phases.test.ts`, `scripts/compat-smoke-behavioral.mjs`, and `docs/subagent-compat.md`.

## Acceptance criteria
- A bounded rollup of declined/classified binds emitted at session end, reset per session, registered in the phase-name sweeps above.
- Closing this closes the last open item on #2130.

Refs #2130, PR #2241.

AngriestBird:
## Summary

Declined session starts now get a session-end summary: one `concurrent_session_bind_rollup`
row counting declined binds by classification (`concurrent-secondary` / `secondary-root`,
plus a bounded `unclassified` bucket for the legacy `classification?` path), emitted from
the primary's `session_shutdown` at the same placement as the bus-event and
path-attribution rollups, and reset per session. This is the remainder that kept #2130 open
after PR #2241.

Design points worth calling out:

- Counters live behind `getProcessSingleton` (family `concurrent-session-bind-rollup`, v1),
  not a module-scope `let`. The module graph evaluates more than once per process — #2146
  measured nine evaluations for one pid — so a `let` would only ever count the binds routed
  through its own copy (defect-shape 25).
- Reset fires on the primary-continuation `session_start` path only, after the declined
  bind's own `return` at `index.ts:1855`. It re-arms the counters for a primary that
  crashed or was killed before reaching `session_shutdown` (defect-shape 17), without
  erasing a live sibling's tally.
- Emit is a no-op when nothing was declined, matching the sibling rollups' "no noise on an
  ordinary session" shape.

Closes #2249

### Fix round 1 (review)

`oxfmt` was failing on four of the touched files. Formatting-only commit; no behavior
change.

## Tests

- EDIT `tests/clients/session-start-observability.test.ts` — the unit seam. Covers: each
  classification increments its own bucket; an unknown/absent `classification` lands in
  `unclassified` rather than being dropped; emit is a no-op at zero; emit clears the
  counters; and a `vi.resetModules()` case that re-imports the module and proves the count
  survives, which is the assertion that would red if the counters regressed to a
  module-scope `let`.
- EDIT `tests/index-integration.test.ts` — the wiring seam. Proves the rollup is emitted
  from `session_shutdown` and reset from the primary `session_start`, i.e. that the two
  calls are actually reached, not merely exported.
- EDIT `tests/scripts/latency-log-phases.test.ts` — registers
  `concurrent_session_bind_rollup` in the phase-name sweep, so the new phase cannot be a
  phase no sweep knows about.
- Red-first, recorded by the authoring run: reverse-applying the implementation patch
  failed 6 of 7 unit tests and 3 of 3 new integration tests.
- Re-run in this worktree after `npm run build`: 67/67 green across the three files.
  `npm run lint` and `npm run fmt:check` clean. Full suite deferred to CI per #1112.
- Screens: the `vi.resetModules()` case exists specifically to defeat the parallel-path
  failure (a second module evaluation with its own counters); the integration test pins the
  call site rather than the export (wrong-layer pin); the zero-total case makes an
  invisible skip visible by distinguishing "nothing declined" from "emit never ran".

## Test assessment

Per touched test file:

- `session-start-observability.test.ts` — uniquely pins the counter semantics
  (bucketing, unclassified fallback, no-op at zero, clear-on-emit) and the
  process-singleton backing. Nothing here is covered elsewhere; the singleton conformance
  suite pins `getProcessSingleton` itself, not this family's use of it.
- `index-integration.test.ts` — uniquely pins that emit and reset are wired at the right
  two places in `index.ts`. The unit file cannot prove this; it only proves the functions
  behave when called.
- `latency-log-phases.test.ts` — uniquely pins phase-name registration. One line, and it
  is the file's whole purpose.

Nothing became redundant, and no test was removed or loosened. The overlap between the unit
and integration files is deliberate and narrow: the integration test asserts *reachability*
and the unit test asserts *behavior*, so neither subsumes the other.

## Blast radius

Touched production modules and their dependents (`rg` over non-test, non-compiled sources):

| Module | Production dependents |
| --- | --- |
| `clients/session-start-observability.ts` | `index.ts` only (plus `config/dependency-cruiser-eager-allowlist.json`, which lists it, and `docs/subagent-compat.md`) |
| `clients/latency-logger.ts` | Broad, but the change is one string added to the `LAST_PHASE_EXCLUDED` set — no signature, no export, no control-flow change |
| `index.ts` | Entry point; no dependents |

Entry points touched: the `session_start` handler (one added call on the
primary-continuation path) and the `session_shutdown` handler (one added call, alongside
the two existing rollups).

Hot-path cost: `logConcurrentSessionBind` gains one `getProcessSingleton` lookup plus one
integer increment. That function only runs on a *declined* bind — at most a handful of
times per process — so this is not a per-spawn, per-file or per-render path. The
session-end emit is one `logLatency` row, and only when a decline actually occurred.

`module_report` with `blastRadius: true` was run on
`clients/session-start-observability.ts` and returned `"blastRadius":"none"` with
`provenance.usedBy: "none"` — the review-graph cache is cold in this worktree, so the
transitive walk is unavailable. Recorded rather than reported as empty; the `rg` sweep
above is the substitute.

## Observability

The record that proves this works in production is the new
`concurrent_session_bind_rollup` latency phase (`clients/latency-logger.ts` sink,
`filePath: <projectRoot>`, `durationMs: 0`), whose metadata carries the three per-
classification counts. Its absence on a session where no bind was declined is itself the
designed signal, matching `session_end_bus_rollup` and
`path_attribution_verified_rollup`.

The phase is registered in all three places a new phase has to be registered:
`tests/scripts/latency-log-phases.test.ts`, `scripts/compat-smoke-behavioral.mjs`, and
`docs/subagent-compat.md`. It is added to `LAST_PHASE_EXCLUDED` so a zero-duration
session-end summary can never take the blame for a stall it did not cause.

Gap named honestly: `scripts/compat-smoke-behavioral.mjs` assertion 6 still records this as
NOT ASSERTED. Observing either the per-bind phase or the rollup requires reproducing the
in-process double-`bindExtensions()` model with a real `createAgentSession()`, which needs
model/provider config that cannot be cheaply stubbed. This PR corrects that assertion's
stale claim (it said `decideSessionStart` had no call site; it has had one since PR #477)
but does not close the gap. That is a documented TODO in `docs/subagent-compat.md`, not new
debt from this change.

## Class sweep

Defect shape 25 — session-scoped counters held in module-scope state, which undercounts
whenever the module graph is evaluated more than once per process (#2146: nine evaluations
for one pid).

Whole-tree sweep, `clients/` plus `tools/`, `mcp/`, `scripts/`, `index.ts`, for the
`*_rollup` counter family:

| Member | Backing | Verdict |
| --- | --- | --- |
| `clients/session-start-observability.ts` (`concurrent_session_bind_rollup`) | `getProcessSingleton` | This PR. Correct by construction |
| `clients/bus-events-logger.ts` (`session_end_bus_rollup`) | module-scope `const eventRollupCounts = new Map(…)` (line 110) | Pre-existing exposure to the same shape. NOT fixed here — out of scope for #2249, and correcting it changes counts other tests pin |
| `clients/path-attribution-telemetry.ts` (`path_attribution_verified_rollup`) | module-scope `let verifiedGuessCount = 0` (line 7) | Same verdict |

This is explicitly not a class of size 1: two sibling rollups carry the same latent shape.
This PR deliberately does not convert them — that is a separate change with its own
red-first proof and its own blast radius — and the new code takes the correct form rather
than copying theirs, so the family does not grow.

Consolidation verdict: the family stays distributed for now. Folding all three onto one
"session rollup counter" seam is the right end state and would let a single
`getProcessSingleton` family serve every rollup, but it needs the two siblings converted
first. Filing that as a follow-up rather than smuggling it into a #2249 PR.

## Review round

Maintainer delta 2f755c80/4461a882 on three review findings, contributor commits untouched: the sequential-replacement session_start now emits the pending rollup before resetting (a crashed prior primary's tally was silently discarded; red proof `expected [] to have a length of 1`); the phase-sweep test drives the real production functions through the real log sink instead of literal objects (red proof: TypeError on the pre-feature base); and the new LAST_PHASE_EXCLUDED entry is pinned (deleting it reds `expected 'concurrent_session_bind_rollup' to be 'provider_request'`). 112/112 targeted + 101/101 governance green. The review also surfaced a structural gap filed as #2319: getProcessSingleton-backed session state with closure-inline resets is invisible to the session-state conformance sweep — not this PR's to fix.

### DIFF
--- a/.changelog/feat-2249-declined-bind-rollup.md
+++ b/.changelog/feat-2249-declined-bind-rollup.md
@@ -1,0 +1,5 @@
+---
+section: Added
+---
+
+- **Roll up declined concurrent session starts (refs #2130, closes #2249)** — `logConcurrentSessionBind` (#473) previously wrote one `concurrent_session_bind` record per declined bind with nothing to aggregate them. `index.ts`'s `session_shutdown` handler now also logs one bounded `concurrent_session_bind_rollup` row per primary session, counting declines by classification (`concurrent-secondary` / `secondary-root`), a no-op when none occurred. The counter lives behind `getProcessSingleton` with its own family version, not a module-scope `let` — pi evaluates the pi-lens module graph more than once per process, and a bare `let` would only ever see the binds routed through its own copy. It resets both at the primary's own `session_start` (so a crashed prior primary can't leak a stale tally) and after each emission.
--- a/clients/latency-logger.ts
+++ b/clients/latency-logger.ts
@@ -119,6 +119,11 @@ let recentPhases: Array<{ phase: string; ts: string }> = [];
  * usage records. It reports no new work and therefore cannot own last-phase
  * stall attribution.
  *
+ * #2249: `concurrent_session_bind_rollup` is the same shape as
+ * `session_end_bus_rollup`/`path_attribution_verified_rollup` above — a
+ * zero-duration session-end summary of records already logged individually,
+ * not new work of its own.
+ *
  * #2044: `test_runner_failed_target_state` is a zero-duration decision after a
  * bounded filesystem probe. The surrounding turn-end test-selection phase owns
  * any real work, so this row must not replace it in stall attribution.
@@ -143,6 +148,7 @@ const LAST_PHASE_EXCLUDED = new Set([
 	"lsp_notify_write_late_landed",
 	"degradation_ledger",
 	"path_attribution_verified_rollup",
+	"concurrent_session_bind_rollup",
 ]);
 
 /**
--- a/clients/session-start-observability.ts
+++ b/clients/session-start-observability.ts
@@ -91,9 +91,7 @@ export function getConcurrentSessionBindRollupCounts(): Record<
  * A no-op when nothing was ever declined this session, matching those same
  * rollups' "no noise on an ordinary session" shape.
  */
-export function emitConcurrentSessionBindRollupAtSessionEnd(
-	cwd: string,
-): void {
+export function emitConcurrentSessionBindRollupAtSessionEnd(cwd: string): void {
 	const counts = bindRollupCounts();
 	const total =
 		counts["concurrent-secondary"] +
--- a/clients/session-start-observability.ts
+++ b/clients/session-start-observability.ts
@@ -1,10 +1,52 @@
 import { logLatency } from "./latency-logger.js";
+import { getProcessSingleton } from "./process-singletons.js";
 
 /**
  * The complete observability path for a concurrent secondary. It deliberately
  * emits one bind record only: secondary sessions skip every primary reset and
  * hydration phase.
  */
+
+// #2249: bounded per-session rollup of declined binds, counted at this shared
+// seam so every decline the process's PRIMARY has seen this session —
+// `concurrent-secondary` and `secondary-root` alike — is summarized in one row
+// at session end instead of requiring a reader to hand-count
+// `concurrent_session_bind` lines. Process-singleton backed (AGENTS.md catalog
+// shape 25), NOT a module-scope `let` like the other `*_rollup` counters
+// (bus-events-logger.ts, path-attribution-telemetry.ts): a `let` here exists
+// once per module EVALUATION, not once per process, and #2146 measured one pid
+// evaluating this exact graph nine times — a bare `let` would only ever see
+// the binds routed through its own copy, undercounting every rollup that
+// isn't reached through the first evaluation. `decideSessionStart` never
+// returns a declined classification outside `concurrent-secondary`/
+// `secondary-root` (session-lifecycle.ts), so those two are the whole
+// contract; `unclassified` is a bounded third bucket for the legacy callers
+// this function's `classification?` param still accepts.
+const BIND_ROLLUP_FAMILY = "concurrent-session-bind-rollup";
+const BIND_ROLLUP_VERSION = 1;
+
+type BindRollupClassification =
+	| "concurrent-secondary"
+	| "secondary-root"
+	| "unclassified";
+
+function bindRollupCounts(): Record<BindRollupClassification, number> {
+	return getProcessSingleton(BIND_ROLLUP_FAMILY, BIND_ROLLUP_VERSION, () => ({
+		"concurrent-secondary": 0,
+		"secondary-root": 0,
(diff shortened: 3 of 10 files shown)


apmantza:

apmantza:
Review complete: MERGE. The maintainer delta closed all three findings and the verify round re-proved each with independent mutations — the emit-before-reset fires exactly once on the replacement path and never on a clean session, the phases test now reds behaviorally when the real emit is neutered (not just on a missing export), and the LAST_PHASE_EXCLUDED entry reds on deletion. 112/112 targeted, 221/221 governance, Unit tests and Lint green on 4461a882. Your commits are untouched.

This is your draft, so it merges whenever you mark it ready — no rush from our side.

One heads-up worth fixing before you flip it: your commits are authored as t <t@t.local>, which GitHub maps to no account, so the merge would not credit you. Adding t@t.local to your GitHub email addresses (or re-authoring the two commits) preserves your contribution graph credit.

As always: AI-generated review, maintainer-supervised.

AngriestBird:
The t.local was me testing a new workflow and missed that they were not reset. Will fix shortly.

AngriestBird:
Marked ready — thanks for the maintainer delta and the independent re-verification of all three findings.

On the authorship note: my commits are authored as `t <t@t.local>`, which maps to no GitHub account, so the merge will not credit me. That is a local git identity I am not going to change from inside a session; I will sort it out on my side. Do not hold the merge for it.

PR #2313, opened 2026-08-27
fix: add a size axis to the five mtime-only content memos (#2300)

### LINKED ISSUE #2300: Class sweep remainder: five mtime-only content memos without a size axis (from #2297)
# Evidence

PR #2297 (closes #2285) fixed the authoritative-snapshot cache's missing size axis. Its class sweep named five further plausible members of the mtime-only-content-memo class, verified as plausible by the review but not fixed there:

- `mcp/build-staleness.ts:93`
- `clients/lsp/diagnostic-binding.ts:466`
- `clients/lsp/workspace-diagnostics-cache.ts:436`
- `clients/recent-touches.ts:277` — short-circuits on `_lastSeenMtimeMs === mtimeMs` with no size axis at all (reviewer-confirmed)
- `clients/installer/index.ts` (lead only)

# Root cause / class

Stat-memo gates on mtime alone; a same-mtime-bucket external rewrite with different content is invisible. AGENTS.md catalog shape: stat-memo needs mtime+size (+content only where the tier's cost budget allows — see #2292's lesson against unbounded content hashing on hot paths).

# Acceptance criteria

- Each named site gets a per-member verdict: fixed with a size axis, or exempted with a recorded reason (immutability, downstream self-guard, cost).
- Fixed members carry a regression test that varies SIZE (never same-length content — the NTFS mtime-granularity collision, #2287 N1).
- Work-count any added stats; no content hashing on per-edit paths.

# Observability record

None of these sites logs a stale-serve; the defect is visible only as wrong behavior. Per-site fix should note whether a degradation record is warranted.

Refs #2285 (parent, closed by #2297), PR #2297 (sweep source), #2262/#2287 (sibling class fix in the rules cache).

AngriestBird:
## Summary

All five sites from #2297's class sweep now carry a size axis (or a recorded exemption), so
a same-mtime-bucket external rewrite with different bytes is no longer invisible. Every fix
reuses a stat call already in flight — zero added syscalls, and no content hashing on any
per-edit path.

Both persisted caches (`workspace-diagnostics-cache`, installer probe cache) add
`sizeBytes` as **optional** and fail OPEN (mtime-only) for pre-existing entries — the
`depIndexAtScan` precedent — self-healing on the entry's next write, and for the probe
cache also within its 24h TTL. No cache-version bump, no forced re-scan on upgrade.

Closes #2300

### Per-site verdicts

| Site | Verdict |
| --- | --- |
| `mcp/build-staleness.ts` | Fixed — `BuildStamp`/`StatFn` carry `size` (in-memory only, so required, no compat shim) |
| `clients/lsp/diagnostic-binding.ts` | Fixed — the `diskHashByPath` memo keys on mtime AND size, because it guards a content-hash recompute and a stale memo therefore serves a stale hash |
| `clients/lsp/workspace-diagnostics-cache.ts` | Fixed (own-file check). Dep-freshness loop exempted — it is a one-sided threshold against `scannedAt`, not an equality memo, and stores no per-dep value to extend |
| `clients/recent-touches.ts` | Fixed — `_lastSeenSizeBytes` alongside the mtime watermark |
| `clients/installer/index.ts` | Was a lead only; the sweep found a real member — probe-cache `ProbeCacheEntry` gains `sizeBytes`. Install-lock age check and PATH-hash memo were reviewed and are not members |

### Fix round 1 (CI)

The three same-mtime-bucket regression tests asserted the forced mtime against the
*nominal* `Date.now()` value they passed to `utimesSync`. `mtimeMs` is derived from a
nanosecond timestamp, so a whole-millisecond `Date` reads back as e.g.
`1787865854006.999` on ext4 — green locally, red on CI:

```
AssertionError: expected 1787865854006.999 to be 1787865854007 // Object.is equality
 ❯ tests/clients/recent-touches.test.ts
AssertionError: expected 1787865954578.999 to be 1787865954579 // Object.is equality
 ❯ tests/clients/lsp/diagnostic-binding.test.ts
```

They now read the pinned value back once and compare the post-rewrite stat against *that*.
Two reads of the same stored value are exact by construction, so the assertion is
platform-independent while still proving the same-bucket collision it exists to prove.
`workspace-diagnostics-cache.test.ts` carried the same latent assertion (it happened to
pass) and was corrected too. Test-only change; no production code touched in this round.

## Tests

New or edited, with what each pins:

- `tests/mcp/build-staleness.test.ts` (EDIT) — a rebuild at the same mtime with a different
  size reads `stale: true`. Regression proof for the in-memory site.
- `tests/clients/lsp/diagnostic-binding.test.ts` (EDIT) — `same-mtime-bucket rewrite with a
  different size is not masked by the mtime-only memo (#2300)`. Pins that the memo forces a
  re-hash, so `boundToCurrentDisk` returns `false` against a stale binding instead of
  reusing the old content's hash.
- `tests/clients/lsp/workspace-diagnostics-cache.test.ts` (EDIT) — end-to-end through a
  real context: `record`'s `sizeBytes` reaches `isEntryFresh` via `lookup`, and a
  same-bucket rewrite invalidates. Also pins the fail-open path for an entry with no
  `sizeBytes`.
- `tests/clients/recent-touches.test.ts` (EDIT) — a same-bucket append with a longer JSON
  body is not dropped by the mtime watermark. Uses a real `setTimeout` gap so the second
  entry's `ts` beats the consumed cursor, and forces the mtime back with `utimesSync` so
  there is no timing race in the test itself.
- `tests/clients/installer/probe-cache.test.ts` (EDIT) — a same-bucket binary replacement
  misses the cache; a pre-#2300 entry without `sizeBytes` still hits (fail-open).

Every regression case varies **size**, never same-length content — same-length is the
#2287 N1 case a size axis provably cannot catch, and pinning it here would be a loose
bound dressed up as coverage.

- Red-first, recorded by the authoring run: source-only `git apply -R`, rebuild, rerun — 8
  new/updated assertions failed pre-fix while 129 pre-existing tests in the same files
  stayed green; 137/137 after re-apply.
- Sibling sweep over files referencing the changed symbols: 33 files, 497 tests, green.
- Re-run in this worktree after the fix-round-1 test correction and `npm run build`:
  117/117 green across the five files. `npm run lint`, `npm run fmt:check` and
  `npm run changelog:check` clean. Full suite deferred to CI per #1112.
- Screens: each new case drives the real cache object rather than asserting on the entry
  shape (wrong-layer pin); the fail-open cases make the pre-#2300-entry path visible
  instead of letting it skip silently; the `utimesSync` pinning removes env-dependent
  timing, and fix round 1 removed the last env-dependent assertion.

## Test assessment

Per touched test file:

- `build-staleness.test.ts` — uniquely pins the in-memory stamp. It is the only site with
  no persistence, so it is also the only one where `size` could be made required.
- `diagnostic-binding.test.ts` — uniquely pins the *memo* layer. The workspace-cache test
  cannot reach it; they are different caches with different lifetimes.
- `workspace-diagnostics-cache.test.ts` — uniquely pins `record` → `lookup` →
  `isEntryFresh` threading, which no unit-level assertion on `isEntryFresh` alone would
  catch (a `record` that never wrote `sizeBytes` would still pass those).
- `recent-touches.test.ts` — uniquely pins the module-scope watermark, including the
  reset in `_resetRecentTouchesForTests`.
- `probe-cache.test.ts` — uniquely pins the fail-open contract for pre-existing persisted
  entries. This is the one assertion that would catch an over-eager fix that made
  `sizeBytes` required and invalidated every cache on upgrade.

Nothing became redundant; nothing was removed or loosened. The five cases look alike but
sit on five different caches with different persistence and lifetime rules, so none
subsumes another. The dep-freshness loop in `workspace-diagnostics-cache.ts` has no new
test because it is a recorded exemption, not a fix — adding a test there would pin
behavior the PR deliberately did not change.

## Blast radius

Touched production modules and their production dependents (`rg` over non-test,
non-compiled sources):

| Module | Dependents |
| --- | --- |
| `mcp/build-staleness.ts` | `mcp/server.ts:84` |
| `clients/lsp/diagnostic-binding.ts` | 9 importers, but the changed symbol is `createDiskBindingCache`, whose only two call sites are `clients/lsp/workspace-diagnostics-cache.ts:624` and `clients/lsp/index.ts:1254`. The other seven import `hashDiagnosticContent` / `touchCoverageGap`, neither of which changed |
| `clients/lsp/workspace-diagnostics-cache.ts` | `clients/lsp/index.ts:132`, `tools/lsp-diagnostics.ts:24`, `clients/lsp/client.ts:69` |
| `clients/recent-touches.ts` | `index.ts:105`, `clients/bus-publish.ts:22` |
| `clients/installer/index.ts` | installer entry points only; `ProbeCacheEntry` is module-private |
| `clients/lsp/index.ts`, `tools/lsp-diagnostics.ts` | Call-site updates (pass `size` through); no exported signature change |

Signature changes, all additive:

- `WorkspaceDiagnosticsCacheContext.record` gains a 6th **optional** parameter. Existing
  callers compile unchanged; both in-tree callers were updated to pass it.
- `StatFn` / `BuildStamp` gain a required `size`. In-memory only, and `realStat` already
  returns a full `fs.Stats`, so the only work was at the test doubles.

Hot paths touched, with the measured cost delta: **zero added syscalls at every site.**
Each one reads `.size` off a `Stats` object the code had already obtained —
`isEntryFresh`'s own-file `statSync`, `boundToCurrentDisk`'s `statSync`, the LSP sweep's
per-file scan stat, `readCrossProcessTouchesForTurnStart`'s per-turn `stat`, and
`checkProbeCache`'s existing `stat`. `clients/lsp/index.ts` previously called
`statSync(...).mtimeMs` inline and now binds the same call's result, which is the same
single syscall. The added work per file is one integer comparison.

`module_report` with `blastRadius: true` returns `"blastRadius":"none"` with
`provenance.usedBy: "none"` in this worktree — the review-graph cache is cold, so the
transitive walk is unavailable. Recorded rather than reported as empty; the `rg` sweep
above is the substitute.

## Observability

No new degradation records, and none warranted: each corrected verdict feeds a mechanism
that is already observed.

- `mcp/build-staleness.ts` — the existing `stale` routing in `mcp/server.ts`. A
  same-bucket rebuild now routes to fresh instead of being served warm.
- `clients/lsp/diagnostic-binding.ts` — the existing #1095 binding demote pipeline. The
  entry now reads `false` instead of a stale `true`.
- `clients/lsp/workspace-diagnostics-cache.ts` — the existing cache-miss path; the file is
  re-scanned rather than replayed.
- `clients/recent-touches.ts` — the existing `agent_nudge` phase, whose
  `originCrossProcess` count now includes entries the mtime watermark used to drop.
- `clients/installer/index.ts` — the existing `logSessionStart` line, whose miss message
  now reads `miss (mtime/size changed)` rather than `miss (mtime changed)`. That wording
  change is the one directly greppable proof in a log.

Gap named: nothing distinguishes a *size*-triggered invalidation from an *mtime*-triggered
one except that installer message. Adding a discriminator to the other four would mean a
new field on hot-path records to observe an event that is rare by construction, so it is
deliberately not done.

## Class sweep

Defect shape: an **equality memo keyed on mtime alone**, where the guarded value depends on
file *content*. Coarse filesystem mtime granularity (the NTFS case named in #2287 N1) lets
a rewrite land in the same bucket, and the memo then serves a value derived from the old
bytes.

This PR *is* the class sweep for #2297's finding — the whole-tree sweep across `clients/`,
`tools/`, `mcp/`, `scripts/` and `index.ts` produced the five-member population in the
Per-site verdicts table above, with a per-member verdict for each. Restated as the two
exemptions, since a verdict of "exempt" needs its reason on the record:

- The **dep-freshness loop** in `workspace-diagnostics-cache.ts` compares each import's
  mtime against `entry.scannedAt` — a one-sided threshold, not an equality memo. There is
  no stored per-dep value to extend with a size, and a size axis would not make a
  threshold comparison more correct.
- The installer's **install-lock age check** and **PATH-hash memo** are time-window and
  string-derived respectively; neither guards a content-derived value.

Explicitly **not** claimed: a size axis does not close the same-length-rewrite case. Two
files with identical length and different bytes in the same mtime bucket still collide.
Closing that needs a content hash, which is the `contentHash` tier this PR deliberately
leaves alone on the per-edit paths for cost reasons. The size axis narrows the window; it
does not eliminate it.

Consolidation verdict: the family stays distributed. The five sites are five different
caches with different lifetimes, persistence models and compat constraints (two persisted
and fail-open, three in-memory), and the shared logic is one integer comparison. Folding
them onto a common "freshness stamp" seam would cost more than it saves, and the real
single-source-of-truth already exists at the layer that matters — every site reads its
axes from one `Stats` object obtained by one stat call.

## Review round

The verify round confirmed all five members with independent mutations (7 of 9 red; the two survivors were the unpinned production wirings), endorsed the exemption arguments, and re-measured the red run (figures above corrected from the authoring run's stale totals). Maintainer delta f7d2d029 closes the C1 gap: both production `record()` call sites now carry end-to-end pins that read the persisted entry back through `loadWorkspaceDiagnosticsCache` — each reds when its wiring argument is neutered to `undefined` (expected undefined to be 13 / to be 2), so a dropped `sizeBytes` can no longer hide behind the legacy fail-open.

### DIFF
--- a/.changelog/2300-mtime-size-axis-sweep.md
+++ b/.changelog/2300-mtime-size-axis-sweep.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Add a size axis to five mtime-only content memos (closes #2300)** — The MCP warm-build staleness gate, the LSP diagnostic-binding content-hash memo, the workspace-diagnostics per-file freshness cache, the cross-process recent-touches watermark, and the installer's tool-path probe cache now all validate byte size alongside mtime, so an external rewrite landing in the same coarse-granularity mtime bucket with a different length is no longer served as unchanged. Every fix reuses a stat call already being made — no added per-edit I/O.
--- a/clients/installer/index.ts
+++ b/clients/installer/index.ts
@@ -1613,6 +1613,16 @@ export function resetResolvedPathCache(): void {
 interface ProbeCacheEntry {
 	path: string;
 	mtimeMs: number;
+	/**
+	 * #2300: byte size at the same stat call as `mtimeMs`. A binary replaced
+	 * in-place (an upgrade landing in the same coarse-granularity mtime tick)
+	 * would otherwise pass the mtime-only freshness check unnoticed. Optional
+	 * so a pre-#2300 persisted entry doesn't need a version bump to stay
+	 * loadable — `checkProbeCache` fails OPEN (mtime-only) when absent; it
+	 * self-heals on the entry's next `updateProbeCache` write or 24h TTL
+	 * expiry, whichever comes first.
+	 */
+	sizeBytes?: number;
 	cachedAt: number;
 	/**
 	 * True when the `getToolPath` resolution that produced `path` saw a
@@ -1931,9 +1941,16 @@ export async function checkProbeCache(
 	try {
 		await fs.access(entry.path);
 		const stat = await fs.stat(entry.path);
-		if (stat.mtimeMs !== entry.mtimeMs) {
+		// #2300: size alongside mtime, from the same stat — a binary replaced
+		// in-place within the same coarse-granularity mtime tick would otherwise
+		// pass this check unnoticed. `entry.sizeBytes` absent (pre-#2300 entry)
+		// fails open to the mtime-only check (see the field's doc comment).
+		if (
+			stat.mtimeMs !== entry.mtimeMs ||
+			(entry.sizeBytes !== undefined && stat.size !== entry.sizeBytes)
+		) {
 			logSessionStart(
-				`auto-install probe-cache ${toolId}: miss (mtime changed)`,
+				`auto-install probe-cache ${toolId}: miss (mtime/size changed)`,
 			);
 			delete cache[toolId];
 			markProbeCacheChange(toolId, null);
@@ -1970,6 +1987,7 @@ export async function updateProbeCache(
 		const entry: ProbeCacheEntry = {
 			path: resolvedPath,
 			mtimeMs: stat.mtimeMs,
+			sizeBytes: stat.size,
 			cachedAt: Date.now(),
 			...(transient && { transient: true }),
 		};
--- a/clients/lsp/diagnostic-binding.ts
+++ b/clients/lsp/diagnostic-binding.ts
@@ -447,30 +447,39 @@ export function createDiskBindingCache(): DiskBindingCache {
 	// the same path (`SUB\a.ts` vs `sub/a.ts`) can't produce a duplicate memo or a
 	// false miss. Ephemeral (slash-fold + win32-lowercase, no realpath I/O) — the
 	// keys are file paths this process is already stat'ing on the hot read path.
-	const diskHashByPath = new PathKeyedMap<{ mtimeMs: number; hash: string }>(
-		normalizeEphemeralMapKey,
-	);
+	// #2300: the memo guards a CONTENT HASH recompute, so a stale memo serves a
+	// stale hash — `size` is the cheap second axis alongside `mtimeMs`, from the
+	// SAME stat call, so a same-mtime-bucket external rewrite is not masked.
+	const diskHashByPath = new PathKeyedMap<{
+		mtimeMs: number;
+		size: number;
+		hash: string;
+	}>(normalizeEphemeralMapKey);
 	return {
 		boundToCurrentDisk(filePath, stored) {
 			// No fingerprint captured (version-less server) → unknown, never false.
 			if (stored.contentHash === undefined) return "unknown";
-			let mtimeMs: number;
+			let stat: fs.Stats;
 			try {
-				mtimeMs = fs.statSync(filePath).mtimeMs;
+				stat = fs.statSync(filePath);
 			} catch {
 				// Can't stat (deleted/unreadable): can't disprove the binding either,
 				// so stay honest — "unknown", never a manufactured false.
 				return "unknown";
 			}
 			let cached = diskHashByPath.get(filePath);
-			if (!cached || cached.mtimeMs !== mtimeMs) {
+			if (
+				!cached ||
+				cached.mtimeMs !== stat.mtimeMs ||
+				cached.size !== stat.size
+			) {
 				let diskHash: string;
 				try {
 					diskHash = hashDiagnosticContent(fs.readFileSync(filePath, "utf-8"));
 				} catch {
 					return "unknown";
 				}
-				cached = { mtimeMs, hash: diskHash };
+				cached = { mtimeMs: stat.mtimeMs, size: stat.size, hash: diskHash };
 				if (diskHashByPath.size >= DISK_BINDING_MEMO_MAX) {
 					diskHashByPath.clear();
 				}
--- a/clients/lsp/index.ts
+++ b/clients/lsp/index.ts
@@ -7632,6 +7632,10 @@ export class LSPService {
 		// mtime it was ACTUALLY scanned at (not re-stat'd after the fact, which
 		// could race a concurrent edit).
 		const scannedMtimeByFile = new Map<string, number>();
+		// #2300: size from the SAME stat call as the mtime above — zero extra
+		// syscalls — so the cache write below can give `isEntryFresh` a size
+		// axis alongside the mtime.
+		const scannedSizeByFile = new Map<string, number>();
 
 		// Group files by their primary language server (#387, extracted as
 		// `groupFilesByPrimaryServer` for #631). tsserver — and most servers — is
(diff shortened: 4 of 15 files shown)


apmantza:

apmantza:
Merged — thank you, @AngriestBird. The review verified all five members with independent per-member mutations, endorsed both exemptions on their mechanisms, and praised the read-back mtime assertions as stronger than what we required on the sibling #2297 fix. The C1 gap (unpinned production sizeBytes wirings) was closed as a maintainer delta with your commits untouched. #2300 closes with this. As always: AI-generated review, maintainer-supervised. Process note: this and two sibling PRs were still marked draft when we merged them after clean reviews — that was our overreach, and going forward we will leave draft PRs to you to mark ready, however green the review.

PR #2322, opened 2026-08-28
test(dispatch): pin availability-lane cooldown consult (closes #2309)

### LINKED ISSUE #2309: Unpinned cooldown consult in availability-probe lane survives neuter across 126 tests (from #2308 review)
# Evidence

From PR #2308's review: a full-tree grep of `isInSpawnTimeoutCooldown(` finds a fourth consult site the #2301 sweep never enumerated — `clients/dispatch/runners/utils/runner-helpers.ts:1059`, the availability-probe lane. Neuter-probing it survives across 8 test files / 126 tests (runner-helpers + runner-helpers-generation-ledger, 56 passed; the six probe-timeout availability suites, 70 passed). Unpinned — the #2301 class.

# Root cause

Same vacuous/mutation-survivable-guard shape, but NOT redundant: unlike the consult #2308 deleted, this one does real work — it emits `noteDecision(... outcome: "transient", cause: "probe-timeout", classifiedBy: "caller")`. Deleting it would change an observable record; it needs a pinning test, not deletion.

# Acceptance criteria

- A test reds when the runner-helpers.ts:1059 consult is neutered (`if (false && ...)`), asserting both the early return AND the noteDecision record it emits.
- Red-first proven on current master; green after.
- The sweep record: re-grep `isInSpawnTimeoutCooldown(` tree-wide and state per-member verdicts, so the population is closed rather than named-surfaces-only (the under-scoping that hid this member through #2301).

# Observability record

The noteDecision transient/probe-timeout record is the observable this guard feeds; the pinning test should assert it so the record cannot silently vanish.

Refs #2301 (parent class, closed by #2308), PR #2308 (where found).

apmantza:
## Summary

Closes #2309. This adds a mutation-proof pin for the availability-probe lane's `isInSpawnTimeoutCooldown()` consult. While the resolved command is cooling down, the checker returns unavailable without spawning and emits the observable transient `probe-timeout` decision classified by the caller. The consult remains because deleting it would remove that decision record.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch
- [x] area:tests

## Tests

- `tests/clients/dispatch/runners/runner-helpers.test.ts` — added `pins the availability-lane cooldown consult and its transient decision (#2309)`, which pins both the early return/no-spawn behavior and the `availability_decision` record (`transient`, `probe-timeout`, `caller`, command/status evidence).
- Red-first proof: with the guard changed to `if (false && isInSpawnTimeoutCooldown(cmd))`, the test failed at the early-return assertion because `safeSpawnAsync` was called once. The captured output is in `.red-proof-2309.txt`.
- Targeted green: 8 files, 138 tests passed.
- The test covers the invisible-skip and observability screens: it asserts the returned result, the absence of the side effect, and the emitted record.

## Test assessment

The existing runner-helper tests uniquely cover the availability checker and now also pin the cooldown consult. No existing test became redundant; no tests were removed.

## Blast radius

The touched production module is `clients/dispatch/runners/utils/runner-helpers.ts`; its `createAvailabilityChecker()` callback is used by dispatch runner availability checks, including the availability-probe lane. The change is test-only because the intended production behavior already exists. `module_report` was unavailable in this checkout, so the blast-radius map is recorded from the direct import/call seam and the targeted runner suite. No hot-path cost delta: the added test does not change runtime code.

## Observability

The pinned production record is the `availability_decision` latency event emitted by `noteDecision()` in `runner-helpers.ts`, with `outcome: "transient"`, `cause: "probe-timeout"`, `classifiedBy: "caller"`, and evidence containing the resolved command and null status.

## Class sweep

Defect shape: an unpinned `isInSpawnTimeoutCooldown()` consult that can be neutered without a test proving its branch contract. Whole-tree grep across `clients/`, `tools/`, `mcp/`, `scripts/`, and `index.ts` found these five members (including the exported predicate declaration):

- `clients/file-utils.ts:961` — autofix lane consult; existing cooldown seam tests pin its zero-result/no-spawn behavior. Keep.
- `tests/support/session-state-registry.ts:679` — test-support registry predicate used to report whether the cooldown latch is armed; not a production dispatch consult. Keep.
- `clients/dispatch/runners/markdownlint.ts:139` — lint lane consult; existing markdownlint timeout tests pin its skip behavior. Keep.
- `clients/dispatch/runners/utils/runner-helpers.ts:1059` — availability-probe lane consult; this PR adds the missing pin for both skip behavior and its transient decision record. Keep.
- `clients/spawn-timeout-cooldown.ts:148` — exported predicate declaration, not a caller/member. Keep.

The production population is three distributed cross-lane callers (`file-utils`, `markdownlint`, and `runner-helpers`) plus one test-support observer; all are accounted for. They stay distributed because each caller owns a different failure-budget lane, while the shared cooldown module remains the single state seam.

## Checklist

- [x] Issue acceptance criteria complete, including red-first proof and whole-tree per-member sweep.
- [x] `npm ci`, `npm run build`, targeted tests, `npm run lint`, and `oxfmt` completed.
- [x] Changelog fragment added with front matter.

## Review round

The verify round proved the pin whole (my exact #2308 neuter reds; each half — early return, noteDecision emission — reds independently; all three discriminating record fields are individually pinned) and independently confirmed the five-member sweep table with zero line drift. Two deltas landed after review: a local `logLatencySpy.mockReset()` (the new test's `toHaveLength(1)` read the whole accumulated spy history and redded under `--sequence.shuffle`; five seeds now pass 54/54 with the neuter still red), and the committed `.red-proof-2309.txt` scratch paraphrase is removed — the red proof, quoted: neutering the consult reds `expected "vi.fn()" to not be called at all, but actually been called 1 times` (spawn half) and `expected [] to have a length of 1 but got +0` (record half).

### DIFF
--- a/.changelog/fix-2309-availability-cooldown-pin.md
+++ b/.changelog/fix-2309-availability-cooldown-pin.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Pin the availability-probe cooldown consult (closes #2309)** — Preserve the transient timeout decision and skip redundant probes while a command is cooling down.
--- a/.red-proof-2309.txt
+++ b/.red-proof-2309.txt
@@ -1,4 +1,0 @@
-Mutation: clients/dispatch/runners/utils/runner-helpers.ts
-  if (false && isInSpawnTimeoutCooldown(cmd))
-
-Result: the pinning test failed at the early-return assertion because safeSpawnAsync was called once.
--- a/.red-proof-2309.txt
+++ b/.red-proof-2309.txt
@@ -1,0 +1,4 @@
+Mutation: clients/dispatch/runners/utils/runner-helpers.ts
+  if (false && isInSpawnTimeoutCooldown(cmd))
+
+Result: the pinning test failed at the early-return assertion because safeSpawnAsync was called once.
--- a/tests/clients/dispatch/runners/runner-helpers.test.ts
+++ b/tests/clients/dispatch/runners/runner-helpers.test.ts
@@ -117,6 +117,36 @@ describe("runner-helpers availability checker", () => {
 		}
 	});
 
+	it("pins the availability-lane cooldown consult and its transient decision (#2309)", async () => {
+		const safeSpawnMod = await import("../../../../clients/safe-spawn.js");
+		const cooldownMod =
+			await import("../../../../clients/spawn-timeout-cooldown.js");
+		const command = "cooldown-probe-tool";
+
+		cooldownMod.resetSpawnTimeoutCooldowns();
+		cooldownMod.noteSpawnTimeout({
+			tool: command,
+			command,
+			phase: "availability",
+		});
+
+		const checker = createAvailabilityChecker(command);
+		expect(await checker.isAvailableAsync(process.cwd())).toBe(false);
+		expect(safeSpawnMod.safeSpawnAsync).not.toHaveBeenCalled();
+		expect(availabilityDecisions()).toHaveLength(1);
+		expect(availabilityDecisions()[0]?.metadata).toMatchObject({
+			tool: command,
+			verdict: "unavailable",
+			outcome: "transient",
+			cause: "probe-timeout",
+			classifiedBy: "caller",
+			evidence: {
+				command,
+				status: null,
+			},
+		});
+	});
+
 	it("falls back to global command when no local node_modules binary exists", () => {
 		const env = setupTestEnvironment("pi-lens-node-bin-global-");
 		try {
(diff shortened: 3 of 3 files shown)


apmantza:
## Post-push broader test run

`npm run test:unit` was started after push but stopped before Vitest's final aggregate summary because the run accumulated broad, unrelated environment-sensitive failures. The observed failures were across LSP grace/lifecycle, filesystem/worktree and temp-directory tests, formatter/graph tests, diagnostics, and ast-grep suites; several emitted `EPERM` while cleaning temporary directories. The run was not a clean pass.

The issue-scoped verification remains green: the eight targeted availability suites passed 138/138 tests, and `npm run lint` passed before push. The mutation red proof is quoted in the PR description and retained in `.red-proof-2309.txt`.

PR #2340, opened 2026-08-28
test(lsp): exercise the sweep bracket end to end (closes #2332)

apmantza:
## Summary

Exercise the workspace-diagnostics sweep bracket through the existing programmable stdio LSP fixture.

The fixture can advertise workspace pull, return a delayed report, or return a JSON-RPC error. The integration test enters through the real `LSPService.runWorkspaceDiagnostics` path and keeps client creation, framing, capability negotiation, workspace pull, fallback, and phase state real. The phase0 production gate stays unchanged.

This change closes #2332.

## Tests

- `opens one attributable phase pair for one real workspace pull sweep` drives a delayed `workspace/diagnostic` request over real stdio. It asserts the live sweep phase, one closed bracket after one sweep, a confirmed result, and post-sweep window attribution.
- `continues on the real document pull when the workspace pull rejects` makes the server return a JSON-RPC error. It asserts that the production client falls back to a real `textDocument/diagnostic` request and completes the sweep with one closed bracket.
- `closes the phase when warm-up throws out of the server-group runner` lets the real-wire pull error reach production fallback, then makes `ensureWarmForSweep` reject. It asserts that the rejection escapes `runPerServerGroups` while `finally` closes exactly one bracket and leaves no live phase.
- `npm run build` passed after the master rebase.
- The targeted attribution files passed: 9 tests across the unit and integration suites.
- The integration file runs in the serialized `lsp-spawn-heavy` Vitest lane.
- `npm run lint` passed.
- `npm run fmt:check` passed.
- Post-push `npm run astgrep:self-scan` passed with 0 findings.
- Post-push `npm test` ran 10,801 tests: 10,737 passed, 34 skipped, and 30 failed across eight unrelated files. Eighteen failures came from the host Git guard refusing temporary-repository `git init`; the remaining failures were unrelated contention timeouts or environment-sensitive assertions. The targeted attribution suites remained green.

### Mutation proof

I replaced the production `finally` with a trailing `phaseFinished` call, rebuilt, and ran the integration file in the `lsp-spawn-heavy` lane. The new escaping-error case failed on the live-phase assertion, while the success and fallback-continuation cases passed:

```text
AssertionError: expected { …(2) } to be undefined
Received: { phase: lsp_workspace_diagnostics_touch, ... }
tests/clients/lsp/workspace-diagnostics-sweep-attribution.integration.test.ts:141:37
Test Files  1 failed (1)
Tests  1 failed | 2 passed (3)
```

[...]

Issue #2324, opened 2026-08-28
ast-grep findings lost when the aux LSP stays silent: napi skips, nothing recovers (29.7% silence over 46h)


apmantza:
# Evidence (live logs, 46h window to 2026-08-28T07:32Z — predates this week's merges)

- Gate B (#239) makes ast-grep-napi skip when the ast-grep LSP is "available". Live: `{"runnerId":"ast-grep-napi","durationMs":1970,"status":"skipped","diagnosticCount":0}` followed by `{"phase":"lsp_aux_wait_outcome","metadata":{"outcomes":[{"serverId":"ast-grep","outcome":"silent","budgetMs":1800,"elapsedMs":794}]}}` and `{"phase":"late_auxiliary_findings","metadata":{"pending":1,"pendingAfter":1,"delivered":0,"rearmed":1,"stuckPairs":[{"filePath":".../src/legacy.ts","serverId":"ast-grep"}]}}`. Nothing re-runs napi; the findings are lost.
- Rate: ast-grep n=202, answered 141 (69.8%), silent 60 (29.7%). Cold start is not the whole cause: first-3 waits 13/29 silent (44.8%) vs later 47/173 (27.2%).
- Concrete race: the LSP finished spawning at 07:32:21.928; napi skipped at 07:32:21.860 — 68 ms earlier.

# Acceptance criteria

- On `outcome:"silent"` the skipped runner re-runs, OR the loss becomes a bounded degradation record naming file and server (`aux-runner-findings-lost`, subject <serverId>) — today `stuckPairs` names them and nothing acts.
- Gate B does not skip on a server that has not completed its first publication for the root.
- Test reds on pre-fix code.
- The warm-path 27% silence gets its own measurement before the fix design is fixed (measure first; the cold/warm split above is the baseline).

# Observability record

The three quoted records; the gap is that `late_auxiliary_findings.delivered` stays 0 with no compensating record.

Refs #239 (Gate B origin), the 2026-08-28 dogfood report §1b.

Issue #1777, opened 2026-08-19
Dispatch collapses ast-grep severities to two tiers; hint isolation is illusory


apmantza:
The dispatch pipeline collapses ast-grep rule severities to two tiers, so the hint tier the #1727 anti-slop rules were designed around does not exist where it matters.

`clients/dispatch/runners/ast-grep-napi.ts:671-672`:

```ts
const severity = rule.severity === "error" ? "error" : "warning";
const semantic = severity === "error" ? "blocking" : "warning";
```

Every `warning`, `hint`, and `info` YAML severity becomes `"warning"` here. Consequences, verified at file:line by a 2026-08-20 audit:

- The 7 hint-tier #1727 rules render identically to true warning-tier rules in every normal turn. Their rule notes state they ship at hint "so they don't dominate the diagnostic stream" — an isolation the runtime does not provide. Live calibration: on plegma, `no-any-type` (hint, 125 hits) and `no-non-null-assertion` (warning, 84 hits) land in the same bucket.
- `clients/code-quality-warnings.ts:162-167` carries hint/info branches that are dead code for any ast-grep-napi-sourced diagnostic.
- `clients/actionable-warnings.ts:243` cannot distinguish a hint-tier rule with a fix from a warning-tier one.
- The blocking-only gate at `ast-grep-napi.ts:579` reads the raw severity and is correct; only the post-gate stream loses the tiers.
- The on-demand MCP scan path (`clients/ast-grep-client.ts:812-819`) preserves the 3-way split — the two paths disagree today.

## Acceptance criteria

1. Decide: preserve `rule.severity` 4-valued through `Diagnostic.severity` in the dispatch path (reviving the dead branches, adjusting `actionable-warnings.ts:243`), or keep the 2-tier collapse intentionally. Either way the decision is recorded, and rule notes stop assuming isolation that does not exist.
2. If preserved: hint/info tiers get distinct rendering or volume treatment (the point of the tier), with the widget/nudge blast radius measured.
3. Write the severity policy down: a short subsection near AGENTS.md's "Actionable warnings routing" table stating the de facto rule the #1727 notes follow (error only with a documented zero-false-positive audit; opinions at hint) plus whatever criterion 1 decides.
4. Decision item: demote `no-non-null-assertion` from warning to hint — it is a style opinion (narrow-after-check is often legitimate) at opinion-density volume, inconsistent with the FP-census discipline newer rules follow.

[...]

Issue #239, opened 2026-06-17
feat(lsp): ast-grep LSP as the rule engine — auxiliary adopter (Phase 1) then consolidate/retire napi runner (Phase 2)


apmantza:
## Goal

Consolidate pi-lens''s rule-based ast-grep onto a single full-engine path: the **`ast-grep lsp`** server, driven by the repo''s own `sgconfig.yml` when present, else by pi-lens''s shipped rules via `--config`. This retires the napi **subset interpreter** (which ignores `field:`, has no bare-regex/anon-kind candidates) for *rule scanning*, expands language coverage (~7 napi file types → ~15 LSP languages), and unlocks engine-driven autofix via codeAction.

**Verified feasibility:** `ast-grep lsp -c, --config <CONFIG_FILE>` exists (checked the installed binary). So the no-sgconfig fallback — launch the LSP pointed at a pi-lens-synthesized sgconfig of our shipped rules — is supported. pi-lens already synthesizes an sgconfig into a session dir (`sg-runner.ts:470`), so the machinery exists.

## NOT in scope
- The `pilens_ast_grep_search` / `pilens_ast_grep_replace` MCP tools stay on napi (ad-hoc *pattern* queries; the LSP doesn''t expose those). "Replace the runner" != "remove napi."

## Phase 1 — ast-grep LSP as auxiliary-diagnostic-LSP adopter (sgconfig repos only) — LOW RISK, additive
Reuse the Opengrep auxiliary-LSP capability (`clients/dispatch/auxiliary-lsp.ts`, `role:"auxiliary"`, `with-auxiliary` touchFile scope, profile registry, `allowBlocking(cwd)`).
- Register `ast-grep lsp` server gated on repo-root `sgconfig.y[a]ml` (its `root_markers` + `workspace_required`).
- Profile: map LSP `source` -> `tool:"ast-grep"` + semantic policy; **blocking-eligible** (sgconfig = curated repo rules, same logic as Opengrep''s curated-rules gate).
- Wire ast-grep rule `fix:` -> LSP `codeAction` -> pi-lens autofix / fixSuggestion -> actionable-warnings routing.
- Smoke: add a fixture to `scripts/smoke-tools.mjs --lsp` (`auxiliaryServerIds`) — sgconfig + a rule + a violating file; assert the ast-grep-sourced diagnostic returns alongside the primary.
- **Does not touch the napi runner** — purely adds the team-rules surface for sgconfig repos.

[...]

PR #2343, opened 2026-08-28
fix(dispatch): run the napi fallback at the ast-grep LSP severity floor (closes #2336)

### LINKED ISSUE #2336: napi has never emitted a dispatch finding: resolve blockingOnly vs clean-files; dedupe path unvalidated in production
# Evidence (live dogfood, 2026-08-28, post-#2329 session on f0559217)

- Across the ENTIRE retained log history, the napi runner has never reported a single diagnostic in dispatch: 0 of 255 runner records with diagnosticCount > 0 (statuses: skipped 230, pending 21, succeeded 4 — all four successes post-#2329).
- The #2329 dedupe path (napi vs late-aux mutual exclusion) therefore has ZERO production validation — nothing could ever collide. Its correctness rests on unit coverage alone.
- Monitor-tested hypotheses, two REJECTED with evidence: severity floor (normalizeRuleSeverity keeps hint; #1777 revived hint/info) and suppressLinterOverlap (requires hasEslintConfig; pi-free has none). Surviving unproven candidate: blockingOnly filtering (ast-grep-napi.ts:752 `if (blockingOnly && rule.severity !== "error") continue`) — napi ran in the runner-collect-later tier while the LSP's one real finding was severity hint.

# Why this matters

If blockingOnly is the mechanism, "#2324 findings survive a silent aux LSP" holds only for ERROR-severity rules — narrower than the changelog claim. If files were simply clean, the claim stands. One measurement decides.

# Acceptance criteria

- Determine which filter suppresses napi findings in the collect-later tier (instrument or unit-drive the real dispatch tier with a hint-severity rule violation).
- If blockingOnly: either run the fallback at the LSP-equivalent severity floor (napi exists to SUBSTITUTE for the LSP, so matching its floor is the coherent design — argue cost), or correct the changelog/AGENTS.md claim to error-severity scope and record the decision.
- Either way: one dispatch-path test where napi DOES emit a finding, so the dedupe seam gains its first non-vacuous production-shaped exercise.

Refs #2324/#2329 (the claim), the 2026-08-28 dogfood verdict §2.

apmantza:
## Summary

`ast-grep-napi` now evaluates the full bundled rule catalog when it runs as the ast-grep LSP's substitute. Outside that role its floor is unchanged.

One line carries the fix (`clients/dispatch/runners/ast-grep-napi.ts`):

```ts
const runningAsLspSubstitute = astGrepLspEnabled && !astGrepLspPublished;
...
blockingOnly: runningAsLspSubstitute ? false : ctx.blockingOnly,
```

## The proven mechanism

The issue named `blockingOnly` filtering as the surviving unproven candidate. It is the mechanism, and the suppressing line is `ast-grep-napi.ts:752`:

```ts
if (blockingOnly && rule.severity !== "error") continue;
```

The trace, end to end:

1. Production per-edit dispatch enters at `clients/pipeline.ts:1439`, which calls `dispatchLintWithResult` without a `blockingOnly` option.
2. `clients/dispatch/integration.ts:2625` defaults that option to `true`.
3. `createDispatchContext` stores it as `ctx.blockingOnly`, and the runner passed it straight through to `evaluateAstGrepRules`.
4. Line 752 then dropped every rule whose declared severity is not `error`.

Line 816's complexity bound is not involved. `no-any-type` scores exactly 8 against `MAX_BLOCKING_RULE_COMPLEXITY = 8`, so `complexity > 8` is false and line 752 is the sole suppressor for the fixture used here.

The two hypotheses the monitor rejected stay rejected. `normalizeRuleSeverity` keeps `hint` (#1777), and `suppressLinterOverlap` needs `hasEslintConfig`.

## Severity census: the size of the gap

Counted over the two directories `getAstGrepRuleSources` actually loads.

| Source | Rules | error | warning | hint | info | no severity key |
|---|---|---|---|---|---|---|
| `rules/ast-grep-rules/rules` | 297 | 101 | 155 | 38 | 3 | 0 |
| `rules/ast-grep-rules/coderabbit/rules` | 184 | 0 | 181 | 0 | 2 | 1 |
| **Total** | **481** | **101** | **336** | **38** | **5** | **1** |

The substitute ran 101 of 481 rules, or 21 percent. A rule with no `severity` key is also dropped: line 752 tests the raw field, while `normalizeRuleSeverity` would map it to `warning`.

## Why the substitute role, and not everywhere

`ast-grep-napi` became a fallback in #239 Phase 2. It kept the per-edit floor its former always-on role justified. Three facts decide the scope:

- The server it replaces has an explicit severity contract. `clients/dispatch/auxiliary-lsp.ts:179-183` states it: "the rule severity is deliberate, so preserve ast-grep's severity semantics: ERROR can block, WARNING/INFO stay advisory." A substitute that delivers a fifth of that is not a substitute.
- Nothing downstream needs the filter for noise control. `dispatcher.ts:1195-1201` shows only `semantic: "blocking"` diagnostics inline and routes warning-tier ones to `lens_diagnostics`. The sibling structural runner already relies on that: `tree-sitter.ts:525` runs its whole query set under `blockingOnly` for the same stated reason.
- Outside the substitute role the extra work buys nothing. Either the LSP has published, in which case Gate B skipped napi before this line, or the user retired ast-grep with `no-ast-grep`.

## Cost

Measured on this repo with `evaluateAstGrepRules`, warm rule cache, mean of 5 runs, `blockingOnly: true` then `false`:

| Fixture | Lines | Before | After |
|---|---|---|---|
| `clients/dispatch/collect-later-tier.ts` | 68 | 2.4 ms | 22.8 ms |
| `clients/sgconfig.ts` | 341 | 9.2 ms | 63.4 ms |
| `clients/dispatch/dispatcher.ts` | 1275 | 19.0 ms | 188.2 ms |

Six to ten times more per file, so roughly 20 ms to 170 ms added to a dispatch. This lands only when napi is the sole ast-grep surface, where the alternative is 21 percent coverage. The collect-later tier (`dispatcher.ts:822-930`) already absorbs a runner that outgrows the inline budget by delivering at turn end, and the issue's own evidence observed napi in that tier, so a slower napi degrades to late delivery rather than to a stall.

## Tests

Every new test was proven red on pre-fix code. Verbatim from the pre-fix run (`git checkout origin/master -- clients/dispatch/runners/ast-grep-napi.ts`, rebuild, run):

```
 FAIL  |default| tests/clients/dispatch/runners/ast-grep-napi-fallback-severity-floor.test.ts > ast-grep-napi fallback severity floor (#2336) > emits a hint-severity finding on the production per-edit floor

 FAIL  |default| tests/clients/dispatch/runners/ast-grep-napi-fallback-severity-floor.test.ts > ast-grep-napi / late-aux dedupe with a real finding (#2329, #2336) > delivers the finding once through napi and suppresses the late-aux mark

 Test Files  1 failed (1)
      Tests  2 failed | 3 passed (5)
```

The three passing tests in that same pre-fix run are the non-vacuity controls. The fixture does fire under an explicit non-blocking floor, so the two red assertions measure the floor and not a dead rule.

Mutation proof for the new conditional. Replacing `runningAsLspSubstitute ? false : ctx.blockingOnly` with a bare `false` reds the opposite direction:

```
 × keeps the blocking floor when the ast-grep auxiliary is disabled 843ms
AssertionError: expected [ { …(14) } ] to have a length of +0 but got 1
 Test Files  1 failed (1)
      Tests  1 failed | 4 passed (5)
```

Deleting the fix reds tests 1 and 4; neutering the condition reds test 2. The guard cannot be vacuous in either direction.

### The dedupe seam's first finding-bearing exercise

`tests/clients/dispatch/runners/ast-grep-napi-fallback-severity-floor.test.ts` drives both orderings of the #2329 mutual exclusion with a real finding present:

- napi first: the real dispatch emits `no-any-type` once, and `napiFallbackCoveredSince(filePath, touchStartedAt)` returns true. That is the exact predicate the aux-grace producer evaluates before marking a pending pair (`clients/lsp/index.ts`, the `o.serverId === "ast-grep" && napiFallbackCoveredSince(...)` exclusion), so the late-aux surface is suppressed.
- LSP first: with the server already published for the root, Gate B skips napi, napi contributes zero diagnostics for the same rule and line, and no coverage is recorded.

Scope statement, honestly: this exercises the producer's predicate and Gate B, not the turn-end consumer in `runtime-turn.ts`. Driving the consumer needs a live LSP client cache; `tests/clients/lsp/service-aux-grace.test.ts` owns that half and still passes.

### Test assessment

| File | What it uniquely pins | Redundant now |
|---|---|---|
| `tests/clients/dispatch/runners/ast-grep-napi-fallback-severity-floor.test.ts` (new) | The only coverage of napi's severity floor through the real dispatch path at the production `blockingOnly: true`. Also the only exercise of the #2329 mutual exclusion with a finding actually present, in both orderings. | Nothing. The existing `ast-grep-napi.test.ts` dedupe cases all pass `blockingOnly: false` and drive the runner directly, so they pin the gate logic, not the floor. Both files are needed. |
| `tests/clients/dispatch/runners/ast-grep-rule-ignores.test.ts` (timeouts) | Unchanged assertions. The budget only stops a real-binary spawn from being reported as an assertion failure. | Nothing removed. |
| `tests/clients/dispatch/ast-grep-plan-css.test.ts` (timeout) | Unchanged assertions. Same reason. | Nothing removed. |

No test was deleted or weakened. No assertion was relaxed.

### Suites run

| Suite set | Result |
|---|---|
| New file, `ast-grep-napi-fallback-severity-floor.test.ts` | 5 passed |
| 21 ast-grep, dispatch, and pipeline files (every test referencing the changed symbols) | 196 passed |
| 13 governance sweeps (`delivery-surface-ratchet`, `finding-delivery-gate`, `session-state-conformance`, `bounded-telemetry-sweep`, `bus-producer-coverage`, `deps-centralization`, `freshness-sweep`, `managed-tool-seam-coverage`, `profiling-coverage`, `module-instance-coverage`, `sweep-floor-coverage`, `vacuous-skip-coverage`) | 179 passed |
| `npm run lint`, `npm run fmt:check` | clean |

The full suite is deferred to CI.

### One test-only change beyond the fix

Four cases ran on vitest's default 5 s budget while doing work that does not fit in it: three in `ast-grep-rule-ignores.test.ts` spawn the real ast-grep binary on a 60 s spawn budget, and the CSS write-plan case runs a real full-catalog dispatch. On a loaded worker pool they timed out mid-flight. They now carry explicit budgets, matching `ast-grep-rule-tests.test.ts:268` and `ast-grep-playground-verify.test.ts`. This is a latent flake this branch's added CPU exposed, not a behavior change: the CSS case passes `no-ast-grep` and `blockingOnly: false`, so its code path is byte-identical before and after the fix.

## Class sweep

**Pattern sweep** over `clients/`, `index.ts`, `tools/`, `mcp/`, and `scripts/`: 12 `blockingOnly` sites. Only three read it as a filter.

| Site | Verdict |
|---|---|
| `ast-grep-napi.ts:752` (severity floor) | fixed |
| `ast-grep-napi.ts:816` (complexity bound) | fixed by the same change; the LSP runs complex rules too |
| `tree-sitter.ts:538` (review-tier queries) | already correct; runs all severities under `blockingOnly` by design, `tree-sitter.ts:525` |
| `tree-sitter.ts:596` (modified-ranges gate) | correct; a range gate, not a severity collapse |
| `integration.ts:2625/2728`, `mcp/analyze.ts:428`, `scripts/smoke-tools.mjs:2173` | entry-point defaults, not filters |

**Population sweep** over the runner family, 52 runners in `clients/dispatch/runners/`. Members whose role is to substitute for another surface:

| Runner | Role | Verdict |
|---|---|---|
| `ast-grep-napi` | substitutes for the ast-grep auxiliary LSP (Gate B, `ast-grep-napi.ts:930-959`) | **the defect; fixed** |
| `pyright` | `PRIORITY.LSP_FALLBACK`, `pyright.ts:31` | already correct at `pyright.ts:164-206`; never reads `ctx.blockingOnly` and maps pyright's own severity vocabulary through, so its floor already equals the LSP's |
| `tree-sitter` | not a substitute; always-on structural runner | already correct at `tree-sitter.ts:525-538` |

`mode: "fallback"` runner groups (`plan.ts:40,51,52,62,81,213` — `ruff-lint`, `mypy`, `golangci-lint`, `rubocop`, `detekt`) are ordering semantics, stop-at-first-success, not role substitution. None of their runners reads `blockingOnly`, so none can carry this defect. Nothing left deliberately unfixed.

The shape is recorded as AGENTS.md defect shape 26, "A substitute surface still applying its old role's filter".

## Blast radius

- **Callers of the changed code.** `evaluateAstGrepRules` has three callers: this runner, `clients/project-diagnostics/scanner.ts`, and `clients/mcp/analyze.ts`. Only the runner's call site changed. The other two already pass `blockingOnly: false`, so their behavior is untouched.
- **Behavior widening.** More diagnostics reach `DispatchResult.diagnostics` in the substitute role, all at `semantic: "warning"` because only `error` maps to `blocking` (`ast-grep-napi.ts:871`). Inline agent output is unchanged: `dispatcher.ts` formats blockers only. The new findings surface through `lens_diagnostics` and the fixable-warning worklog append, which is where the ast-grep LSP's own warning-tier findings already land.
- **Hot path.** Measured, see the cost table. 20 ms to 170 ms added per substitute dispatch, zero when the LSP has published, zero under `no-ast-grep`.
- **No new process spawns.** The work is in-process napi matching.
- **Caps hold.** `MAX_MATCHES_PER_RULE = 10` and `MAX_TOTAL_DIAGNOSTICS = 50` are unchanged, so per-file output stays bounded.
- **Claim correction.** "#2324 findings survive a silent aux LSP" previously held only for error-severity rules. The changelog fragment says so.

## Observability

The record that proves this in production is the dispatcher's per-runner latency record for `ast-grep-napi` in `latency.log` — the same `diagnosticCount` field that read 0 across all 255 retained records. A non-zero count on a substitute run is the fix working.

That field alone cannot say which floor produced a given count, so the substitute path now also emits one bounded `logLatency` phase record, `astgrep_napi_substitute_floor`, carrying `floor`, `ctxBlockingOnly`, `diagnosticCount`, and a `bySeverity` tier breakdown. One line per substitute run, the same cadence as the runner record it annotates. A future "napi found nothing" reading can then tell an empty catalog pass from a filtered one.

## Merge order

No open PR touches `clients/dispatch/runners/ast-grep-napi.ts`. Two open PRs edit AGENTS.md: #2340 at line 2689 and #2339 at line 2571. This branch appends one line at line 225, so there is no textual overlap and no merge order is required. If a later branch appends to the same defect-shape list, the resolution is additive, keep both.

Rebased on `3c3c2f40`.


apmantza:
Merged on the verify verdict — every link of the mechanism chain, the census (481 = 101/336/38/5+1 keyless), the boundary-aware fixture (no-any-type scoring exactly 8 vs MAX 8), the 6-10x cost band, and the class sweep all independently reproduced, and the #2329 dedupe seam passed its first finding-bearing exercise across all three arms. Two record corrections from the review, for the archive: (1) the fix's magnitude was UNDERSTATED — with the complexity bound also active under blockingOnly, the pre-fix substitute evaluated ~38 of 481 rules (~8%), not 21%; (2) the substitute window is not always transient — under the never-publishing LSP #2324 measured at 27% of warm waits, the role is permanent for that file and the ~0.2s per-edit cost recurs; judged acceptable since the alternative is 8% coverage, but stated here so the trade is on the record.

Issue #687, opened 2026-07-16
ast-grep: no-javascript-url has no sink context + LSP surface never skips test files


apmantza:
Found via live dogfooding on \`pi-free\` (following up on the ESM-import false-positive fixed in #686).

## Bug 1: \`no-javascript-url\` matches any string literal, no sink context

\`rules/ast-grep-rules/rules/no-javascript-url.yml\` fires on any string literal starting with \`"javascript:\`, with zero constraint on where that string is used. Confirmed false positive:

\`\`\`ts
// tests/open-browser.test.ts:82 (pi-free)
expect(openBrowser("javascript:alert(1)")).toBe(false);
\`\`\`

This is a negative test asserting the guard *rejects* a \`javascript:\` URL — not a navigation sink. The rule can't tell "assigned to \`.href\`/\`.src\`" from "string literal used anywhere, including as adversarial test input."

## Bug 2: ast-grep's \`skipTestFiles\` only applies to the napi fallback runner, not the LSP surface

\`clients/dispatch/runners/ast-grep-napi.ts\` declares \`skipTestFiles: true\` (rule: this runner + tree-sitter-query-loader intentionally skip ast-grep findings in test files). But when the ast-grep LSP binary is installed (the default/common path), \`ast-grep-napi\`'s \`run()\` skips ENTIRELY in favor of the LSP surface (\`clients/dispatch/runners/lsp.ts\`, merging via \`AUXILIARY_LSP_PROFILES\` in \`clients/dispatch/auxiliary-lsp.ts\`) — and that path has NO test-file gating at all. So \`skipTestFiles: true\` silently stops applying the moment the binary is present, which is the common case. Confirmed by dogfooding: \`cline-xml-bridge.test.ts:48\` and \`open-browser.test.ts:82\` (pi-free) both got ast-grep findings despite being test files.

Related but distinct from #576 (broader "should test files get full dispatch by default" assessment) — this is a narrower, concrete inconsistency: one specific runner's own declared policy (test-file skip) silently doesn't survive its own LSP-backed code path.

## Fix
- Scope \`no-javascript-url\`(-js) to an actual sink (assignment to \`.href\`/\`.src\`, or passed to \`window.open\`/\`location.assign\`/\`location.replace\`), matching the pattern used to fix #686's ESM-import rule.
[...]

Issue #209, opened 2026-06-14
test: verify every supported tool installs & runs (generalizes the LSP smoke harness)


apmantza:
Generalizes the LSP-only smoke idea to **every tool pi-lens claims to support** — the recurring failure class behind #208 (a tool we wire that doesn't actually come up) isn't LSP-specific. A first LSP-only cut was dropped before 3.8.51 because it false-failed on Windows (a hand-rolled handshake bypassing `vscode-jsonrpc` framing missed typescript, which responds in ~120 ms) and used bare-project fixtures. This issue is that harness, done right and broadened.

## Scope — the full tool surface
- **~50 dispatch runners** (`clients/dispatch/runners/`): biome, ruff, oxlint, eslint, rust-clippy, golangci-lint, go-vet, mypy, pyright, rubocop, shellcheck, shfmt, hadolint, ktlint, detekt, swiftlint, stylelint, markdownlint, yamllint, sqlfluff, taplo, tflint, actionlint, semgrep, ast-grep-napi, tree-sitter, fact-rules, …
- **43 installable tools** (`clients/installer/index.ts` registry): 24 npm, 4 pip, 1 gem, 14 github-binary (linters, formatters, LSP servers, gitleaks).
- **Standalone analysis clients**: knip, jscpd, madge (dependency-checker), type-coverage, prettier / format-service.
- **35 LSP servers** (`clients/lsp/server.ts` `LSP_SERVERS`).

## Per-tool-type verification (drive pi-lens's REAL code paths, not reimplementations)
- **LSP servers** → install via the real `ensureTool`/spawn path, then a real `initialize` handshake through pi-lens's `vscode-jsonrpc` client (NOT a hand-rolled one — that was the false-fail).
- **Linters / formatters / checkers (dispatch runners)** → install via `ensureTool`, then run the runner's real `run(ctx)` against a tiny per-language fixture; assert it executes and returns structured output (not that it finds N issues).
- **Structural (ast-grep-napi, tree-sitter)** → run a trivial query against a fixture; assert match/parse, not crash.
- **knip / jscpd / madge / type-coverage** → run against a tiny fixture project; assert they execute.

## Requirements (carried from the first cut)
1. **Toolchain-gated skip**: absent toolchain → SKIP (not fail). Obtainable-but-won't-run → FAIL. Distinguish "installed but broken" from "not obtainable here."
[...]

PR #2349, opened 2026-08-28
fix(dispatch): enforce skipTestFiles on write-group path (closes #2337)

### LINKED ISSUE #2337: skipTestFiles declared on napi runner but .test.ts files run anyway
# Evidence (live dogfood, 2026-08-28)

The napi runner declares `skipTestFiles: true`, yet ran to `succeeded` twice on `providers/kiro/kiro-desktop-auth.test.ts` (session 101064). canHandle checks extension only, so the test-file skip must live dispatcher-side — either it is not applied on this path, or the declaration is dead.

# Acceptance criteria

- Locate where skipTestFiles is enforced for dispatch runners; determine whether the napi path bypasses it and whether that is intended (a test file CAN legitimately want rule findings — if so, the declaration is the stale half and should go).
- Align declaration and behavior, with a test pinning whichever direction wins (red on the mismatch).
- Cost note: if the skip should apply, quantify the wasted evaluations (bounded — napi runs are ms-scale — so P3).

Refs #2329 (where observed), the 2026-08-28 dogfood verdict §5.

AngriestBird:
## Summary

`skipTestFiles` was enforced only in `RunnerRegistry.getForKind`, which the per-edit path uses. The plan/write-group path (`getToolPlan` → `runGroup`) resolves runners by id via `registry.get()` and never consulted the flag, so ast-grep-napi (`skipTestFiles: true`) ran and reported findings on `.test.ts`/`.spec.ts` files (the #2329 dogfood evidence). `runGroup` now applies the same `isTestFile(ctx.filePath)` gate before the `when` precondition and records a distinct `test_file_skipped` latency status instead of silently dropping the runner. Design decision: kept the declaration and aligned behavior to it, and did not reuse the post-execution `RUNNER_SKIP_REASONS` vocabulary, which is reserved for runner-reported skips.

Direction ratified by the maintainer in review: the blanket skip is intended. Suppressing all 297 napi rules on everything `isTestFile` matches is acceptable, including the testing-tagged `go-test-functions.yml`, and no narrower predicate is wanted. That settles acceptance criterion 1's fork in favor of the PR as written.

All three acceptance criteria are met (enforcement located and fixed; test pins the direction; cost note below). Closes #2337.

Cost note: napi evaluates its catalog at 2-23ms/file (#2336), so the leaked test-file evaluations were bounded — correctness/noise, P3.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch

## Checklist

- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds if I changed code under `clients/`
- [ ] `AGENTS.md` update — not needed; no documented behavior/convention changes
- [x] `.changelog/fix-2337-napi-skiptestfiles.md` has one valid entry

## Tests

- `tests/clients/dispatch/dispatcher-flow.test.ts`, new `describe("skipTestFiles (dispatcher-level gate, #2337)")`, 4 cases:
  - skips a skipTestFiles runner via the group/plan path on a test file (regression proof; also asserts the `test_file_skipped` latency status, so neutering the gate or the latency record reds it)
  - still runs a skipTestFiles runner on a non-test file (mutation-proof against over-broad gating)
  - still runs a runner that does not declare skipTestFiles on a test file (review round F1: mutation-proof in the over-correction direction — deleting the `runner.skipTestFiles &&` half of the gate reds this one)
  - does not fire onRunnerResult for a skipTestFiles runner on a test file (contract seam with the #209 result sink)

Red on pre-fix code (build + targeted file):

```
× skips a skipTestFiles runner via the group/plan path on a test file
  AssertionError: expected [ { id: 'napi-finding', … } ] to have a length of +0 but got 1
× does not fire onRunnerResult for a skipTestFiles runner on a test file
  AssertionError: expected [ 'skip-test-runner' ] to deeply equal []
Test Files  1 failed (1)
     Tests  2 failed | 22 passed (24)
```

Red on the over-correction mutation (gate rewritten to `if (isTestFile(ctx.filePath))`, rebuilt):

```
× still runs a runner that does not declare skipTestFiles on a test file
  AssertionError: expected [] to deeply equal [ 'plain-finding' ]
Test Files  1 failed (1)
     Tests  1 failed | 24 passed (25)
```

Screens: the first test is the parallel-path screen made explicit (per-edit vs plan path); the sink test guards the invisible-skip shape (skip must not masquerade as an executed runner).

## Blast radius

Touched modules: `clients/dispatch/dispatcher.ts` (`runGroup`, `RunnerLatency.status` union, `buildCoverageNotice`) and `clients/mcp/analyze.ts` (`lsp.ran`). Dependents: every plan-driven dispatch consumer (`integration.ts` `dispatchLint`, `mcp/analyze.ts`); the only behavioral delta is for runners declaring `skipTestFiles: true` on test files — today that is ast-grep-napi alone. Hot path cost: one boolean field read plus one `isTestFile` call per runner per dispatch, both already paid on the per-edit path; no measurable delta.

Latency-status consumers (review round F4): both predicates that enumerate skip statuses now include `test_file_skipped`. `buildCoverageNotice`'s `allPrimarySkipped` would otherwise read a file whose only primary runner was gated as covered-and-clean rather than uncovered, and `analyze.ts`'s `lsp.ran` would report a gated runner as having run. Neither can fire today — no runner in `PRIMARY_DISPATCH_GROUPS` declares `skipTestFiles` — so neither has a regression test; the change closes the trap for whoever sets the flag on a primary runner next.

## Observability

`latency.log` runner records now carry `status: "test_file_skipped"` for the gated skip (`logLatency` type `runner`), so the skip is visible rather than silent. Post-fix, napi runner records for `*.test.ts` paths disappear from dogfood logs — that absence plus the new status is the production proof.

## Class sweep

Shape: declared-but-unenforced runner capability flag (a config field with only one of two consumer paths honoring it).

The pattern sweep (`rg -n "skipTestFiles"` over `clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts`, tests) found every dispatch runner except ast-grep-napi declaring `skipTestFiles: false`. That sweep was too narrow: the population is the fields of `RunnerDefinition` (`clients/dispatch/types.ts:129`), because a sibling flag spells the same defect differently. The review round swept that population and found two more, both proven with a probe:

- `appliesTo` — enforced in `getForKind` (`dispatcher.ts:99`), never on the group path. `runGroup`'s `filterKinds` filter (`dispatcher.ts:753`) checks only `ctx.kind` against `group.filterKinds` and never reads `runner.appliesTo`, so a runner declaring `appliesTo: ["python"]` in a group with `filterKinds: ["jsts"]` runs on `a.ts` and emits. Filed as #2353.
- `enabledByDefault` (`types.ts:133`) — required on all runners and read by nothing in dispatch. A runner with `enabledByDefault: false` runs on both paths. Only the separate auxiliary-LSP profile flag is enforced, at `auxiliary-lsp.ts:251`. Filed as #2354.

Both are latent today and neither belongs in this PR, so this is not a class of 1: it is a class of 3, one fixed here and two carried as follow-ups.

The ast-grep auxiliary LSP path has its own separate, working skipTestFiles gate (#687/#692, `tools/lsp-diagnostics.ts`, `clients/lsp/index.ts:7767`) — verified untouched and covered by its existing tests. `tree-sitter` deliberately sets `false` and filters per-rule (`skip_test_files`) instead. Consolidation verdict: enforcement now lives at the single `runGroup` seam the plan path funnels through; `getForKind` keeps its copy for the per-edit path — folding both onto one seam is a refactor of `getForKind`'s contract, better done alongside #2353.

## Test assessment

New coverage pins the plan-path gate, the non-test passthrough, the no-flag passthrough, and the sink contract. Not covered: a live napi run on a real `.test.ts` fixture (the mock runner stands in); the existing napi integration tests cover its real execution path and were part of the 199-test sibling sweep run green. Also not covered: the two F4 status tokens, which no reachable production state can exercise today (see Blast radius).


apmantza:
Thank you for this. The PR body is unusually thorough: the red-run transcript, the blast-radius note, and the latency-status observability record all made the review faster.

Disclosure: this review is AI-generated. A maintainer supervises the process and reads every verdict before acting on it.

Verdict: close, but it needs changes before it is ready. The core fix is correct and I reproduced your red run exactly. Two items need work, and two are follow-ups you do not have to carry.

## Red run and CI, verified

I reverted `clients/dispatch/dispatcher.ts` to `origin/master`, kept your tests, rebuilt, and got your exact transcript: `Tests 2 failed | 22 passed (24)`, with the same two assertions and messages. CI on 3408a87 is fully green. Unit tests and Lint and type-check both executed. `npx tsc --project tsconfig.json --noEmit` is clean.

## 1. The new guard is not mutation-proof in the over-correction direction

`clients/dispatch/dispatcher.ts:793`

I deleted the `runner.skipTestFiles &&` half of the condition, so every runner skips on every test file, then rebuilt and ran the tests. `dispatcher-flow.test.ts` stayed green at 24 of 24. The whole `tests/clients/dispatch` directory also stayed green: 1599 passed, 4 skipped, and one unrelated `helm-render` 5-second timeout that passes in isolation.

Your second case, "still runs a skipTestFiles runner on a non-test file", cannot catch this. It uses `auth.ts`, and the mutated gate lets non-test files through too. The missing case is a runner that does not declare `skipTestFiles`, dispatched on a `.test.ts` path, asserted to still run. Add that and the mutation goes red.

## 2. The catalog argues against the direction you chose

Acceptance criterion 1 in #2337 asks you to decide which half is stale, and notes that a test file can legitimately want rule findings. The PR body says napi's catalog is not written for test files. The catalog does not support that.

`rules/ast-grep-rules/rules/go-test-functions.yml` exists only for test code. Its message is "Go test function detected, verify it follows the project's test naming convention" and it is tagged `testing`. Several correctness rules in the same directory apply to test code as much as to source: `no-async-promise-executor`, `mutex-unlock-mismatch`, `loop-var-capture`, `incomplete-string-escaping`.

Scope matters here too. `isTestFile` (`clients/file-utils.ts:995`) matches far more than `.test.ts`: any path under `test/`, `tests/`, or `__tests__/`, any path containing `test-utils`, `.fixture.`, or `.mock.`, and any relative path starting with `test-`. The gate suppresses all 297 napi rules across all of that.

The fix is still right for the dogfood evidence in the issue. What is missing is the decision record. Either state why suppressing `go-test-functions` on Go test files is acceptable, or narrow the predicate. A maintainer should ratify the direction either way.

## 3. Class sweep: two siblings the pattern grep could not see

Your sweep was `rg -n "skipTestFiles"`, which is a pattern sweep. The population sweep is over the fields of `RunnerDefinition` (`clients/dispatch/types.ts:129`), because a sibling flag spells the same defect differently. Two turned up. I proved both with a throwaway probe.

`appliesTo` has your exact shape: enforced in `getForKind` at `dispatcher.ts:99`, never on the group path. `runGroup`'s `filterKinds` filter at `dispatcher.ts:753` checks only `ctx.kind` against `group.filterKinds` and never reads `runner.appliesTo`. A runner declaring `appliesTo: ["python"]`, listed in a group with `filterKinds: ["jsts"]`, ran on `a.ts` and emitted its diagnostic.

`enabledByDefault` (`types.ts:133`) is required on all runners and read by nothing in dispatch. A runner with `enabledByDefault: false` ran on both paths and appeared in `getForKind` output. Only the separate auxiliary-LSP profile flag is enforced, at `auxiliary-lsp.ts:251`.

Both are latent today. Neither belongs in this PR. Please replace the "class of 1" conclusion with these two named as follow-ups, and a maintainer will file them.

## 4. Two latency-status consumers worth one token each

`dispatcher.ts:610` computes `allPrimarySkipped` from `"skipped"` and `"when_skipped"` only. A primary runner recorded as `test_file_skipped` makes that false, so `buildCoverageNotice` returns undefined and a file with zero coverage reads as clean. That is the "an empty result must distinguish clean from errored" shape in AGENTS.md. `mcp/analyze.ts:540` has the same gap: `lsp.ran` would be true for a skipped runner.

You are right that neither can fire today. I verified it: no runner in `PRIMARY_DISPATCH_GROUPS` (`clients/language-policy.ts:91`) declares `skipTestFiles: true`. But both predicates enumerate skip statuses, and you added one. Adding it to both is one token each and closes the trap for whoever sets the flag on a primary runner next.

## What I cleared

The gate sits before the `when` precondition, matching `getForKind`'s ordering. `continue` preserves fallback-mode semantics, so a skipped runner does not consume the fallback slot. Runner execution funnels through exactly two seams, `getForKind` at `dispatcher.ts:1429` and `integration.ts:2785`, and `runGroup` at `dispatcher.ts:1091`, so no third path bypasses the fix. The auxiliary-LSP `skipTestFiles` gate at `auxiliary-lsp.ts:330` and `:365` is untouched. The changelog fragment passes `scripts/check-changelog-fragments.mjs`. The build and the typecheck are clean.

Items 1 and 2 are what stand between this and ready. Item 1 is a test addition. Item 2 may be a PR-body paragraph rather than a code change. Items 3 and 4 are cheap, and 3 is only a wording change to your sweep section. Mark the PR ready when you are done and a maintainer will take it from there.

apmantza:
Maintainer ratification on the review's F2: the blanket skip is the intended direction. Skipping all napi rules on test files is acceptable, including the testing-tagged rules; no narrower predicate needed. That resolves acceptance criterion 1's fork in favor of the PR as written.

Remaining before ready: F1 (add the missing mutation case — a runner without skipTestFiles on a .test.ts path must still run, so deleting the flag check reds a test) and optionally the two F4 status tokens. F3's siblings are filed as #2353 and #2354 and do not block this PR.

AngriestBird:
Fix round pushed as `c4c533c0`. Thanks for the ratification on F2 — the blanket skip stays as written.

## F1 — the missing mutation case (fixed)

Added `still runs a runner that does not declare skipTestFiles on a test file`: a plain runner (no `skipTestFiles`) dispatched on `auth.test.ts`, asserted to emit its diagnostic and to land `succeeded` in the latency report. Proven red by deleting the `runner.skipTestFiles &&` half of the gate at `dispatcher.ts:796`, rebuilding, and running the file:

```
 ❯ tests/clients/dispatch/dispatcher-flow.test.ts (25 tests | 1 failed)
       × still runs a runner that does not declare skipTestFiles on a test file

AssertionError: expected [] to deeply equal [ 'plain-finding' ]
- Expected
+ Received
- [ "plain-finding" ]
+ []
 ❯ tests/clients/dispatch/dispatcher-flow.test.ts:851:48

 Test Files  1 failed (1)
      Tests  1 failed | 24 passed (25)
```

Gate restored, rebuilt, 25/25 green.

## F4 — both latency-status consumers (fixed)

- `dispatcher.ts:609` `allPrimarySkipped` now includes `test_file_skipped`.
- `clients/mcp/analyze.ts:539` `lsp.ran` now excludes it too.

Honest note: neither is covered by a new test. You are right that no runner in `PRIMARY_DISPATCH_GROUPS` declares `skipTestFiles`, and there is no seam to inject one without editing `clients/language-policy.ts`, so a test would have to assert on a state production cannot currently reach. Both changes are one token each and are documented in the changelog fragment instead. If you would rather have the coverage than the token, say so and I will add a policy-injection seam.

## F3 — sweep wording

Rewritten in the PR body: the "class of 1" conclusion is replaced with the population sweep over `RunnerDefinition`'s fields, naming `appliesTo` and `enabledByDefault` as siblings you filed as #2353 and #2354, and stating that neither belongs in this PR.

## Verification

`npm run build` clean, `npm run lint` clean, `npm run fmt:check` clean. `tests/clients/dispatch/dispatcher-flow.test.ts` + `tests/clients/mcp/analyze.test.ts`: 46/46. The whole `tests/clients/dispatch` directory: green.

Marking ready for review.

Issue #2140, opened 2026-08-26
Availability probe searches PATH only: managed-dir binaries latch a false unavailable every session


apmantza:
Evidence (dogfood pass 2, 2026-08-26, 3.0h window)
- Every session start on the plegma host probes opengrep, gets ENOENT, writes a LATCHED unavailable verdict; ~900ms later the caller overrides with available, evidence.source: managed-dir. The binary sat in ~/.pi-lens/bin/opengrep.exe the whole time (instances.json spawns it from there). 7 unavailable/available pairs in 3h, gaps 625-2500ms; typos shows the same shape twice.
- Quoted pair: availability_decision verdict=unavailable cause=not-found latched=true classifiedBy=probe (ENOENT) at 10:07:30.158Z → verdict=available cause=ok classifiedBy=caller evidence.install=succeeded source=managed-dir at 10:07:31.053Z.

Root cause (hypothesis, labeled): the probe resolves through PATH and never consults the managed-binary directory the installer writes; the fast path knip/jscpd take (findManagedNodeToolBinary, clients/knip-client.ts:261 / clients/jscpd-client.ts:96) has no equivalent for the ~/.pi-lens/bin family.

Acceptance criteria
- The probe resolves managed binaries from ~/.pi-lens/bin before declaring not-found.
- Session start with opengrep.exe present emits exactly ONE availability_decision for opengrep, verdict available.
- Binary absent: unavailable then the install-path override still both emit (the negative case is not swallowed).
- Regression test red on pre-fix code.

Observability: availability_decision in latency.log; post-fix the paired records for a present binary disappear.

Class sweep: every tool resolved from ~/.pi-lens/bin (opengrep, typos-lsp, marksman, siblings in the installer's binary list) — per-member verdict; the knip/jscpd fastPath is the precedent shape. Cost today: ~900ms per session start plus a false latched verdict any consumer reading between the pair sees as authoritative. refs #2131 (adjacent probe machinery — compose with PR #2137's single-flight, don't collide).

apmantza:
The security-scan managed-directory resolution is fixed in PR #2148. The typos-lsp and marksman auxiliary-LSP resolution remains in `clients/lsp/auxiliary-lsp/server.ts` and is intentionally out of scope; this issue stays open for that remainder.

apmantza:
[...]

Issue #1657, opened 2026-08-18
Latent: a compensating-row emission with no prior latch burns the shared once-per-correction memo — cross-seam silencing when a tool joins both seams


apmantza:
From #1650's verify round (maintainer-supervised AI review); latent today, live the moment one toolId is served by both availability seams.

Three small related residuals at the shared emitCompensatingAvailableRow seam:

1. (F2, the substantive one) verifyOrInstallCommand emits an available row even when NO prior latched unavailable exists (two of 15 runners — biome-check.ts:92 unconditional, oxlint.ts:84 FactStore-gated — have no checker probe), and that emission burns correctedAvailabilityByCwd, so a LATER genuine latch-then-recover through the #1612 seam for the same (cwd, toolId) is silenced (probe-proven: verdicts after genuine latch+recover = ["unavailable"] — the #1606 defect re-reachable). No toolId overlaps both seams today; the guard should key the memo on correction-actually-happened (prior latch existed) or per-seam.
2. (F1) the comment at runner-helpers.ts:1105 claims a prior probe wrote the latched row — false for those two runners; correct the comment.
3. (F4) findManagedNodeToolBinary uses bare existsSync while the installer's managed-first branch runs verifyToolBinary — a broken managed shim now shadows a working PATH binary (self-recovering via ensureTool, but costs a failed probe + an unavailable row). Use the same verify.

## Acceptance criteria
- Memo burned only on genuine corrections (red-first: no-prior-latch emission then genuine latch+recover → the recovery row EMITS).
- Comment corrected; managed-binary verify aligned with the installer's own branch.
- The cross-seam scenario added as a test so the latency becomes impossible, not just unlikely.

Refs #1650, #1636, #1612, #1606.

Issue #1606, opened 2026-08-18
gitleaks availability log latches 'unavailable' then ensureTool succeeds silently — the durable record says the lane is off when it is on


apmantza:
Found by a maintainer-supervised AI log-forensics pass on the 2026-08-18 dogfood session. Denominator: 2/2 gitleaks availability records across both log rotations show the same wrong verdict; recurs every session.

`ensureViaInstaller` (clients/security-scan-client.ts:258) probes bare `gitleaks` on PATH first, writes a latched `unavailable` availability_decision (ENOENT, tool-not-found), THEN calls `ensureTool`, which finds the managed binary at ~/.pi-lens/bin/gitleaks.exe 162 ms later and updates probe-cache.json — but no compensating `available` row is ever written:

```
21:08:18.779Z availability_decision {"tool":"gitleaks","verdict":"unavailable","outcome":"missing","cause":"not-found","latched":true,"classifiedBy":"probe","evidence":{"failure":"spawn","spawnFailureKind":"tool-not-found","errno":"ENOENT"}}
probe-cache.json: "gitleaks": {"path":"...\bin\gitleaks.exe","cachedAt":1787000898941}  → 21:08:18.941Z
```

Not the unspawnable-prober-durable-verdict bug (the latch re-arms at session_start) — but the same family: the record that proves the lane's state is wrong. Anyone auditing the logs concludes the secrets lane is off when it is on.

## Acceptance criteria

- After a PATH probe fails but ensureTool succeeds, emit a second availability_decision with verdict:"available", cause:"ok", classifiedBy:"caller", and the resolved binaryPath in evidence.
- Regression test: a SecurityScanClient subclass whose probeVersion rejects ENOENT and whose ensureTool resolves must produce two availability_decision rows, the last one available. Prove red on pre-fix code (pre-fix emits one row, unavailable).
- Class sweep: any other probe-then-installer seam with the same one-sided logging.

Refs #1574, #1605 (lane 6).

apmantza:
[...]

PR #2351, opened 2026-08-28
feat(lsp): emit availability_decision on the managed-bin launch path (refs #2140)

AngriestBird:
## Summary

Implements the remainder of #2140 that merged PRs #2148 (CLI-scan probe managed-dir resolution) and #2266 (LSP launch candidate ordering) left open, and which #2266's own body named verbatim: no `availability_decision` record existed on the LSP launch path at all, so the LSP half of the managed-binary story stayed invisible in `latency.log`. `resolveAndLaunch` now emits two decisions mirroring `SecurityScanClient.probeVersion`/`ensureViaInstaller`'s exact shape: an `available` (`classifiedBy: "probe"`, `evidence.source: "managed-dir"`) when the managed-dir candidate itself launches, and an `unavailable` (`probe`, `cause: not-found`) + compensating `available` (`caller`) pair around the existing `ensureTool` install step. Gotcha handled: `evidence.source: "managed-dir"` on the install path is only asserted after a fresh `findManagedToolBinary` re-check confirms the install landed there — never fabricated for npm/pip installs (pinned by a dedicated test).

Refs #2140 — all four literal acceptance criteria now read satisfied end-to-end across the three PRs, but the maintainer twice deferred closing as their call, so this PR does not presume it. Issue comment posted naming what this PR adds.

## Type of change

## Area

- [x] area:lsp
- [x] area:observability

## Checklist

- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds (code under `clients/` changed)
- [ ] `AGENTS.md` update — not needed; no documented invariant changes
- [x] `.changelog/fix-2140-lsp-managed-bin-telemetry.md` has one valid entry

## Tests

- `tests/clients/lsp/managed-bin-fastpath.test.ts` (4 new cases beside 3 pre-existing):
  - emits exactly ONE `availability_decision` (verdict=available) when the managed binary is present (regression proof for the issue's "exactly one decision" criterion)
  - the negative case is not swallowed: `unavailable` then the install-path `available` both emit when the binary is absent (regression proof)
  - a PATH-resolved copy with no managed binary emits no decision (mutation-proof against over-broad gating on `command === managedCandidate`)
  - does not assert `evidence.source` when the install lands outside the managed dir (pins the no-fabricated-evidence rule for npm/pip strategies)
- `tests/clients/availability-classifiedby-ok-sweep.test.ts`: governance floor bumped 18→20 for the two new correctly-stamped `logAvailabilityDecision` sites.

Red on pre-fix code (only `clients/lsp/server.ts` reverted via patch, tests untouched):

```
× emits exactly ONE availability_decision (verdict=available) when the managed binary is present
× does not swallow the negative case: unavailable then the install-path override both emit when the binary is absent
  AssertionError: expected [] to have a length of 2 but got +0
Test Files  1 failed (1)
     Tests  2 failed | 4 passed (6)
```

Screens: the sweep-floor bump is the loose-bound screen (exact count, not >=); the PATH-copy test is the parallel-path screen.

## Blast radius

Touched production module: `clients/lsp/server.ts` (`resolveAndLaunch`) — the single shared launch seam for all 26 `managedToolId` LSP servers. Behavioral delta is telemetry-only: two `logAvailabilityDecision` calls, no control-flow change. Hot path cost: the fast-path decision is one function call after a successful spawn (session-start frequency, not per-file); the install-path re-check adds one `findManagedToolBinary` stat only on the already-slow install fallback. Verification: 20 test files / 317 tests green across sibling `resolveAndLaunch` consumers, `security-scan-probe-telemetry.test.ts`, and the fixer-playbook governance sweeps (delivery-surface-ratchet, session-state-conformance, bounded-telemetry-sweep, bus-producer-coverage, managed-tool-seam-coverage, and siblings).

## Observability

This PR IS the observability record: `latency.log` gains `availability_decision` rows on the LSP launch path — fast path `verdict=available cause=ok classifiedBy=probe evidence.source=managed-dir`; fallback `verdict=unavailable cause=not-found classifiedBy=probe` followed on successful install by `verdict=available classifiedBy=caller`. Post-fix dogfood proof: the issue's paired false-unavailable records for a present managed binary disappear, replaced by the single fast-path row.

## Class sweep

Shape: telemetry seam present on one resolution path (CLI scan) but absent on its sibling (LSP launch) — the cross-path coverage gap AGENTS.md's parallel-path screen targets. Population: all 26 `managedToolId` servers inherit from the one `resolveAndLaunch` seam — 13 github/maven/archive-strategy servers (clojure-lsp, cue, deno, expert, gleam, marksman, opengrep, rust-analyzer, taplo, terraform-ls, typos-lsp, zizmor, zls — matching PR #2266's precedent table) get the fast-path decision; the 13 npm/pip-strategy servers get the negative/compensating pair with `evidence.source` correctly omitted. Whole-tree sweep for other `logAvailabilityDecision` writers: `SecurityScanClient` (fixed by #2148) and the installer's `ensureViaInstaller` were already correct; the classifiedBy-ok sweep test enumerates every site and its floor is pinned at 20. Consolidation verdict: already folded — one seam per process family, no distributed members remain.

## Test assessment

Scope note flagged for review: the telemetry extension covers all 26 managed servers, not only the issue's literal 13 — deliberate, because gating by install strategy would recreate the hand-maintained-allowlist defect. Covered: presence, absence, PATH-copy, non-managed-install evidence. Not covered: a live spawn of a real managed binary (mocked `launchLSP`); the pre-existing integration suites cover real launches.

### DIFF
--- a/.changelog/fix-2140-lsp-managed-bin-telemetry.md
+++ b/.changelog/fix-2140-lsp-managed-bin-telemetry.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Emit availability_decision telemetry on the LSP-launch managed-bin fast path (refs #2140)** — `resolveAndLaunch`'s `~/.pi-lens/bin` resolution (clojure-lsp, cue, deno, expert, gleam, marksman, opengrep, rust-analyzer, taplo, terraform-ls, typos-lsp, zizmor, zls) now logs the same `availability_decision` record `SecurityScanClient`'s CLI-scan probe already writes: exactly one `available` row when the managed binary launches directly, and the `unavailable`-then-install-override pair when it doesn't, so the LSP-server launch path stops being invisible in `latency.log`.
--- a/clients/dispatch/runners/utils/availability-policy.ts
+++ b/clients/dispatch/runners/utils/availability-policy.ts
@@ -247,6 +247,8 @@ export function describeInstallAttempt(
 
 export interface AvailabilityDecision {
 	tool: string;
+	/** Producer scope for independent availability evidence (#2351). */
+	producer?: "security-scan" | "lsp-launch";
 	verdict: "available" | "unavailable";
 	outcome: AvailabilityOutcome;
 	cause: AvailabilityCause;
@@ -844,6 +846,9 @@ export function logAvailabilityDecision(
 		durationMs: Math.round(decision.elapsedMs),
 		metadata: {
 			tool: decision.tool,
+			...(decision.producer !== undefined && {
+				producer: decision.producer,
+			}),
 			verdict: decision.verdict,
 			outcome: decision.outcome,
 			cause: decision.cause,
--- a/clients/lsp/index.ts
+++ b/clients/lsp/index.ts
@@ -93,6 +93,7 @@ import {
 	hasProjectBoundaryMarker,
 	isDirectLspCommandTemporarilyUnavailable,
 	resetClassicTsRepairGuard,
+	resetLspLaunchAvailabilityGeneration,
 } from "./server.js";
 import {
 	classifyCascadeWaitTier,
@@ -8946,6 +8947,10 @@ export async function notifyExternalFileChange(
 }
 
 export function resetLSPService(options: LSPShutdownOptions = {}): void {
+	// Invalidate availability publication started by the retiring service before
+	// any asynchronous teardown. The launch seam checks this generation after
+	// every managed lookup, install, and process launch (#2351).
+	resetLspLaunchAvailabilityGeneration();
 	// A new session must get its own classic-tsserver-repair attempt: the
 	// guard is a process-lifetime flag (see resetClassicTsRepairGuard), so a
 	// repair that failed transiently in an earlier session must not stay
--- a/clients/lsp/server.ts
+++ b/clients/lsp/server.ts
@@ -35,6 +35,7 @@ import {
 	getToolEnvironment,
 	getToolPath,
 } from "../installer/index.js";
+import { logAvailabilityDecision } from "../dispatch/runners/utils/availability-policy.js";
 import { resolveOpengrepConfig } from "../opengrep-config.js";
 import {
 	isZizmorAuditTarget,
@@ -536,6 +537,7 @@ export async function resolveAndLaunch(
 	// `findManagedNodeToolBinary`'s npm-managed fast path (runner-helpers.ts)
 	// and `SecurityScanClient.probeVersion`'s equivalent fix for the CLI-scan
 	// half of this same issue (#2140, landed in PR #2148/#2137).
+	const managedProbeStartedAt = Date.now();
 	const managedCandidate = spec.managedToolId
 		? await findManagedToolBinary(spec.managedToolId)
 		: undefined;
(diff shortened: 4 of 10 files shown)


apmantza (commented):
Thank you for the contribution — this closes a real observability gap, and the test that pins the no-fabricated-source claim is exactly the right instinct.

Disclosure: this review is AI-generated (Claude Code), supervised by the maintainer.

**Verdict: needs changes.** One real gap in the shape this PR exists to close, plus three record-accuracy points. Verification first, so the findings have context.

## Verification

- Proved the new tests red on pre-fix code: reverted `clients/lsp/server.ts` to the merge base (`bb733dd3`), kept the PR's tests, rebuilt. Result: `Tests 5 failed | 11 passed` — the three telemetry assertions and the sweep count fail; the no-emission guard passes pre-fix as designed. Restored the PR head: `16 passed (16)`.
- CI on the current head: Unit tests and Lint both registered and passed (run 33182624084). Not a DIRTY skip.
- Confirmed no `resolveAndLaunch` spec passes both `managedToolId` and `runtimeInstall`, so step 4 cannot leak an unpaired record. The force-reinstall path can — see F1.

## Findings

**F1 (medium) — the force-reinstall recovery path leaves the `unavailable` record unpaired.** The compensating `available` row is emitted only on the first `launchLSP(installed, …)` success. When that launch fails and the force-reinstall retry succeeds (`lsp_launch_managed_force_reinstall_success`, ~line 798), the function returns `source: "managed"` with the earlier `latched: true` `unavailable` row still standing. That is the exact #1606 shape the PR body claims to handle: the durable record says the tool is off while it is running. The retry success path needs the same compensating row (with the same fresh `findManagedToolBinary` re-check).

**F2 (low) — `evidence.failure: "tool-not-found"` misuses the field.** In `ProbeEvidence`, `failure` is safeSpawnAsync's structured failure reason (`"timeout"`, `"aborted"`); `"tool-not-found"` is a `spawnFailureKind`. A reader grepping `spawnFailureKind` for not-found rows misses these, and `failure` grows a value no producer emits. Use `spawnFailureKind`, or better, derive the whole triple — see F3.

**F3 (low) — `classifiedBy: "probe"` on the `unavailable` row is actually a caller assertion.** The contract in `availability-policy.ts` says `probe` means `classifyProbeFailure` derived the outcome from the evidence. Here the call site asserts `missing`/`not-found` from the `hasOnlyRepairableCandidateFailures` gate. The classification is materially right (the gate admits only `tool-not-found` failures), but the honest spelling is to run the last candidate failure through `classifyProbeFailure` and log what it returns — which also fixes F2 for free, since it produces the evidence via `describeProbeEvidence`.

**F4 (low) — `latched: true` has no latch behind it.** `resolveAndLaunch` stores nothing; every relaunch (idle-reset recycle, next session touch) re-decides and re-emits. The `SecurityScanClient` precedent bounds latched rows to one per session via its `availabilityLatch`, and `logAvailabilityDecision`'s doc claims "bounded by construction". If the justification is that the launched process itself persists for the session, say so in a comment at the site; otherwise `latched: false` describes these rows more honestly.

**F5 (info) — the compensating row omits `correctsLatchedRow`.** The #1657/#1674 field exists precisely to distinguish a row that cleared a latched `unavailable` from one that corrected nothing. The PR body claims an exact mirror of the precedent's shape; carrying `correctsLatchedRow: true` here would make that claim true and keep the once-per-correction tooling able to read these rows.

The maintainer decides the merge; happy to re-verify a fix round.

apmantza:
Thank you again for the contribution. I pushed maintainer fix commit `8c7dabd1` to the PR branch.

Disclosure: the fix implementation and verification were AI-assisted under maintainer supervision.

Fix-round disposition:

- F1 fixed: force-reinstall success emits the compensating available record.
- F2/F3 fixed: unavailable classification and evidence now derive from the real probe failure.
- F4/F5 fixed: rows no longer claim a nonexistent latch, and correction fields match that model.
- Producer composition fixed: security-scan and LSP-launch rows carry explicit producer identity; a real-sink composition test proves exact-once-per-producer behavior.
- Session ownership fixed: stale managed lookup, install, and launch completions cannot publish into a replacement session.
- Installer evidence fixed: cached/discovered paths no longer claim a fresh install.
- Managed-directory attribution fixed: source is asserted only when normalized result paths match.
- Test contract fixed: telemetry assertions read real `latency.log` bytes, and the managed-failure/PATH-success mutation case is covered.

Verification on the fix tree: build, lint, format, changelog, and ast-grep passed. Focused suites passed 146 tests, runtime-install discovery passed 10 tests, and the managed-bin suite passed 15 tests. Neutering the shared stale-generation guard failed all three interleaving regressions.

apmantza:
Maintainer workflow record — 2026-08-29

Thank you for allowing maintainer edits. The same Luna reviewer verified the fix-round tree that became 943ff40dd8788facf6c2bd9f2def77fb82ce12d4.

- Stale managed-launch rejection side effects: fixed in 943ff40d. The reset interleaving and guard mutation both fail without the fix.
- Installer-attempt evidence overwrite: fixed in 943ff40d. Evidence is snapshotted at the ensureTool boundary and survives a concurrent later attempt.
- Real-seam test requirement: fixed in 943ff40d. The installer test uses real installer state and a temporary PI_LENS_HOME; only process spawning is mocked.

Local verification passed build, lint, formatting, ast-grep self-scan, 46 affected tests, and 97 governance tests. The next owner is the same reviewer after required CI finishes on 943ff40d. Autocomplete/merge remains disabled until that exact-head confirmation.

This verification is AI-generated and maintainer-supervised.

Issue #2132, opened 2026-08-26
memory_sample subsystem accounting explains under 1% of RSS


apmantza:
Evidence (live dogfood forensics, 2026-08-26)
- Across 11 samples: accounted subsystem bytes 5.7-6.8MB against RSS 478-970MB (0.7-1.1%). RSS moved 279MB while accounted bytes moved 1MB.
- reviewGraph reports node/edge counts with no byte estimate (#2114 tracks the estimator).
- The sampler that would attribute the rest is itself degraded: resource-sampler-query-failed x3 on windows-process-table / windows-descendant-process-table.

Acceptance criteria
- reviewGraph.residentBytes estimator in #2115's style (fold into #2114 if that lands first — cross-link, don't duplicate).
- The Windows process-table sampler failure is fixed or documented with its fallback, with a bounded degradation record either way.
- After the additions, accounted-vs-RSS coverage is re-measured and the residual named.

Class links
- refs #2114 #2115. The attribution-first philosophy of #2042 applied to process memory.

apmantza:
Addition from dogfood pass 2 (2026-08-26): the #2098 retention cap remains unexercised after 3h (peak 102/128 aggregate entries, 4.46/64MiB) AND the eviction loop at clients/lsp/client.ts:1907-1935 emits NO record if it ever fires — if the cap is hit we learn nothing. Proposed sibling criterion for this issue: a bounded incrementalTextEvicted record (incrementDegradationCount, keyed client+path count) at the eviction site, so the next long dogfood can prove the cap's behavior from logs.

apmantza:
PR #2213 addresses the memory-reporting portion of this issue in the combined #2132/#2114 branch. It adds bounded byte fields and preserves the bounded Windows fallback record.

The remaining item is #2114 acceptance criterion 1: a three-run heap census isolating `nodes`, `edges`, `edgesByFrom`, and `edgesByTo`. The repository graph has 18,695 nodes and 48,815 edges, but raw `heapUsed + external` deltas also include parser/runtime allocation and GC noise, ranging from negative values to 191.3 MiB. The PR records this limitation and does not claim those deltas are graph-only memory.

Refs #2213

apmantza:
Remainder tracking (PR #2213 landing): two items remain open under refs, not closes:

[...]

PR #2352, opened 2026-08-28
fix(dispatch): bound sessionFacts per-file key growth (closes #2282)

### LINKED ISSUE #2282: sessionFacts grows per file touched: session.baseline.* keys are never evicted (refs #2240)
## Evidence

`FactStore.sessionFacts` is described in several places as bounded by a fixed fact-id vocabulary. It is not. Two families mint one session-fact key per file touched:

- `clients/dispatch/dispatcher.ts:1058-1059` — ``session.baseline.${normalizeMapKey(ctx.filePath)}`` and ``session.baseline.${normalizeMapKey(relativeKey)}``, so a file contributes up to two keys.
- `clients/dispatch/integration.ts:2329` — ``session.baseline.cascade.${normalizeMapKey(neighbor.filePath)}``, one key per cascade neighbor.

The store that holds them is module-scope: `const sessionFacts = new FactStore()` at `clients/dispatch/integration.ts:183`. It lives for the whole process. Nothing removes a baseline key. `clearFileFactsFor` only touches `fileFacts`; the sole path that drops a session fact is `clearAll` on an explicit session reset.

So a batch touching N distinct paths leaves one to three permanent entries per path, each holding a `Diagnostic[]`.

## Why this is a #2240 sibling, not a duplicate

#2240 named two unbounded dispatch stores. `fileFacts` was bounded by #2243 with an LRU plus a pin set. #2255 bounds the review-graph VALUE on `session.reviewGraph`. Neither reaches this: the growth here is in the KEY COUNT of `sessionFacts`, which both of those changes deliberately left alone on the stated basis that the fact-id vocabulary is fixed. That basis is wrong, and this issue records where.

Severity is lower than either sibling. A `Diagnostic[]` for one file is small next to a project graph, so this is accumulation rather than the multi-gigabyte abort. It is still unbounded in batch size.

## Root cause

Hypothesis, not yet measured: per-file baseline keys were modelled as session-scoped because a baseline is conceptually per session, but their identity is per FILE, so their cardinality follows the batch, not the session.

## Acceptance criteria

1. Measure the retained cost of `sessionFacts` across a several-hundred-file batch and state the number, separating baseline keys from fixed-vocabulary keys.
2. Bound the baseline families, or state with evidence why the measured cost does not need bounding.
3. Whatever the verdict, correct the "fixed fact-id vocabulary" claim wherever it appears so the next reader is not misled. It currently appears in `clients/review-graph/builder.ts` (corrected there by #2255's review round) and in #2243's class-sweep note.
4. Any bound reuses the store-level discipline #2243 established rather than adding a third mechanism.

## Observability

`latency.log`'s `memory_sample` has no `sessionFacts` member today, so this growth is invisible from logs. Either add a bounded entry-count field alongside the existing subsystems, or record the gap. This is the same reporting hole #2114 closed for the review graph.

## Cross-links

- #2240 — the parent OOM issue naming the unbounded dispatch stores.
- #2243 — bounds `fileFacts`; its class sweep states the vocabulary claim this issue corrects.
- #2255 — bounds the review-graph value on `session.reviewGraph`; its review round found this.

AngriestBird:
## Summary

`FactStore.sessionFacts` was described as bounded by a fixed fact-id vocabulary; it is not. Three families mint one permanent key per touched file — `session.baseline.*` (abs+rel, `dispatcher.ts`), `session.baseline.cascade.*` (`integration.ts`), and (found by the required sweep) review-graph `entitySnapshot:*`/`changedSymbols:*` (`review-graph/service.ts`) — so retained cost scales with batch size for the process lifetime. Fix: a bounded sibling map on `FactStore` (`setBoundedSessionFact`/`getBoundedSessionFact`, `MAX_SESSION_FACT_RECORDS = 4096`) reusing #2243's `fileFacts` LRU-touch + count-cap + eviction-reporter discipline rather than a third mechanism; the reporter axis union gains `"session-count"` and the existing degradation-ledger wiring needed zero changes. Fixed-vocabulary facts stay on the plain map.

Design note: no pin exemption (unlike `fileFacts`) — every per-file session fact is read once at dispatch start and written once at dispatch end, so mid-batch eviction just means "no previous baseline". That equivalence holds only where an absent value means "report everything", which is true of the diagnostic baselines and false of the entity snapshot: there, an absent snapshot diffed as an empty one puts every symbol in `added` and schedules a blast-radius run for a file that never changed. So eviction leaves a keys-only tombstone (`wasBoundedSessionFactEvicted`, same cap, FIFO) and `recordEntitySnapshotDiff` re-seeds and reports no diff instead of inverting.

Closes #2282. All four acceptance criteria are met, and the `cascadeDiagnosticBaselines` shadow map originally carried as a remainder is delivered here too: it dual-wrote the same keys the bounded map holds and was cleared alongside it, so eviction freed nothing while the cascade family still spent budget. Removed.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch
- [x] area:perf
- [x] area:observability

## Checklist

- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds (code under `clients/` changed)
- [ ] `AGENTS.md` update — not needed; the store discipline it documents (#2243) is reused, not changed
- [x] `.changelog/2282-session-fact-baseline-bound.md` has one valid entry

## Tests

`tests/clients/dispatch/fact-store-session-bound.test.ts` (new, 10 cases): cap enforcement + LRU order, read-touch recency, eviction-reporter emission on the new axis, `clearAll` coverage, the measurement case (500 files → 1,000 entries → 439,560 bytes, exact-asserted), the production-level regression (below), and three from the review round — an evicted entity snapshot reporting no diff rather than every symbol, the re-seed that keeps the next real edit diffable, and the changed-symbols write landing on the key `builder.ts` reads.

`tests/clients/memory-sampler.test.ts`: `estimateDispatchCacheBytes` asserted directly, so the session-fact term cannot be dropped silently.

Edits: `dispatcher-flow.test.ts`, `rule-policy-dispatcher.test.ts`, `review-graph.service.test.ts` — test helpers that poke `FactStore` directly moved to the bounded API to match production (contract-seam alignment, no assertions weakened).

Red on pre-fix code (via `git apply -R`/`git apply`, no stash): 6 of 7 original cases fail. The key production-level case (uses only `dispatchForFile`/public baselines, no new API):

```
stops retaining an early file's delta baseline once a large batch exceeds the session-fact cap
AssertionError: expected 1 to be +0
```

The 5 API-shape cases fail with `TypeError: store.setBoundedSessionFact is not a function` (API absent pre-fix, same category as #2243's proof). The measurement case passes pre and post by design — it documents scale, not the fix.

Review-round cases, each proven red against its own mutation:

```
× reports no diff for a snapshot the cap evicted, not every symbol      (eviction branch removed)
  AssertionError: expected { …(3) } to deeply equal { Object (added, removed, ...) }
  - "added": [],
  + "added": [ "function:alpha", "function:beta", "class:Gamma" ],

× writes changedSymbols under the key builder.ts reads                  (normalizeMapKey dropped from the write)
  AssertionError: expected undefined to deeply equal [ 'alpha', 'beta', 'Gamma' ]

× attributes bytes to session-fact entries, not just the neighbor cache (session-fact term deleted)
  AssertionError: expected 640 to be 440640
```

Screens: the production-level case is the wrong-layer-pin screen satisfied (asserts through the dispatch surface, not store internals); the exact byte assertion is the loose-bound screen. The changed-symbols case asserts `normalizeMapKey(path) !== path` before asserting the read hit, so it cannot go vacuously green — `normalizeMapKey` leaves a plain POSIX path alone, and only rewrites Windows-shaped or backslash-separated ones.

## Blast radius

Touched production modules: `clients/dispatch/fact-store.ts` (new API plus the eviction tombstone), `clients/dispatch/dispatcher.ts` + `clients/dispatch/integration.ts` + `clients/review-graph/service.ts` + `clients/review-graph/builder.ts` (call-site migration, read+write pairs kept symmetric), `clients/memory-sampler.ts` (entry count and its byte attribution).

Dependents: every dispatch consumer reads baselines through these seams; behavior is identical below the 4096-entry cap. Above it, a diagnostic baseline degrades to "first dispatch" semantics with a once-per-session ledger record, and an entity snapshot degrades to "no change" — deliberately the under-reporting direction, because over-reporting there schedules real background work off a signal that is not a change. Tombstones are keys only and carry the same cap, so when the coldest one falls off, that key reverts to first-dispatch semantics.

Hot path cost: `getBoundedSessionFact` adds one Map delete+set per read (LRU touch) — the same O(1) pattern `fileFacts` already pays per dispatch — plus one Set lookup on the entity-snapshot miss path; no added scans. Removing `cascadeDiagnosticBaselines` removes a Map write per cascade neighbor.

## Observability

`latency.log`: `memory_sample.dispatchCaches.sessionFactEntries` exposes the combined fixed + bounded entry count, closing the issue's "invisible from logs" gap, and `dispatchCaches.estimatedBytes` now carries bytes for it (440 per entry, from this PR's 1,000-entry / 439,560-byte measurement) rather than leaving the count unbacked. Evictions surface through the existing degradation ledger via the `CapacityEvictionReporter` on the new `"session-count"` axis (once-per-session record, #2243 wiring).

## Class sweep

Shape: per-file-minted keys on a session-lifetime store (AGENTS.md unbounded-growth shape; #2240 parent, #2243/#2255 siblings). Population sweep of every `setSessionFact` call site across the whole tree (`clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts`): per-member verdicts — tool-availability vocabulary: fixed cardinality, stays plain; `session.reviewGraph` singleton: bounded by #2255, stays plain; baseline abs/rel, cascade baseline, review-graph entitySnapshot/changedSymbols: per-file, all migrated to the bounded map (the last family was not named by the issue — identical defect, same store, fixed under the sweep obligation). Vocabulary-claim grep: no "fixed fact-id vocabulary" claim remains; `review-graph/builder.ts`'s "that is #2282" comment updated to state the bound.

Consolidation verdict: folded onto the one `FactStore` bounded seam. The surviving distributed member, `cascadeDiagnosticBaselines` (`integration.ts`), turned out to be pure redundancy rather than a second store worth bounding — it held the same keys, was written and cleared alongside the bounded map, and its `??` fallback was unreachable — so it is deleted here rather than carried.

A second class question the review round raised and this PR answers: a bounded store is only safe for readers whose "absent" branch is a safe default. The entity snapshot was the one migrated family where it is not, which is why the tombstone exists rather than a blanket exemption.

## Test assessment

Covered: cap/LRU/reporter mechanics, production-surface regression, measured scale, the memory_sample field and its byte attribution, evicted-snapshot honesty and its re-seed, and read/write key symmetry for changed symbols. Not covered: a byte-axis bound (deliberate — the issue frames growth as entry-count with small values; count-only matches "bound the axis that actually grows"), and tombstone overflow behavior, which needs 8,192+ distinct evicted keys to reach and degrades to the already-covered first-dispatch path.


apmantza:
Thank you for this. The class sweep that found the third family, the measured 500-file number, and the decision to reuse #2243's discipline instead of inventing a third mechanism are all the right calls. This review is AI-generated, with the maintainer supervising the process.

Verdict: needs changes before this leaves draft. The mechanism is correct and the guard is real, but three findings sit on top of it. Two of them mean roughly 40% of the new 4096-entry budget is spent on entries nothing ever reads.

I reproduced your red run exactly. Reverting the six production files to master and keeping the tests gives 6 of 7 red, including the verbatim `AssertionError: expected 1 to be +0` on the production-surface case. Neutering the guard (removing the `this.evictColdBoundedSessionFacts()` call in `setBoundedSessionFact`) reds 5 of 7. The guard is not vacuous. CI on `f1921ff` is green with Unit tests and Lint both executed.

## F1 (high) Entity-snapshot eviction inverts to "every symbol in the file changed"

`clients/review-graph/service.ts:57` falls back to an empty Map when the snapshot is evicted. Every entity then lands in `added`, so `clients/dispatch/runners/tree-sitter.ts:696-703` sees a non-empty `changedEntityKeys` and schedules a blast-radius run, and `changedSymbolsByFile` receives the whole file.

Probe, run against the built code:

```
const store = new FactStore("dispatch");
recordEntitySnapshotDiff(store, target, snap);        // added: 3  (correct, first sight)
recordEntitySnapshotDiff(store, target, snap);        // added: 0  (correct, unchanged)
for (let i = 0; i < 4200; i++) store.setBoundedSessionFact(`session.baseline./repo/src/f${i}.ts`, []);
recordEntitySnapshotDiff(store, target, snap);        // added: ["function:alpha","function:beta","class:Gamma"]
```

The file did not change. The design note in `clients/dispatch/fact-store.ts:66-80` says an eviction "only means the next dispatch of that path sees no previous baseline, the same safe state as its first ever dispatch." That holds for `session.baseline.*`, where an absent baseline means "report everything", which is the honest answer. It does not hold for the entity snapshot: a first-ever dispatch reporting everything as added is correct by construction, while an evicted one reporting everything as added is a false change signal that drives real downstream work. Either exempt the entity-snapshot family from the bound, or bound it separately with a distinguishable "evicted" marker so the consumer can tell "no prior snapshot" from "no change".

## F2 (medium-high) The cascade migration buys nothing and spends budget

`clients/dispatch/integration.ts:2413-2420` reads `cascadeDiagnosticBaselines.get(key) ?? sessionFacts.getBoundedSessionFact(key)` and then writes both. The two maps are written together on the same key and cleared together in `resetDispatchBaselines` (`integration.ts:515` and `integration.ts:523`). Eviction removes from the bounded map only. So the shadow map is a strict superset of the bounded map for this key family at all times, and the right side of the `??` can never return a value the left side missed.

The result is that the cascade family gains zero memory, because the shadow map still retains every entry, while its bounded writes evict entries from the two families that are read. Your issue comment already calls the shadow map "fully redundant". Removing it here is about four lines and turns this finding into the memory win the PR is for.

## F3 (medium) The changedSymbols write key and read key never match

`clients/review-graph/service.ts:85` writes `${CHANGED_SYMBOLS_PREFIX}${filePath}` with the raw path. `clients/review-graph/builder.ts:3624` reads `${CHANGED_SYMBOLS_PREFIX}${normalizeMapKey(filePath)}`. `normalizeMapKey` lowercases and forward-slashes, so any path with an uppercase character produces two different keys:

```
writeKey  session.reviewGraph.changedSymbols:/home/me/repo/src/Target.ts
readKey   session.reviewGraph.changedSymbols:/home/me/repo/src/target.ts
read hits write?  false
```

This predates your PR. It becomes yours to answer because the sweep enumerated this family, migrated it, and the PR body states "read+write pairs kept symmetric". They are not symmetric. A write-only family is the worst thing to put in a capped shared budget. Normalize at the write in `service.ts`, or note it as a second remainder on the issue.

Taken together, F2 and F3 mean two of the five keys minted per file are never read on the path that matters, so the real coverage of the 4096 cap is well under what it looks like.

## F4 (medium) `sessionFactEntries` is a count with no bytes behind it

`clients/memory-sampler.ts:288` still computes `estimatedBytes: dispatchCaches.recentlyCleanNeighborCacheSize * 320`. The new entry count contributes nothing to it, so the byte attribution for `dispatchCaches` still omits the whole `sessionFacts` footprint. The registry sweep at `tests/clients/memory-sampler.test.ts:267` checks that each subsystem exposes a byte field, so it passes and cannot see a count-only field added inside a subsystem that already has one. That is the #2114 and #2132 hole recurring one level down. Fold an estimate for the session-fact entries into `estimatedBytes`, using the 439,560-byte measurement you already have.

## F5 (low) State the file count the bound implies

`MAX_SESSION_FACT_RECORDS = 4096` has no stated derivation. With all families writing, a probe shows 5,000 dispatched files leave 819 of them still holding an absolute baseline. `fileFacts` carries a count cap and a 64 MiB byte cap; this map carries a count cap only, and entity snapshots hold up to 200 entries per query definition, so their per-entry weight is far above a two-diagnostic baseline. Put the implied file count in the doc comment at `fact-store.ts:81` so the next reader can judge whether 4096 still fits their batches.

## Cleared under attack

Session re-arm: `clearAll` clears `boundedSessionFacts` and `resetDispatchBaselines` calls it, so nothing latches past `session_start`. Ledger discrimination: the `session-count` axis gets its own subject, and I confirmed the file-count axis does not consume its slot. LRU read-touch: correct, and the read of the absolute key before the relative key leaves the relative key colder, which is the right victim. Changelog fragment front matter: valid. Targeted suites: 189 passed across nine files, and 67 across seven more, with two failures that reproduce as green in isolation and were parallel-run contention.

Findings F1 through F3 are worth a fix round. When you push, the maintainer will re-verify with the same probes.

AngriestBird:
Fix round pushed as `2431a946`. All five findings addressed. One correction to F3's stated mechanism below.

## F1 — evicted entity snapshot no longer inverts (fixed)

Your read is right and the probe reproduces. The gap is that "absent means report everything" is honest for the diagnostic baselines and dishonest for the entity snapshot, and the design note treated the two families as one.

Rather than exempt the snapshot family (which would give back most of the memory win, since snapshots are the heaviest entries), the store now makes eviction distinguishable:

- `FactStore` keeps `evictedSessionFactKeys`, a Set of keys capacity eviction dropped and nothing has rewritten since. `setBoundedSessionFact` clears a key's tombstone; `clearAll` clears the set. Tombstones carry the same `MAX_SESSION_FACT_RECORDS` cap in FIFO order, so they are keys-only and bounded — when the coldest tombstone falls off, that key reverts to first-dispatch semantics, which is the honest state by then.
- `wasBoundedSessionFactEvicted(factId)` exposes it.
- `recordEntitySnapshotDiff` (`review-graph/service.ts`) branches on it: an evicted snapshot re-seeds from `nextSnapshot` and returns `{added: [], removed: [], modified: []}`. No false change signal, no blast-radius run, and the next real edit diffs normally against the re-seeded snapshot.

Your probe, post-fix:

```
recordEntitySnapshotDiff(store, target, snap);        // added: 3   (first sight)
recordEntitySnapshotDiff(store, target, snap);        // added: 0   (unchanged)
for (let i = 0; i < 4600; i++) store.setBoundedSessionFact(`session.baseline./repo/src/f${i}.ts`, []);
recordEntitySnapshotDiff(store, target, snap);        // added: 0   (evicted, not "everything")
```

Two regression tests in `fact-store-session-bound.test.ts`. Red with the branch removed:

```
× reports no diff for a snapshot the cap evicted, not every symbol
AssertionError: expected { …(3) } to deeply equal { Object (added, removed, ...) }
- "added": [],
+ "added": [ "function:alpha", "function:beta", "class:Gamma" ],
```

The second test (`re-seeds the evicted snapshot so the next real edit still diffs`) guards the re-seed inside the branch: drop the `setBoundedSessionFact(snapshotKey, …)` line and the following edit reports `modified: []` instead of `["function:beta"]`.

## F2 — cascade shadow map removed (fixed)

Confirmed exactly as you described: same key, written together, cleared together, evicted from one side only, so the `??` right side was unreachable and the family gained nothing. `cascadeDiagnosticBaselines` is gone (declaration, the `resetDispatchBaselines` clear, and both accesses in `applyCascadeDeltaBaselines`); the bounded map is now the only store. Four lines, and the cascade family's win is real.

## F3 — write key normalized (fixed), with a correction

Fixed: `service.ts` writes `${CHANGED_SYMBOLS_PREFIX}${normalizeMapKey(filePath)}`, matching `builder.ts:3624`.

The correction: `normalizeMapKey` does not lowercase on POSIX. `normalizeFilePath` (`clients/path-utils.ts:95`) converts backslashes to forward slashes, then returns the path unchanged when `process.platform !== "win32"` and the path is not Windows-shaped. So on Linux `/home/me/repo/src/Target.ts` normalizes to itself, and your read/write pair actually matched there. The asymmetry bites on Windows, and on any path carrying a backslash separator (the #1150 shape) on any platform.

That matters for the test: an uppercase-path assertion is vacuously green on CI. The regression test uses `/repo/src\nested\Target.ts` and asserts `normalizeMapKey(path) !== path` first, so it cannot go vacuous. Red on the unnormalized write:

```
× writes changedSymbols under the key builder.ts reads
AssertionError: expected undefined to deeply equal [ 'alpha', 'beta', 'Gamma' ]
```

The finding stands; only its blast radius was narrower than stated. The entity-snapshot key is deliberately left unnormalized — it is written and read only inside `recordEntitySnapshotDiff`, so it is self-consistent, and lowercasing it on Windows would collide two genuinely distinct POSIX paths for no gain.

## F4 — bytes behind the count (fixed)

`estimateDispatchCacheBytes` is now an exported pure function in `memory-sampler.ts` and `dispatchCaches.estimatedBytes` calls it: `recentlyCleanNeighborCacheSize * 320 + sessionFactEntries * 440`, the 440 taken from this PR's own 1,000-entry / 439,560-byte measurement. Extracting it is what makes it mutation-provable — the registry sweep cannot see a count-only field inside a subsystem that already has a byte field, exactly as you said, so the test asserts the arithmetic directly. Red with the session-fact term deleted:

```
× attributes bytes to session-fact entries, not just the neighbor cache
AssertionError: expected 640 to be 440640
```

## F5 — the implied file count (fixed)

The `MAX_SESSION_FACT_RECORDS` doc comment now states that five per-file keys are minted per dispatched file (absolute and relative baseline, cascade baseline, entity snapshot, changed symbols), so 4096 records covers roughly 800 distinct files before the coldest file's baseline degrades. It also records the eviction-honesty split: the "same safe state as a first-ever dispatch" equivalence holds only where absent means "report everything", which is why the snapshot family gets a tombstone.

## Verification

`npm run build` clean, `npm run lint` clean, `npm run fmt:check` clean. `fact-store-session-bound.test.ts` 10/10, `memory-sampler.test.ts` 24/24, plus `review-graph.service.test.ts`, `dispatcher-flow.test.ts`, `rule-policy-dispatcher.test.ts`, the whole `tests/clients/dispatch` directory, and the two cascade suites, all green.

The shadow-map remainder named in the issue comment is now delivered rather than deferred, so the PR body's remainder note is updated accordingly.

AngriestBird:
CI on `2431a946` caught one thing the targeted suites did not: `tests/clients/session-state-conformance.test.ts` pins a stateful-symbol count per module, and removing `cascadeDiagnosticBaselines` took `dispatch/integration.ts` from 9 to 8.

`573473fb` repins it to 8 and drops the map from the `dispatch-integration:sessionCaches` registry entry's state list. Conformance suite 86/86, lint and fmt clean. That is the whole delta; no production code changed.

Full run on the previous head was 10,774 passed / 1 failed / 61 skipped, the failure being that pin alone.

PR #2373, opened 2026-08-29
fix(dispatch): restore embedded-script napi coverage (closes #2347)

### LINKED ISSUE #2347: napi runner has no embedded-<script> coverage that ast-grep LSP has
# Evidence

On the same `tmp-google-page.html` dispatch (2026-08-28 12:33, project
`pi-webaio`, pi-lens 4.1.2), the two ast-grep engines disagreed:

    12:33:16.849 astgrep_napi_unsupported_rules_skipped
                 "mismatch:javascript->html":{count:79, route:"napi"}
                 "mismatch:typescript->html":{count:134, route:"napi"}

The napi runner skipped all 221 JS/TS/TSX rules and returned zero
diagnostics. The ast-grep LSP server on the same content returned 161
ast-grep findings from embedded `<script>` blocks — ast-grep 0.45.1 resolves
HTML script injections itself and runs `language: JavaScript` rules inside
them (verified by direct CLI repro against the same binary and generated
baseline config).

# Root cause

`clients/dispatch/runners/ast-grep-napi.ts:676-687` drops any rule whose
language differs from the file's language. That guard is correct for plain
mismatches but has no concept of embedded injections, so napi gives zero
embedded-`<script>` coverage where the LSP gives full coverage.

The comment at `clients/lsp/server.ts:3444-3446` claims rule semantics are
identical across the three ast-grep routes. This falsifies it.

# Impact

Under Gate B fallback (ast-grep binary absent, napi substituting), all
embedded-JS findings in HTML silently vanish. Low severity today because the
LSP route is primary; the divergence only bites in the substitute role.

# Acceptance criteria

- Decide and record the intended contract: either napi gains embedded-script
  coverage for HTML, or the divergence is documented as accepted and the
  `server.ts:3444-3446` comment is corrected to name it.
- If coverage is added: a regression test runs a JS rule against an HTML
  fixture with an embedded `<script>` through the napi runner and asserts the
  finding; proven red on pre-fix code.
- If accepted as-is: the skip record distinguishes "mismatch on a plain file"
  from "mismatch on a file with embedded injections" so the coverage gap is
  countable in production.

# Observability

The `astgrep_napi_unsupported_rules_skipped` record exists and fired; it does
not distinguish injection-bearing files. That distinction is the gap.

# Cross-links

- #2326 — napi/LSP capability divergence on catalog languages: different
  axis (whole-file language support), same engine pair.
- The minified-HTML classification issue filed alongside this one: fixing it
  removes this file class from the scanners entirely, but the divergence
  remains for hand-authored HTML with inline scripts.

apmantza:
## Summary

The napi ast-grep fallback skipped every JS/TS/TSX rule as a `language` mismatch on HTML files, returning zero embedded-`<script>` findings while the ast-grep LSP and CLI resolve script bodies and run `language: JavaScript` rules inside them (161 findings against 221 skipped rules on the reported dispatch). Closes #2347 by giving napi the same embedded coverage: `evaluateAstGrepRules` reparses each `script_element` body with the addon's `js` grammar and runs JS rules there, translating findings back to file coordinates via the body's byte offset and a file line-start table.

Contract pinned against the real ast-grep 0.45.1 CLI before coding:
- Every `<script>` body is injected, unconditionally. `src`, `type="text/template"`, `type="application/json"`, and `type="application/ld+json"` all still scanned.
- Only `language: JavaScript` rules enter scripts. The TypeScript twin never fires on script content.
- `<style>` bodies are NOT injected. Accepted divergence, recorded in the server.ts comment.

The shared `evaluateAstGrepRules` seam lifts both the per-edit runner and the project scanner. Acceptance criterion "coverage added" is met, so the regression test path applies. `<style>` is the documented residual divergence, matching the CLI. The issue's alternative branch (accept-as-is with a distinguishing skip record) is superseded: the mismatch record now carries `htmlInlineScriptCount` on every HTML entry, so the residual "no injection target" state is countable anyway.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch

## Checklist

- [x] I have read CONTRIBUTING.md and AGENTS.md (the test-authoring screens, the severity policy, the real-runner rule, the red-first and mutation-proof rules, and the PR contract)
- [x] The change has tests (regression, contract pins, position mapping, observability): tests/clients/dispatch/runners/ast-grep-napi-html-embed.test.ts
- [x] Targeted test files for the touched seams pass locally after `npm run build`; the full suite also ran locally once (850 files / 10878 tests, with the pre-existing Windows `sh`-less hooks test as the only red).
- [x] Every NEW regression test is proven RED on pre-fix code; the red output is quoted below
- [x] Every new guard/branch/filter is mutation-proof: the `runsInsideHtmlScript` guard and the position translator each red on a proven mutation
- [x] PR title carries the conventional prefix and the issue ref (closes #2347)
- [x] `npm run lint` passes
- [x] `npm run build` passes (in-place emit keeps the vitest stale-build guard green). `build:dist` is the publish-lane build; it is not part of this PR's verification because no dist artifact ships from a worktree.
- [x] `package-lock.json` is in sync (checked via `npm run check:lockfile`); no dependency changes
- [x] `AGENTS.md` is updated: a new paragraph under the ast-grep-rules section documents the embedded-`<script>` contract and its caller requirements
- [x] `.changelog/2347-embedded-script-napi.md` has one valid Fixed entry in this PR
- [x] Commit subject includes the issue number `(closes #2347)` (subject: `fix(dispatch): restore embedded-script coverage (closes #2347)`)

## Tests

NEW file: `tests/clients/dispatch/runners/ast-grep-napi-html-embed.test.ts` (13 tests).

Part 1, real-seam regression and contract:
- `runs a language: JavaScript rule inside an inline script (red on pre-fix)` — `no-global-eval-js` must fire inside a script at the exact file line and column. This is the primary regression proof.
- `translates a same-line inline script back to the file column` — line 2, column 9 for `<script>eval(...)` on one line.
- `runs JS rules in every script block with per-block positions` — two scripts yield findings on lines 6 and 9.
- `injects a src-bearing script body, matching the CLI (unconditional)` — a `src` attribute does not suppress injection.
- `injects a type=application/json body, matching the CLI (type-agnostic)` — `no-dupe-keys-js` fires inside a JSON-typed script.
- `does not run TypeScript rules inside scripts, matching the CLI` — the `-js` twin fires, the TypeScript twin never does.
- `produces no embedded finding for an HTML file without scripts` — scriptless HTML still yields zero JS findings and zero injections.
- `drives the full runner against the real fixture` — the per-edit substitute runner path returns a blocking `no-global-eval-js` finding.

Part 2, extraction unit pins:
- `extracts each script body with its file byte offset`, `skips whitespace-only and empty script bodies`, and `handles an HTML file with no script elements at all`.

Part 3, observability through the REAL `latency.log` sink (#1742 real-sinks rule, `PI_LENS_TEST_MODE=0` scoped):
- `marks a scriptless HTML file's javascript mismatch with htmlInlineScriptCount 0`.
- `emits javascript->html mismatch only when no script ran, and counts scripts otherwise`.

Test-authoring screens satisfied: parallel path (the driver enters at `evaluateAstGrepRules`, the exact function both production callers invoke; one test drives the full `astGrepNapiRunner.run`), wrong-layer pin (the real catalog and real grammars, no hand-supplied rule), real seams (the real loaded addon, real files, real latency bytes), not-throw avoided. `assertGrammarAvailable`-style vacuity is blocked by a `beforeAll` hard failure when the addon does not load.

RED proof (main regression test against pre-fix production code, same test file):

```
 FAIL  tests/clients/dispatch/runners/ast-grep-napi-html-embed.test.ts > embedded <script> coverage (#2347)
 > runs a language: JavaScript rule inside an inline script (red on pre-fix)
    99|   );
   101|   expect(finding).toBeDefined();
```

All 13 tests failed on pre-fix code; the assertion failed for the defect's reason (the finding is absent), not a setup error.

Mutation proofs:
- Nuetering the guard (`const runsInsideHtmlScript = false`) reds `runs a language: JavaScript rule inside an inline script` with `expected undefined to be defined`.
- Nuetering the position translator (return body-relative instead of file-absolute) reds `translates a same-line inline script back to the file column` at `expect(finding?.line).toBe(2)` (received 1).

Both mutations were built, run, and reverted; quotes captured above and in the red-run transcript.

### Test assessment

`ast-grep-napi-html-embed.test.ts` uniquely pins embedded-`<script>` coverage: the injection contract, file-absolute position mapping, and the `htmlInlineScriptCount` skip-record field. Nothing else pins it, and no existing test is made redundant (the sibling napi suites assert disjoint behavior: language routing, utils, tsx, css, Gate-B coverage accounting).

## Blast radius

`module_report` with `blastRadius` is unavailable in this session (no pi host tool surface). Manual dependency map, traced by grep:

- `clients/dispatch/runners/ast-grep-napi.ts` — `evaluateAstGrepRules` gains two optional options (`content`, `sgModule`) and a new exported helper (`collectHtmlScriptInjections`). Existing callers that omit the options keep byte-identical behavior (verified: the file-root scope's id and diagnostic fields are bit-for-bit the pre-change ones). Callers: the per-edit substitute runner `run()` (now passes both) and the project scanner `scanAstGrepFile` (now passes both). The `astgrep_napi_unsupported_rules_skipped` record gains `htmlInlineScriptCount` on HTML files only.
- `clients/project-diagnostics/scanner.ts` — passes `content` and `sgModule` through to the shared seam; no other behavior change. Scanner suites green.
- `clients/lsp/server.ts` — doc comment only.
- Hot path: napi evaluations only happen under Gate B when the LSP has not published. For HTML files the added cost is one `findAll` over `script_element` nodes plus a JS parse per inline body (bounded by the existing 1 MiB file cap and per-rule/match limits). Non-HTML files pay zero (the injection branch is gated on `fileLang === "html"` before any call).
- Tests touching the touched seams (napi runner family, scanner suites, governance) all pass; the diagnostic `id` shift was avoided by keeping the file-root path identical.

## Observability

`astgrep_napi_unsupported_rules_skipped` (latency.log) now carries `htmlInlineScriptCount` on every HTML entry, so a `mismatch:javascript->html` row is distinguishable as "no injection target" and TS/TSX rows on the same file report the script bodies they could not enter. The substitute-floor record `astgrep_napi_substitute_floor` already logs the per-run diagnostic count, which now includes embedded findings. Test pins both fields through the real latency sink.

## Class sweep

Defect class: a rule-engine surface silently lacking a coverage axis its sibling surface has, because it cannot represent embedded content as a rule target. This is a variant of the AGENTS.md substitute-surface discipline (shape 26) and the "silencing counted as fixing" family (shape 10); the concrete shape is the napi fallback disagreeing with the ast-grep LSP/CLI on HTML, which the issue names.

Pattern sweep (grep `script_element`, `raw_text`, `html.parse`, embedded/injection handling across `clients/`, `tools/`, `mcp/`, `index.ts`): the only HTML-parsing engine surfaces inside pi-lens are the napi runner (now covered) and the ast-grep CLI/LSP (the reference, already covered). The tree-sitter engine has no HTML query dir, so no tree-sitter embedded axis exists to mirror. `scripts/smoke-tools.mjs` ast-grep fixtures exercise the CLI/LSP route, which already injects.

Population sweep: parse-into-HTML runtime consumers are exactly two code paths: the per-edit runner and the project scanner, both funneled through `evaluateAstGrepRules` (fixed once). `<style>` bodies are an accepted divergence (CLI 0.45.1 does not inject them either, verified by direct `no-important` probe; recorded in the server.ts comment and AGENTS.md paragraph). No parser or rule surface is left with the gap.

Consolidation verdict: the two consumers were already one seam, so no new fold is invented; the AGENTS.md record names the shared seam as the carry surface for any future ast-grep embedded-content axis.


apmantza:
Adversarial review round 2 dispositions, on the exact reviewed head `3d2d29f2` and re-verified on the rebased current head `8bf8ed56`:

**F1 — HIGH Unicode coordinates (fixed in 8bf8ed56).** `@ast-grep/napi`'s `Pos.index`/`column` are UTF-16 code-unit offsets (probed against 0.45.1 with `é` and an emoji); the round-1 translation added them to a UTF-8 byte line table. The line table is now built from the JS string (`computeLineStartsUtf16`), so one unit carries the whole mapping. Regression tests pin exact coordinates with a multibyte char before the script on the previous line (`é` and `😀`), and multibyte text inside the script body; all three went red on the reverted byte table.

**F2 — HIGH unbounded script parse work (fixed in 8bf8ed56).** Embedded evaluation is `rules × bodies × body-size`; measured on this seam at ~8.5 ms per body, ~9 s at 500 blocks. `MAX_SCRIPT_BODIES_EVALUATED` (64) and `MAX_SCRIPT_BODY_BYTES_EVALUATED` (128 KiB) now cap the evaluated set before any grammar parse, with a bounded `ast-grep-napi-html-script-budget` degradation record naming the file and the dropped total. 2000 script blocks now evaluate in ~626 ms. The first body is always admitted so a single over-cap inline script matches its file's own root-evaluation cost. The cap guard is mutation-proven.

**F3 — MEDIUM silent parser failures (fixed in 8bf8ed56).** Missing JS grammar, HTML `script_element` scan failure, and JS-body parse failure each dropped embedded coverage with no record. `collectHtmlScriptInjections` now preserves the raw `scriptElementCount` and returns per-failure counts; `evaluateAstGrepRules` emits four new bounded `DegradationKind`s (`ast-grep-napi-html-*`) keyed by file path, using `recordDegradationOnce` / `incrementDegradationCount` so exact totals survive the ledger's per-kind window. Emission and collection failures are mutation-proven.

Verification: `npm run lint`, `fmt:check`, `changelog:check`, `check:lockfile`, `depcruise:check`, `astgrep:self-scan`, `npm run check`, rule-catalogs, governance (`tests/config`), deps-centralization, and the affected napi/scanner suites all pass on the rebased head. The full local suite is green except the pre-existing Windows-only `pre-push-targeted-tests` (`sh` absent) and one contention flake in `formatters.test.ts` (passes in isolation; seam untouched). CI on the new head is authoritative.

No material deferral: `<style>` bodies remain the sole accepted divergence (CLI 0.45.1 does not inject them), as recorded in the round-1 notes.

apmantza:
Luna review round 3 disposition, verified on the exact reviewed head `8bf8ed56` and fixed on `2516193f`:

**F4 — HIGH first embedded body bypasses MAX_SCRIPT_BODY_BYTES_EVALUATED (fixed in 2516193f).**
The round-2 budget loop admitted the FIRST script body regardless of size, so a single oversized inline body ran the whole catalog synchronously (bounded only by the runner's 1 MiB file gate). Reproduced on the reviewed head: a 1,047,735-byte dense first body evaluated in **19.5 s with zero findings** (the reviewer's 13.1 s class; content-shape dependent, denser content is worse).

Fix: the byte cap now binds EVERY body, first included. The `firstBody` bypass is removed; an over-cap body is omitted and counted in `truncatedBodies`, which feeds the existing bounded `ast-grep-napi-html-script-budget` degradation record — so a file that loses embedded coverage always carries the record naming why.

Red-first regression (real seam, `ast-grep-napi-html-embed.test.ts`): one over-cap first body must yield `bodiesEvaluated: 0`, `truncatedBodies: 1`, the budget record (`0/1 bodies evaluated`), and zero diagnostics. Against the reinstated pre-fix bypass it fails for the defect's reason: `expected 1 to be +0` on `bodiesEvaluated` (the body WAS evaluated). A boundary test pins the opposite side: a sole under-cap body still evaluates with no record, so the hard cap does not over-restrict.

Guard mutation proof: reintroducing the first-body bypass (the pre-fix loop, byte-for-byte) reds the oversized-body test; the assertion and the red quote are above.

Post-fix measurement: the same 1,047,735-byte dense fixture now completes the collection in **4 ms** with `bodiesEvaluated 0 / truncatedBodies 1` — the seconds-long synchronous stall is gone.

Sweep of embedded body-entry paths: there is exactly ONE admission path (the `selected` loop in `collectHtmlScriptInjections`) and the `firstBody` special-case was its only bypass; after removal every candidate passes the same count+bytes gate. Verdict per member: candidate admission = fixed, no other path parses a body. Observed sibling (out of scope): the file-root path evaluates a plain 1 MiB `.js` file with the same catalog — a pre-existing, non-HTML cost shape not part of this PR's seam; recorded here rather than changed.

Verified on the pushed head: `npm run build`, lint, fmt:check, astgrep:self-scan, changelog:check, check:lockfile, depcruise:check, `npm run check`, governance (`tests/config`), deps-centralization, and the focused napi/scanner/degradation suites (147 + 1711 tests) all pass. Skipped: the full local suite is not re-run for this delta (CI Unit tests on the pushed head are authoritative, as in the prior rounds); the pre-existing Windows-only `pre-push-targeted-tests` (`sh` absent) remains the only known local red, unrelated to this change.

Preserved behavior: Unicode UTF-16 coordinates, multi-body count/byte budgets, parser-failure records (`ast-grep-napi-html-script-parse-failed`), language selection (only `language: JavaScript` enters scripts), and unconditional `type`/`src` injection all retain their prior verified tests, which pass unchanged.

CI on `2516193f` will confirm; no merge or automerge enabled.

Issue #2074, opened 2026-08-25
One-file review-graph rebuild is O(graph), not O(changed): 350ms at 8k nodes, 1,900ms at 32k (refs #1318)


apmantza:
The #1318 umbrella asks for an O(changed) rebuild. Measurement says the rebuild is O(graph): changing **one** file costs 350 ms on an 8,000-node graph and 1,900 ms on a 32,000-node graph, on the seq fast path that exists specifically to be incremental. The per-changed-file work is 36 ms; everything else is fixed cost proportional to the whole graph.

## Evidence — the rebuild growth curve

Windows dev box, compiled runtime, synthetic trees, each file importing two others. Cold build and rebuilds measured through `buildOrUpdateGraph`.

Without a seq hint (the shape of 6 of the 7 production call sites — only `clients/dispatch/integration.ts:1021` passes one):

| project files | nodes / edges | cold build | warm, nothing changed | rebuild, 1 file | rebuild, 10 files | rebuild, 500 files |
|---|---|---|---|---|---|---|
| 2,000 | 8,000 / 18,000 | 27,748 ms | 1,058 ms | 2,074 ms | 3,017 ms | 19,786 ms |
| 8,000 | 32,000 / 72,000 | 110,810 ms | 3,468 ms | 6,885 ms | 11,017 ms | — |

Four times the files gives 4.0x the cold build and 3.3x the one-file rebuild. The 2,000-file row also separates the two components: fixed cost ~1.9 s, marginal cost ~36 ms per changed file.

With a seq hint, so the #451 fast path fires (`mode=seq-fastpath` on every iteration, no fallback):

| project files | nodes / edges | 1-file rebuild, steady state |
|---|---|---|
| 2,000 | 8,000 / 18,000 | 334-478 ms |
| 8,000 | 32,000 / 72,000 | 1,590-2,638 ms |

Same one changed file. Four times the graph, 5.4x the cost.

## Evidence — where the fixed cost goes

Isolated cost of each O(graph) phase on the 32,000-node / 72,000-edge graph, measured against the live graph object:

| phase | cost per rebuild | repeating axis | site |
|---|---|---|---|
| `countRetainedSourceFiles` realpath loop | ~1,540 ms | per project source file | `builder.ts:1650` |
| serialize + gzip the whole graph for persist | 543 ms (944 KiB out) | per node + edge | `persistGraph` |
[...]

Issue #1976, opened 2026-08-21
Ignore matcher recompiles patterns per file; memoize compilation per pattern and base


apmantza:
Evidence (from #1974's profiling, yams repo, 45,378-file warmup walk): ~60% of the walk's CPU was inside minimatch pattern COMPILATION (regExpEscape 22.5%, unescape 10.7%, toRegExpSource/make/parseGlob), not matching. isIgnored measured 0.68-0.93ms per unique file. The patterns are identical for every file in a walk, yet the matcher recompiles them per call.

Root cause: the existing memo is keyed by file PATH, which a walk visits exactly once each, so it never hits. The repeating axis is the pattern. The compile-once form makes every check a cheap regex.test.

Why this is separate from #1974's order swap: the swap only helps files an extension gate can discard. Sites that must consult ignore for every file get nothing from it — pipeline.ts's snapshot walk (cleared in #1974 as legitimately gate-less) and any future walker. Memoization is the structural fix for all fourteen isIgnored call sites at once.

The correctness trap, stated up front: gitignore patterns are directory-relative. A nested .gitignore scopes its patterns to its subtree, and pattern semantics depend on the base directory. The memo key must be (pattern, base context, flags), or the compiled regex must be provably context-free. A pattern-only key returns wrong answers for nested ignore files.

Acceptance criteria:
- Pattern compilation happens once per (pattern, base, flags) per matcher instance; repeated isIgnored calls reuse it.
- Correctness proven against nested .gitignore fixtures: a subtree pattern must not leak to sibling trees through the cache; red-first tests for the trap above.
- Measured: re-run #1974's yams-shape fixture walk (or the synthetic wal/ fixture from its PR) with the order swap already in place; the remaining isIgnored cost drops by the compilation share. Assert on call/compile counts, not wall time (#1920).
- The per-path memo is removed or justified: if it serves the per-event call sites (dispatch/integration, lsp/index — repeated edits to one file), keep it alongside and say so; if it serves nothing, delete it.
- Bounded: the cache lives per matcher instance with matcher lifetime; no process-lifetime growth across sessions (catalog shape 17 screen).

Observability record: the warmup and scan phase durations in sessionstart.log/latency.log prove the field effect, as in #1974.

Refs #1974 (the profiling and the order-swap fix this composes with).

apmantza:
[...]

Issue #1920, opened 2026-08-20
Real-clock timing tests flake under worker contention (startup-overhead, scan-cache)


apmantza:
Two test files assert wall-clock budgets. They fail under high test-file parallelism and pass solo. Both sides of a controlled comparison (same 7-file batch, branch and master builds) show the same failures, verified during the #1919 review:

- `tests/clients/startup-overhead.test.ts`, "quick mode self-reports within 500ms": measured 659-2321ms under contention on both master and branch. Green solo, 3 of 3 runs. Same magnitude both sides, so not a regression.
- `tests/clients/runtime-session-scan-cache.test.ts`: flaked in the #1919 fixer's worktree under parallelism, independent of the diff. Verified by substituting unrelated files.

Defect shape: a real-clock budget assertion in a machine-shared parallel vitest run measures scheduler contention, not code speed.

Acceptance criteria:

- Both tests use fake timers or logical clocks, assert on recorded spans instead of live elapsed time, or run in an isolated serial pool with a calibrated budget.
- Sweep: grep tests/ for other real-clock `toBeLessThan(<ms>)` assertions. List them in a comment. Fix here or file siblings.

Refs #1919 (review evidence). Note: #1919 raised the quick-mode observability test to `INTEGRATION_TIMEOUT_MS` as an interim fix for one instance.

apmantza:
Family addition from the #1930 review: tests/clients/astgrep-self-scan (scripts/lib/astgrep-self-scan.mjs) fails intermittently under vitest with exit status 6 and no output, on Windows with a junctioned node_modules. Running the scanner directly with node_modules/.bin on PATH is clean (1180 files, 13 rules, 0 findings); without PATH it fails as spawn ast-grep ENOENT. The failure tracks binary resolution across the junction, not source content or timing budgets. Different mechanism than the wall-clock entries above, same symptom class: an environment-coupled test that fails under the runner and passes in isolation. Acceptance addition: the self-scan resolves the binary by absolute path instead of PATH lookup.

apmantza:
Candidate addition from the #1933 verify round: tests/clients/lsp/workspace-diagnostics-sweep-batch-open.test.ts ('a pre-open call that never resolves does not hang the whole sweep') failed 2 of 3 parallel-batch runs and passed 3 of 3 in isolation, on a tree that changes nothing it imports. Per-file-deadline timing under machine contention, the same shape as the entries above. Not yet controlled-verified against master; treat as a candidate pending the protocol.

[...]

PR #2378, opened 2026-08-29
perf(review-graph): batch owned-edge removal in multi-file rebuilds (refs #2074)

AngriestBird:
## Summary

`removeFileOwnedGraphData` did a full `graph.nodes` scan and a full `graph.edges.filter()` once per changed file, at both call sites — so a multi-file rebuild cost changedFiles x graph, the compounding scan #2121's own class sweep re-homed to #2074 without quantifying. It now reads a file's owned nodes from the authoritative `fileNodes`/`symbolNodesByFile` indexes and its owned edges from `edgesByFrom`/`edgesByTo`, and both callers compact `graph.edges` exactly once per batch.

Non-obvious gotchas handled: compaction lands before `pruneOrphanNonFileNodes`/`restoreValidIncomingEdges` read `graph.edges`; the checkpoint-resume path is safe because the loaded graph is fully indexed (`rebuildIndexes` at load) before eviction; each touched adjacency bucket is filtered once per batch, so high-fan-in hubs avoid repeated linear indexOf/splice removal while within-batch position deduplication preserves correctness.

Refs #2074 — AC1/AC2 already landed via #2092/#2121; AC3 (delta persist) and AC4 (2x growth bound) remain and need a persist-format/edge-store redesign, not another incremental fix. Remainder noted on the issue.

## Tests

- `tests/clients/review-graph/rebuild-cost.test.ts` (edit) — the multi-file ring test holds collection visits flat (`766` pre-fix versus `190`), and the high-fan-in test pins bucket-removal work at `408` positions for 200 callers and `808` for 400 callers. Reverting to per-edge unindexing produces `82,214` positions against the `63,342` bound. Both tests use production counters, not wall-clock time.
- No edits to existing suites: review-graph 145/145 (2 pre-existing skips), seq-fastpath 8/8, checkpoint-resume 5/5, and the randomized `incremental-equivalence.test.ts` (its edit sequence includes a multi-file batch step comparing full vs incremental graph shape) all green unmodified — the correctness contract for the restructure.

## Test assessment

Against the six screens: no parallel path (the counter test drives `updateGraphFiles` on a real graph, and the untouched randomized equivalence suite is the independent correctness oracle); no invisible skip; not a wrong-layer pin (the counter sits inside the production function, exported through the existing `_getReviewGraphRebuildCountersForTests` seam #2074 already uses); no ambient-inspection double; env leakage covered by the existing counter reset helper; the flat-counter assertion is an exact count, and the loose-bound screen is deliberately avoided per the file's #1920 wall-clock rationale.

## Blast radius

Touched: `clients/review-graph/builder.ts` only — `removeFileOwnedGraphData` and its two callers (`updateGraphFiles`, `tryResumeFromCheckpoint`). Hot path (per incremental rebuild): edge-array compaction drops from N scans to 1 per batch; owned-edge collection from O(graph.edges) to O(file fan-in/out); measured by production counters: 766 → 190 visits on the ring fixture, and 15,453/75,953 → 408/808 bucket positions for 200/400-caller fan-in hubs. `module_report` blastRadius: unavailable this session (pi-lens MCP failed to connect).

Investigated, deliberately not touched: `resolveDeferredSymbolEdges`' full edge map (runs once per batch, no compounding; needs a name→placeholder index that doesn't exist), delta persist (existing `semanticFingerprint` skip only helps no-op rebuilds), `cloneGraph` shallow-copy (would alias mutable bucket arrays with the cached graph).

## Class sweep

Class: per-item full-collection scan inside a batch loop (the #1976 re-derive-per-item shape, edge-array variant). Swept `clients/review-graph/` plus `clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts` for `graph.edges.filter`/full-scan calls inside per-file loops: after #2092/#2121, `removeFileOwnedGraphData` was the last member on the incremental path; `resolveDeferredSymbolEdges` scans once per batch (not per item) and is named above as the remaining single-scan cost. No other per-item full scans found; consolidation verdict: the family now shares the adjacency indexes as its one seam.

## Observability

The `_rebuildCounters` instrumentation (existing #2074 seam) gains `removeOwnedEdgeVisits`, exported through `_getReviewGraphRebuildCountersForTests` — the count-based production-shaped signal the issue asks for. Rebuild wall time remains visible via the existing review-graph phase logs.

### DIFF
--- a/.changelog/2074-batch-rebuild-edge-removal.md
+++ b/.changelog/2074-batch-rebuild-edge-removal.md
@@ -2,4 +2,4 @@
 section: Changed
 ---
 
-- **Bound a multi-file review-graph rebuild's edge removal (refs #2074)** — `removeFileOwnedGraphData` now finds a changed file's owned edges through `edgesByFrom`/`edgesByTo` instead of scanning `graph.edges`, and a batch of changed files compacts the edge array once instead of once per file. A multi-file rebuild previously scanned the whole edge array `changedFiles` times; it now scans it once.
+- **Bound a multi-file review-graph rebuild's edge removal (refs #2074)** — `removeFileOwnedGraphData` now finds a changed file's owned edges through `edgesByFrom`/`edgesByTo`, batches adjacency-index removal once per touched bucket, and compacts the edge array once per batch. A multi-file rebuild previously scanned the whole edge array `changedFiles` times and rescanned high-fan-in buckets once per edge; it now keeps both costs proportional to touched buckets.
--- a/.changelog/2074-batch-rebuild-edge-removal.md
+++ b/.changelog/2074-batch-rebuild-edge-removal.md
@@ -1,0 +1,5 @@
+---
+section: Changed
+---
+
+- **Bound a multi-file review-graph rebuild's edge removal (refs #2074)** — `removeFileOwnedGraphData` now finds a changed file's owned edges through `edgesByFrom`/`edgesByTo` instead of scanning `graph.edges`, and a batch of changed files compacts the edge array once instead of once per file. A multi-file rebuild previously scanned the whole edge array `changedFiles` times; it now scans it once.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1242,7 +1242,9 @@ Incremental review-graph updates resolve deferred symbol edges after restoring
 preserved incoming edges. The live indexes must remain synchronized, and
 `dedupeResolvedEdges` removes only post-resolution duplicates from affected
 target buckets. Do not add a second whole-graph dedupe pass to the hot path;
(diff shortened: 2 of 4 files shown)


apmantza:
Thanks for the contribution. This review was AI-generated and was supervised by a maintainer. The implementation and correction cleared adversarial review at exact head 656248215aeab953841e65164c544f4ebfea4c43. The high-fan-in removal probe now scales from 408 bucket positions at 200 callers to 808 at 400 callers, and the mutation returns to 82,214 positions. Focused tests, governance suites, build, lint, format, ast-grep self-scan, merge-tree composition, and both required CI checks passed. Verdict: approve.

PR #2379, opened 2026-08-29
fix(lsp): stop starving auxiliary resync after a slow cold primary spawn (closes #2239)

### LINKED ISSUE #2239: Auxiliary LSP resync queueWaitMs starves after a 5s+ cold primary spawn
## Summary

`LSPService.touchFile`'s auxiliary-resync queue wait
(`clients/lsp/index.ts` around line 4042) computes
`queueWaitMs = min(budget, max(0, options.maxClientWaitMs - elapsed))`.
`options.maxClientWaitMs` is the CALLER's originally-passed value (the
dispatch lsp-runner passes the flat `LSP_SPAWN_BUDGET_MS` constant, 5s),
not the per-server effective wait floor `getClientForFile` computes
internally from `LSPServerInfo.clientWaitTimeoutMs`.

## Evidence

Found during PR #2233's fix round (F2), which raised
`clientWaitTimeoutMs` for Bash, JSON, Vue, Svelte, and Prisma so their
cold spawns can exceed 5 seconds. Any of those five files touched with
`clientScope: "with-auxiliary"` (an enabled auxiliary scanner, e.g.
opengrep) will see the primary's cold spawn consume most or all of the
elapsed time before the auxiliary resync gate even runs. Since
`options.maxClientWaitMs` stays at the caller's original 5s regardless
of how long the primary's own spawn actually took,
`options.maxClientWaitMs - elapsed` goes negative and clamps to 0 —
the auxiliary gets zero queue time and is reported uncovered
(`notifyDeferredServerIds`) on every such touch, not only the first
cold one.

This is a pre-existing shape, not introduced by #2233 — Ruby's
existing 30s `clientWaitTimeoutMs` already triggers it. #2233 widens
the set of servers that can trip it.

## Scope

Audit whether `queueWaitMs` should derive from the SAME effective
per-server wait floor `getClientForFile` uses internally, rather than
the caller's original flat budget. Any fix must not widen the
auxiliary wait unconditionally — that would reintroduce a version of
the same per-edit latency problem #1459 bounded in the first place.

Refs #2169, #2176, #2233

AngriestBird:
## Summary

`touchFile`'s auxiliary-resync `queueWaitMs` charged elapsed time against the caller's flat `LSP_SPAWN_BUDGET_MS` (5s), while `getClientForFile` races the primary spawn against a larger per-server floor (`clientWaitTimeoutMs`: Ruby 30s, plus the #2233 Bash/JSON/Vue/Svelte/Prisma overrides). A cold primary legitimately running past 5s but under its own floor drove `options.maxClientWaitMs - elapsed` negative, clamped queue time to 0, and reported the auxiliary uncovered via `notifyDeferredServerIds` on every such touch — not only the first cold one.

Fix: extracted `primaryServerWaitFloorMs` (the exact servers-filter-then-reduce `getClientForFile` already did inline) and used it on both sides, so the acquisition race and the queue-wait computation read one source of truth. The outer `Math.min(budget, …)` is untouched — the Scope's anti-widening guard (#1459): the fix raises what "already spent" is measured against without letting any wait exceed one write budget.

Closes #2239 — the Scope's audit directive and anti-widening constraint are both satisfied and tested.

## Tests

- `tests/clients/lsp/service-scanner-coverage-gap.test.ts` — new: "an auxiliary resync is not starved when a cold primary spawn exceeds the caller's flat budget (#2239)". RED pre-fix (with only the no-op refactor in place): `expected [ { type: 'phase', …(4) } ] to have a length of +0 but got 1` on `lsp_notify_resync_deferred`.
- Same file — new guard test pinning that a naively widened fix (server floor as the outer cap) would still defer a wedged 2s write; passes pre- and post-fix by design, reds the wrong implementation (verified by hand-calculation of the deliberately-wrong shape).
- `makeServer` gained an optional `extra` override param (non-breaking, mirrors the sibling test file's pattern).

## Test assessment

Against the six screens: no parallel path (both tests drive `touchFile` through the suite's existing fake-server harness); no invisible skip; not a wrong-layer pin (the starvation is asserted on the emitted `lsp_notify_resync_deferred` record, the production symptom); no ambient-inspection double; env leakage covered by the suite's per-test service teardown; the guard test exists specifically so the fix cannot drift into an unconditionally-widened wait — mutation-proofing the `Math.min(budget, …)` cap.

## Blast radius

Touched: `clients/lsp/index.ts` — one extracted helper, two consumers (`getClientForFile`, the notify-write `queueWaitMs`). The `getClientForFile` side is a pure refactor (same filter+reduce). Behavior change is confined to files whose primary server sets `clientWaitTimeoutMs` above the caller's budget, touched `with-auxiliary`. Per-touch cost: one extra `getServersForFileWithConfig` call on the resync path (config lookup, no spawn). `module_report` blastRadius: unavailable this session (pi-lens MCP failed to connect). Verified: coverage-gap suite 26/26; touch-collect, notify-inflight-throttle (x2), aux-grace (77); every other file referencing `getClientForFile`/`queueWaitMs`/`notifyWriteBudgetMs`/`clientWaitTimeoutMs` (212 tests) green.

## Class sweep

Class: downstream consumer budgeting against a caller-declared constant while the upstream seam races against a derived effective value (split-brain budget). Swept `clients/lsp/` and the dispatch runners for other consumers of `options.maxClientWaitMs`/`LSP_SPAWN_BUDGET_MS` measured against elapsed time: the resync `queueWaitMs` was the only site subtracting elapsed from the flat caller value; `getClientForFile` derives the floor itself and the dispatch lsp-runner passes the constant through without subtracting. Class of size 1; both sides now share `primaryServerWaitFloorMs`, which is the consolidation verdict (one seam).

## Observability

Existing records prove the fix in production: `lsp_notify_resync_deferred` (the starvation signature this removes for slow-cold-spawn servers) and `lsp_scanner_coverage_gap` frequency on Ruby/Bash/JSON/Vue/Svelte/Prisma files touched with-auxiliary. No new record needed — the defect was already visible in these; their post-merge drop is the proof.

### DIFF
--- a/.changelog/2239-resync-queue-wait.md
+++ b/.changelog/2239-resync-queue-wait.md
@@ -1,0 +1,11 @@
+---
+section: Fixed
+---
+
+- **Stop starving auxiliary LSP resyncs behind a cold primary spawn (closes #2239)** —
+  the auxiliary resync queue wait now derives from the same effective
+  per-server wait floor `getClientForFile` uses internally, instead of the
+  caller's flat budget alone. A cold primary spawn that legitimately runs
+  past that flat budget (Ruby's 30s override, and the Bash/JSON/Vue/Svelte/
+  Prisma overrides from #2233) no longer zeroes the auxiliary's queue time
+  and reports it uncovered on every such touch.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -516,7 +516,11 @@ and only a scanner that cannot accept a write inside the budget makes a waiter
 give up. A write that lands after its deadline but inside the wedge window
 retracts the timeout it was charged for (slow is not broken); one nothing
 accepts for the whole wedge window keeps its strike and demotes the server, so
-the gate cannot defer a dead input path forever. A DEFERRED server is neither
+the gate cannot defer a dead input path forever. Its queue wait uses the
+effective `primaryServerWaitFloorMs`, which includes the caller's
+`maxClientWaitMs` and any primary server `clientWaitTimeoutMs` override, so a
+cold primary's configured wait does not make the remaining queue budget appear
+to be already exhausted. A DEFERRED server is neither
 waited on nor read from — its version cannot advance, so waiting only burns its
 budget and would flip the touch to `inconclusive`, and its diagnostics cache
 still holds the previous content's findings because the resync that would have
--- a/clients/lsp/index.ts
+++ b/clients/lsp/index.ts
@@ -4107,9 +4107,10 @@ export class LSPService {
 			if (!notifySkipped) {
 				const budget = notifyWriteBudgetMs();
 				// #1459: how long a queued auxiliary may wait for its resync slot. Bounded
-				// by the write budget AND by whatever the caller already declared it is
-				// willing to spend on this touch (`maxClientWaitMs` — cascade's cold
-				// snapshot passes 1000ms), minus what the client wait above already spent.
+				// by the write budget and by the effective primary wait floor, minus what
+				// which includes the caller's `maxClientWaitMs` and any primary
+				// server `clientWaitTimeoutMs` override. The elapsed client wait is
+				// subtracted below.
 				// A flat write-budget wait would tax a caller that asked for less than one
 				// budget in total. Non-positive means "no time left to queue": the server
 				// is reported as uncovered immediately.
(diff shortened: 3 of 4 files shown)


apmantza:
Thanks for the contribution. This review was AI-generated and was supervised by a maintainer. The implementation and issue acceptance criteria cleared adversarial review at exact head 8f4b5ec5134b91bd31f5b768bd60d01e865989a8. The documentation correction now describes the effective primary-server wait floor accurately. Focused tests, build, lint, format, ast-grep self-scan, and both required CI checks passed on that head. Verdict: approve.

Issue #2385, opened 2026-08-30
autofix (biome arm) rewrites value imports to import type in emitDecoratorMetadata repos — runtime crash


Macrox:
## Outcome

Prevent pi-lens's bundled Biome fallback from rewriting runtime-required value imports into type-only imports. PR #2386 implements the accepted conservative fallback policy.

## Evidence

A consumer repository uses `experimentalDecorators` and `emitDecoratorMetadata`. Biome's `useImportType` safe fix changed a value import used by emitted decorator metadata into `import type`. TypeScript erased the binding, `design:type` became `undefined`, and the application crashed during provider lookup.

The affected pi-lens path was:

1. A JavaScript or TypeScript file had no project Biome configuration.
2. Smart-default formatter policy selected the bundled Biome fallback.
3. The fallback enabled recommended lint rules, including `style/useImportType`.
4. Deferred autofix applied the import rewrite without knowing that decorator metadata required the runtime value.

## Root cause

`useImportType` assumes imports used only in type positions have no runtime footprint. Repositories using emitted decorator metadata violate that assumption because metadata generation can require the runtime value binding.

## Accepted design

Maintainer decision recorded on 2026-08-31: use a conservative bundled fallback instead of decorator-specific detection.

- Disable `style/useImportType` for both lint and autofix when pi-lens supplies its bundled Biome configuration.
- Keep explicit user Biome configuration authoritative, including configurations that enable the rule.
- Avoid a second TypeScript decorator parser and source-pattern heuristic in the dispatch path.

## Acceptance criteria

- [x] The bundled Biome fallback disables `style/useImportType` on lint and autofix surfaces.
- [x] Explicit user Biome configuration remains unchanged and takes precedence.
- [x] A real Biome and TypeScript regression proves decorated metadata retains its runtime value binding.
- [x] Mutation coverage proves re-enabling the fallback rule reproduces the failure.
- [x] Documentation and changelog text describe the fallback policy accurately.

## Non-goals

- Detect decorator metadata from `tsconfig` or source patterns.
- Emit a decorator-specific autofix skip reason.
- Override rules chosen by an explicit user Biome or ESLint configuration.

## Defect shape and class sweep

This belongs to the tool-default-overrides-project-semantics class, also represented by #1144. The implementation swept both bundled Biome fallback consumers: lint and autofix. Explicit-config paths remain outside the fallback and retain user policy.

## Implementation

[...]

Issue #1247, opened 2026-08-10
Autofix surfaces run tools with default config — divergent from lint surfaces' package-owned fallback configs


apmantza:
## Defect shape

A tool whose **lint** surface applies a package-owned fallback config (when the project has none) runs its **autofix** surface with **no config at all** → the tool's *default* ruleset applies during `--fix`, so autofix rewrites files according to rules the lint surface never endorsed (and the package config explicitly disables).

## Confirmed members (lint config vs autofix invocation)

1. **markdownlint** — lint: `--config config/markdownlint/core.json` (MD013 off, MD024 siblings_only) when `!hasMarkdownlintConfig(cwd)` (`clients/dispatch/runners/markdownlint.ts`); autofix: bare `markdownlint-cli2 --fix <file>` (`clients/pipeline.ts` `tryMarkdownlintFix`) → default all-rules-on config → **whole-file rewrites of CHANGELOG.md / AGENTS.md** and `#946` → `# 946` heading-spacing corruption (MD018/MD022 applied under the default config's heading parse).
2. **ruff** — lint: `--config config/ruff/core.toml` when `!hasRuffConfig(cwd)` (`clients/dispatch/runners/ruff.ts`); autofix: bare `ruff check --fix <file>` (`clients/ruff-client.ts` `fixFileAsync`).
3. **biome** — lint: `--config-path=<user-config-or-core.jsonc>` — **always** passes one, user's or the package fallback (`clients/dispatch/runners/biome-check.ts`); autofix: bare `biome lint --write <file>` (`clients/biome-client.ts` `fixFileAsync`) → with a user config present, autofix ignores it and uses biome's default.

## Empirical finding — the config seam is necessary but NOT sufficient

[...]

Issue #833, opened 2026-07-26
fix(markdownlint): use a valid bundled default config for unconfigured Markdown


apmantza:
## Summary

The Markdown dispatch runner produces a large amount of noisy output for unconfigured projects and fails to apply its intended default suppression for line-length warnings.

Dogfood evidence (2026-07-26): editing `CHANGELOG.md` produced 1,590 warnings. Latency showed the `markdownlint` runner produced all diagnostics; Marksman was skipped. The findings were primarily:

- `MD013/line-length` — default 80-column line limit, unsuitable for long release-note bullets.
- `MD024/no-duplicate-heading` — repeated `### Added`/`### Fixed` headings, which are intentional across changelog sections.

## Root cause

`clients/dispatch/runners/markdownlint.ts` intends to disable MD013 when no project config exists, but invokes:

```text
markdownlint-cli2 --disable MD013 <file>
```

`markdownlint-cli2` has no `--disable` option; its CLI accepts glob arguments plus `--config`, `--fix`, and related options. The attempted suppression is therefore ignored.

## Goal

Make the unconfigured markdownlint path use a bundled, package-owned default config that disables only MD013, while preserving project-owned configuration exactly as the user specified it. Do not globally disable MD024: duplicate headings should remain useful outside changelogs, and project-specific configs can opt out when appropriate.

Also recognize all supported markdownlint-cli2/config filenames, including `.markdownlint-cli2.*` and `.markdownlint.cjs/.mjs`, so an existing project config is not accidentally treated as unconfigured.

## Constraints

- Resolve the bundled config robustly from the installed package; do not assume the source-tree cwd.
- Keep project config precedence and user overrides intact.
- Add runner-level regression coverage proving the default `--config` is passed and the invalid `--disable` arguments are gone.
- Add config-discovery coverage for the newly recognized filenames.
- Update docs/AGENTS.md and the Unreleased changelog entry as appropriate.

apmantza:
Implemented locally via Luna subagent.

- Added bundled config/markdownlint/core.json disabling only MD013.
- Replaced invalid `--disable MD013` invocation with valid `--config <package config>`.
- Expanded config discovery for markdownlint-cli2 JSONC/YAML/CJS/MJS variants.
- Added runner/tool-policy regression tests.

Validation: 63 targeted tests, build, and lint pass. Commits: 7059aba0 plus follow-up 1fb397ab. Pending review/PR.

apmantza:
[...]

PR #2386, opened 2026-08-30
fix(biome): preserve decorator metadata in fallback (refs #2385)

apmantza:
## Summary

The bundled fallback Biome config no longer converts value imports to `import type`. It disables `style/useImportType`, the rule behind #2385.

The rule's fix is marked safe, and it rewrites a value import used only in type positions into `import type`. Under `experimentalDecorators` + `emitDecoratorMetadata` that import still has a runtime footprint: the emitted `design:type` metadata needs the class value. Real-binary reproduction on Biome 2.5.9 plus tsc 7.0.2:

- Before any rewrite: `__metadata("design:type", model_1.CounterModel)`, runtime reports `design:type=CounterModel`.
- After `biome lint --write` with the old bundled config: `import type { CounterModel }`, emitted `__metadata("design:type", Function)`, runtime reports `design:type=Function`.

Biome's own `biome explain useImportType` documents this exact hazard ("Caveat with TypeScript experimental decorators") and recommends disabling the rule for such repos.

**Change:** one rule override, `"useImportType": "off"`, in the `style` group of `config/biome/core.jsonc`, with a comment citing the hazard. Both surfaces (lint and `lint --write` autofix) consume it through the shared `biomeConfigArgs` seam, so neither drifts.

**Acceptance rescope (why `refs`, not `closes`):** the maintainer decision, recorded on 2026-08-31 at issue #2385 (comment 5477218580), limits the fix to the bundled fallback config and disables style/useImportType on both the lint and autofix surfaces. Issue AC1's detection (tsconfig flag primary, source-pattern fallback) and its `decorator_metadata_guard` `skipReason` are deliberately not implemented: source-pattern detection was rejected, and per-edit tsconfig discovery would add I/O and caches the decision rules out. AC3's documentation requirement is covered in `docs/features.md` and AGENTS.md. Consequence, stated plainly: import-type conversion stops in all unconfigured fallback projects, including decorator-free ones. Explicit project `biome.json(c)` configs are untouched; a user who enables `useImportType` keeps it, verified by test.

**Why no `skipReason`:** a skip reason exists when a known action is withheld. Here no code path skips anything ├ö├ç├Â the rule simply never fires, the same way a rule absent from any config never fires. A `skipReason` on every clean biome run would be an unobservable discriminator, and deriving it would require exactly the per-edit detection this fix avoids.

Thank you, @Macrox, for the detailed chain and the `latency.log` evidence ├ö├ç├Â the reproduction made the diagnosis unambiguous.

## Tests

New: `tests/clients/biome-config-decorator-metadata.test.ts`. All three drive the real pinned Biome 2.5.9 binary; the first two enter at the production entry point `BiomeClient.fixFileAsync` (hermetic project-local `node_modules/.bin` shim; no PATH, network, or managed tools tree):

- **bundled fallback keeps the value import and its design:type metadata** ├ö├ç├Â runs `fixFileAsync` on a decorator fixture, asserts the import is unchanged, and independently compiles with the real tsc and executes the emitted JS, asserting the reported `design:type` is `CounterModel`. Positive controls: `result.success === true` and `result.changed === false`, so an unchanged file is evidence about this rule, not a silently failed run.
- **an explicit project biome.json stays authoritative** ├ö├ç├Â fixture with its own `biome.json` enabling `useImportType`; asserts `biomeConfigArgs` returns no flag and the real binary still converts the import (metadata becomes `Function`). Proves pi-lens does not override user configs and that the conversion machinery works.
- **other recommended rules still fire under the bundled fallback config** ├ö├ç├Â derives the production argv from `biomeConfigArgs`, runs the real binary read-only, and asserts `noDebugger` still reports. Proves the override is one rule, not a gutted ruleset.
- **real Biome and TypeScript binaries are installed in CI** ÔÇö guard test following the #448 convention: the suite skips visibly when a devDependency binary is absent, but fails loud in CI so a missing binary can never silently vanish the real-binary proof.

Mutation proof (red-first, both vectors):

1. Override removed (commented out in `core.jsonc`): the first test fails at `expect(result.changed).toBe(false)` ├ö├ç├Â the real binary rewrote the import. Restored, green.
2. Fallback flag dropped (no `--config-path`): verified separately that Biome's default ruleset also converts the import, so a `biomeConfigArgs` mutation turns the same test red.

Sibling suites run green: biome-autofix-argv, biome-client-dedupe, biome-client-cwd, biome-install-evidence, biome-check, biome-check-runner, tool-policy-config-args, autofix-policy-consistency, formatter-exit-code-posture (10 files, 67 tests), runtime-agent-end (37 tests), plus the new suite.

Gates: `npm run build`, `npm run lint` (tsc), `npm run fmt:check`, `npm run astgrep:self-scan` (0 findings), `npm run changelog:check` (23 entries validated).

## Blast radius

`module_report` was unavailable in this worker session (no MCP tool surface); inspected manually. Consumers of `config/biome/core.jsonc`: the `biomeConfigArgs` seam (`clients/tool-policy.ts:2332`) feeding `BiomeClient.fixFileAsync` (autofix, immediate and deferred) and the `biome-check` lint runner; `scripts/benchmark-biome.ts` (perf measurement only); tests and docs. No other reader exists (repo-wide grep, excluding node_modules and dist). The override removes one advisory rule from unconfigured projects; both surfaces lose it together by design (#1247's one-seam contract). `docs/features.md` documents the fallback posture and the ruff analogue.

## Class sweep

Hazard class: a bundled safe autofix whose rewrite changes runtime semantics. Population: every package-owned fallback config that reaches an autofix spawn ├ö├ç├Â the three `configArgsWithFallback` call sites (`clients/tool-policy.ts`). Per-member verdicts:

- `config/biome/core.jsonc` ├ö├ç├Â member, fixed here.
- `config/ruff/core.toml` ├ö├ç├Â not a member, verified with real ruff 0.16.4: `check --fix` with the bundled config on a value import used only in an annotation leaves the file untouched ("All checks passed!"). Structurally: the flake8-type-checking (TC) family, whose fixes hoist imports into `TYPE_CHECKING`, is not selected; `ruff-client.ts` never passes `--unsafe-fixes`, so only safe fixes apply.
- `config/markdownlint/core.json` ├ö├ç├Â not a member by construction: markdown content formatting, no runtime semantics (config pinned by #833).

The eslint arm ships no bundled config (config-first contract, project-owned fixers), so it is outside the population. Sweep coverage: `rg "configArgsWithFallback\("` and `rg "resolvePackagePath\(import\.meta\.url, \"config/"` across `clients/` ├ö├ç├Â three hits, all triaged above.

## Observability

Existing records already prove the behavior: `latency.log` autofix phases show when the biome arm runs, and the report #2385 quotes them. No new telemetry is added because no new decision exists ├ö├ç├Â the config gains one rule override, and nothing at runtime observes or branches on it. The regression evidence lives in the test suite, not a log.

## Test assessment

- `tests/clients/biome-config-decorator-metadata.test.ts` (new): what each case pins is described above; skip is `describe.skipIf` on real-binary presence; the suite title names the dependency so a skipped report reads as a dependency skip, and a TOP-LEVEL guard test, placed outside the skipped describe so it still runs when the suite skips, fails loud in CI when a devDependency is missing (#448 convention; mutation-proven with CI=1 and an absent binary: the guard fails with the dependency named while the behavior suite reports skipped).
- No existing test edited. The config-args builder tests (`tool-policy-config-args.test.ts`) pin argv shapes only and are unaffected: the fix changes config content, not argv. `biome-check.test.ts:291`'s argv pin likewise unaffected.


apmantza:
Review round 1 dispositions (fix round pushed as f7f35ca2):

1. Changelog surface claim (`.changelog/fix-2385-biome-bundled-config.md` said the rule "stays active on the lint surface") � **fixed in f7f35ca2**. The fragment now states truthfully that the bundled fallback disables `style/useImportType` on both surfaces (lint and autofix share the one `biomeConfigArgs` seam), matching the maintainer decision recorded at issue #2385 (comment 5477218580).

2. Visible dependency skip for the real Biome + TypeScript suite � **fixed in f7f35ca2**. The suite title now names the dependency ("real @biomejs/biome + typescript devDeps"), so a skipped report reads as a dependency skip rather than a pass, and a guard test (`real Biome and TypeScript binaries are installed in CI`) fails loud in CI when a devDependency is missing, following the #448 convention from `ast-grep-catalog-rules.test.ts`. No production code changed.

Verification on f7f35ca2: new suite 4/4 green, 10 sibling Biome/autofix files (101 tests) green, `npm run build`, `npm run lint`, `fmt:check`, `astgrep:self-scan` (0 findings), `changelog:check` (23 entries) all clean. Rescope, user-config precedence, real-binary mutation proof, and the `refs` (not `closes`) claim re-checked in the updated PR body.

apmantza:
Review round 2 dispositions (fix round + master merge pushed, head e1551f29):

1. HIGH: the CI binary-presence guard sat inside `describe.skipIf`, so when a devDependency binary was absent the guard itself was skipped � vacuous exactly when it mattered. **Fixed in fac2c5d2.** The guard is now an unskipped top-level test (`real Biome and TypeScript binaries are installed in CI`) that fails loud in CI with a named dependency reason; the real-binary behavior cases stay inside the visibly titled conditional describe for local environments. Mutation proof: with the biome path mutated to a nonexistent location and `CI=1`, the top-level guard FAILS ("real @biomejs/biome and typescript devDeps are required...") while the behavior suite reports skipped; restored, 4/4 green.

2. Master merge: origin/master (5 commits through #2387) merged in e1551f29. The single AGENTS.md conflict (my #2385 bundled-config invariant vs master's #2384 template-formatter invariant at the same insertion point after the #833 markdownlint paragraph) was resolved by keeping BOTH invariants verbatim in their correct dispatch standing-invariants section; nothing dropped, nothing duplicated.

Verification on e1551f29 (post-merge head): focused + sibling Biome/autofix suites 11 files / 105 tests green (covers the master-adjacent tool-policy seams from #2384); all 14 `tests/config/` governance meta-sweeps (68 tests) green after the merge per the re-run-governance-after-master-merge rule; changelog/host-sdk/packaging/finding-delivery-gate suites green; `npm run build`, `npm run lint`, `fmt:check`, `astgrep:self-scan` (0 findings), `changelog:check` (29 entries incl. master's fragments) all clean. Required CI gates (Lint & type-check, Unit tests) re-run on this head; advisory install-matrix checks not waited on.

PR #2412, opened 2026-09-01
fix(read-guard): preserve partial apply outcomes (closes #2402)

### LINKED ISSUE #2402: Read-guard: edit applies to disk but reports oldText-not-found; identical retry re-applies and duplicates content
## Summary

A wrapped `edit` call can **apply the write to disk** while the tool result reports `oldText ... was not found in the current file content` (`RETRYABLE` / `RE-READ REQUIRED`). The agent retries the identical payload, the retry **re-executes the write**, and the file ends up with duplicated content (observed: 15 copies of one import line after repeated retry loops).

## Evidence

The read-guard's own recovery hint contradicts the headline error. From a session transcript:

```
edits[0].oldText ("import { FilterOption, GroupsIOMailingList } from '@lfx-one/") was not found
in the current file content. This is attempt #4 — this text does not appear anywhere in the file...

Did you mean one of these current lines?
        17 │ import { FilterOption, GroupsIOMailingList, MailingListTableRowVm } from '@lfx-one/shared/interfaces';
        18 │ import { getEntityCommands } from '@lfx-one/shared/utils';
```

Line 17 already contains the edit's `newText` (`MailingListTableRowVm` was the added identifier) — the edit had applied. The error text says "not found" because `oldText` was consumed by the first (successful) application, and `trackOldTextFailure` in `clients/read-guard-tool-lines.js` then escalated each identical resubmission as "attempt #2…#4" instead of recognizing the already-applied state.

## Root cause (as observed)

The edit seam appears to return the post-write pipeline/guard error **after the write has landed**, so the tool result says failure while the bytes are on disk. Because retries with the identical payload re-execute the write, the failure mode compounds: duplicates → real compiler errors (TS2300 duplicate identifier) → cleanup churn. In the observed session this retry loop burned tens of thousands of tokens.

## Expected behavior

1. **Applied-then-errored must be distinguishable from not-applied.** If the write landed, the result must not report oldText-not-found regardless of what the post-write analysis does.
2. **Identical-payload retries should be idempotent.** If every `newText` is already present verbatim in the target file (or the "Did you mean" recovery would quote the *new* text), the result should be a success note — "already applied, do not retry" — not an escalating error.
3. Bonus: if `newText` is present **more than once**, warn that a retry re-applied it and the file needs deduping — that's the state the loop leaves behind.

## Workaround (deployed locally)

Disabled the surface via `~/.pi-lens/config.json`:

```json
{ "readGuard": { "enabled": false }, "contextInjection": { "enabled": false } }
```

plus a local `tool_result` extension that rewrites oldText-mismatch errors to "already applied" when the `newText` is found on disk. Both are mitigations; the seam itself should be fixed here.

## Environment

- pi 0.84.4, pi-lens 4.1.3
- macOS (arm64), Node v22.19.0
- Repo context: Yarn 4 monorepo in a fresh git worktree (no `node_modules` at session start, so the LSP was also reporting stale module-resolution errors throughout — likely the trigger for the post-edit analysis erroring after the write)

Possibly related: #1659 (read-guard seam umbrella), #1053 (partial-apply seam).

apmantza:
## Summary

Closes #2402. Preserve the side-effect and result contract for mixed-validity edit batches.

Final head: 7fc00f540433946ae45fe9ec06c8ecc05a853633.

The read guard carries one raw-byte snapshot identity and an authoritative LF-view span for every accepted normalization tier. If a valid edit cannot be represented, the whole batch closes, including mixed native-range and oldText requests.

The apply seam validates the complete batch under the file lock before writing. It rejects stale, drifted, out-of-bounds, or overlapping input without disk, sequence, or pipeline effects. Atomic writes preserve the existing mode and follow a leaf symlink.

The retry store keeps fixed-size SHA-256 pair digests for exact submitted oldText/newText strings, plus the post-commit raw-byte file hash. Recognition requires this session/process provenance and unchanged content. Session reset clears the store.

The unsupported blanket 15-copy attribution remains excluded. This PR implements the confirmed committed-write/error mismatch and corrected retry acceptance.

## Tests

All bounded verification passed at final head: 621 tests across 22 files, with one Windows-inapplicable POSIX symlink test skipped.

- `tests/clients/partial-edit-apply.test.ts`: edited success, stale snapshot, span, overlap, line-ending, BOM, pipeline-failure, and write-failure cases; added exact trailing-whitespace pair identity, mode preservation, invalid-UTF8 raw-byte drift, and real atomic-writer coverage. The failure case injects only the low-level rename boundary.
- `tests/clients/read-guard-tool-lines.test.ts`: edited normalized-only expectations and added lone-CR, Unicode dash, Unicode quote, missing-edit, mixed-range closure, span, retry, contained-edit, and overlap coverage.
- `tests/clients/runtime-tool-call.test.ts`: added a real `handleToolCall` normalized mixed-validity batch proof alongside the existing mixed-validity, post-edit failure, exact retry, native retry, and bookkeeping cases. Only the pipeline process boundary is mocked.
- `tests/clients/runtime-tool-result.test.ts`: added a real failed-native-edit case proving no applied-pair record, change-log bookkeeping, or pipeline delivery; added a read-call budget proving one post-result hash is reused for three native edits.
- `tests/clients/file-time.test.ts`: queued a waiter while a rejecting holder remained suspended, then proved waiter progress after release.
- `tests/clients/read-guard-logger.test.ts`: wrote rejection, retry, and post-edit failure events to the real NDJSON sink at default verbosity.
- `tests/clients/runtime-coordinator.test.ts`: retained reset/session lifecycle coverage.
- `tests/clients/atomic-write-sweep.test.ts`, `tests/clients/delivery-surface-ratchet.test.ts`, `tests/clients/finding-delivery-gate.test.ts`, `tests/clients/session-state-conformance.test.ts`, `tests/clients/bounded-telemetry.test.ts`, `tests/clients/bounded-telemetry-sweep.test.ts`, `tests/config/dependency-boundaries.test.ts`, `tests/config/sweep-floor-coverage.test.ts`, `tests/config/vacuous-skip-coverage.test.ts`, `tests/scripts/check-pr-body.test.ts`, `tests/scripts/changelog.test.ts`, `tests/scripts/changelog-entries.test.ts`, and `tests/scripts/rollup-changelog.test.ts`: passed overlap and governance verification.

## Test assessment

 tests reach the real read-guard, RuntimeCoordinator, filesystem, atomic-writer, NDJSON sink, and FileTime seams. Mocks remain at process boundaries only. The reviewer-named guards were mutation-tested: failed-result routing, lock rejection swallowing, raw-byte snapshot validation, exact content evidence, overlap exclusion, normalized-span carrying, and native hash reuse each produced a red focused test when neutered.

Skipped full suite: the fork-heavy full suite was not run. Focused touched and adjacent suites plus every required governance lane passed.

## Blast radius

The change affects read-guard oldText preflight, mixed-range admission, partial apply, native edit result recording, RuntimeCoordinator retry lifecycle, FileTime locking, read-guard telemetry, and the atomic-write exemption sweep.

The dependent entry points are `handleToolCall`, `getTouchedLinesForGuard`, `applyPartiallyApplicableEdits`, `handleToolResult`, and `RuntimeCoordinator.resetForSession`. Normal native edit dispatch and post-edit pipeline behavior remain unchanged.

`module_report` is unavailable in this worker environment. Direct import and call-site tracing covered the affected modules. The correction delta is 559 insertions and 133 deletions against reviewed head 4a2b9d1e. The full branch diff versus origin/master is 2,398 insertions and 315 deletions; the recovered baseline before this correction was 1,947 insertions and 290 deletions.

## Class sweep

The sweep covered all PartiallyApplicableEdit producers and consumers, applied-pair record writes and reads, FileTime.withLock callers, read-guard batch summary constructors, atomic-write policy and exemption maps, delivery-surface registration, path normalization, session reset, and runtime finding delivery.

Partial apply and successful native edits are the only applied-pair producers. Records use normalized file keys, fixed-size pair and content digests, bounded per-file/store retention, and session reset. No unrelated sibling behavior was broadened.

## Observability

Rejection, applied-retry, post-edit failure, and batch-summary events use the bounded read-guard logger. Records retain normalized file identity, edit indexes, rejection kind, commit status, post-edit status, and correlation identity within existing caps.

The real-sink test confirms default-verbosity emission for the new outcomes. No raw per-occurrence failure logger was added. Pair digests and raw content hashes intentionally avoid storing submitted text while retaining discriminating evidence.

## Non-goals

- Do not attribute the unsupported 15-copy report without correlated session evidence.
- Do not change native host edit semantics or add rollback after a committed write.
- Do not add project-wide edit behavior or bypass the read guard.
- Do not merge this pull request.

Final body verification: this description matches the pushed head 7fc00f540433946ae45fe9ec06c8ecc05a853633.

---

## Review round 1 (head 45bbff22)

Three non-blocking findings from the merge-ready review, implemented on the same branch.

**Finding 1 (correctness) — retry recognition survives a post-commit formatter rewrite.**
`isExactAppliedRetry` previously recognized an already-applied edit only when the current file hash equaled the exact post-commit hash. When the afterWrite pipeline (a formatter) rewrote the file between commit and an identical retry, the hash gate hard-failed, recognition fell back to oldText resolution, and a surviving-oldText retry re-applied (the #2402 duplication class). Fix: `AppliedPartialEditRecord` now carries `afterWriteContentHash` alongside `contentHash`; `applyPartiallyApplicableEdits` stamps it after `afterWrite` when the pipeline actually changed the bytes, and the native-edit path stamps the post-pipeline state at the existing final-state hash. `isExactAppliedRetry` accepts EITHER recorded state. A file in any other (unrecognized third-party) state matches neither and stays fail-safe — it falls back to normal resolution rather than being silently accepted, so the coincidental-newText-presence guard remains live. Red-first proof (pre-fix source, committed tests): `isExactAppliedRetry > accepts the post-afterWrite formatter state` expected true, got false; `read-guard-tool-lines > recognizes an identical retry after a formatter rewrote the file post-commit` — `afterWriteContentHash` expected defined, got undefined.

**Finding 2 (standards) — cross-separator path-key coverage for the store.** Added two tests to the `PartialApplyRecordStore` suite: record with `/`, `find` with `\` (and reverse); and `noteAfterWriteHash` stamped under one separator, read back under the other. Red-first: `stamps an afterWrite hash under the other path separator` — `TypeError: store.noteAfterWriteHash is not a function` on pre-fix source.

**Finding 3 (nit) — collapsed the redundant post-result read.** The success path read+hashed the file twice per edit result (the applied-record hash and the pipeline dedup key `initialStateHash`); nothing between them rewrites the file. Hoisted one `postWriteStateHash`, reused for both. The native-multi-edit read-budget test drops from 3 raw-byte hashes to 2 (one post-write, shared by all records and the dedup key; one post-pipeline). Red-first: the read-budget test expected 2, got 3 on pre-fix source.

**Changelog.** `.changelog/2402-partial-apply-contract.md` updated to state recognition accepts either the post-commit or post-afterWrite state, and to document the fail-safe residual (an unrecognized third-party state falls back rather than being silently accepted).

**Verification.** `npm run build` and `npm run lint` clean. Touched suites green in isolation: `partial-edit-apply`, `read-guard-tool-lines`, `runtime-tool-result`, `runtime-tool-call`, `runtime-coordinator`, `file-time`, `read-guard-logger` (218 passed, 1 Windows-skipped). Governance sweeps green: `atomic-write-sweep`, `bounded-telemetry`(+sweep), `bus-producer-coverage`, `delivery-surface-ratchet`, `finding-delivery-gate`, `managed-tool-seam-coverage`, `profiling-coverage`, `session-state-conformance`, `dependency-boundaries`, `sweep-floor-coverage`, `vacuous-skip-coverage`. The full local suite showed only resource/timing/spawn-contention flakes under parallel fork load (cap/eviction/atomic/installer/mcp files taking 50–96s; "Worker exited unexpectedly"); each re-ran green single-file (e.g. `atomic-write` 22/22 in 3.3s, `fact-store-bound` 17/17). None touch the changed code paths. The full suite is delegated to CI as authoritative.


apmantza:
Adversarial review at `4a2b9d1e72403dc13bde6da99fbf6dc0b059e5b1` requested changes. This review is AI-generated, and a maintainer supervises the process.

Material findings:

- Mixed batches can count lone-CR, Unicode-normalized, or range edits as valid while omitting them from partial apply.
- Retry identity strips trailing whitespace, so distinct payloads can be suppressed as identical.
- Atomic replacement can drop executable mode bits and replace a leaf symlink.
- Snapshot identity hashes decoded text instead of raw bytes.
- Native multi-edit bookkeeping hashes the whole file once per edit.
- Applied-pair records are count-bounded but retain unbounded strings.
- The FileTime rejection guard and failed-native-result suppression have surviving mutants.
- Success tests mock the atomic writer, and new failure events lack real-sink coverage.

The same fixer is addressing the consolidated findings. The PR remains held despite green required CI. The same primary reviewer will verify the corrected exact head.

Issue #2573, opened 2026-09-03
Install lanes red: npm 10.9.8 crashes (edgesOut) resolving the published manifest's devDependencies peer graph after @vitejs/devtools 0.7.1 / vitest 5.0.0


apmantza:
Since 2026-09-03 11:42 UTC every `Install test` lane on master fails at the host-provided-deps supply step (`npm install --no-save` into the installed extension, the way pi does — #1926) with `npm error Cannot read properties of null (reading 'edgesOut')` in arborist `#loadPeerSet`. Last green: e62ebff2 at 10:52 UTC, same npm 10.9.8. Trigger: `@vitejs/devtools@0.7.1` (10:28 UTC) and `vitest@5.0.0` (12:24 UTC) published a peer graph npm 10 cannot load; the resolver reaches it because the PUBLISHED package.json still carries devDependencies. npm 11 unaffected. Proven locally with the last-green tarball: crash with the block, `added 4 packages in 3s` without it. User impact: an npm-10 user installing pi-lens today hits the same crash at pi's supply step. Fix: strip devDependencies at pack time (prepack/postpack) + a real-npm-pack test guarding the published manifest (PR #2572).

PR #2577, opened 2026-09-04
fix(rules): extend Python SQL safe-builder recognition without narrowing it (#2576)

### LINKED ISSUE #2576: fix: eliminate SQLAlchemy and psycopg typed-builder false positives
## Reproduction

A real FastAPI, SQLAlchemy, and PostgreSQL consumer reports false positives from three bundled Python tree-sitter rules.

1. python-hallucinated-import flags valid SQLAlchemy imports such as Mapped, mapped_column, relationship, Session, and select because sqlalchemy is treated as a suspicious-module whitelist while its modern API is not represented exhaustively.
2. python-cross-language-method flags valid sa.select calls because the blacklist treats select as exclusively Ruby. The emitted message also leaks literal METHOD and LANG placeholders.
3. python-sql-injection flags SQLAlchemy Session.query expression-builder calls because the receiver exemption currently applies only to execute, and flags psycopg typed identifier composition with sql.SQL.format and sql.Identifier even when the identifier is separately validated.

Consumer evidence:
- app/learning/service.py lines 36 and 61
- scripts/disposable_db.py lines 61 and 86
- app/assets/router.py lines 26 and 32
- app/learning/router.py lines 57 and 89
- five model/import sites using valid SQLAlchemy APIs

## Root cause

The rules use finite method and module name assumptions without sufficient typed-builder or context guards. Two messages use single-brace placeholders although the renderer and catalog contract requires doubled braces.

## Bounded fix

- Remove SQLAlchemy from the hallucinated-import suspicious whitelist and fix message placeholders.
- Remove ambiguous Python-valid select from the cross-language blacklist and fix its placeholders.
- Extend SQL-injection safe-builder recognition narrowly for SQLAlchemy session query and psycopg sql.SQL.format with sql.Identifier; retain true-positive raw and dynamic SQL coverage.
- Add real-runner positive and negative regression tests for all three rules.
- Regenerate the tracked tree-sitter catalog.
- Apply required patch-version and AGENTS issue-ledger updates.

## Acceptance

- Build, rule audits, focused runner tests, and formatting pass.
- Full-suite result introduces no failures versus clean main.
- Source-build dogfood on the consumer emits none of the listed false positives.
- No consumer repository or global install is modified by the fix branch.

AcademyKUB8:
## Summary

Fixes the false positives reported in #2576 in three bundled Python tree-sitter
rules, **without narrowing any suppression that ships today**.

The first round of this PR traded two shipped exemptions for a structural
provenance check, so canonical SQLAlchemy 2.x started firing a *blocking*
sql-injection finding that master keeps quiet. The maintainer review
(CHANGES_REQUESTED, 2026-09-04) is applied in full on top of the contributor's
three commits (authorship preserved):

**Blocker 1 — extend, not narrow.** Both pre-existing exemptions are restored
verbatim: the receiver-name exemption for unannotated `session.execute(...)`,
and the `select/insert/update/delete(...)` builder-call check on the argument.
The provenance summary is now layered *on top* of them: when the receiver is
proven `sqlalchemy.orm.Session` or `sqlalchemy.ext.asyncio.AsyncSession`
(parameter annotation, including FastAPI's `db: Session = Depends(get_db)`),
`query` is suppressed outright and `execute` is suppressed when its argument is
a builder call **or a name bound exactly once to one** — `stmt = select(User);
db.execute(stmt)`, the row master genuinely got wrong. `text(...)` counts only
with a literal template, so `db.execute(text("..." + uid))` stays diagnostic.

`scalar`/`scalars` are **no longer sinks**. They were added by the first round,
are not requested by #2576, and made `client.scalar(stmt)` a new blocking
finding on any object with a `.scalar(x)` method — the same class of noise as
blocker 1, one step removed. Removing them satisfies the review table's
"`db.scalars(select(User))` quiet" rows by construction; re-adding them later is
a one-line change plus a sink test.

| snippet (all with `from sqlalchemy.orm import Session` + `db: Session`) | master d39fc68e | reviewed head 80be3998 | this head |
|---|---|---|---|
| `db.execute(select(User)).all()` | quiet | **FIRES** | quiet |
| `db.scalars(select(User)).all()` | quiet | **FIRES** | quiet |
| `db.scalar(select(User))` | quiet | **FIRES** | quiet |
| `await db.execute(select(User))` on `AsyncSession` | quiet | **FIRES** | quiet |
| `stmt = select(User); db.execute(stmt)` | **FIRES** | **FIRES** | quiet |
| `db: Session = Depends(get_db)`, `db.execute(stmt)` | **FIRES** | **FIRES** | quiet |
| `db.execute(text("SELECT 1"))` | **FIRES** | **FIRES** | quiet |
| unannotated `session.execute(stmt)` | quiet | **FIRES** | quiet |
| `conn.execute(select(User))` (unprovable receiver) | quiet | **FIRES** | quiet |
| `cursor.execute("..." + uid)` | FIRES | FIRES | FIRES |
| `cursor.execute("..." % uid)` | FIRES | FIRES | FIRES |
| `db.execute(text("..." + uid))` | FIRES | FIRES | FIRES |
| `db.execute(text("...".format(uid)))` | FIRES | FIRES | FIRES |
| `db.execute(opaque_parameter)` | FIRES | FIRES | FIRES |
| `stmt = tainted; stmt = select(User); db.execute(stmt)` | FIRES | FIRES | FIRES |

**Blocker 2 — version bump reverted.** `package.json` and `package-lock.json`
are byte-identical to master (`git diff origin/master -- package.json
package-lock.json` is empty); package.json stays tab-indented (#2573).
Releases are maintainer-cut.

**Structural ask — rule data back in YAML.** `python-hallucinated-import.yml`
keeps its `match_captures` post-filter and simply drops `sqlalchemy` from the
MODULE regex, which is what #2576 itself proposed. The
`LEGACY_PYTHON_HALLUCINATED_IMPORT_MODULES/NAMES` constants, the
`isKnownPythonHallucinatedImport` helper and the `py_hallucinated_import`
post-filter branch are deleted from `clients/tree-sitter-client.ts` (-80 lines
of client code, net -1 post-filter name).

*Decision on the `sqlalchemy` → `JSONResponse` pair: dropped, no second rule.*
`match_captures` is a cross-product of two independent regexes, so admitting
`sqlalchemy` as a module admits all 35 names — `Session`, `Column`, `Integer`,
`String`, `Model`, `delete`, `get`, `post` are all real SQLAlchemy APIs, and
that cross-product **is** the reported false positive. A dedicated second rule
(new YAML + catalog row + audit entry + tier) to catch one hallucination that
`ImportError`s on the first run does not earn its rule budget. `from requests
import JSONResponse` and `from requests import Session` still fire (pinned in
`tree-sitter-879-post-filters`).

**`clients/python-provenance.ts`: 797 → 781 lines** while *adding* the
Session-statement route (~90 lines). Trimmed: three binding-name classifiers
plus the `BindingNameDecision` dispatch engine folded into one recursive
`scanBindingNames`; six import-parsing functions and the
`ParsedImportStatement` type folded into one recorder plus
`parseImportBinding`; the `tainted` set deleted (it recomputed a check
`provenanceFor` can make directly from the binding counts); the PEP 695
`type_parameter`/`type_parameter_list` recorders deleted as the only entries in
`SUMMARY_NODE_RECORDERS` no test reached. Every remaining recorder is reached by
a runner test — the "keeps every statically discoverable function-local Python
binder diagnostic" table in `tree-sitter-security-gap-rules` covers
destructuring, annotated, augmented, walrus, import, for, async-for, with,
except, except-star, `del`, function, class, type-alias, match-capture,
match-as, lambda, and the `global` case is covered by "keeps production
provenance fail-closed across namespace and binder hazards".

One acceptance criterion of #2576 I could not verify myself: the source-build
dogfood against the consumer repository. Its eleven reported sites are encoded
as production-runner fixtures instead ("suppresses the four production consumer
shapes but not unsafe lookalikes").

Fixes #2576

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [x] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [x] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `.changelog/<branch-or-slug>-<short-desc>.md` has one valid entry **in this PR** for any user-facing change

## Tests

Run after `npm run build`, in the foreground.

**NEW — `tests/clients/dispatch/runners/tree-sitter-python-rule-regressions.test.ts`**
(production `treeSitterRunner.run` path, real grammar, real YAML):

- `keeps canonical SQLAlchemy 2.x statement execution quiet (#2577 review)` —
  nine fixtures, one per review-table row plus the two exemptions each guard
  owns. `expect.soft` so one run reports the whole table. Regression proof.
- `keeps composed SQL strings diagnostic in SQLAlchemy sinks (#2577 review)` —
  six fixtures: `+` concatenation, `%` formatting, `text("..." + uid)`,
  `text("...".format(uid))`, an opaque parameter, and a twice-bound `stmt`.
  Over-suppression guard for the new branch.

Screens (AGENTS.md test-authoring): not a *parallel path* (drives the real
runner, not the predicate), no *invisible skip* (no `skipIf`; grammar presence
asserted in `beforeAll`), not a *wrong-layer pin* (asserts fired rule ids, not
predicate return values), no *ambient-inspection double* (real fixtures on
disk), no *env leakage* (temp cwd from `makeRealRunnerEnv`), no *loose bound*
(exact rule id membership), not *all-mocks* (only `recordEntitySnapshotDiff` is
mocked, as this helper requires), not *not-throw*, not an *implementation
mirror* (fixtures are Python source, not our data structures), no snapshots.

### RED on the reviewed head `80be3998` (contributor sources, these tests)

```
=== RED on the reviewed PR head 80be3998 (contributor sources, current tests) ===
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: db.execute(select(User)): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db.scalars(select(User)): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db.scalar(select(User)): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: await db.execute(select(User)) on AsyncSession: expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: stmt = select(User); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db: Session = Depends(get_db); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: unannotated session.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: conn.execute(select(User)) on an unprovable receiver: expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db.execute(text("SELECT 1")): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)
```

### Same tests on master `d39fc68e` (which rows are regressions vs. real bugs)

```
=== SAME TESTS on master d39fc68e (pre-PR baseline; provenance module does not exist there) ===
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: stmt = select(User); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db: Session = Depends(get_db); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db.execute(text("SELECT 1")): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)
```

Six of the nine rows were quiet on master and red on the reviewed head — those
are the regression. Three were red on master and are what this PR actually
fixes.

### Mutation proofs (neutered in the built `clients/*.js`, rebuilt after each)

```
### M1  delete the restored receiver-name exemption (tree-sitter-client.js:3582-3585)
AssertionError: unannotated session.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)

### M2  delete the restored builder-expression-call exemption (tree-sitter-client.js:3586-3588)
AssertionError: conn.execute(select(User)) on an unprovable receiver: expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)

### M3  force the proven-Session branch off (tree-sitter-client.js:3593)
AssertionError: stmt = select(User); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db: Session = Depends(get_db); db.execute(stmt): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: db.execute(text("SELECT 1")): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)

### M4  accept any text(...) call, not only a literal template (python-provenance.js:539)
AssertionError: db.execute(text("..." + uid)): expected [] to include 'python-sql-injection'
AssertionError: db.execute(text("...".format(uid))): expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)

### M5  drop the single-binding requirement on a bound statement name (python-provenance.js:417)
AssertionError: stmt bound twice — tainted, then select(): expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 8 skipped (10)

### M6  put sqlalchemy back into the hallucinated-import MODULE regex (rule YAML)
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: expected 1 to be +0 // Object.is equality
AssertionError: expected [ { …(6) }, { …(6) } ] to have a length of +0 but got 2
      Tests  2 failed | 62 passed (64)
```

M1/M2 are the *restored* exemptions: the first round removed exactly these two
lines, which is why the review table went red. Each now owns a fixture that
cannot be satisfied by any other branch (M2's first attempt was green because
the proven-Session branch masked it; `conn.execute(select(User))` on an
unprovable receiver was added to make it discriminate).

### Suite runs

```
Test Files  60 passed (60)
     Tests  904 passed (904)
```

60 files = the mechanical governance selection
`ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts`
plus every `tests/config/*.test.ts` (54, including `tests/config/tracked-control-bytes.test.ts`
from #2584) plus the six touched
files and `tree-sitter-python-rules`. Also green: `npm run lint`,
`npx oxfmt --check` (0.65.0, the pinned devDependency, over `clients/`,
`tests/clients/`, `rules/` — 1276 files), `node
scripts/check-changelog-fragments.mjs` (114 entries), `node
scripts/audit-tree-sitter-rules.mjs` (170 files, 0 missing defect_class /
inline_tier), `node scripts/validate-rule-catalog.mjs` (exit 0), `node
scripts/gen-rule-catalogs.mjs` (no further diff — the tracked catalog is
current), `npm run depcruise:check`, `node scripts/check-lockfile-sync.mjs`,
`npm run build:dist`. The full suite is delegated to CI.

### Test assessment

Deletion accounting: **every test this PR deletes was also added by this PR.**
All four (`keeps arbitrary object scalar and scalars calls diagnostic`,
`flags JSONResponse imported from SQLAlchemy`, `uses double-brace captures in
its production message`, `keeps only METHOD in the rendered rule message`)
appear in the contributor's own diff and in **no** `origin/master` file
(`git show origin/master:<file> | grep -c` → 0 for each). Net test count vs
master is +2 restored, +0 removed.

- `tests/clients/dispatch/runners/tree-sitter-python-rule-regressions.test.ts` —
  uniquely pins the three Python rules through the production runner (rendered
  message text included). Edited three pre-existing cases that could not
  distinguish a guard from its absence: the FastAPI row used `scalars`, which is
  no longer a sink (now `execute`, and M3 reds it); the rebound-statement row
  read the last binding either way (now `stmt = tainted; stmt = select(User)`,
  and M5 reds it); `boundedAliases` was built but asserted by no index (now
  asserted). Deleted the trailing `from sqlalchemy import JSONResponse`
  assertion inside "keeps production provenance fail-closed…" — it pinned rule
  data this PR removes; survivor for the true-positive half is
  "runs the production YAML rules and renders their capture messages"
  (`from requests import JSONResponse`), which reds if the MODULE regex loses
  `requests`. Removed the two `client.scalar/scalars` fixtures (sink set).
- `tests/clients/tree-sitter-security-gap-rules.test.ts` — uniquely pins the
  post-filters at the query layer (`runQueryOnFile`) including the binder table.
  **Round 2 restores two master tests the branch had replaced in place**, byte
  identical to `origin/master`: `does not match SQLAlchemy session.execute(stmt)`
  and `does not match SQLAlchemy expression-builder execute calls`. They are the
  query-layer pins for exemptions A and B; their runner-layer survivors are the
  M1 fixture (`unannotated session.execute(stmt)`) and the M2 fixture
  (`conn.execute(select(User))` on an unprovable receiver), but restoring them
  costs nothing and keeps the layer's own coverage whole.
  **Deleted** `keeps arbitrary object scalar and scalars calls diagnostic`: it
  asserted `client.scalar(statement)` produces 2 matches, which is exactly the
  new false-positive surface this round removes. No survivor needed — the
  behavior it pinned is gone; the runner rows `db.scalar/scalars(select(User))`
  now pin the opposite. `keeps raw execute calls even when the receiver is
  Session-annotated` is **kept unchanged**: its argument is an opaque parameter,
  not a builder or a name bound to one, so it fires on master and on this head
  alike — it never pinned the regression.
- `tests/clients/tree-sitter-slop-rules.test.ts` — uniquely pins the slop rules
  at the query layer. **Deleted** `flags JSONResponse imported from SQLAlchemy`
  (pins removed rule data; survivor `tree-sitter-879-post-filters` >
  `keeps hallucinated-import pairs exact` now asserts that pair is 0 and reds
  under M6). **Deleted** two vacuous cases — `uses double-brace captures in its
  production message` and `keeps only METHOD in the rendered rule message` —
  which asserted a constant read straight out of the YAML (defect shape 33,
  source-text assertion in place of a runtime probe). Survivor:
  `tree-sitter-python-rule-regressions` > `runs the production YAML rules and
  renders their capture messages`, which asserts both *rendered* messages
  through `treeSitterRunner` and therefore reds on the same mutation (reverting
  either message to single braces) and on more.
- `tests/clients/tree-sitter-879-post-filters.test.ts` — uniquely pins that every
  shipped `post_filter` name has an implementation, and pair exactness. Updated
  the sqlalchemy row to 0; reds under M6.
- `tests/clients/dispatch/runners/tree-sitter-runner.test.ts` — uniquely pins the
  runner's `{{CAPTURE}}` interpolation. Unchanged this round.
- `tests/clients/tree-sitter-python-rules.test.ts` — not touched; run as a
  neighbour of the changed rules (green).

## Blast radius

`module_report` with `blastRadius: true` was not available in this environment;
dependents enumerated by grep over the whole tree instead (recorded explicitly
per template).

- `clients/python-provenance.ts` — importers: `clients/tree-sitter-client.ts`
  only (`grep -rln python-provenance clients/ mcp/ tools/ scripts/ tests/
  index.ts`). New module in this PR; no other entry point.
- `clients/tree-sitter-client.ts` — importers:
  `clients/dispatch/runners/tree-sitter.ts`, `clients/read-expansion.ts`,
  `clients/tree-sitter-logger.ts`, `clients/tree-sitter-shared.ts`,
  `clients/tree-sitter-symbol-extractor.ts`, plus
  `scripts/check-grammar-load.mjs`, `scripts/scan-tree-sitter-rules.mjs`,
  `scripts/run-ts-rules-pi-lens.mjs`, `scripts/run-all-ts-rules-posthog.mjs`.
  The only changed surface is the `py_sql_injection_sink` post-filter branch and
  the removal of `py_hallucinated_import`; the batched and single-query walkers
  are untouched apart from the contributor's `tree.rootNode` hoist.
- Hot path: the post-filter runs per query match on Python files. The added work
  is bounded and cached — `getPythonProvenanceSummary` memoizes one summary per
  root node in a `WeakMap` and is only reached when the sink-method check and
  both cheap name checks have already failed; the summary walk carries the
  pre-existing 50k-visit / 128-depth caps and fails closed (`invalid`) on any
  cap hit or parse error. Proven by test: the parse-error bail (MU-F4a) and
  `carriesComposedSql`'s `EXPRESSION_DEPTH_CAP` fail-closed branch (round 3
  pins it with a generated nine-deep wrapper — MU-N3). NOT proven by test: the
  50k-visit / 128-depth summary traversal caps, which would need a generated
  multi-megabyte fixture; the claim there is code-reading, not evidence. The new `isSqlAlchemyStatementArgument` is a
  string-set lookup plus at most one map lookup. No measured regression in the
  runner suites (60 files, 4.17s wall).
- Rule-data changes affect only Python `python-hallucinated-import` (strictly
  fewer findings), `python-cross-language-method` (strictly fewer: `select`
  removed from the blacklist, message placeholders fixed) and
  `python-sql-injection`.

## Observability

No new record. Post-filter suppressions are not logged anywhere today — the
tree-sitter runner logs the diagnostics it *emits* (`~/.pi-lens/extension.log`
via `clients/tree-sitter-logger.ts`, and phase timings in
`$PI_LENS_HOME/latency.log`), so a suppressed match leaves no trace by design;
production evidence for this change is the *absence* of `python-sql-injection`
and `python-hallucinated-import` findings on the consumer's SQLAlchemy sites in
the extension log. Naming the gap rather than adding a per-match log line: a
suppression counter on a per-match path is exactly the hot-path cost this rule
cannot afford, and no acceptance criterion asks for one.

## Class sweep

**Class:** a rule suppression narrowed while being extended — a shipped
exemption removed in the same change that adds a stricter one, so users who
depended on the old exemption regress (this is the shape the review caught, and
the sibling of AGENTS.md shape 17 "guard tightened past the reported defect").
Plus the secondary class: **rule data living in client code instead of the rule
YAML** (single-source-of-truth rule).

Sweep for suppression/exemption predicates that key off a *name* in the
tree-sitter post-filter dispatch, across the whole tree
(`clients/`, `tools/`, `mcp/`, `scripts/`, `scripts/lib/`, `tests/support/`,
`index.ts`) — `grep -rn "isLikely\|Receiver\|_SAFE_\|EXEMPT"`:

- `isLikelySqlAlchemyReceiver` (python) — restored; per-member verdict: keep,
  it is the only thing standing between unannotated `session.execute(...)` and
  a blocking finding, and M1 proves a test owns it.
- `isSafeSqlAlchemyExpressionCall` (python) — restored; verdict: keep, M2 owns
  it.
- `ts_ssrf_sink` / `unsafe_regex_dynamic_identifier` (typescript) — same shape
  (name-based safe-value recognition), untouched here and already covered by the
  `#963`/`#1000`/`#1008` cases in `tree-sitter-security-gap-rules`. No
  consolidation: they key off different languages' grammars and share no data.

Sweep for hardcoded rule data in post-filters — `grep -rh "^post_filter:"
rules/ | sort | uniq -c` gives 3 `match_captures` rules
(`python-hallucinated-import`, `python-cross-language-method`,
`ts-hallucinated-react-import`); after this PR all three carry their regexes in
`post_filter_params` in the YAML, and
`grep -rn "LEGACY_PYTHON_HALLUCINATED\|py_hallucinated_import"` over
`*.ts *.mjs *.yml *.md *.json` (node_modules excluded) returns nothing.
**Class of size 1, now zero** — the only code-resident copy of rule data was the
one this PR deletes. Consolidation verdict: the family stays on the existing
`match_captures` seam; no new seam needed.

Sweep for the values this PR introduces: `PYTHON_SQLALCHEMY_RECEIVER_NAMES`,
`PYTHON_SQLALCHEMY_STATEMENT_BUILDERS`, `isSqlAlchemyStatementArgument`,
`"sqlalchemy.ext.asyncio:AsyncSession"` — each appears only in
`clients/python-provenance.ts` and (for the two exported sets)
`clients/tree-sitter-client.ts`; no duplicate list exists anywhere in the tree.
An unreachable third set (`PYTHON_SQLALCHEMY_STATEMENT_METHODS`, gated by a sink
set it could never intersect beyond `execute`) was found by this sweep during
review of my own diff and deleted in `400c7714`.

## Merge order

Rebuilt on `origin/master` `d39fc68e` (includes #2584's AGENTS.md shape 37 and
`tests/config/tracked-control-bytes.test.ts`, which passes). The AGENTS.md hunk
is in the tree-sitter invariant section and does not touch the recurring-defect
catalog, so it composes with any pending catalog edit. No other open PR touches
`clients/tree-sitter-client.ts` or the Python rule YAMLs.

## Review round 2

One root cause behind both high findings: **a suppression that never looked at
its argument.** Every probe below was reproduced through the production
`treeSitterRunner.run` on `origin/master` and on the reviewed head `2f8c263b`
*before* any code changed.

### Per-probe results (production runner, real grammar, real YAML)

| probe | master `d39fc68e` | reviewed head `2f8c263b` | this head |
|---|---|---|---|
| c01 `db.query("SELECT ... id=" + uid)` | FIRES | **quiet** | FIRES |
| c02 `db.query(text("... id=" + uid))` | FIRES | **quiet** | FIRES |
| c03 `db.query("... id=%s" % uid)` | FIRES | **quiet** | FIRES |
| c04 `db.query(User)` | FIRES | quiet | quiet |
| c05 `db.query(User).filter(User.id == uid)` | FIRES | quiet | quiet |
| c06 `db.query(q)` (opaque parameter) | FIRES | **quiet** | FIRES |
| w1 `db.execute(repo.update("..." + uid))` | FIRES | **quiet** | FIRES |
| w2 `db.execute(repo.delete("..." + uid))` | FIRES | **quiet** | FIRES |
| w3 `stmt = repo.update("..." + uid); db.execute(stmt)` | FIRES | **quiet** | FIRES |
| w4 `db.execute(sa.select(User))` | FIRES | quiet | quiet |
| w5 `conn.execute(repo.update("..." + uid))` (unproven receiver) | FIRES | FIRES | FIRES |
| w6 `db.execute(sa.select("..." + uid))` | FIRES | **quiet** | FIRES |
| w7 `db.execute(repo.update(value))` (unproven module) | FIRES | **quiet** | FIRES |
| `db.execute(sa.text("..." + uid))` | FIRES | FIRES | FIRES |

c04/c05/w4 are the three rows this PR is *for*: they fire on master and are
quiet here. Every other row now matches master.

### F1 (high) — `query` suppressed regardless of argument

`clients/tree-sitter-client.ts` did `if (fn === "query") return false;` under a
proven Session receiver. Fixed by `isSqlAlchemyEntityQueryArgument`
(`clients/python-provenance.ts`): `Session.query` still does **not** require a
builder argument — `db.query(User)` is #2576's own case — but it suppresses only
what cannot be SQL text. A `string`, `concatenated_string`, `binary_operator`
(`+` and `%`), an f-string, a `"...".format(...)` chain, or a `text(...)` whose
template is not a static literal all stay diagnostic; so does **any** name the
enclosing function binds — parameter or local, whatever it was assigned (round 3
N-2) — while a free or module-level entity name stays quiet.

### F2 (high) — any object's `.update(...)` read as a statement builder

`isStatementBuilderCall` took the last attribute segment and ignored the
argument. Now an attribute callee names a builder only when its object resolves
through the module's own `expressionProvenance` to a proven `sqlalchemy` import
(`import sqlalchemy` / `import sqlalchemy as sa`, via a new `sqlalchemy-module`
provenance in `PLAIN_PACKAGE_PROVENANCE`), and **no** builder call may wrap
composed SQL. A bare-name builder (`select(...)`, `update(...)`) is still
name-only — that is master's shipped `isSafeSqlAlchemyExpressionCall`, which
short-circuits before this branch; matching master there is deliberate, and
`db.execute(update("..." + uid))` is quiet on master and quiet here (unchanged
pre-existing hole, not a new one).

### F3 (medium) — deletion accounting

Both master tests the branch replaced in place are **restored byte-identical**
(diff against `origin/master` is empty), and the Test assessment section now
carries the full accounting: all four deleted cases were added by this PR's own
commits and exist in no master file.

### F4 (low) — mutation-vacuous guards

Pinned the parse-error bail with a `parse-error.py` fixture (a syntax error next
to a proven-Session sink must not suppress) and the `nonlocal_statement`
recorder with a `nonlocal.py` fixture; both now red under mutation (MU-F4a,
MU-F4b). The traversal caps stay unpinned and the Blast radius section says so
instead of claiming them.

### Red-first, round 2

```
=== ROUND-2 PROBES on the reviewed head 2f8c263b ===
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: c01 db.query('...' + uid): expected [] to include 'python-sql-injection'
AssertionError: c02 db.query(text('...' + uid)): expected [] to include 'python-sql-injection'
AssertionError: c03 db.query('...' % uid): expected [] to include 'python-sql-injection'
AssertionError: c06 db.query(opaque_parameter): expected [] to include 'python-sql-injection'
AssertionError: w1 db.execute(repo.update('...' + uid)): expected [] to include 'python-sql-injection'
AssertionError: w2 db.execute(repo.delete('...' + uid)): expected [] to include 'python-sql-injection'
AssertionError: w3 stmt = repo.update('...' + uid); db.execute(stmt): expected [] to include 'python-sql-injection'
AssertionError: w7 db.execute(repo.update(value)) — unproven module: expected [] to include 'python-sql-injection'
AssertionError: w6 db.execute(sa.select('...' + uid)): expected [] to include 'python-sql-injection'
      Tests  2 failed | 10 skipped (12)

=== SAME PROBES on origin/master d39fc68e (no provenance module there) ===
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 2 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: c04 db.query(User): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: c05 db.query(User).filter(...): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: w4 db.execute(sa.select(User)): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  2 failed | 10 skipped (12)
```

### Mutation proofs, round 2

```
### MU-F1  force the Session.query argument check true (tree-sitter-client.js:3596)
AssertionError: c01 db.query('...' + uid): expected [] to include 'python-sql-injection'
AssertionError: c02 db.query(text('...' + uid)): expected [] to include 'python-sql-injection'
AssertionError: c03 db.query('...' % uid): expected [] to include 'python-sql-injection'
AssertionError: c06 db.query(opaque_parameter): expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)

### MU16  require a statement builder for query instead (reds #2576's own db.query(Model) cases)
AssertionError: expected [ { …(6) } ] to have a length of +0 but got 1
AssertionError: expected [ { …(6) } ] to have a length of +0 but got 1
AssertionError: expected [ { …(6) }, { …(6) } ] to have a length of +0 but got 2
AssertionError: expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: c04 db.query(User): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
AssertionError: c05 db.query(User).filter(...): expected [ 'python-sql-injection' ] to not include 'python-sql-injection'
      Tests  6 failed | 70 passed (76)

### MU-F2a (with w7)  let any attribute callee name a builder (python-provenance.js:575)
AssertionError: w7 db.execute(repo.update(value)) — unproven module: expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)

### MU-F2b  let a builder wrap composed SQL (python-provenance.js:599)
AssertionError: w6 db.execute(sa.select('...' + uid)): expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)

### MU-F4a  delete the parse-error bail (python-provenance.js:376)
AssertionError: expected [] to include 'python-sql-injection'
      Tests  1 failed | 11 skipped (12)

### MU-F4b  delete the nonlocal_statement recorder (python-provenance.js:349)
AssertionError: expected [] to include 'python-sql-injection'
      Tests  1 failed | 11 skipped (12)
```

MU16 is the reviewer's own counter-probe: requiring a statement builder for
`query` reds seven cases including #2576's `db.query(Model)` shape, which is why
`query` keeps a different predicate from `execute`.

### New tests, round 2

- `keeps composed and opaque Session.query arguments diagnostic (#2577 round 2)`
  — c01-c06 in one `expect.soft` sweep.
- `keeps builder-shaped laundering diagnostic (#2577 round 2)` — w1-w7 plus
  `sa.text("..." + uid)`.
- `keeps production provenance fail-closed across namespace and binder hazards`
  gains `nonlocal.py` and `parse-error.py`.
- `tests/clients/tree-sitter-security-gap-rules.test.ts` regains master's two
  exemption tests.

Rebased on `origin/master` `ada10ece` (#2589). Suites after the merge: 60 files
/ 904 tests green, plus lint, `oxfmt --check` (1276 files), changelog fragments
(115), rule audit, catalog validate, `gen-rule-catalogs` (no diff),
`depcruise:check` (929 modules, 0 new violations).

## Review round 3

Three prescribed items, applied exactly; nothing else changed. Rebased on
`origin/master` `9be58ca1` (#2583, shape 39) — the tree-sitter section is
untouched by that merge and the invariant paragraph was re-read after it.

### N-1 — the AGENTS.md invariant described round-1 behaviour

`AGENTS.md:1067` still read "`query` is suppressed outright", the exact defect
round 2 fixed, and named neither F2 clause. Rewritten in place, same format: it
now says both suppressions read the ARGUMENT and never the receiver alone; that
`query` does not require a builder but stays diagnostic for strings,
concatenation, `%`/`.format`, f-strings, non-literal `text(...)` and any
function-bound name; that `execute` needs a builder call or a name bound once to
one; that an attribute callee is a builder only when its object resolves to a
proven `sqlalchemy` import; and that no builder call may wrap composed SQL.

### N-2 — the unreachable bound-value branch, deleted

`isSqlAlchemyEntityQueryArgument` followed a name to its single assignment. The
branch was reachable by no test and cost one true positive. Deleted, so "a name
the enclosing function binds stays diagnostic" is what the code does; the
now-vacuous recursion (`isEntity` closure and its `depth` parameter, which could
no longer recurse) went with it.

Accepted consequence, per the maintainer's decision: `n1` fires. A model built
inside the function is statically indistinguishable from the laundering case,
and following the assignment loses `x2`. The fixture carries that in its name.

```
=== N-2 RED on head a070e0f5 (bound-value branch still present) ===
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
AssertionError: x2 q = build_sql(uid); db.query(q): expected [] to include 'python-sql-injection'
AssertionError: n1 User = get_model(); db.query(User) — accepted noise: expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)
```

| probe | master | head `a070e0f5` | this head |
|---|---|---|---|
| x2 `q = build_sql(uid); db.query(q)` | FIRES | **quiet** | FIRES |
| n1 `User = get_model(); db.query(User)` | FIRES | quiet | FIRES *(accepted noise)* |

c01-c06, w1-w7, the `sa.text` row and c04/c05/w4 are unchanged by round 3 — all
twelve still hold (`Tests 12 passed`).

### N-3 — pinned rather than disclaimed

I pinned it. `carriesComposedSql`'s `EXPRESSION_DEPTH_CAP` fail-closed branch now
has fixture `d1`: nine generated wrappers put a concatenation one level past the
cap inside `sa.select(...)`. The Blast-radius sentence is narrowed to the summary
traversal caps, which stay unclaimed.

### Mutation proofs, round 3

```
### MU-N2  treat every query argument name as a module-level entity (python-provenance.js:638)
AssertionError: x2 q = build_sql(uid); db.query(q): expected [] to include 'python-sql-injection'
AssertionError: n1 User = get_model(); db.query(User) — accepted noise: expected [] to include 'python-sql-injection'
AssertionError: c06 db.query(opaque_parameter): expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)

### MU-N3  make the expression depth cap fail OPEN (python-provenance.js:548-549)
    if (depth > EXPRESSION_DEPTH_CAP)
        return false;
AssertionError: d1 composed SQL nested past EXPRESSION_DEPTH_CAP: expected [] to include 'python-sql-injection'
      Tests  1 failed | 1 passed | 10 skipped (12)
```

Suites after round 3: 60 files / 904 tests green (the same mechanical governance
selection plus the six touched files), 186 tests across the six touched suites
alone, `npm run lint`, `oxfmt --check` (1276 files), changelog fragments (116),
rule audit, catalog validate, `gen-rule-catalogs` (no diff).

---

Maintainer takeover of @AcademyKUB8's PR: the contributor's three commits are
preserved with their authorship; the review is applied in five commits on top.


VladimirGVP:
Final follow-up is ready on `80be3998775eab92b25bc50182e68698dfdaa5f8`.

- SonarCloud: quality gate **OK**
- unresolved new-code issues: **0**
- new-code duplication: **0.0%**
- focused provenance suites: **149/149**
- build, lint, and formatting: passing
- independent staged review and final techlead gate: accepted

The PR is mergeable at the Git object level, but branch policy requires a maintainer action and the fork author cannot enable auto-merge. @apmantza, please review/merge when convenient.

apmantza (changes requested):
Thanks for the thorough writeup and the real-runner tests — the placeholder fix, the `select` removal from the cross-language blacklist, and the psycopg `sql.SQL(...).format(sql.Identifier(...))` recognition are all right. Two blockers and one structural ask before this can merge.

### Blocker 1 — this regresses every SQLAlchemy 2.x codebase

I built this branch and master locally and ran the production runner over the canonical 2.x idioms:

| snippet (all with `from sqlalchemy.orm import Session` + `db: Session`) | master | this PR |
|---|---|---|
| `db.execute(select(User)).all()` | quiet | **FIRES (blocking)** |
| `db.scalars(select(User)).all()` | quiet | **FIRES** |
| `db.scalar(select(User))` | quiet | **FIRES** |
| `await db.execute(select(User))` on `AsyncSession` | quiet | **FIRES** |
| `stmt = select(User); db.execute(stmt)` | FIRES | FIRES |
| `cursor.execute('...' + uid)` | FIRES | FIRES |

The PR removes the two existing exemptions (`isLikelySqlAlchemyReceiver` for `execute`, and the `select/insert/update/delete(...)` expression-call check), adds `scalar`/`scalars` as new sinks, and then only re-admits `Session.query`. `tests/clients/tree-sitter-security-gap-rules.test.ts` ("keeps raw execute calls even when the receiver is Session-annotated") pins this regression as intended behaviour. `session.execute(select(...))` is *the* SQLAlchemy 2.x API — a blocking sql-injection finding on it is a far larger false-positive surface than the four Neuroshift sites this fixes, and it re-introduces exactly the noise the removed exemption was added for.

Issue #2576's own scope says "extend safe-builder recognition narrowly … retain true-positive coverage". Extend, not narrow. The provenance summary is a reasonable way to get there: when the receiver is *proven* `Session`/`AsyncSession`, suppress `execute`/`scalar`/`scalars`/`query` whose argument is a `select/insert/update/delete/text(...)` call or a name bound to one — and keep the existing receiver-name exemption for unannotated `session.execute(...)` so current users don't regress. The last row of the table above (`stmt = select(User); db.execute(stmt)`) is the case that is actually worth fixing and this PR leaves it firing.

### Blocker 2 — drop the version bump

`package.json` 4.1.3 → 4.1.4 must not be in a contributor PR. Merging it fires `release.yml` on master, which would then fail because the PR's own `.changelog/` entry is still pending (releases run `changelog:release` first). Releases are cut by the maintainer; please revert `package.json` and `package-lock.json`.

### Structural ask — keep rule data in the YAML

`python-hallucinated-import.yml` used a `match_captures` post-filter with the module/name regexes as rule data. The PR moves those lists verbatim into `LEGACY_PYTHON_HALLUCINATED_IMPORT_MODULES/NAMES` constants in `tree-sitter-client.ts` plus a hardcoded `sqlalchemy → JSONResponse` special case. The fix the issue itself proposes is simply removing `sqlalchemy` from the MODULE regex (its pairs were the false positives); that keeps the catalog data-driven and deletes ~80 lines of new client code. If a `sqlalchemy`/`JSONResponse` pair is genuinely worth keeping, it belongs as a second `match_captures` rule, not a code branch.

Relatedly: `clients/python-provenance.ts` is 797 lines for a bounded fix. It's fine to keep the binding/shadow tracking if the Session-provenance route is what we take for blocker 1, but please make sure every branch is reached by a test (the aliased-chain/`del`/`match`-pattern/type-alias handlers in particular) or trim to what the two predicates actually need.

### Smaller

- The AGENTS.md invariant paragraph should be rewritten to describe the final behaviour once blocker 1 is resolved.
- "39 failures match clean main" — master is fully green on CI at 8a18586e, so those are local-environment failures; worth checking your `npm run build` ran before the suite (stale compiled `.js` is the usual cause).
- Fork workflows need maintainer approval before CI runs here; I'll flag that to the maintainer once the above is addressed.

Local verification on this end: build + lint pass, the 6 touched test files pass (185 tests), no flake-shape ratchet impact.

apmantza:
Merged, thank you @AcademyKUB8 — your placeholder fix, the cross-language `select` removal and the psycopg `sql.SQL(...).format(sql.Identifier(...))` recognition all shipped as you wrote them; the maintainer follow-up commits restored the SQLAlchemy 2.x exemptions and layered the provenance proof on top so `session.execute(select(...))` stays quiet while `stmt = select(User); db.execute(stmt)` and the `Depends(get_db)` case now go quiet too. Ships in the next release.

Issue #2582, opened 2026-09-04
Consolidate the 21 hand-rolled LSPService test doubles behind one factory in tests/support (refs #2540)


apmantza:
## Problem
21 test files each hand-roll a partial `LSPService` double (`supportsLSP` / `touchFile` / `openFile` / `getAllDiagnostics` …) with `vi.fn()`:

```
tests/clients/cascade-compute.test.ts (28 doubles), tests/index-integration.test.ts (10), tests/index-lsp-idle-reset.test.ts (4),
tests/clients/pipeline-lsp-sync.test.ts (2), and 17 files with one each (pipeline, lazy-install-retry, observed-mutation-integration,
read-expansion-enrichment, reverse-deps-cache, runtime-session-* ×8, session-start-total-classification, shared-checkout-guard-wiring,
word-index-per-edit, dispatch/runners/lsp-inferred-warm-attach)
```
(list from `grep -rlE 'touchFile: vi\.fn' tests/`).

Every method added to the service surface is a latent break across all of them. PR #2568 (#2540) showed the cost: dropping an optional call on `getAuxiliaryClientsForFile` — asked for in review because the production service always has it — threw in `tests/clients/pipeline.test.ts` whose double lacked it, redding the unit job (run 33856858492) and needing a maintainer trailing commit (91535417). The production code had carried `?.()` only to accommodate partial doubles — a test-shape leaking into the hot path (AGENTS.md shape 7 family: vacuous / mock-shaped fixtures).

## Ask
- One `makeLspServiceDouble(overrides?)` in `tests/support/` (next to `fake-lsp-server.ts` / `pi-mock.ts`) that returns a double covering the full method surface the pipeline and runtime-session paths touch, with sensible resolved defaults; per-test overrides for the methods a test actually asserts.
- Migrate the 21 files (net-count rule: the factory lands with its siblings folded in the same PR, not as a follow-up).
- One guard, the recurrence itself: a sweep that reds on a fresh hand-rolled service double outside the factory (same pattern as `tests/config/sweep-floor-coverage.test.ts`). No prototype-completeness guard — speculative, dropped per the AGENTS.md ladder-first rule (2026-09-04).
- Delete the vacuous fixtures the migration exposes (tests whose double never reached the code under test).

## Acceptance
- `grep -rlE 'touchFile: vi\.fn' tests/` → only the factory.
[...]

PR #2583, opened 2026-09-04
fix(test-runner): honor uv environments (closes #2580)

### LINKED ISSUE #2580: Test runner: pytest runs with the wrong interpreter for uv-managed projects, reported as “Pytest configuration error”
### What happened?

The turn-end test runner reports Pytest configuration error for a project
whose tests pass when run normally with uv run pytest.
Root cause is the interpreter resolution, not pytest. resolveExec() asks
detectPythonEnvironment(cwd) for the project interpreter
(clients/test-runner-client.ts:2758-2770):

`if (runner === "pytest") {
    const pythonEnvironment = await detectPythonEnvironment(cwd);
    if (pythonEnvironment) {
        return {
            command: pythonEnvironment.pythonPath,
            args: config.args(testFile, cwd),
            env: augmentPythonEnvironment(process.env, pythonEnvironment),
        };
    }
}
`

and detectPythonEnvironment only knows four candidates
(clients/python-environment.ts:24-32):

`const candidates: Array<{ root: string | undefined; source: PythonEnvironmentSource }> = [
    { root: process.env.VIRTUAL_ENV,  source: "virtual-env" },
    { root: process.env.CONDA_PREFIX, source: "conda" },
    { root: path.join(projectRoot, ".venv"), source: "project-dot-venv" },
    { root: path.join(projectRoot, "venv"),  source: "project-venv" },
];
`

There is no uv awareness anywhere in the codebase — the only occurrences of
“uv” are uv.lock as a generated artifact (clients/generated-artifacts.ts:58,
clients/lsp/server.ts:121) and a vulture install hint
(clients/dead-code-client.ts:400). uv’s default layout (<root>/.venv)
happens to be matched by the third candidate, so the common case works by
coincidence; every non-default uv layout misses:
	•	UV_PROJECT_ENVIRONMENT pointing outside the project root
	•	a uv workspace, where the single environment lives at the workspace root
while the pi session cwd is a member package (the candidate list is flat —
there is no walk up to a parent directory)
	•	PEP 723 script environments (uv run --script), which live in uv’s cache
On a miss, resolveExec falls through to the runner default
(clients/test-runner-client.ts:241-246, returned at :2806):

`pytest: {
    configFiles: ["pytest.ini", "pyproject.toml", "setup.cfg", "tox.ini"],
    command: "python",
    args: (testFile, _cwd) => ["-m", "pytest", testFile, "--tb=short", "-q"],
    parseJson: false,
},
ˋ

i.e. a bare python off the host PATH. That interpreter has neither the
project’s dependencies nor its pytest plugins, so pytest aborts during
collection and exits 2.
Secondary bug, same chain: clients/test-runner-client.ts:1872 maps that
exit code to a misleading label.

`
error: exitCode === 2 ? "Pytest configuration error" : undefined,
`

In pytest’s ExitCode enum, 2 is INTERRUPTED; the usage/configuration
error is 4 (USAGE_ERROR). INTERRUPTED is what a collection error
produces (ImportError, missing plugin, failing conftest.py). So the one
symptom the user sees actively points away from the real cause — a wrong
interpreter — and toward their pytest config, which is fine.

### Steps to reproduce

Verified with uv 0.8.17. Any uv layout that is not <cwd>/.venv reproduces
it; the shortest is UV_PROJECT_ENVIRONMENT:

`$ mkdir uvtest && cd uvtest
$ printf '[project]\nname="uvtest"\nversion="0.1.0"\nrequires-python=">=3.9"\n\n[tool.pytest.ini_options]\ntestpaths=["tests"]\n' > pyproject.toml

$ export UV_PROJECT_ENVIRONMENT=/tmp/altenv
$ uv sync
$ ls -a
.  ..  pyproject.toml  uv.lock          # <- no .venv in the project root

# what uv actually uses:
$ uv run --no-sync python -c "import sys; print(sys.executable)"
/tmp/altenv/bin/python3

# what pi-lens falls back to (all four candidates missed):
$ python -c "import sys; print(sys.executable)"
/usr/local/bin/python
`

Add a tests/test_x.py importing any project dependency, edit a .py file so
the turn-end runner fires, and the run comes back as
Pytest configuration error while uv run pytest passes.
Note that the runner is still selected in this state: detectRunner’s
priority 1 matches on the presence of [tool.pytest.ini_options] in
pyproject.toml (clients/test-runner-client.ts:627-634), and its priority 5
fallback accepts a merely global which pytest
(clients/test-runner-client.ts:743-750). Nothing in detection checks that
the interpreter that will run pytest can actually import the project.

### Expected behavior

pytest runs against the environment uv would use, so the turn-end result
matches uv run pytest. Failing that, pi-lens should stay silent rather than
report a Pytest configuration error that does not exist.

### pi-lens version

4.1.3 (commit aeff0a88d2a1082b2f38e95ed806f31612912745)

### Relevant logs

_No response_

### Area

area:tests

apmantza:
## Summary

Pytest runs through the environment uv selects. `clients/python-environment.ts` resolves candidates in this order: `UV_PROJECT_ENVIRONMENT`, an explicit uv workspace `.venv`, `VIRTUAL_ENV`, `CONDA_PREFIX`, `<root>/.venv`, `<root>/venv`. Pytest exit code 4 reports `Pytest configuration error`; exit code 2 reports `Pytest interrupted`. Child `PATH` and `VIRTUAL_ENV` stay process-scoped. PEP 723 `uv run --script` cache environments remain out of scope: they are content-keyed with no stable discoverable path.

**Review round 2** reshaped the uv half after six findings; **review round 3** found that round 2 had gated on the wrong axis and fixed it. The behaviour that ships:

- The two uv candidates exist **only when the passed root IS a uv project root** — its nearest `pyproject.toml` is the directory itself. A `pyproject.toml` merely sitting *above* the root makes that directory a subdirectory of a project, not a project, and **nothing below a project root inherits anything from it**. An exported `UV_PROJECT_ENVIRONMENT` (uv's documented CI/Docker recipe) therefore captures neither an unrelated checkout nor a subdirectory of an unrelated project.
- A relative `UV_PROJECT_ENVIRONMENT` is anchored at the member's workspace root, else at the project root — which, given the gate, is the passed root itself. Never the cwd.
- Only a **declared, non-excluded member** of an *explicit* workspace inherits that workspace's `.venv`. The workspace root itself is a project, not a "member": it reaches its own `<ws>/.venv` through the ordinary project candidate at the same path.
- The `pyproject.toml` walk stops at an **injected** `homeDir`.
- The uv glob dialect is pinned to upstream, not guessed.

Closes #2580 — every acceptance criterion is met.

Merge-order note: this branch merges `origin/master` twice — at `70bb559b` (picks up #2584 / AGENTS.md shape 37) and at `340a9d71` (picks up #2587's `pi.skills` packaging bullet, semantically disjoint from this diff). It adds **AGENTS.md defect shape 39**; shape 38 belongs to #2585, so the numbers do not collide. No other open PR touches `clients/python-environment.ts`, `clients/cargo-manifest.ts`, or `tests/clients/test-runner-python-environment.test.ts`.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:lsp
- [x] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [x] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] Every new guard/branch/filter is mutation-proof: deleting or neutering it reds at least one test — all ten transcribed below; round 3 added vectors for the two that were previously unproven (T2)
- [x] `npm run lint` passes
- [x] `.changelog/<branch-or-slug>-<short-desc>.md` has one valid entry **in this PR** for any user-facing change

## Review round 3 — state space (written BEFORE the code, per the round-count rail)

Round 2 fixed F1/F2 on the wrong axis: it gated the uv candidates on "a `pyproject.toml` exists at **or above** root" when the rule needed is "root **is** the project". That left S1 and introduced S2. This section is the state-space step AGENTS.md requires once a seam reaches this round count; the code that follows is written to it, and every cell below has a test.

### Invariants

- **I1** — uv's project settings (`UV_PROJECT_ENVIRONMENT`, the workspace `.venv`) apply to a directory only when that directory **is** a uv project root (its nearest `pyproject.toml` is at the directory itself), or is a declared, non-excluded member of an explicit workspace — which is itself a project root.
- **I2** — nothing **below** a project root inherits anything from above it. This is the F3 decision from round 2, applied consistently.
- **I3** — a relative `UV_PROJECT_ENVIRONMENT` is anchored at `memberWorkspaceRoot ?? projectRoot`. Wherever a uv candidate exists, `projectRoot === root`, so the anchor is root itself or the explicit workspace root.
- **I4** — candidate precedence is fixed and total: `uv-project-environment` > `uv-workspace` > `VIRTUAL_ENV` > `CONDA_PREFIX` > `<root>/.venv` > `<root>/venv`. A candidate whose interpreter file is missing falls through to the next; exhausting them yields `undefined`.
- **I5** — the `pyproject.toml` walk stops at the injected `homeDir` (`isAtOrAboveHomeDir`).
- **I6** — nothing mutates `process.env`; the environment is built for the child only.

### Writers × axes

| Writer | What it sets | Axis |
| --- | --- | --- |
| The user's repo layout | `pyproject.toml` placement; `[tool.uv.workspace] members` / `exclude` | **A** (root position) |
| The process environment (CI image, `Dockerfile`, shell activation) | `UV_PROJECT_ENVIRONMENT`, `VIRTUAL_ENV`, `CONDA_PREFIX` | **U**, **V** |
| The filesystem | `<root>/.venv`, `<root>/venv`, and whether the dir named by `UV_PROJECT_ENVIRONMENT` exists | **D** |
| pi-lens callers | the `projectRoot` argument — `clients/test-runner-client.ts:2767` (pytest cwd), `clients/dispatch/runners/pyright.ts:83` (dispatch cwd), `clients/lsp/server.ts:2277` / `:2434` (LSP root) | **A**, and this is what makes A2 reachable: `PythonServer`'s `NearestRoot` detector can return `<mono>/frontend` off a `requirements.txt` while the `pyproject.toml` sits at `<mono>` |

Axis values:

- **A** — `A1` root-is-project (implicit single project) · `A2` root-below-project · `A3` root-is-declared-member of an explicit workspace · `A4` root-excluded (a project the workspace's `exclude` matches) · `A5` no-pyproject anywhere · `A6` root-is-explicit-workspace-root
- **U** — `U-abs` (absolute) · `U-rel` (relative) · `U-unset`
- **V** — `V-set` · `V-unset`
- **D** — `D-present` (`<root>/.venv` exists) · `D-absent`

`CONDA_PREFIX` is held unset across the grid (it is an unconditional pass-through already covered by `tests/clients/lsp/server-policy.test.ts`), and `<root>/venv` is held absent; both sit below `.venv` in I4 and add no new interaction.

### Derivation

Per I1, the two uv candidates exist iff **A ∈ {A1, A3, A4, A6}** (`isStartDir`); the `uv-workspace` candidate additionally requires **A3** (declared, non-excluded member). Everything else falls out of I4's precedence. Two consequences worth stating, because both are cells this PR changes:

- **A6** (the explicit workspace root itself) is a project, not a "member": it gets `uv-project-environment`, and its own `<ws>/.venv` is found by the ordinary `project-dot-venv` candidate — same directory, honest source label. This is what deleting S2's `projectRoot === workspace.root` branch buys.
- **A4** (an excluded project) still gets `uv-project-environment` anchored at itself — `exclude` removes it from the *workspace*, not from being a uv project.

### Cell list — 72 cells, expected source for every one

```
A1 — root-is-project (implicit single project)
  U-abs    V-set    D-present  -> uv-project-environment
  U-abs    V-set    D-absent   -> uv-project-environment
  U-abs    V-unset  D-present  -> uv-project-environment
  U-abs    V-unset  D-absent   -> uv-project-environment
  U-rel    V-set    D-present  -> uv-project-environment
  U-rel    V-set    D-absent   -> uv-project-environment
  U-rel    V-unset  D-present  -> uv-project-environment
  U-rel    V-unset  D-absent   -> uv-project-environment
  U-unset  V-set    D-present  -> virtual-env
  U-unset  V-set    D-absent   -> virtual-env
  U-unset  V-unset  D-present  -> project-dot-venv
  U-unset  V-unset  D-absent   -> undefined

A2 — root-below-project (a subdir of a project root)
  U-abs    V-set    D-present  -> virtual-env
  U-abs    V-set    D-absent   -> virtual-env
  U-abs    V-unset  D-present  -> project-dot-venv
  U-abs    V-unset  D-absent   -> undefined
  U-rel    V-set    D-present  -> virtual-env
  U-rel    V-set    D-absent   -> virtual-env
  U-rel    V-unset  D-present  -> project-dot-venv
  U-rel    V-unset  D-absent   -> undefined
  U-unset  V-set    D-present  -> virtual-env
  U-unset  V-set    D-absent   -> virtual-env
  U-unset  V-unset  D-present  -> project-dot-venv
  U-unset  V-unset  D-absent   -> undefined

A3 — root-is-declared-member of an explicit workspace
  U-abs    V-set    D-present  -> uv-project-environment
  U-abs    V-set    D-absent   -> uv-project-environment
  U-abs    V-unset  D-present  -> uv-project-environment
  U-abs    V-unset  D-absent   -> uv-project-environment
  U-rel    V-set    D-present  -> uv-project-environment
  U-rel    V-set    D-absent   -> uv-project-environment
  U-rel    V-unset  D-present  -> uv-project-environment
  U-rel    V-unset  D-absent   -> uv-project-environment
  U-unset  V-set    D-present  -> uv-workspace
  U-unset  V-set    D-absent   -> uv-workspace
  U-unset  V-unset  D-present  -> uv-workspace
  U-unset  V-unset  D-absent   -> uv-workspace

A4 — root-excluded (a project the workspace excludes)
  U-abs    V-set    D-present  -> uv-project-environment
  U-abs    V-set    D-absent   -> uv-project-environment
  U-abs    V-unset  D-present  -> uv-project-environment
  U-abs    V-unset  D-absent   -> uv-project-environment
  U-rel    V-set    D-present  -> uv-project-environment
  U-rel    V-set    D-absent   -> uv-project-environment
  U-rel    V-unset  D-present  -> uv-project-environment
  U-rel    V-unset  D-absent   -> uv-project-environment
  U-unset  V-set    D-present  -> virtual-env
  U-unset  V-set    D-absent   -> virtual-env
  U-unset  V-unset  D-present  -> project-dot-venv
  U-unset  V-unset  D-absent   -> undefined

A5 — no-pyproject anywhere at or above root
  U-abs    V-set    D-present  -> virtual-env
  U-abs    V-set    D-absent   -> virtual-env
  U-abs    V-unset  D-present  -> project-dot-venv
  U-abs    V-unset  D-absent   -> undefined
  U-rel    V-set    D-present  -> virtual-env
  U-rel    V-set    D-absent   -> virtual-env
  U-rel    V-unset  D-present  -> project-dot-venv
  U-rel    V-unset  D-absent   -> undefined
  U-unset  V-set    D-present  -> virtual-env
  U-unset  V-set    D-absent   -> virtual-env
  U-unset  V-unset  D-present  -> project-dot-venv
  U-unset  V-unset  D-absent   -> undefined

A6 — root-is-explicit-workspace-root
  U-abs    V-set    D-present  -> uv-project-environment
  U-abs    V-set    D-absent   -> uv-project-environment
  U-abs    V-unset  D-present  -> uv-project-environment
  U-abs    V-unset  D-absent   -> uv-project-environment
  U-rel    V-set    D-present  -> uv-project-environment
  U-rel    V-set    D-absent   -> uv-project-environment
  U-rel    V-unset  D-present  -> uv-project-environment
  U-rel    V-unset  D-absent   -> uv-project-environment
  U-unset  V-set    D-present  -> virtual-env
  U-unset  V-set    D-absent   -> virtual-env
  U-unset  V-unset  D-present  -> project-dot-venv
  U-unset  V-unset  D-absent   -> undefined
```

Verifier's note, recorded as expected rather than as a defect: probe E′ (an implicit project called with `<imp>/tests`) returns `undefined` on master too. That is cell **A2 / U-unset / V-unset / D-absent → undefined** — the F3 decision, not a regression.

### Findings this round, reproduced first

Both reproduced on the built `clients/python-environment.js` at `70bb559b` via `detectPythonEnvironment`, `PI_LENS_HOME`/`PILENS_DATA_DIR` pinned to `<worktree>/.probe-home`:

```
R1 (below project, .venv, UV_PROJECT_ENVIRONMENT exported)   want project-dot-venv: source=uv-project-environment root=R/ci-env
R2 (below project, VIRTUAL_ENV + UV_PROJECT_ENVIRONMENT)     want virtual-env    : source=uv-project-environment root=R/ci-env
W3 (workspace root's docs/ subdir, not a project)            want undefined      : source=uv-workspace root=W/ws/.venv
W4 (same, and docs/ is in exclude)                           want undefined      : source=uv-workspace root=W/wsx/.venv
G  (the workspace root itself)                               want project-dot-venv: source=uv-workspace root=W/ws/.venv
```

- **S1 (leftover F1)** — R1/R2 are cells A2 / U-abs / {V-unset,V-set} / D-present, expected `project-dot-venv` and `virtual-env`. Violates I1 and I2.
- **S2 (new in round 2)** — W3/W4 are A2-shaped directories (`<ws>/docs` is not a project) that the `projectRoot === workspace.root` shortcut promoted to members. Violates I1; `exclude` was bypassed entirely.
- **T1** — AGENTS.md, this body and the changelog fragment claim behaviour that is false under R1/R2/W3/W4. All three are being rewritten to the shipped rule.
- **T2** — `{ dot: true }` and `path.resolve(projectRoot)` need vectors or the mutation-proof checklist box is false as written.

### Planned remedy (the maintainer's design decision)

`findUvWorkspace` returns `{ projectRoot, isStartDir, … }`. The uv candidates exist only when `isStartDir` (a declared member is itself a project root, so "member" is a strict subset that additionally unlocks the workspace `.venv` and re-anchors a relative `UV_PROJECT_ENVIRONMENT`). `isUvWorkspaceMember`'s `projectRoot === workspace.root` branch is deleted. The round-2 test at `:246` encodes the rejected premise — a directory *below* a project root resolving the project's `UV_PROJECT_ENVIRONMENT` — and is rewritten so the passed root **is** the project, with a sibling asserting the below-project directory gets its own `.venv` / `VIRTUAL_ENV` / `undefined`.

### Round 3 — results

**Remedy shipped** (`8fd1fa8b`): `findUvWorkspace` now reports `isStartDir` alongside `projectRoot`, and `isStartDir` is the gate for both uv candidates. Membership is a strict refinement of it (a declared member is itself a project root) that additionally unlocks the workspace `.venv` and re-anchors a relative `UV_PROJECT_ENVIRONMENT`. `isUvWorkspaceMember`'s `projectRoot === workspace.root` shortcut is deleted; the relative-path guard already returns the right answer for the root itself.

**Probes after the fix** (same harness, same pinned `PI_LENS_HOME`) — all ten match the cell list:

```
R1 (below project, .venv, UV_PROJECT_ENVIRONMENT exported)   want project-dot-venv: source=project-dot-venv root=R/mono/frontend/.venv
R2 (below project, VIRTUAL_ENV + UV_PROJECT_ENVIRONMENT)     want virtual-env    : source=virtual-env root=R/activated
C  (below project, no own .venv, nothing set)                want undefined      : source=undefined root=undefined
W3 (workspace root's docs/ subdir, not a project)            want undefined      : source=undefined root=undefined
W4 (same, and docs/ is in exclude)                           want undefined      : source=undefined root=undefined
G  (the workspace root itself)                               want project-dot-venv: source=project-dot-venv root=W/ws/.venv
A  (no pyproject, own .venv, UV_PROJECT_ENVIRONMENT exported) want project-dot-venv: source=project-dot-venv root=A/proj/.venv
B  (no pyproject, VIRTUAL_ENV + UV_PROJECT_ENVIRONMENT)       want virtual-env    : source=virtual-env root=R/activated
F  (declared member, relative UV_PROJECT_ENVIRONMENT)         want uv-project-environment ws/.uv-env: source=uv-project-environment root=F/ws/.uv-env
E' (implicit project, called with <imp>/tests)                want undefined (F3)  : source=undefined root=undefined
```

**Tests.** `tests/clients/test-runner-python-environment.test.ts` now holds 93 cases: the 21 seam tests (spawn-level, through `TestRunnerClient.runTestFileAsync`) plus the 72-cell `python environment state space (review round 3)` grid. The grid's `expectedSource` is computed from I1–I4 alone — the contract, not the implementation — so a behaviour change that is not also a contract change reds rather than silently re-baselining (the "implementation mirror" screen). Each `uv-project-environment` cell also asserts the resolved ROOT: the fixture materializes the environment **only** at I3's anchor, plus a **decoy** at the anchor a violating resolver would use (A2), so those cells discriminate instead of passing on a missing directory.

W3/W4 are A2-shaped (a `docs/` under a workspace root is not a project). The ancestor's *kind* is not an axis, because once root is not the project I1 makes it irrelevant — so they ride as named regression tests rather than widening the grid.

New/edited this round:

1. `gives a directory below a project root none of that project's uv environment` — S1/R1.
2. `keeps an activated VIRTUAL_ENV for a directory below a project root` — S1/R2.
3. `does not give a non-project subdirectory of a workspace root the workspace .venv` — S2/W3.
4. `does not give an EXCLUDED non-project subdirectory the workspace .venv` — S2/W4.
5. `keeps the workspace root itself on its own .venv` — S2/G.
6. `matches a dot-directory member the way uv's MatchOptions do` — T2 vector for `{ dot: true }`.
7. `resolves a relative root to an absolute interpreter path` — T2 vector for the seam-entry `path.resolve`. `process.chdir` is a stated boundary decision (a relative root is only meaningful against a cwd) and is restored in `finally`.
8. **EDITED** — round 2's `resolves a relative UV_PROJECT_ENVIRONMENT from the discovered project root` encoded the rejected premise (its fixture passed a directory *below* the project). Rewritten as `... from the project root itself`, with tests 1 and 2 as the siblings covering the below-project case.
9. **EDITED** — `stops the uv workspace walk at the home directory`. The old fixture put a bare ancestor project above `$HOME`; the new `isStartDir` gate discards that on its own, so the test stopped discriminating (mutation M4 went green). The fixture now puts an explicit `[tool.uv.workspace]` above `$HOME` that claims the project as a **member** — the only shape that can still change the answer — and M4 reds again.
10. The env save/restore hooks moved to **file level**: scoped to one `describe`, the grid leaked `VIRTUAL_ENV` into the other (caught by 7 spurious reds during the red-first run; the "env leakage" screen).

**Red-first**, final test file against pre-round-3 `clients/python-environment.ts` (restored with `git checkout 70bb559b -- clients/python-environment.ts` after committing tests + fix, then rebuilt; the source was restored from `HEAD` and rebuilt afterwards, `git status` clean):

```
 FAIL  ... > gives a directory below a project root none of that project's uv environment
 FAIL  ... > keeps an activated VIRTUAL_ENV for a directory below a project root
 FAIL  ... > does not give a non-project subdirectory of a workspace root the workspace .venv
 FAIL  ... > does not give an EXCLUDED non-project subdirectory the workspace .venv
 FAIL  ... > keeps the workspace root itself on its own .venv
 FAIL  ... > A2 U-abs V-set   D-present -> virtual-env      (root-below-project)
 FAIL  ... > A2 U-abs V-set   D-absent  -> virtual-env      (root-below-project)
 FAIL  ... > A2 U-rel V-set   D-present -> virtual-env      (root-below-project)
 FAIL  ... > A2 U-rel V-set   D-absent  -> virtual-env      (root-below-project)
 FAIL  ... > A2 U-abs V-unset D-present -> project-dot-venv (root-below-project)
 FAIL  ... > A2 U-rel V-unset D-present -> project-dot-venv (root-below-project)
 FAIL  ... > A2 U-abs V-unset D-absent  -> undefined        (root-below-project)
 FAIL  ... > A2 U-rel V-unset D-absent  -> undefined        (root-below-project)
 FAIL  ... > A6 U-unset V-set   D-present -> virtual-env      (root-is-explicit-workspace-root)
 FAIL  ... > A6 U-unset V-unset D-present -> project-dot-venv (root-is-explicit-workspace-root)
 Test Files  1 failed (1)
      Tests  15 failed | 78 passed (93)
```

(The full run prints the `FAIL` lines with the full file path; they are elided to `...` here for width. Post-fix: `Tests 93 passed (93)`.)

**Mutation proof.** T2 is closed: both previously-unproven guards now have vectors, so the checklist box is true as written.

```
=== N1 (S1) gate on "a pyproject at or above root" again (the round-2 shape) ===
 FAIL  ... > gives a directory below a project root none of that project's uv environment
 FAIL  ... > keeps an activated VIRTUAL_ENV for a directory below a project root
 FAIL  ... > A2 U-abs V-set D-present / D-absent, A2 U-rel V-set D-present / D-absent,
             A2 U-abs V-unset D-present / D-absent, A2 U-rel V-unset D-present / D-absent
      Tests  10 failed | 83 passed (93)

=== N2 (S2) restore the projectRoot === workspace.root shortcut ===
 FAIL  ... > keeps the workspace root itself on its own .venv
 FAIL  ... > A6 U-unset V-set D-present, A6 U-unset V-unset D-present
      Tests  3 failed | 90 passed (93)

=== N3 (T2) delete { dot: true } from the uv member matcher ===
 FAIL  ... > matches a dot-directory member the way uv's MatchOptions do
      Tests  1 failed | 92 passed (93)

=== N4 (T2) delete the seam-entry path.resolve ===
 FAIL  ... > resolves a relative root to an absolute interpreter path
      Tests  1 failed | 92 passed (93)

=== M4 (round 2, F6) delete the home-directory ceiling ===
 FAIL  ... > stops the uv workspace walk at the home directory
      Tests  1 failed | 92 passed (93)

=== M5 (round 2) delete the uv member-glob normalization ===
 FAIL  ... > matches a uv member glob written with a leading ./
      Tests  1 failed | 92 passed (93)

=== M6 (round 2, F5) weaken hasTomlTable presence to a truthy test ===
 FAIL  |default| tests/clients/cargo-manifest.test.ts > hasCargoWorkspaceTable (#2498) > is true for a present-but-empty [workspace] table (review round 3, F1 shape)
      Tests  1 failed | 24 passed (25)
```

Built output restored from a byte-for-byte backup after every mutation (`diff -q` clean; verified after each batch).

**Honest note on M4.** The round-2 ceiling test was *not* vacuous when it shipped — it reddened M4 then. Round 3's `isStartDir` gate subsumed its fixture, which is exactly the "a guard's test stops discriminating because an unrelated change now short-circuits it" trap; it was caught by re-running the round-2 mutations rather than assuming they still held, and the fixture was strengthened. Re-running the previous round's mutation set after a gate change is the practice that caught it.

**T1 — the three surfaces.** AGENTS.md's Python-project-environments bullet, `.changelog/fix-2580-uv-pytest-environment.md`, and this body's Summary all previously claimed behaviour that was false while S1/S2 were live. All three now state the shipped rule: the passed root must BE the project; nothing below a project root inherits from it; the workspace root is a project rather than a member. AGENTS.md additionally names both ways this repo already got it wrong, and points at the 72-cell grid so the next change extends it instead of adding a one-off case.

**Catalog.** This round adds **AGENTS.md defect shape 39 — "a walk-up result used as an eligibility gate"**, the class that produced both S1 and S2 (a "which project owns this directory" walk answering "does this directory get that project's environment", and an `===` shortcut standing in for the same predicate). It carries the screen, the state-space pairing, and two detect greps, and cites #2583. Numbered 39 because 38 belongs to #2585.

**Suites (round 3, post-merge).** `npm run build` first. **70 files** selected mechanically by the same procedure as round 2 (symbol grep over `tests/`, extended with `isUvWorkspaceMember` and `hasTomlTable`; `ls` of every `tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts`; every `tests/config/*.test.ts`; plus `test-runner-client`, `flake-shape-ratchet`, `lsp/server-policy`, the pyright dispatch runners, and `tests/packaging.test.ts` for the merged #2587 change):

```
 Test Files  70 passed (70)
      Tests  1325 passed | 1 skipped (1326)
```

`npm run lint` clean · `npx oxfmt --check` on the touched files: `All matched files use the correct format.` (it flagged two files first; they were formatted and re-tested) · `node scripts/check-changelog-fragments.mjs`: `changelog fragments OK (115 entries in .changelog/)` · `scripts/check-pr-body.mjs`: `PR body OK: 2583`.

## Review round 2

Every finding was reproduced first against the built `clients/python-environment.js` through `detectPythonEnvironment`, with `PI_LENS_HOME`/`PILENS_DATA_DIR` pinned to `<worktree>/.probe-home`.

### F1 (high) — exported `UV_PROJECT_ENVIRONMENT` hijacked non-uv projects

Reproduced (pre-fix build):

```
A (no pyproject, own .venv, UV_PROJECT_ENVIRONMENT exported) expect project-dot-venv: source=uv-project-environment root=A/ci-env
B (VIRTUAL_ENV activated + env var, no pyproject) expect virtual-env: source=uv-project-environment root=B/ci-env
```

Remedy: both uv candidates are gated on `uvWorkspace !== undefined`, i.e. a discovered `pyproject.toml`. After the fix:

```
A (no pyproject, own .venv, UV_PROJECT_ENVIRONMENT exported) expect project-dot-venv: source=project-dot-venv root=A/proj/.venv
B (VIRTUAL_ENV activated + env var, no pyproject) expect virtual-env: source=virtual-env root=B/activated
```

The shipped test at `:146` did pin the wrong behaviour (`createProject` writes no `pyproject.toml`). It now writes one and is renamed `uses an absolute UV_PROJECT_ENVIRONMENT path for a uv project`; probes A and B are new red-first tests.

### F2 (medium-high) — relative `UV_PROJECT_ENVIRONMENT` resolved from the wrong base

Remedy as prescribed: `uvWorkspace.projectRoot` replaces `path.resolve(projectRoot)`. The whole expression collapses to `memberWorkspaceRoot ?? uvWorkspace.projectRoot`, and `path.resolve(base, value)` replaces the `isAbsolute` ternary.

**Premise note, reported honestly.** The reviewer's probe varied the *cwd*, which only changes the answer when `projectRoot` itself is relative. All four production call sites pass an absolute path (`clients/test-runner-client.ts:2767`, `clients/dispatch/runners/pyright.ts:83`, `clients/lsp/server.ts:2277`, `:2434`), so I did **not** build machinery for a cwd collision (the #2490 shape). The cwd-independent defect underneath it is real and is what the red-first test drives: the resolver is handed a dispatch cwd / LSP root that can sit **below** the `pyproject.toml`, and the old code anchored the relative value at that directory. One line — `const root = path.resolve(projectRoot)` at the seam entry, matching `findUvWorkspace`'s own existing `path.resolve` — makes the relative-input variant behave identically rather than leaving a second answer for the same question. Post-fix probes, including the reviewer's cwd shapes and probe E:

```
F1 (cwd=nested/tests) expect nested/.uv-env: source=uv-project-environment root=F/ws/tools/standalone/.uv-env
F2 (cwd=nested) expect nested/.uv-env: source=uv-project-environment root=F/ws/tools/standalone/.uv-env
F3 (projectRoot=nested/tests, a subdir of the project) expect nested/.uv-env: source=uv-project-environment root=F/ws/tools/standalone/.uv-env
E (cwd=nested/tests, relative projectRoot arg) expect nested/.venv: source=project-dot-venv root=E/ws/tools/standalone/.venv
E' (absolute projectRoot arg) expect nested/.venv: source=project-dot-venv root=E/ws/tools/standalone/.venv
```

Probe E's pre-fix answer was not `undefined` on my build but garbage — `source=project-dot-venv root=../../home/akis/Desktop/pi-lens/.claude/worktrees/.venv` — the same relative-input root cause, now fixed.

### F3 (medium) — ancestor-`.venv` inheritance

**Chosen: DROPPED**, per the maintainer's stated preference. #2580 asks for uv workspaces, not generic ancestor inheritance. The branch (and the now-unused `UvWorkspace.startIsProject` field, and the `workspaceRoot !== path.resolve(projectRoot)` guard the reviewer noted was implied) is gone. Probe C now matches master:

```
before: C (poetry root .venv, cwd=mono/frontend): source=project-dot-venv root=C/mono/.venv
after:  C (poetry root .venv, cwd=mono/frontend): source=undefined root=undefined
```

A red-first test pins it, and mutation M3 (restoring the branch) reds it.

### F4 — AGENTS.md

The Python-project-environments bullet is rewritten to the final behaviour: candidate order, the discovered-project gate with F1's reason, the workspace-root resolution rule, the no-ancestor-inheritance rule with the poetry case named, the injected home ceiling, and the pinned uv glob dialect.

### F5 — duplication

**Folded in this PR:**

- `hasTomlTable(content, tableName)` on the existing TOML seam in `clients/cargo-manifest.ts`. `hasCargoWorkspaceTable` is now a one-line caller; the private `hasUvWorkspaceTable` is gone.
- The hand-rolled minimatch options block is gone. Reading upstream (below) showed `nocase` was not merely duplicated but **wrong**: uv matches members case-sensitively on every platform. The remaining call passes `{ dot: true }`, which is minimatch's default dialect plus the one flag upstream's `require_literal_leading_dot: false` needs.

**NOT folded, with the reason and a follow-up: #2591.** Sibling list: `clients/lsp/server.ts:2500` `matchesCargoWorkspacePattern` + `:2533` `cargoWorkspaceDeclaresMember` (+ the private `segmentGlobToRegExpSource` at `:2475`); `clients/python-environment.ts` `isUvWorkspaceMember`; `clients/review-graph/workspace-modules.ts:124` `expandWorkspacePattern`. Reason (dialect divergence, now evidenced rather than asserted — see Class sweep): cargo never matches `**` and requires equal segment counts; uv matches `members` with `MatchOptions { require_literal_separator: true, ..MatchOptions::new() }` but `exclude` with `Pattern::matches_path`, i.e. `require_literal_separator: false` — **two different option sets inside one tool**; npm globs the filesystem instead of matching. Folding cargo onto minimatch would silently start honouring `**` in Rust-LSP workspace hoisting, inside a Python PR.

### F6 — depth cap and hardcoded `os.homedir()`

The `depth++ >= 64` cap is deleted (it named no recurrence and could not fire before the home/root ceiling). `findUvWorkspace` now uses the plain `walkUpDirs` + `isAtOrAboveHomeDir` shape, and `detectPythonEnvironment(projectRoot, homeDir = os.homedir())` injects the ceiling the way `configSearchDirs` (`clients/config-locations.ts:182`) and `localBinMatches` (`clients/package-manager.ts:674`) take it. A new red-first test proves the ceiling and reds mutation M4.

### Nit — test count

Correct: the runner prints **14 passed** for this file now (8 before this round). The red-first transcript is pasted verbatim below rather than paraphrased. I have also **removed** the previous body's "Real uv 0.10.8 probe selected the same workspace interpreter" line: `uv` is not installed in this worktree (`which uv` is empty) and I cannot reproduce that claim.

## Tests

All in `tests/clients/test-runner-python-environment.test.ts`, all driving the production seam (`TestRunnerClient.runTestFileAsync` → `detectPythonEnvironment`), except the home-ceiling case which calls the exported `detectPythonEnvironment` directly because `homeDir` is its injected parameter. No fake timers or wall-clock waits are involved (no timing seam is touched), so no `flake-shape` header is needed; `tests/clients/flake-shape-ratchet.test.ts` passes.

NEW (6):

1. `ignores an exported UV_PROJECT_ENVIRONMENT outside a uv project` — F1 probe A. Regression proof.
2. `keeps an activated VIRTUAL_ENV over an exported UV_PROJECT_ENVIRONMENT outside a uv project` — F1 probe B. Regression proof.
3. `resolves a relative UV_PROJECT_ENVIRONMENT from the discovered project root` — F2. Explicit workspace, non-member nested project, `projectRoot` = its `tests/` subdirectory. Regression proof.
4. `does not inherit an ancestor project's .venv from a subdirectory` — F3 probe C (poetry root, `cwd=mono/frontend`). Regression proof.
5. `stops the uv project walk at the home directory` — F6. A `pyproject.toml` above the injected `$HOME` must not be discovered. Contract seam.
6. `matches a uv member glob written with a leading ./` — upstream contract vector (see Class sweep). Contract seam.

EDITED (1): `uses an absolute UV_PROJECT_ENVIRONMENT path` → `... for a uv project`, now writing a `pyproject.toml` (it pinned the F1 behaviour). Four pre-existing uv tests were rewritten onto new `createEnvironment` / `createTestFile` / `runPytest` helpers — same assertions, ~90 fewer lines of fixture boilerplate.

AGENTS.md test-authoring screens: **ambient-inspection double** — tests 1, 2 and 5 assert the spawned command and `options.env` the runner actually hands `safeSpawnAsync`, not an internal resolver return; **env leakage** — `VIRTUAL_ENV`/`CONDA_PREFIX`/`UV_PROJECT_ENVIRONMENT` are saved and restored per test by the existing `beforeEach`/`afterEach`, and test 5 injects `homeDir` rather than mutating the real one; **implementation mirror** — test 6's expectation comes from upstream uv's source, not from this repo's matcher.

### Red-first proof (pre-fix `clients/python-environment.js`, verbatim)

```
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > ignores an exported UV_PROJECT_ENVIRONMENT outside a uv project
AssertionError: expected '/tmp/pi-lens-uv-ci-env-HnZgiY/bin/pyt…' to be '/tmp/pi-lens-pytest-environment-Kdvsx…' // Object.is equality
Expected: "/tmp/pi-lens-pytest-environment-Kdvsxl/.venv/bin/python"
Received: "/tmp/pi-lens-uv-ci-env-HnZgiY/bin/python"
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > keeps an activated VIRTUAL_ENV over an exported UV_PROJECT_ENVIRONMENT outside a uv project
AssertionError: expected '/tmp/pi-lens-uv-ci-env-hAta7o/bin/pyt…' to be '/tmp/pi-lens-activated-env-A64GX6/bin…' // Object.is equality
Expected: "/tmp/pi-lens-activated-env-A64GX6/bin/python"
Received: "/tmp/pi-lens-uv-ci-env-hAta7o/bin/python"
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > resolves a relative UV_PROJECT_ENVIRONMENT from the discovered project root
AssertionError: expected 'python' to be '/tmp/pi-lens-pytest-environment-ALLOw…' // Object.is equality
Expected: "/tmp/pi-lens-pytest-environment-ALLOwZ/tools/standalone/.uv-env/bin/python"
Received: "python"
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > does not inherit an ancestor project's .venv from a subdirectory
AssertionError: expected '/tmp/pi-lens-pytest-environment-T8o4f…' to be 'python' // Object.is equality
Expected: "python"
Received: "/tmp/pi-lens-pytest-environment-T8o4fD/.venv/bin/python"
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > stops the uv project walk at the home directory
AssertionError: expected { …(4) } to be undefined
- Expected:
+ Received:
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > matches a uv member glob written with a leading ./
AssertionError: expected 'python' to be '/tmp/pi-lens-pytest-environment-GhEtS…' // Object.is equality
Expected: "/tmp/pi-lens-pytest-environment-GhEtSz/.venv/bin/python"
Received: "python"
 Test Files  1 failed (1)
      Tests  6 failed | 8 passed (14)
```

(The full-output form of the `stops the uv project walk at the home directory` failure, from the same run:)

```
AssertionError: expected { …(4) } to be undefined
+ Received:
{
  "binDir": "/tmp/pi-lens-uv-home-ceiling-hf19GO/.uv-env/bin",
  "pythonPath": "/tmp/pi-lens-uv-home-ceiling-hf19GO/.uv-env/bin/python",
  "root": "/tmp/pi-lens-uv-home-ceiling-hf19GO/.uv-env",
  "source": "uv-project-environment",
}
```

### Mutation proof (each guard neutered in the built `clients/*.js`, verbatim)

```
=== M1 (F1) delete the discovered-project gate on the uv candidates ===
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > ignores an exported UV_PROJECT_ENVIRONMENT outside a uv project
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > keeps an activated VIRTUAL_ENV over an exported UV_PROJECT_ENVIRONMENT outside a uv project
 Test Files  1 failed (1)
      Tests  2 failed | 12 passed (14)

=== M2 (F2) resolve a relative UV_PROJECT_ENVIRONMENT from the passed dir again ===
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > resolves a relative UV_PROJECT_ENVIRONMENT from the discovered project root
 Test Files  1 failed (1)
      Tests  1 failed | 13 passed (14)

=== M3 (F3) restore the ancestor-.venv inheritance branch this PR deletes ===
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > does not inherit an ancestor project's .venv from a subdirectory
 Test Files  1 failed (1)
      Tests  1 failed | 13 passed (14)

=== M4 (F6) delete the home-directory ceiling ===
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > stops the uv project walk at the home directory
 Test Files  1 failed (1)
      Tests  1 failed | 13 passed (14)

=== M5 delete the uv member-glob normalization ===
 FAIL  |default| tests/clients/test-runner-python-environment.test.ts > pytest project environment > matches a uv member glob written with a leading ./
 Test Files  1 failed (1)
      Tests  1 failed | 13 passed (14)

=== M6 (F5) weaken hasTomlTable presence to a truthy test ===
 FAIL  |default| tests/clients/cargo-manifest.test.ts > hasCargoWorkspaceTable (#2498) > is true for a present-but-empty [workspace] table (review round 3, F1 shape)
 Test Files  1 failed (1)
      Tests  1 failed | 24 passed (25)
```

The built output was restored from a byte-for-byte backup after each mutation (`diff -q` clean).

### Suites run

- `npm run build` before every run.
- **69 test files** selected mechanically, not from memory: the symbol grep over `tests/` for `cargoWorkspaceDeclaresMember|hasCargoWorkspaceTable|extractTomlTableSection|parseTomlStringArray|detectPythonEnvironment|detectPythonVenv|python-environment|cargo-manifest|augmentPythonEnvironment|pythonEnvironmentToolCandidates|isAtOrAboveHomeDir|walkUpDirs` (12 files), **plus** `ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts` and **every** `tests/config/*.test.ts` (54 files), plus `test-runner-client`, `flake-shape-ratchet`, and the pyright dispatch-runner suites.

```
 Test Files  69 passed (69)
      Tests  1214 passed | 1 skipped (1215)
```

- `npm run lint` — passes (tsc + oxlint, no output).
- `npx oxfmt --check` on the three touched source files — `All matched files use the correct format.`
- `node scripts/check-changelog-fragments.mjs` — `changelog fragments OK (114 entries in .changelog/)`.
- `npm run changelog:check` — `Validated 113 changelog entries.` (run pre-merge).
- The full suite is delegated to CI's Unit tests lane: it is serialized machine-wide (#1112) and this box is shared with other agent lanes.

### Test assessment

`tests/clients/test-runner-python-environment.test.ts` uniquely pins the real pytest execution seam — which interpreter `TestRunnerClient` spawns and what child `PATH`/`VIRTUAL_ENV` it hands it. **No test in it is vacuous and none is removed**; each has a named survivor mutation, demonstrated rather than asserted:

```
=== V1: member resolves UV_PROJECT_ENVIRONMENT from its own root, not the workspace root ===
 FAIL  ... > resolves a relative UV_PROJECT_ENVIRONMENT from the workspace root
      Tests  1 failed | 13 passed (14)

=== V2: drop the uv-workspace .venv candidate ===
 FAIL  ... > matches a uv member glob written with a leading ./
 FAIL  ... > uses the uv workspace .venv for a member package
      Tests  2 failed | 12 passed (14)

=== V3: treat every explicit-workspace descendant as a member (drop the members glob check) ===
 FAIL  ... > resolves a relative UV_PROJECT_ENVIRONMENT from the discovered project root
 FAIL  ... > keeps an independent nested project on its own .venv
      Tests  2 failed | 12 passed (14)

=== V4: ignore the workspace exclude list ===
 FAIL  ... > honors uv workspace exclusions over member globs
      Tests  1 failed | 13 passed (14)
```

Together with M1–M5 above, every one of the file's 14 cases reds on at least one mutation. `tests/clients/cargo-manifest.test.ts` is not edited; its existing `is true for a present-but-empty [workspace] table` case is the survivor that guards the `hasTomlTable` fold (M6).

## Blast radius

`module_report` is unavailable in this worktree, so the dependent set was mapped by grep over `clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts`:

- `clients/python-environment.ts` → `detectPythonEnvironment` has three callers (`clients/test-runner-client.ts:2767` pytest, `clients/dispatch/runners/pyright.ts:83` standalone Pyright, `clients/lsp/server.ts:2277` Python LSP) plus `detectPythonVenv` (`clients/lsp/server.ts:2434`, re-exported at `:26`). All four pass an absolute root and are unaffected by the new optional `homeDir` parameter, which defaults to `os.homedir()`.
- `clients/cargo-manifest.ts` → `hasCargoWorkspaceTable`'s behaviour is unchanged (same body, one indirection); its callers are `clients/cargo-manifest.ts:430,450`, `clients/review-graph/workspace-modules.ts:97`, `clients/lsp/server.ts:2604`. `tests/clients/cargo-manifest.test.ts`, `review-graph-workspace-modules-cargo`, `formatters-rustfmt-edition`, `formatter-unavailable-outcome` and `lsp/server-policy` all pass.
- `clients/test-runner-client.ts` → only the pytest exit-code label; `tests/clients/test-runner-client.test.ts` passes.

Not a hot path: `detectPythonEnvironment` runs once per pytest spawn / per LSP root resolution, not per file or per render. The walk is strictly shorter than before (the deleted ancestor branch removed one `access` probe; the depth cap never fired).

## Observability

No new failure path, ledger record, or unbounded surface. A wrongly-resolved environment surfaces the same way it did before: the spawned interpreter path in the pytest command, and `TestResult.error` for exit codes 2/4 — both bounded and already recorded. The narrowing changes are *removals* of candidates, so their production signal is the generic-`python` fallback already covered by `keeps the generic Python fallback when no project environment exists`. Gap named honestly: there is no per-resolution log line saying *which* candidate won, so a mis-selection in the field is diagnosed from the spawn record rather than from a dedicated event; adding one was out of this PR's scope and is not needed for the tests above.

## Class sweep

**Shapes cited:** AGENTS.md shape 13 (an ambient/global signal outranking the specific one — F1), the minimalism-ladder / single-source-of-truth rule (F5), the guard-names-its-recurrence rule (F6, #2582), and the "external contracts are fetched, never paraphrased" rule (#2432).

**Upstream contract, fetched not paraphrased.** The uv glob dialect was read from `astral-sh/uv` at pinned SHA `3c979abda4530fe9bf3d92e9bcf5c5575e3b3126` (tag `0.12.10`), `crates/uv-workspace/src/workspace.rs`:

- `is_included_in_workspace` matches `members` with `MatchOptions { require_literal_separator: true, ..MatchOptions::new() }` → case-**sensitive** on every platform, `*`/`?` confined to one path component, no literal-leading-dot requirement, `**` crosses components.
- Patterns are run through `normalize_path` first, so a leading `./` is not part of the pattern (upstream fixture `exclude_package_with_normalized_glob_and_escaped_root` uses `"./packages/./excluded-glob-*"`).
- `WorkspaceExclusions::matches` uses `Pattern::matches_path`, i.e. `MatchOptions::new()` defaults where `require_literal_separator` is **false**.

This corrected two guesses in the round-1 code: `nocase: process.platform === "win32"` (upstream is always case-sensitive) is deleted, and pattern normalization is added, pinned by the new test vector citing the SHA. The `exclude` separator divergence is not expressible in minimatch and is documented as a known limitation in the code, the same way `matchesCargoWorkspacePattern` documents cargo's `**` limitation; it can only under-exclude, and both outcomes fall back to the project's own `.venv`.

**Population sweep — `[table]`-presence checks (name + literal value), whole tree.** `grep -rn 'extractTomlTableSection('` over `clients tools mcp scripts tests index.ts`: 10 sites, all in `clients/cargo-manifest.ts` except the one `clients/python-environment.ts:70` section *extraction* (not a presence check). Verdict: **folded** onto `hasTomlTable`; `hasUvWorkspaceTable` deleted; `grep -rn hasUvWorkspaceTable` now returns only the doc comment recording why it is gone. The existing `tests/clients/cargo-manifest.test.ts` governance sweeps (`no hand-rolled [workspace]-table regex outside cargo-manifest.ts`, `every workspace-table presence check names hasCargoWorkspaceTable`) both pass.

**Population sweep — hand-rolled minimatch options blocks, whole tree.** `grep -rn 'nocase: process.platform'`: was 4 sites, now 3 — `clients/file-utils.ts:587`, `clients/lens-engine.ts:402`, `clients/path-utils.ts:636` (`nameMatchesMarkerGlob`, the shared marker-glob seam). Per-member verdict: all three match a file/entry NAME or an ignore path against project globs, where the win32 filesystem's case-insensitivity is the correct semantics; `python-environment.ts`'s fourth copy matched a *uv* pattern, where upstream is case-sensitive, so it was not a candidate for `nameMatchesMarkerGlob` and is now deleted rather than re-homed. Verdict: **stays distributed**, and the class is smaller by one.

**Population sweep — workspace-member glob matchers, whole tree.** `grep -rn 'declaresMember|DeclaresMember'` and `'WorkspacePattern|workspacePattern'` over `clients tools mcp scripts tests index.ts AGENTS.md`: three members — `clients/lsp/server.ts:2500` `matchesCargoWorkspacePattern` (+ `:2533` `cargoWorkspaceDeclaresMember`, + the private `segmentGlobToRegExpSource` at `:2475`), `clients/python-environment.ts` `isUvWorkspaceMember`, `clients/review-graph/workspace-modules.ts:124` `expandWorkspacePattern`. **Consolidation verdict: stays distributed for now, tracked in #2591** — the reason is the three-way dialect divergence evidenced above, and folding cargo's matcher would change Rust-LSP workspace hoisting inside a Python PR. Recorded in AGENTS.md's "one Cargo.toml reader" paragraph next to the `hasTomlTable` fold, so the next reader does not have to re-derive it.


apmantza:
Orchestration handoff for head `b9168e249fa9f61f76a1e073db42d95eca1c1cef`:

- Fixer: Luna fixer in registered worktree `C:\tmp\pi-lens-2580`.
- Verdict: implementation and local verification complete; branch clean and synchronized.
- Warden state: `CI_PENDING`.
- Required checks: `Lint & type-check` passed; `Unit tests` is running on this head.
- Next owner: orchestrator CI watcher.
- Next action: verify the terminal Unit-tests verdict on this SHA, rerun the warden audit, then delegate a separate pi-lens reviewer.

apmantza:
Warden re-audit for head `b9168e249fa9f61f76a1e073db42d95eca1c1cef`:

- State: `UNREVIEWED`.
- Required CI: `Lint & type-check` and `Unit tests` passed on this exact head.
- Worktree and remote: clean and synchronized.
- Next owner: separate pi-lens reviewer.
- Next action: adversarial review against this SHA, followed by a warden audit that routes approval or correction.

apmantza:
Pi-lens reviewer verdict for exact head `b9168e249fa9f61f76a1e073db42d95eca1c1cef`: **REQUEST_CHANGES**.

## Spec-compliance finding

**HIGH — Explicit uv workspace selection ignores membership and exclusions.**

`clients/python-environment.ts` applies an ancestor workspace `.venv` to every descendant with a `pyproject.toml`, without checking `[tool.uv.workspace].members` or `exclude`. A real uv 0.10.8 probe placed an independent project beneath the workspace but outside its member glob. `uv run --no-sync` selected the nested project's own `.venv`; pi-lens selected the ancestor workspace `.venv`. Pytest, standalone Pyright, and Python LSP can therefore use the wrong interpreter.

Require workspace membership/exclusion semantics, or stop ancestor selection at the nearest independent project, with a real regression for a non-member descendant.

## Standards-compliance findings

None.

Reviewer verification passed focused suites (6 + 188 tests), build, lint, diff check, and ast-grep self-scan. The reviewer worker could not submit its formal review because the integration returned HTTP 403, so the orchestrator recorded the exact-head verdict here.

apmantza:
Reviewer F1 (ancestor uv workspace inherited by non-members): fixed in `8471b830`.

The resolver now parses workspace `members` and `exclude` through the shared TOML seam. It inherits the ancestor workspace environment only for declared, non-excluded members; independent nested projects retain their own environment. Red-first cases failed for a non-member and an excluded member on the reviewed head. Workspace-membership and exclusion mutants reddened those cases. Build, lint, formatting, changelog validation, ast-grep self-scan, and a 331-test matrix pass locally. Required CI is restarting on the corrected head.

apmantza:
Post-correction warden handoff for head `8471b8305553312dafcab391e4d4f4a6d399d7de`:

- State: `CI_PENDING`.
- Reviewer F1 disposition is recorded above.
- Required checks: `Lint & type-check` passed; `Unit tests` is running on this exact head.
- Worktree and remote: clean and synchronized.
- Next owner: orchestrator CI watcher.
- Next action: verify Unit tests on this SHA, rerun the warden, then return the delta to the same pi-lens reviewer.

PR #2584, opened 2026-09-04
test(config): guard tracked control bytes (closes #2571)

apmantza:
## Summary

Closes #2571. Add a byte-level governance sweep over every tracked `*.ts`, `*.mjs`, and `*.md` file. Reject each byte below `0x20` except tab, LF, and CR. Failures name the path, byte, code point, offset, and escaped remediation such as `\^@`. Replace seven existing literal controls with equivalent source escapes. No ast-grep or oxlint rule is added because neither mechanism scans raw bytes across Markdown and binary-classified files.

## Tests

- `tests/config/tracked-control-bytes.test.ts` (new, hardened in review round 2) — scans the real Git-tracked population and filesystem bytes; asserts a non-empty 1,788-file population (1,514 `.ts` / 99 `.mjs` / 175 `.md`); reports NUL, 0x01, 0x1B, and 0x1F with the exact escaped remediation text (backslash-caret-at for NUL, backslash-u0001/u001b/u001f for the others); asserts 0x20 (space) is never flagged; asserts the drop cap is enforced across files, not reset per file; and allows only tab, LF, and CR.
- Mutation proof — `byte < 0x20 && false` reddened the NUL assertion. The restored guard passes. See "Review round 2" below for three additional mutation proofs added in this round.
- `tests/clients/widget-state.test.ts` (edited) — preserves ANSI regex behavior while replacing literal ESC bytes with source escapes.
- `tests/mcp/analyze-cli.test.ts` (edited) — preserves control-stripping behavior while replacing literal BEL/NUL fixture bytes with source escapes.
- Initial real-population run found seven violations. The final sweep scans 1,788 files with zero violations.
- Affected sibling suites — 177 passed.
- Governance suites — 11 passed.
- `npm run build`, `npm run lint`, `oxfmt --check`, `npm run changelog:check`, and `npm run astgrep:self-scan` passed.

## Blast radius

`clients/runtime-turn.ts` changes only the source spelling of the `deferralEntryKey` separator. Runtime behavior and hot-path cost do not change. Fixture edits preserve their runtime bytes. `module_report` was unavailable; static inspection found the local caller and the full tracked population was scanned.

## Class sweep

[...]

Issue #1766, opened 2026-08-19
Resync timeout log blames a server still spawning as slow/wedged


apmantza:
During the 2026-08-20 plegma dogfood session, writing a brand-new JSON file logged:

```
LSP resync timed out after 3000ms; server slow/wedged
```

What actually happened: the write triggered the first spawn of `vscode-json-language-server`. The spawn took 8915ms total. The 3000ms resync deadline expired mid-spawn. The spawn then completed (`lsp spawn json: success`) and the workspace-diagnostic probe ran normally.

The record blames a server that did not exist yet. "Slow/wedged" is a verdict about a running server; a cold start still in progress is a different state, and an operator reading the log cannot tell them apart.

Proposed fix, in order of preference:

1. When the resync deadline expires while the target server's spawn is still in flight, emit a distinct reason (for example `reason: "spawn-in-flight"`) instead of the slow/wedged wording, or suppress the record and let the spawn outcome record tell the story.
2. At minimum, include the server lifecycle state (spawning, running, restarting) in the timeout record so the two cases are distinguishable.

Screen from the house catalog: an empty or timed-out result must distinguish its causes; a log record is only useful if it preserves the discriminating identity of what failed.

Behavior is correct — the session recovered on its own. This is a log-accuracy defect only.

Origin: plegma dogfood monitor, 2026-08-20 22:13 window, single occurrence.

PR #2585, opened 2026-09-04
test: LSP double factory + semantic recurrence sweep (refs #2582)

apmantza:
## Summary

Consolidate `LSPService` test doubles behind `makeLspServiceDouble`
(`tests/support/lsp-service-double.ts`), replace round 1's regex sweep with a
structural detector, and fix the production ordering defect the review found in
`resyncLspFile`.

The recurrence: a test stubs `getLSPService` with a partial hand-rolled object,
production later calls a method it does not have, and the `TypeError` lands in a
swallow-all catch. It has cost three rounds — #1766 F3 (`isSpawnInFlight`),
#2582 F3 (`getAuxiliaryClientsForFile`), and this branch's own first CI red
(`tests/clients/lsp-lazy-liveness.test.ts:88`).

`refs #2582`, not `closes`. The issue's acceptance ("a sweep that reds on a
fresh hand-rolled service double outside the factory") IS met — proven below by
injecting one into a real test file. But 18 hand-rolled doubles in 17 files on
seams this issue did not scope are pinned in a ratchet baseline rather than
migrated, so the consolidation is partial by an explicit maintainer decision.
That remainder is #2592 and is commented on #2582.

### Review round 2 — per finding

**F1 (blocking) — identifier laundering in `cascade-compute.test.ts`.**
Reproduced first. `sed -i 's/makeTouchFileMock()/vi.fn()/g'` on the pre-fix file,
then the round-1 sweep:

```
- []
+ [
+   "tests/clients/cascade-compute.test.ts:217",
+   "tests/clients/cascade-compute.test.ts:254",
...
+   "tests/clients/cascade-compute.test.ts:3390",
+ ]
 ❯ tests/config/lsp-service-double-sweep.test.ts:50:45
 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)
```

28 violations — the alias existed only to keep the sweep green. Fixed: the
`const makeTouchFileMock = vi.fn;` line is deleted and all 28 sites are back to
`vi.fn()`, which is the factory's documented "override only that method" usage
(`tests/support/lsp-service-double.ts`). The detector no longer forbids it (F2).
The sweep's own red proof no longer launders either: it reads real fixture
files.

**F2 (blocking) — the detector is now semantic.**
`tests/support/lsp-double-gate.ts` parses each file with `@ast-grep/napi` (the
repo's existing seam, same as `tests/support/availability-gate.ts`), starts at
the `getLSPService` stub — `vi.mocked(getLSPService).mockReturnValue(x)`,
`mocks.getLSPService.mockReturnValue(x)`, `getLSPService: () => x`,
`getLSPService: vi.fn(() => x)` — and resolves `x` back to the literal it names
through `as`-casts, local `const`s, local factory functions, `vi.fn()` wrappers
and one hop of `holder.current`. A violation is a resolved object that is NOT
seeded from `makeLspServiceDouble`. The method vocabulary is
`Object.keys(makeLspServiceDouble())` — derived from the factory, never
mirrored.

Seam anchoring is load-bearing, not decoration. A first cut matched "any object
literal with two `LSPService` method names" and found **159 objects in 65
files** — most of them fake LSP *clients* in `tests/clients/lsp/*`, built by
`createLSPClient` mocks, which share `getDiagnostics`/`openFile`/
`documentSymbol` with the service and are a different seam.
`tests/fixtures/lsp-double-gate/compliant-lsp-client-double.ts` pins that.

Red proof is a **real fixture file scan**, one file per shape, read and parsed
through the same path as the population walk (`tests/fixtures/lsp-double-gate/`).
Each case asserts the round-1 regex finds nothing AND the new detector does:

```
 ✓ flags shape-a-shorthand.ts (shorthand properties)
 ✓ flags shape-b-plain-async.ts (a plain async stub, no vi.fn at all)
 ✓ flags shape-c-post-hoc.ts (post-hoc property assignment)
 ✓ flags shape-d-multiline.ts (vi + newline + .fn())
 ✓ flags shape-e-alias-laundered.ts (const makeTouchFileMock = vi.fn)
 ✓ does NOT flag compliant-factory-override.ts (focused overrides on a seeded object)
 ✓ does NOT flag compliant-lsp-client-double.ts (a fake LSP client, a different seam)
```

Shape (d) is the earlier reviewer comment's multiline `vi` + newline + `.fn()`;
it is covered by the same walk.

**F3 (blocking) — production reorder in `resyncLspFile`.**
Reproduced with exactly the reviewer's probe (override only
`getAuxiliaryClientsForFile` with a bare `vi.fn()`), as a red-first test on
pre-fix `clients/pipeline.js`:

```
 FAIL  |wall-clock-budget| tests/clients/pipeline-lsp-sync.test.ts > resyncLspFile — bounded pre-dispatch LSP sync > starts the primary touch and still records the stall when auxiliary acquisition throws
AssertionError: expected "vi.fn()" to be called 1 times, but got 0 times
 ❯ tests/clients/pipeline-lsp-sync.test.ts:269:24
 Test Files  1 failed (1)
      Tests  1 failed | 10 skipped (11)
```

Fix: the primary `touchFile` promise is started FIRST; the `#2540` auxiliary
kick-off moves below it and is wrapped in the repo's `void (async () => …)()`
fire-and-forget idiom, so a synchronous throw out of the call becomes a
rejection the existing `.catch` covers. The `?.` deletion #2582 asked for is
kept. The test asserts touch call ordering, `touchFile` called once, no
`"after autofix error"` in dbg, and `lsp_sync_abandoned` with `reason:
"timeout"` still logged — the #1766 F3 invariant, restated for this method.

**F4 (net-count) — migrate the claimed seams, ratchet the rest.**
Migrated in this round (12 files, all previously hand-rolled):
`runtime-session`, `runtime-session-warm` (4 sites), `runtime-session-warmup-
oneshot`, `-prewarm`, `-supersession`, `runtime-tool-call`,
`write-autofix-attachment-message`, `word-index-lifecycle`, `pi-host-contract`,
`formatters-lazy-liveness`, `dispatch/lazy-liveness`,
`dispatch/runners/runner-status-semantics`. All 12 pass (110 tests).

Baselined, not migrated — 18 doubles in 17 files, pinned in
`tests/support/lsp-double-baseline.json`, listed in the Class sweep below with
the reason. Follow-up: **#2592** (`enhancement`, `area:tests`, `priority:p3`).

**F5 — PR body correction.** The round-1 claim "No optional service-method chain
remains solely for partial doubles" was **false** and is retracted.
`clients/pipeline.ts` still carries `typeof lspService.isSpawnInFlight ===
"function"`. It is NOT deleted. Its comment now names the live recurrence (19
doubles in 18 files, the baseline, #2592) and says to re-evaluate when the
baseline reaches zero. Mutation-proven below.

**F6 — factory typing and `omit`.** `overrides` is now
`Partial<Record<keyof LSPService, unknown>>`, so a typo is a compile error:

```
tests/clients/pipeline-lsp-sync.test.ts(228,31): error TS2561: Object literal may only specify known properties, but 'touchfile' does not exist in type 'Partial<Record<keyof LSPService, unknown>>'. Did you mean to write 'touchFile'?
```

Values stay `unknown`: these doubles are deliberately looser than the production
signatures (`touchFile` resolving `{ diags: [] }`, `supportsLSP` ignoring its
argument), so `Partial<LSPService>` would be a value-migration, not a typo
guard — stated in the factory's doc comment. New
`omit?: readonly (keyof LSPService)[]` makes absence first-class;
`pipeline-lsp-sync`'s `delete (service as any).isSpawnInFlight` is now
`makeLspServiceDouble({…}, { omit: ["isSpawnInFlight"] })`.

### The ratchet's admission state space (rounds 3–4)

Round 2's ratchet had exactly the defect it was built to stop, one level up:
**N1** — a brand-new file with a hand-rolled double plus ONE line added to
`tests/support/lsp-double-baseline.json` passed 12/12, with no reason and no
second gate.

Round 3 answered that with a two-part admission gate **plus an exempt tier**
(`ORIGINAL_POPULATION`, "the debt as minted"), and the verifier walked straight
through the tier: fresh double, one baseline line, **one line added to that
Set** → 29/29 green, no reason, no header. Round 3's own table did not list
"edit `ORIGINAL_POPULATION`" as a writer, and that omission is precisely where
the escape lived.

**Round 4's decision: an exempt tier IS a door, so there is no tier.**
`ORIGINAL_POPULATION` is deleted; the seventeen inherited pins are admitted the
same way as anything added tomorrow. One writer fewer, one path in, one path
out.

#### Inputs

| symbol | source | meaning |
| --- | --- | --- |
| `LIVE` | detector walk over `tests/**/*.ts` | `file → hand-rolled double count` today |
| `BASELINE` | `tests/support/lsp-double-baseline.json` | `file → pinned count` — the COUNT ratchet, nothing more |
| `ADMITTED` | the sweep source | `file → reason`; one entry per pin, reason must name an issue |
| `HEADER(f)` | the file itself | a `// lsp-double: <reason>` line that is a REAL comment (F3) |

#### States of a file

| state | definition |
| --- | --- |
| `S-absent` | not in `BASELINE` |
| `S-admitted` | in `BASELINE` ∧ in `ADMITTED` with a valid reason ∧ `HEADER` present — the only passing pinned state |
| `S-map-only` | in `BASELINE` ∧ in `ADMITTED`, no `HEADER` |
| `S-header-only` | in `BASELINE`, `HEADER` present, not in `ADMITTED` |
| `S-bare` | in `BASELINE`, neither |
| `S-dead-admission` | in `ADMITTED` but not in `BASELINE` |

#### Writers

`W1` add a `BASELINE` key · `W2` remove a `BASELINE` key · `W3` add a new test
file carrying a hand-rolled double · `W4` count rises · `W5` count falls ·
`W6` a pinned file stops being flagged · `W7` edit `ADMITTED` · `W8` add or
remove the file's header. (Round 3 had a ninth — edit `ORIGINAL_POPULATION` —
which it failed to write down; there is no such writer now.)

#### Rules

`R1` live file ∉ `BASELINE` → **new-file** · `R2` live > pin → **count-risen** ·
`R3` live < pin → **stale-ceiling** · `R4` pin with no live hit → **vanished-pin** ·
`R5` `BASELINE` key ∉ `ADMITTED` → **never-admitted** ·
`R6` `ADMITTED` key with no real-comment `HEADER` carrying a ≥15-char reason → **missing-header** ·
`R7` `ADMITTED` reason under 15 chars or with no `#NNN` → **reason-not-real** ·
`R8` `ADMITTED` key ∉ `BASELINE` → **dead-admission**.

#### Cell list — every cell has a test

| # | start | writer | verdict | rule | why |
| --- | --- | --- | --- | --- | --- |
| C1 | `S-absent`, unflagged | W3 | **RED** | R1 | the issue's acceptance criterion |
| C2 | `S-absent`, newly flagged | W1 only → `S-bare` | **RED** | R5 | a data edit admits nothing — **N1 and the round-3 bypass, both closed here** |
| C3 | `S-bare` | W7 → `S-map-only` | **RED** | R6 | the map alone is one part of two |
| C4 | `S-bare` | W8 → `S-header-only` | **RED** | R5 | the header alone is the other part of two |
| C5 | `S-map-only` | W8 → `S-admitted` | **PASS** | — | both parts present, reason real |
| C6 | `S-admitted`, empty reason | W7 | **RED** | R7 | a blank reason is not a reason |
| C7 | `S-admitted`, reason without `#NNN` | W7 | **RED** | R7 | an admission must point at tracked work |
| C8 | `S-admitted`, header reason under 15 chars | W8 | **RED** | R6 | a marker is not a reason |
| C9 | the 17 inherited pins, all `S-admitted` | — | **PASS** | — | steady state; asserted by the sweep itself against the real repo |
| C10 | `S-admitted` | W4 | **RED** | R2 | |
| C11 | `S-admitted` | W5 | **RED** | R3 | a stale ceiling silently re-admits regrowth |
| C12 | `S-admitted` | W6 alone | **RED** | R4 | a dead pin reads as a screen |
| C13 | `S-admitted` | W6 + W2 + W7 (+ W8) | **PASS** | — | the burn-down path (#2592) |
| C14 | `S-admitted`, still flagged | W2 alone | **RED** | R1 | deleting a pin does not unflag the file |
| C16/C17 | `S-dead-admission` | W2 without W7, or W7 alone | **RED** | R8 | stale scaffolding |
| C18 | `S-absent` | W8 on an unpinned file | **PASS** | — | deliberate: an inert comment grants nothing (C4 proves it), and a rule here buys nothing R8 does not |
| C19 | `S-admitted` | file deleted, pin kept | **RED** | R4 | no live hit |
| C21 | `S-absent`, unflagged | W1 on a file that is not flagged | **RED** | R4 ∧ R5 | a speculative pin is dead weight twice over |
| C23 | `S-map-only` | W8 with the marker inside a string/template | **RED** | R6 | **F3**: a header must be a real comment |

**Deleted in round 4** (net-count): round 3's C9/C20 covered the exempt tier
and have no referent now, and the hardcoded `migrated` array is gone — a
migrated file that regrows a double is an ordinary `new-file` red, proven by
re-adding one to `runtime-session-warm.test.ts`:

```
+   "tests/clients/runtime-session-warm.test.ts: 1 hand-rolled LSPService double(s) in a file the baseline has never seen — seed from makeLspServiceDouble(), using omit for methods that must be absent; a partial double is how #1766 F3 and #2582 F3 both shipped",
      Tests  1 failed | 27 passed (28)
```

#### N2 — the second machine (the detector's two rules)

The object rule and the post-hoc rule do not share a gate, and round 2's
docstring claimed they did:

| rule | gated by the vocabulary? | probe |
| --- | --- | --- |
| object (any object handed to the `getLSPService` seam that is not seeded from `makeLspServiceDouble()`) | **no**, by design — that IS the recurrence, whatever keys it carries | emptying the vocabulary must leave every object-rule hit unchanged |
| post-hoc (`service.<method> = …` on a binding the seam receives) | **yes**, load-bearing — the property name is the only signal there is | emptying the vocabulary must red shape-c and nothing else |

Round 2 computed `keys` (vocabulary ∩ object keys), stored it on
`HandRolledDouble`, and never gated on it, so emptying the vocabulary left the
whole population scan green and the "derives its vocabulary from the factory"
test asserted the getter rather than the detector. The `keys` plumbing is
deleted, the docstring restated, and both probes above are now tests.

#### Round 3 — reproduction, then result

**N1 reproduced on `977e36d6`.** A brand-new `tests/clients/zz-n1-probe.test.ts`
with a hand-rolled `mockReturnValue({ supportsLSP, touchFile })`, plus one line
in the baseline JSON:

```
 Test Files  1 passed (1)
      Tests  12 passed (12)
```

Twelve of twelve, no reason, no second gate — shape 38 exactly.

**N1 closed.** Same probe against the fix:

```
 FAIL  tests/config/lsp-service-double-sweep.test.ts > … > every pin is either original debt or a two-part admission
+ [
+   "tests/clients/zz-n1-probe.test.ts: pinned in tests/support/lsp-double-baseline.json but never admitted. A pin is not a data edit: add an ADMITTED_AFTER_BASELINE entry naming the issue that tracks it, AND a `// lsp-double: <reason>` header in the file itself. Both parts are required (AGENTS.md shape 38). Or seed from makeLspServiceDouble(), using omit for methods that must be absent.",
+ ]
      Tests  1 failed | 28 passed (29)
```

The escalation, driven through the REAL sweep rather than the unit fixtures —
**C2 → C3 → C5**. Adding only the `ADMITTED_AFTER_BASELINE` entry:

```
+ [
+   "tests/clients/zz-n1-probe.test.ts: admitted in ADMITTED_AFTER_BASELINE but carries no `// lsp-double: <reason>` header. Both parts are required (AGENTS.md shape 38).",
+ ]
```

then adding the file's own header as well:

```
      Tests  29 passed (29)
```

**N2 reproduced on `977e36d6`.** Emptying the vocabulary
(`lspServiceMethodNames()` → `new Set()`):

```
     × flags shape-c-post-hoc.ts (post-hoc property assignment)
     × derives its vocabulary from the factory, not from a hand-copied list
      Tests  2 failed | 10 passed (12)
```

The **population scan stayed green**. So `keys` was dead plumbing, the
docstring's "the vocabulary is a filter on the resolved object" was false, and
the surviving test asserted the getter, not the detector.

**N2 fixed.** `keys` is deleted from `HandRolledDouble`; the object rule is
stated and tested as seam-only; the vocabulary is now a parameter of
`findHandRolledLspDoubles(source, vocabulary?)` so both properties are provable
through a real seam. The same probe against the fix:

```
     × flags shape-c-post-hoc.ts (post-hoc property assignment)
     × the POST-HOC rule DOES read the vocabulary — there it is load-bearing
     × the vocabulary is derived from the factory, not from a hand-copied list
      Tests  3 failed | 26 passed (29)
```

and the object rule's independence test passes under the same mutation, as it
must. Its own mutation (**M11**, make the object rule consult the vocabulary):

```
     × the OBJECT rule does not read the vocabulary — the seam is the anchor
      Tests  1 failed | 28 passed (29)
```

Failure message updated for zero-key hits: a `getLSPService: () => ({})` is as
much a partial double as one naming twelve methods, so every message now ends
"seed from makeLspServiceDouble(), using omit for methods that must be absent".

**`readCachedDiagnosticsForServers` — checked, and it is real.**
`clients/lsp/index.ts:3393`, `async readCachedDiagnosticsForServers(filePath,
serverIds)`. So it joins the factory defaults and
`tests/clients/lsp/late-auxiliary-findings.test.ts` is **migrated**, not
baselined. The baseline is now **17 files / 18 doubles** — and the burn-down
path proved itself on the way: migrating the file turned the sweep red with
`pinned at 1 but the scan no longer flags it` (cell C12) until the pin was
removed with it (cell C13).

#### Round 3 — mutation matrix for the new machinery

| probe | neutered | reds |
| --- | --- | --- |
| M11 | make the object rule consult the vocabulary | `the OBJECT rule does not read the vocabulary` |
| M12 | delete the `unadmitted-pin` branch (R5) | C2, C4 |
| M13 | delete the header requirement (R6) | C3, C8 |
| M14 | stop requiring `#NNN` in the reason (R7) | C7 |
| M15 | delete the dead-admission rule (R8) | C16/C17 |
| empty vocabulary | — | shape-c fixture, the post-hoc rule test, the derivation test |

Round 2's probes were re-run on this head and still discriminate: M1/M2 (F3
ordering and containment), M4–M7 (detector rules), M9 (the `isSpawnInFlight`
guard), M10 (`omit`).

#### Round 3 — boundary map (the verifier's, recorded as a limit)

Now in the module doc of `tests/support/lsp-double-gate.ts`, so a later reader
treats this as a ratchet and never as a proof. All are false NEGATIVES, the
safe direction:

- `Object.assign(base, { touchFile })` handed to the seam — the resolver
  follows `vi.fn(...)` wrappers and local factory functions, not an arbitrary
  builtin call, so the resulting object is invisible;
- a CLASS INSTANCE (`getLSPService: () => new FakeLspService()`) — a
  `new_expression` resolves to no object literal and the class body is never
  walked;
- a double assembled by a helper in another module (the shape-c fixture leans
  on exactly this, which is what gives the post-hoc rule something only it can
  catch);
- indirection deeper than `MAX_RESOLUTION_DEPTH`, and any method installed
  through a computed key.

#### Round 3 — AGENTS.md

Added recurring-defect **shape 38**, "A guard, ratchet or baseline that a
one-line data edit or a convenient fixture satisfies away", citing #2584 r1
(every control-byte fixture `0x00`, so the `< 0x20` boundary went untested),
#2585 r1 (28 renamed call sites), #2585 r2 (one baseline JSON line admits a
fresh double), and #2583 r2 (mutation-vacuous branches under checked boxes).
Screen: run the cheapest evasion yourself before shipping the guard, and
require a second machine-checked part so admission cannot be a data edit.
Numbered 38 per the orchestrator; 39+ is left for other in-flight PRs.

#### Round 4 — reproduction, then result

**F1 — the PR was DIRTY.** #2583 merged and inserted shape 39 at
`AGENTS.md:329`, the line shape 38 occupied. Merged `origin/master`; the
conflict was one hunk, resolved additively as `37 · 38 · 39` with both texts
byte-intact and no other AGENTS.md edit. The merge also brought #2580's
`clients/python-environment.ts` work, which touches no seam of this PR;
`tests/clients/test-runner-python-environment.test.ts` was added to the
verification set and passes.

**F2 reproduced on `ba53342f`.** Fresh hand-rolled double, one baseline JSON
line, one line in `ORIGINAL_POPULATION`:

```
 Test Files  1 passed (1)
      Tests  29 passed (29)
```

**F2 closed.** `ORIGINAL_POPULATION` is deleted; all 17 inherited files carry
their own `// lsp-double: … #2592` header and an `ADMITTED` entry. The same
probe, with no Set left to edit:

```
 × every pin carries a reason and the file carries its header
+ [
+   "tests/clients/zz-f2-probe.test.ts: pinned in tests/support/lsp-double-baseline.json but never admitted. A pin is not a data edit: add an ADMITTED entry naming the issue that tracks it, AND a `// lsp-double: <reason>` header in the file itself. Both parts are required (AGENTS.md shape 38). Or seed from makeLspServiceDouble(), using omit for methods that must be absent.",
+ ]
      Tests  1 failed | 27 passed (28)
```

The round-3 rationale was also inverted, as the verifier said: keeping a
burned-down file in `ORIGINAL_POPULATION` **permitted** a silent re-pin instead
of preventing one — `lsp/late-auxiliary-findings` was migrated in round 3 and
stayed in the Set, so only the separate hardcoded array caught its re-pin. Both
are gone.

**M16 — delete the whole-baseline admission loop** (the new R5–R7 in one
mutation):

```
     × C2: a bare pin — one baseline JSON line and nothing else — reds
     × C3: the ADMITTED map alone, without the file's header, reds
     × C4: the header alone, without an ADMITTED entry, reds
     × C6: an empty ADMITTED reason reds
     × C7: an ADMITTED reason naming no issue reds
     × C8: a header reason too short to be real reds
     × C23: a `// lsp-double:` line inside a string is not a header
     × admits a file that has vanished from disk only as a problem
      Tests  8 failed | 20 passed (28)
```

**F3 reproduced and closed.** The header was matched on raw file text, so a
`// lsp-double:` line inside a template literal or a plain string admitted a
file without ever being a comment — sweep-kit's own comment/string-laundering
attack, one level below the gate it defends. `admissionHeader` now discriminates
with `stripSource(source, { strings: "keep" })`, which blanks comments in place
and leaves literal contents verbatim: a match that comes back blank was a real
comment. Two real fixture files pin both sides
(`header-inside-a-string.ts`, `header-real-comment.ts`). Reverting to raw-text
matching:

```
     × C23: a `// lsp-double:` line inside a string is not a header
AssertionError: expected [] to deeply equal [ StringContaining{…} ]
      Tests  1 failed | 27 skipped (28)
```

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [x] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [ ] `npm run build:dist` succeeds if I changed code under `clients/` … — not run; the change to `clients/pipeline.ts` is a statement reorder plus comments, and `npm run build` + `npm run lint` (tsc over the whole project) both pass. Flagging it rather than claiming it.
- [x] `package-lock.json` is in sync with `package.json` (no dep changes)
- [x] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there — round 3 adds recurring-defect **shape 38** (a guard a one-line data edit satisfies away), numbered after #2584's shape 37 which is merged in here
- [x] `.changelog/fix-2582-lsp-double.md` has one valid entry — `node scripts/check-changelog-fragments.mjs` → `changelog fragments OK (114 entries in .changelog/)`

### New

- **`tests/support/lsp-double-gate.ts`** (analyser, not a test) — the structural
  detector. Contract seam.
- **`tests/config/lsp-service-double-sweep.test.ts`** — rewritten. Two-sided
  ratchet (new file / risen count / stale ceiling / vanished pin), a named guard
  that the 15 migrated files stay on the factory, five shape fixtures, two
  compliant fixtures, a vocabulary-derivation check, and two factory-contract
  cases (`omit`, focused override).
- **`tests/fixtures/lsp-double-gate/*.ts`** — seven real fixture files plus a
  README. Excluded from `tsconfig.json` and `vitest.config.ts`, so they are
  scanned, never compiled or run.
- **`tests/support/lsp-double-baseline.json`** — the 19/18 pin.
- **`tests/clients/pipeline-lsp-sync.test.ts`** — one new case, "starts the
  primary touch and still records the stall when auxiliary acquisition throws".
  Regression proof for F3; red transcript quoted above.

### Edited

- `tests/clients/cascade-compute.test.ts` — alias deleted, 28 sites back to
  `vi.fn()`.
- `tests/clients/pipeline-lsp-sync.test.ts` — the `delete (service as
  any).isSpawnInFlight` case now uses `omit`.
- The 12 migrated files above — hand-rolled literal replaced with a factory call
  carrying the same overrides; no assertion changed.

### AGENTS.md test-authoring screens

- *ambient-inspection double* — the F3 test drives the real `resyncLspFile`
  through the real `getLSPService` seam with the real `bounded()`/latency path;
  it asserts the emitted `lsp_sync_abandoned` record, not an internal flag.
- *implementation mirror* — the detector's fixtures are files, not strings
  concatenated inside the assertion, precisely so the test cannot agree with
  itself. The vocabulary is read from the factory, so the sweep cannot mirror a
  hand-copy of it.
- *not-throw* — the F3 test asserts a positive outcome (touch called, record
  present, ordering) rather than "did not throw".
- *invisible skip* — `assertNonEmptyScan(..., 200)` fails the walk if it ever
  scans fewer than 200 test sources.

### Red-first and mutation transcripts

**Acceptance criterion — a fresh hand-rolled double in a real file reds the
sweep.** Appended to `tests/clients/pipeline.test.ts`:

```
+ [
+   "tests/clients/pipeline.test.ts: 1 hand-rolled LSPService double(s) in a file the baseline has never seen — seed it from makeLspServiceDouble() (tests/support/lsp-service-double.ts) instead of writing the object by hand; a partial double is how #1766 F3 and #2582 F3 both shipped",
+ ]
 ❯ tests/config/lsp-service-double-sweep.test.ts:154:59
 Test Files  1 failed (1)
```

and the migrated-seam guard on the same probe:

```
 ❯ tests/config/lsp-service-double-sweep.test.ts:178:51
    178|   expect(migrated.filter((file) => file in live)).toEqual([]);
 Test Files  1 failed (1)
```

**Ratchet, both directions** (pin `lsp-navigation` 1→5, add a `gone.test.ts` pin):

```
+ [
+   "tests/clients/gone.test.ts: pinned at 3 but the scan no longer flags it — delete the entry from tests/support/lsp-double-baseline.json",
+   "tests/tools/lsp-navigation.test.ts: fell from 5 to 1 — tighten the pin in tests/support/lsp-double-baseline.json to 1",
+ ]
```

**M1 — hoist the auxiliary kick-off back above the primary touch** (F3 ordering):

```
  [
-   "touchFile",
    "getAuxiliaryClientsForFile",
+   "touchFile",
  ]
 ❯ tests/clients/pipeline-lsp-sync.test.ts:281:17
 Test Files  1 failed (1)
```

**M2 — remove the async-IIFE containment, keep the order** (F3 containment):

```
AssertionError: expected 'LSP resync after autofix error: TypeE…' not to contain 'after autofix error'
Received: "LSP resync after autofix error: TypeError: Cannot read properties of undefined (reading 'catch')"
 ❯ tests/clients/pipeline-lsp-sync.test.ts:271:22
 Test Files  1 failed (1)
```

**M7 — neuter the object-literal rule in the detector:**

```
     × no new file, no risen count, no stale pin
     × flags shape-a-shorthand.ts (shorthand properties)
     × flags shape-b-plain-async.ts (a plain async stub, no vi.fn at all)
     × flags shape-d-multiline.ts (vi + newline + .fn())
     × flags shape-e-alias-laundered.ts (const makeTouchFileMock = vi.fn)
      Tests  5 failed | 5 passed (10)
```

**M4 — delete the post-hoc-assignment rule:**

```
     × flags shape-c-post-hoc.ts (post-hoc property assignment)
AssertionError: expected [] to not deeply equal []
      Tests  1 failed | 4 passed | 5 skipped (10)
```

The first attempt at M4 stayed GREEN, because `mockReturnValue(service as
never)` hands the seam an `as_expression` and the binding collector only looked
at bare identifiers — the post-hoc rule was unreachable. That is exactly the
"ticked box, dead guard" failure the mutation rule exists to catch. Fixed
(`bareIdentifier()`), and the shape-c fixture now sources its stub from a second
module so only the post-hoc rule can catch it.

**M5 — stop resolving identifier bindings:**

```
     × flags shape-c-post-hoc.ts (post-hoc property assignment)
      Tests  1 failed | 4 passed | 5 skipped (10)
```

**M6 — stop honouring factory seeding:**

```
     × does NOT flag compliant-factory-override.ts (focused overrides on a seeded object)
      Tests  1 failed | 1 passed | 8 skipped (10)
```

**M8 — delete `assertNonEmptyScan` and empty the walk.** Honest result: with the
floor deleted the suite still reds, because every baseline pin then reads as
"vanished". Restoring the floor gives the specific message:

```
Error: #2582 LSP service-double test walk: scanned/matched 0, below the declared floor of 200. An empty sweep must fail, not read as clean — if the target genuinely went away, delete the sweep instead of letting it pass on nothing.
```

The floor is therefore belt-and-braces over the ratchet's own vanished-pin rule,
kept because `sweep-kit` requires it and its message names the failure directly.

**M9 — delete the `typeof lspService.isSpawnInFlight === "function"` guard**
(F5; the guard stays, this only proves it is live):

```
     × degrades to the old timeout wording, without throwing, when the service lacks isSpawnInFlight
AssertionError: expected 'LSP resync after autofix error: TypeE…' to contain 'slow/wedged'
```

**M10 — neuter `omit`:**

```
     × omit leaves the named method ABSENT, not stubbed
AssertionError: expected true to be false // Object.is equality
      Tests  1 failed | 11 skipped (12)
```

M10's first run was GREEN against `pipeline-lsp-sync` alone, because the
factory's `isSpawnInFlight` default returns `false` and "absent" and
"present-returning-false" produce the same record. The composite probe
(neuter `omit` AND delete the guard) turns M9's red back to `Tests 11 passed`,
which proves `omit` is what makes M9 red — and the direct factory-contract case
above now reds on the mutation on its own.

### Test assessment

Every test file this PR touches, what it uniquely pins, and removal candidates:

- `tests/config/lsp-service-double-sweep.test.ts` — the whole file is replaced.
  Round 1's `findHandRolledTouchFileDoubles` and its "red-proves a fresh
  hand-rolled double" case are **deleted**: the case asserted a
  string-concatenated fixture against a regex that agreed with it by
  construction, which is the vacuous shape AGENTS.md names, and the reviewer's
  F1 showed the whole detector was being kept green by renaming a variable. The
  survivors that cover the same ground on real files are the five
  `flags shape-*` cases and the acceptance probe quoted above. No other case in
  the file predates this PR.
- `tests/clients/cascade-compute.test.ts` — pins cascade compute over LSP
  diagnostics. No case removed; the diff is the alias deletion and 28
  `vi.fn()` restorations. Swept the file for vacuous cases while probing: none
  found (every case reds on at least one of the M-mutations or on a cascade
  behavior change).
- `tests/clients/pipeline-lsp-sync.test.ts` — pins the bounded pre-dispatch LSP
  sync: touch, abort, timeout, spawn-in-flight discrimination, auxiliary
  overlap, and now auxiliary containment. One case added, one edited to use
  `omit`. Nothing removed: the two `isSpawnInFlight` cases discriminate
  different reasons (`spawn-in-flight` vs `timeout`) and the third exercises
  absence — M9 reds only the third, so none is redundant.
- The 12 migrated files — each keeps every case and assertion it had; only the
  double's construction changed. All 12 green (110 tests).

## Blast radius

`module_report` was not run (the tool is unavailable in this worktree — recorded
explicitly rather than claimed). Manual search: `resyncLspFile` has one
production caller, `clients/pipeline.ts`'s own post-autofix path, and is
exercised by `tests/clients/pipeline-lsp-sync.test.ts` and
`tests/clients/pipeline.test.ts` (both green).

The production change is a statement reorder inside `resyncLspFile` plus one
comment correction. No signature, schema, telemetry record, parser, runner
registry or durable format changes. It is on the per-edit path, but the reorder
strictly *reduces* work before the primary `didChange` write goes out — the
auxiliary kick-off's synchronous portion (a `Set` allocation and a call) now
happens after it, not before. No measurable cost delta expected; none measured.

Everything else is test-only: one new support analyser, one rewritten sweep,
one baseline JSON, seven fixture files, and 14 edited test files.

Merge-order note: this branch merges `origin/master` at `dbdc2f84`, which brings
in #2584 (AGENTS.md shape 37 + `tests/config/tracked-control-bytes.test.ts`).
That sweep passes here. No other open PR touches `clients/pipeline.ts`,
`tests/support/lsp-service-double.ts` or `tests/config/lsp-service-double-sweep.test.ts`.

## Observability

**Corrected from round 1** (which said "No production telemetry changes" — true
of the record set, misleading about the seam). This PR changes the ORDER of
operations on the `resyncLspFile` seam in production. The record that proves it
works is `logLatency({ type: "phase", phase: "lsp_sync_abandoned", … })` in
`clients/pipeline.ts` (surfaced in `~/.pi-lens/latency.log`): before this
change, a throw out of auxiliary acquisition suppressed that record entirely
and the stall went completely silent; after it, the record is emitted with
`metadata.reason` of `timeout` / `spawn-in-flight` / `aborted` as before. The
F3 test asserts exactly that record. No new record kinds, no schema change.

## Class sweep

**Defect class:** a partial hand-rolled test double for a production service,
whose absent method throws into a swallow-all catch — AGENTS.md shape 14
(a double that is not production-faithful on the axis under test) compounded by
shape 10 (a guard that reads as clean because nothing can make it fire).
The sweep-evasion half is `sweep-kit.ts`'s own attack catalogue:
comment/string laundering and identifier laundering.

**Whole-tree grep for the symbol NAME and literal VALUE of everything
introduced** (`makeLspServiceDouble`, `makeLspServiceDouble(`,
`findHandRolledLspDoubles`, `lsp-double-baseline.json`, `omit:`,
`makeTouchFileMock`), across `clients/`, `tools/`, `mcp/`, `scripts/`,
`scripts/lib/`, `tests/`, `tests/support/`, `index.ts`:

- `makeTouchFileMock` — 0 occurrences outside the fixture that pins it as an
  attack and the two doc comments that name it.
- `findHandRolledTouchFileDoubles` (round 1's export) — 0 occurrences; deleted
  with its only consumer.
- No parallel roster of `LSPService` method names exists anywhere: the
  vocabulary is `Object.keys(makeLspServiceDouble())`, so there is nothing to
  drift. This is the specific thing #2550 got wrong (a byte-identical
  `REQUIRED_CHECKS` in `scripts/lib/`); `scripts/` and `scripts/lib/` were
  grepped and contain no LSP-service double or method roster.
- Production `getLSPService` callers were checked for other optional-chained
  service methods: `clients/pipeline.ts:~1145`'s `typeof
  lspService.isSpawnInFlight === "function"` is the only one left, and it is
  kept deliberately (F5) with its recurrence named in the comment.

**Population sweep, per-member verdict.** 30 files carried hand-rolled doubles
at the seam before this round (34 doubles). Per-member:

- **Migrated (12 files, 15 doubles):** `runtime-session`,
  `runtime-session-warm` (×4), `runtime-session-warmup-oneshot`,
  `runtime-session-warmup-prewarm`, `runtime-session-warmup-supersession`,
  `runtime-tool-call`, `write-autofix-attachment-message`,
  `word-index-lifecycle`, `pi-host-contract`, `formatters-lazy-liveness`,
  `dispatch/lazy-liveness`, `dispatch/runners/runner-status-semantics`. These
  are the pipeline and runtime-session seams the factory claims.
- **Baselined (18 files, 19 doubles),** pinned in
  `tests/support/lsp-double-baseline.json` and tracked in **#2592**:
  `actionable-warnings-bounds`, `actionable-warnings-deferred-bounds`,
  `actionable-warnings-history`, `actionable-warnings-lsp-cache`,
  `actionable-warnings`, `dispatch/runners/pyright-environment`,
  `lsp-document-symbols` (×2), `lsp/late-auxiliary-findings`, `mcp/analyze`,
  `mcp/session-context-eviction`, `mcp/session-test-findings-retire`,
  `mcp/session`, `warm-attach-confirmation`, `tools/lsp-diagnostics-cache`,
  `tools/lsp-diagnostics-inferred-project`,
  `tools/lsp-diagnostics-per-server-concurrency`, `tools/lsp-diagnostics`,
  `tools/lsp-navigation`.
  **Reason they stay:** different seams — the tools layer, the MCP session
  layer, the actionable-warnings deferred path and one dispatch runner — not
  the pipeline/runtime-session paths #2582 scoped. Migrating them needs a
  per-seam read of what each double's absent methods are load-bearing for, and
  the factory's defaults could silently switch a code path that relies on a
  method's absence. Doing that blind in this round is how the first head shipped
  CI red.
- **Not in the population:** the ~65-file `tests/clients/lsp/*` family the
  vocabulary-only detector flagged. Verdict: NOT this class. They double
  `createLSPClient`, a different seam with its own factory story; pinned as a
  compliant fixture so a future widening cannot quietly absorb them.

**Consolidation verdict:** fold onto one seam — `makeLspServiceDouble` — and the
fold is enforced, not advisory: the ratchet is two-sided, so #2592 can be burnt
down file by file and a stale pin fails as loudly as a new violation.

## Non-goals

No production LSP lifecycle redesign, no prototype-completeness guard, no
migration of the `createLSPClient` doubles, no `Partial<LSPService>` value-type
migration of the factory's overrides.

AGENTS.md sections consulted: minimalism ladder; recurring defect shapes 7, 10,
14, 16, and the 28–36 pre-push screen; Test requirements; Test assessment and
removal; sweep authoring (`tests/support/sweep-kit.ts` attack catalogue).


apmantza:
Orchestration handoff for head `fd3879d614faff3ba7c9305737db2b1158a5fa12`:

- Fixer: Luna fixer in registered worktree `C:\tmp\pi-lens-2582`.
- Verdict: implementation and local verification complete; branch clean and synchronized.
- Warden state: `CI_PENDING`.
- Required checks: `Lint & type-check` passed; `Unit tests` is running on this head.
- Next owner: orchestrator CI watcher.
- Next action: verify Unit tests on this SHA, rerun the warden audit, then delegate a separate pi-lens reviewer.

apmantza:
Warden/CI transition for head `fd3879d614faff3ba7c9305737db2b1158a5fa12`:

- State: `CI_FAILED`.
- Classification: real test failure, not an exit-137 infrastructure kill. The log contains an assertion failure and `[mem-watch] done. exitCode=1`.
- Failure: `tests/clients/lsp-lazy-liveness.test.ts:88` expected `touchFile` to be called.
- Next owner: original Luna fixer.
- Next action: reproduce and diagnose on the branch, compare master under equivalent conditions if flake is suspected, push the smallest correction, and re-run exact-head verification.

apmantza:
F1 (`tests/clients/lsp-lazy-liveness.test.ts:88`): fixed in `36ae3d73`.

The initial literal census missed an alias-based partial `LSPService` double. The strict `getAuxiliaryClientsForFile` call threw before `touchFile`. The fixture now uses `makeLspServiceDouble`. Focused red reproduced on `fd3879d6`; master and corrected head pass. The corrected 23-file lane passes 248 tests, and the acceptance grep still finds only the factory. Required CI is restarting on the new head.

apmantza:
Post-correction warden handoff for head `36ae3d7376eaddf2598b81f6f65c9ac7e67e6b7c`:

- State: `CI_PENDING`.
- Correction: F1 is fixed and disposition is recorded above.
- Required checks: `Lint & type-check` passed; `Unit tests` is running on this exact head.
- Worktree and remote: clean and synchronized.
- Next owner: orchestrator CI watcher.
- Next action: verify Unit tests on this SHA, rerun the warden, then trigger the same separate pi-lens reviewer.

apmantza:
Warden re-audit for corrected head `36ae3d7376eaddf2598b81f6f65c9ac7e67e6b7c`:

- State: `UNREVIEWED`.
- Required CI: `Lint & type-check` and `Unit tests` passed on this exact head.
- Worktree and remote: clean and synchronized.
- F1 disposition is recorded.
- Next owner: separate pi-lens reviewer.
- Next action: adversarial exact-head review, followed by a warden audit that routes approval or correction.

apmantza:
Pi-lens reviewer verdict for exact head `36ae3d7376eaddf2598b81f6f65c9ac7e67e6b7c`: **REQUEST_CHANGES**.

## Spec-compliance finding

**P2 — The recurrence sweep misses multiline `vi.fn()` service doubles.**

`tests/config/lsp-service-double-sweep.test.ts` requires no whitespace between `vi` and `.fn`. A fully hand-rolled `touchFile: vi` followed by a newline and `.fn()` produces zero matches, so the exact recurrence can evade the guard through formatting. Three current multiline occurrences exist in `cascade-compute.test.ts`; those objects spread the factory and are complete, but they prove the detector's formatting gap.

Make detection multiline-resistant or syntax-aware while allowing focused overrides on factory-spread objects. Add a synthetic multiline hand-rolled-double red proof. Do not prohibit legitimate factory overrides merely to simplify the regex.

## Standards-compliance findings

None.

Build, lint, all 23 changed suites (248 tests), governance sweeps, and required CI pass on the exact head. GitHub rejected the worker's formal REQUEST_CHANGES because the authenticated account owns the PR, so the orchestrator recorded the reviewer verdict here.

Issue #2587, opened 2026-09-06
packaging: pi.skills manifest path (../../skills) escapes the published package — the 5 shipped skills never load from npm installs


raphaelbahat:
### Summary

The published package manifest declares:

```json
"pi": {
  "extensions": ["./dist/index.js"],
  "skills": ["../../skills"]
}
```

`../../skills` is a **monorepo-relative path**: it is correct in the source repository (where the package lives at `packages/pi-lens/` and the skills live at the repo root), but it escapes the **published npm package**, so the five skills that the tarball actually ships never register on any npm-installed copy.

### Evidence

- The npm tarball **ships the skills at the package root** (per the jsDelivr flat file listing for `pi-lens@4.1.3`):
  - `/skills/pi-lens-ast-grep/SKILL.md`
  - `/skills/pi-lens-lsp-navigation/SKILL.md`
  - `/skills/pi-lens-write-ast-grep-rule/SKILL.md` + `reference.md`
  - `/skills/pi-lens-write-tree-sitter-rule/SKILL.md`
- The manifest path resolves **outside the package** in every standard install layout:
  - Global install: `~/.pi/agent/npm/node_modules/pi-lens/../../skills` → `~/.pi/agent/npm/skills/` (does not exist)
  - Managed extension caches (e.g. Outfitter's): `<cache>/npm/node_modules/pi-lens/../../skills` → `<cache>/npm/skills/` (does not exist)
- Observed: the extension itself (`dist/index.js`) loads and functions, but none of the five skills appear in sessions on npm-installed copies.

### Root cause

The manifest path was authored for the monorepo layout (`packages/pi-lens/../../skills` → repo-root `skills/`) and was not rewritten for the published artifact — the same monorepo-relative leak as `../../` paths in other package fields.

### Suggested fix

Point the manifest at the shipped directory (self-contained):

```json
"pi": {
  "extensions": ["./dist/index.js"],
  "skills": ["./skills"]
}
```

The tarball already ships `skills/` at the package root, so this is a one-line manifest change plus release.

### Environment

- `pi-lens@4.1.3` (latest), installed via `pi install npm:pi-lens@4.1.3` and via managed extension caches.
- pi-coding-agent 0.85.0.

apmantza:
Triage: confirmed, thank you — behaviour and fix are right; the root cause is slightly different from the monorepo story.

[...]

Issue #335, opened 2026-06-25
Static ESM imports of npm packages fail when loaded by pi Bun-compiled binary


MzaGuille:
### What happened?
pi-lens static ESM imports of npm packages (`typescript`, `minimatch`, `typebox`, `js-yaml`, `@earendil-works/pi-tui`, etc.) fail with `ResolveMessage: Cannot find package 'X'` when pi-lens is loaded by pi's Bun-compiled binary. This affects ALL npm imports in `dist/clients/*.js`, `dist/tools/*.js`, `dist/mcp/*.js`, etc.

The error is not specific to dynamic vs static imports — it also affects static top-level `import { minimatch } from "minimatch"` in `file-utils.js`.

**Root cause:** pi v0.80.x (possibly earlier) is compiled via `bun build --compile`. Buns compiled binary uses a virtual filesystem (`bunfs`) for module resolution. When pi-lens is loaded as an extension from `~/.pi/agent/npm/node_modules/pi-lens/dist/index.js`, Buns resolver looks for npm packages starting from the importing files directory and walking up. However, in a Bun-compiled binary, this upward traversal does not reliably reach the extensions `node_modules/` directory — the resolver is constrained to the bunfs virtual filesystem.

**Confirmed:** The npm packages ARE installed locally via `npm install --production --install-strategy=nested` inside `pi-lens/node_modules/`, but Buns resolver still reports `Cannot find package`.

### Steps to reproduce
1. Install pi-lens: `pi install npm:pi-lens` (on pi v0.80.x)
2. Start pi: any command triggers pi-lens extension loading
3. Observe errors: `ResolveMessage: Cannot find package 'typescript'`, `Cannot find package 'minimatch'`, etc.

Alternatively, run pi with `pi -e 'console.log(1)'` and check stderr for extension loading errors.

### Expected behavior
pi-lens should load without `ResolveMessage` errors when its npm dependencies are properly installed in its `node_modules/` directory.

### Workaround (confirmed working)
Create directory junctions from each subdirectorys `node_modules` to the local package installation:

```powershell
# In pi-lens root
$src = "$(pwd)\node_modules"
foreach ($sub in @('clients', 'tools', 'mcp', 'commands')) {
    $target = "dist\$sub\node_modules"
    New-Item -ItemType Directory -Path $target -Force | Out-Null
[...]

Issue #257, opened 2026-06-20
Align read-guard/edit-autopatch with host edit-tool (vendor edit-diff normalization) + pi SDK integration survey


apmantza:
## Context

Bumped the pi host SDK (`@earendil-works/pi-coding-agent`) `^0.79.4 → ^0.79.8`, which now exports edit/diff primitives from `core/tools/edit-diff` (`generateDiffString`, `generateUnifiedPatch`, `normalizeForFuzzyMatch`, `fuzzyFindText`, `detectLineEnding`/`normalizeToLF`/`restoreLineEndings`, and `Edit`/`FuzzyMatchResult`/`AppliedEditsResult`/`EditToolInput` types). This issue surveys the **full public API** for integration opportunities, led by the read-guard / edit-autopatch finding.

## Hard constraint (shapes every item below)

pi installs extensions with `npm install --omit=dev`, so the host SDK is **NOT in `node_modules` at runtime** — and pulling it in as a real dependency drags the host's transitive tree (provider SDKs, Windows MAX_PATH breakage on `pi update`). So **every SDK import is type-only** (cf. the inlined `isToolCallEventType`).

**Consequence:** the runtime functions (`normalizeForFuzzyMatch`, `generateUnifiedPatch`, EOL helpers) **cannot be statically imported**. Every "use the SDK function" item is really **"vendor/port the ~50 lines so pi-lens matches host behaviour,"** pinned to a host-version floor with a sync test. The value is **behavioural alignment, not code reuse.** Type-only imports (Edit/EditToolInput/…) are free and don't hit this.

---

## #1 — LEAD (verified): the read-guard normalizes `oldText` DIFFERENTLY than the host applies it

The guard is a gate in front of the host's edit tool, but resolves `oldText`→range with a *weaker* normalizer than the host's `normalizeForFuzzyMatch` / `fuzzyFindText` (`edit-diff.js:30-91`). **Confirmed by reading both sides:**

| Transform | Host `normalizeForFuzzyMatch` | pi-lens `normalizeContent` (`read-guard-tool-lines.ts:121`) / `normalizeOldTextForMatch` (`oldtext-autopatch.ts:19`) |
|---|---|---|
| `NFKC` normalize | yes (first) | **no** |
| Strip trailing WS / line | yes | yes |
| Smart quotes to ASCII | yes (baseline) | **no** baseline (only Tier C autopatch) |
| Unicode dashes to `-` | yes (baseline) | **no** baseline (only Tier C) |
[...]

PR #2588, opened 2026-09-06
fix(packaging): widen pi-tui peer range to cover 0.85.x (closes #2586)

### LINKED ISSUE #2586: Published pi-tui peer range excludes Pi 0.85.x
### What happened?

pi-lens 4.1.3 declares `@earendil-works/pi-tui: ^0.84.1`, which excludes Pi 0.85.1's host TUI. The installed package tree contains TUI 0.85.0 and `npm ls --omit=dev --all --json` reports that peer as invalid (exit 1).

Observed on macOS 26.6.2 (25G83), arm64, Node 22.22.3, npm 10.9.8, and Pi 0.85.1. This is a package compatibility declaration issue, not a demonstrated startup failure: an offline Pi catalog check succeeded. The semver mismatch is platform-independent; other operating systems were not tested.

### Steps to reproduce

1. Run `npm view pi-lens@4.1.3 peerDependencies --json`. Observe the TUI range `^0.84.1`.
2. In an existing Pi package tree containing pi-lens 4.1.3 and TUI 0.85.x, run `npm ls --omit=dev --all --json` and inspect the invalid TUI peer entry. Other host-supplied core peers can be reported separately; they are not this issue.

### Expected behavior

Publish a TUI peer range matching tested host versions, or document a concrete incompatibility with Pi 0.85.x. Acceptance: a regression check verifies the declared range accepts each supported, tested TUI version. Do not broaden the range without compatibility testing.

### pi-lens version

4.1.3 (current npm latest at investigation time)

### Relevant logs

Observability: npm's dependency-tree report exposes the invalid peer; a corrected supported pairing should remove this specific entry. No new runtime telemetry is needed for a manifest-range defect. Raw machine logs are omitted for privacy.

Manifest: https://github.com/apmantza/pi-lens/blob/v4.1.3/package.json
Release: https://github.com/apmantza/pi-lens/releases/tag/v4.1.3
Open and closed searches for `pi-tui 0.85` and `pi-tui compatibility` found no equivalent report.

### Area

area:config

apmantza:
## Summary

`peerDependencies["@earendil-works/pi-tui"]` was `^0.84.1`. On a pre-1.0
version, npm's caret only floats the patch (`>=0.84.1 <0.85.0`), so a real
pi-coding-agent/pi-tui 0.85.x host was excluded by declaration — even though
the nightly real-pi compat smoke already runs green against pi-coding-agent
0.85.1, and pi-tui 0.85.1 still exports every symbol `clients/deps/pi-tui.ts`
consumes (`Text`, `truncateToWidth`, `visibleWidth`). No adapter code needed
to change; this is purely a declaration-range defect.

Fix: widen the peer range to `^0.84.1 || ^0.85.0` (explicit about exactly the
two minors this repo has compat evidence for — see "Peer range syntax"
below), bump `devDependencies` for both `@earendil-works/pi-coding-agent` and
`@earendil-works/pi-tui` to `^0.85.1` and update `package-lock.json` via a
real `npm install` against the registry (lockfile now pins 0.85.1 for both),
and add a regression test in `tests/packaging.test.ts` that asserts the
declared range accepts every tested host version, without asserting
acceptance of anything untested.

Every acceptance criterion in #2586 is met.

Closes #2586

### Peer range syntax

Chose `^0.84.1 || ^0.85.0` over `>=0.84.1 <0.86.0`. The `||` form only
endorses the two minors this repo actually has compat evidence for (the
0.84.x line already shipped against, plus 0.85.x verified by the nightly
compat smoke + the export check above); a `>=`/`<` span would silently
accept every 0.84.x patch AND every 0.85.x patch AND ALSO implicitly nothing
past 0.86.0 — which reads as fine here, but the pattern generalizes badly:
next time compat is verified for 0.86.x, an `>=<` span invites bumping just
the upper bound without re-verifying the newly-included minor, while `||`
forces a deliberate `|| ^0.86.0` addition that mirrors the devDependency bump
and the tested-versions list in the new test.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:installer
- [x] area:tests

## Checklist

- [x] The change has tests (regression test for the declared-range defect)
- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds if I changed code under `clients/`, `commands/`, `tools/`, or `index.ts` — N/A by that rule (no files under those directories changed), but verified anyway after merging master's #2594 fix — see "Merge with master" below. `npm run build` (tsc type-check) passes clean.
- [x] `package-lock.json` is in sync with `package.json` (ran `npm install` after each dep change; `tests/scripts/check-lockfile-sync.test.ts` passes)
- [ ] `AGENTS.md` is updated — not needed; no new convention, invariant, or command introduced
- [x] `.changelog/2586-widen-pi-tui-peer-range.md` has one valid entry (section: Fixed)

## Tests

**`tests/packaging.test.ts`** — new `describe("pi-tui peer range covers every tested host version (#2586)")` block, four new `it`s:
- `"lists at least one tested version to guard"` — guards the guard (an emptied `testedVersions` would make the per-version loop vacuously pass).
- `"declares a peer range"` — sanity check the field exists.
- one `it` per tested version (`semver.satisfies` against the declared range) — the actual regression proof. `testedVersions` is `[lockVersion, "0.85.1"]` deduped, where `lockVersion` is read from `package-lock.json` at test time (`node_modules/@earendil-works/pi-tui`'s installed version), so the list cannot silently drift from what the suite actually ran against; `"0.85.1"` is also named literally per the issue's ask, though after the devDependency bump it coincides with the lockfile-derived entry.
- `"does not broaden acceptance past a tested minor (0.86.0 stays out)"` — asserts `semver.satisfies("0.86.0", peerRange)` is `false`, so the fix cannot silently over-broaden past what is actually tested.

This is a regression proof for the exact defect in #2586 (screen: "wrong-layer pin" avoided — the test reads the real `package.json`/`package-lock.json` off disk rather than re-declaring the range as a literal).

Added `semver` + `@types/semver` as devDependencies (grepped the repo first — neither was previously declared or used anywhere).

### Test assessment

`tests/packaging.test.ts` — this PR adds one new `describe` block; nothing existing in the file becomes redundant. No removal candidates.

### Red-run proof (pre-fix `^0.84.1`, verbatim)

Committed the test + fix first, then reverted only `peerDependencies["@earendil-works/pi-tui"]` to `^0.84.1` (leaving the 0.85.1 lockfile/devDependency bump in place) and reran:

```
 ❯ |default| tests/packaging.test.ts (36 tests | 1 failed | 32 skipped) 7ms
     × accepts tested host version 0.85.1 5ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  |default| tests/packaging.test.ts > pi-tui peer range covers every tested host version (#2586) > accepts tested host version 0.85.1
AssertionError: peerDependencies["@earendil-works/pi-tui"] (^0.84.1) must accept 0.85.1: expected false to be true // Object.is equality

- Expected
+ Received

- true
+ false

 ❯ tests/packaging.test.ts:354:6
    352|     semver.satisfies(version, peerRange ?? ""),
    353|     `peerDependencies["@earendil-works/pi-tui"] (${peerRange}) must ac…
    354|    ).toBe(true);

 Test Files  1 failed (1)
      Tests  1 failed | 3 passed | 32 skipped (36)
```

Restored via `git checkout -- package.json` (tests + fix already committed); re-ran green (32 passed | 4 skipped).

### Mutation-proof transcripts (both new guards neutered, both red)

**Guard 1 — the range itself (neutered by broadening `package.json`'s peer range to `>=0.84.1 <0.87.0`, i.e. simulating a broadened-past-tested-versions regression):**

```
 ❯ |default| tests/packaging.test.ts (36 tests | 1 failed | 35 skipped) 7ms
     × does not broaden acceptance past a tested minor (0.86.0 stays out) 6ms

AssertionError: expected true to be false // Object.is equality
 ❯ tests/packaging.test.ts:363:55
    363|   expect(semver.satisfies("0.86.0", peerRange ?? "")).toBe(false);
       |                                                       ^

 Test Files  1 failed (1)
      Tests  1 failed | 35 skipped (36)
```

**Guard 2 — "lists at least one tested version to guard" (neutered by forcing `testedVersions` to `[]` in the test file):**

```
 ❯ |default| tests/packaging.test.ts (35 tests | 1 failed | 34 skipped) 4ms
     × lists at least one tested version to guard 3ms

AssertionError: expected 0 to be greater than 0
 ❯ tests/packaging.test.ts:339:33
    339|   expect(testedVersions.length).toBeGreaterThan(0);
       |                                 ^

 Test Files  1 failed (1)
      Tests  1 failed | 34 skipped (35)
```

Both mutations reverted; `git status`/`git diff` confirmed no stray changes; re-ran green.

## Blast radius

`package.json` (peerDependencies + devDependencies + allowScripts), `package-lock.json`, `.github/workflows/ci.yml` (three call sites), `scripts/supply-host-provided-deps.mjs`, `clients/host-edit-normalize.ts` (SDK floor constant only — the vendored normalization logic itself is unchanged), `tests/packaging.test.ts`, `tests/clients/host-edit-normalize-sync.test.ts`, `tests/clients/flake-shape-ratchet.test.ts` + `tests/support/flake-shape-baseline.json` (admission bookkeeping), `vitest.config.ts` (wallClockBudgetInclude), a new `tests/scripts/supply-host-provided-deps.test.ts`, and the `.changelog/` fragment. No production module under `clients/` other than the one-constant `host-edit-normalize.ts` change, and no `tools/`, `mcp/`, or `index.ts` change — `clients/deps/pi-tui.ts` (the only runtime consumer of the pi-tui import) is unchanged, since pi-tui 0.85.1 still exports the same three symbols. Blast radius is: (1) the published npm package's declared peer-dependency contract — any pi host resolving `@earendil-works/pi-tui@0.85.x` now installs cleanly; (2) `ci.yml`'s three "supply host-provided packages" steps, which now correctly install a space-containing peer range instead of failing; (3) local/CI installs against this lockfile now get libc-correct optional-binary selection. No hot path touched.

## Observability

Not applicable — this is a static package-manifest declaration, not runtime behavior with its own log/ledger record. The pre-existing nightly real-pi compat smoke (cited in the issue triage) is the observability signal that host 0.85.1 actually works; this PR just stops the declared range from contradicting that evidence.

## Class sweep

Defect shape: a hand-written semver range on a published dependency that undercounts the versions actually compatible/tested (a form of the "hard-coded value that mirrors a fact which can drift" class). Swept the whole tree, not just `clients/`:

- `grep -rn "0\.84"` across `.mjs`/`.ts`/`.js`/`.json` (excluding `node_modules`, `package-lock.json`, `dist/`) — the only hits after this change are `package.json`'s own now-correct range and this PR's own test/comments. Nothing else in `scripts/`, `scripts/lib/`, `mcp/`, `clients/`, or fixtures hard-codes the old `0.84` bound.
- `grep -rl "pi-tui"` across `.mjs`/`.ts`/`.js` (excluding `node_modules`/`dist`) found: `scripts/check-prod-install-shape.mjs`, `scripts/supply-host-provided-deps.mjs`, `scripts/install-selftest.mjs`, `scripts/lib/host-provided-deps.mjs`, `tools/render-compact.ts`, `clients/tool-render.ts`, `clients/widget-state.ts`, `clients/test-runner-delivery.ts`, `clients/tui-fit.ts`, `clients/turn-summary-render.ts`, `clients/deps/pi-tui.ts`, plus their test files. **Correction (review F1): the round-1 sweep's conclusion here was wrong.** None of these hard-code a version NUMBER, true — but `scripts/supply-host-provided-deps.mjs` dynamically READS `peerDependencies["@earendil-works/pi-tui"]` at runtime and prints it for `.github/workflows/ci.yml` to `npm install`. The sweep checked "does anything hard-code a stale range" and missed the actual defect class: does anything downstream assume the range is a SINGLE whitespace-free token. It didn't — `--install-args` space-joined its `name@range` entries, and all three `ci.yml` callers word-split that output with `read -ra ... <<< "$(...)"`, so pi-tui's new OR-form range (which itself contains a space) exploded into three argv tokens and broke `npm install` with `notarget No matching version found for undefined@^0.85.0` (reproduced verbatim, and again via `gh run rerun`). Fixed: the script now newline-joins (a newline can never appear inside one `name@range` string), and all three call sites read it with `mapfile -t` (newline-only splitting) instead. Regression-tested in `tests/scripts/supply-host-provided-deps.test.ts`.
- Consolidation verdict: class of size 1 for the stale-range shape (only `package.json`'s `peerDependencies` declares a version range for this package). For the shell-splitting shape: `grep -rn 'read -ra\|read -a ' .github/ scripts/` (excluding `node_modules`) — after this fix, the only hit is `scripts/supply-host-provided-deps.mjs`'s own doc comment WARNING against the pattern; the three former `read -ra` call sites in `ci.yml` were the entire population and all three are now `mapfile -t`. Class fully closed, not deferred.

## Merge order

Issue #2587's `pi.skills` fix (PR #2589) and issue #2590's `bundle-dist.mjs` esbuild-collision fix (PR #2594) both merged to master while this PR was open. Merged `origin/master` into this branch (see "Merge with master" below) — a clean auto-merge with no conflicts; both incoming changes and this PR's own changes verified to coexist correctly.

## Merge with master (2026-09-06)

**Why:** this PR's first CI run failed Unit tests / Lint & type-check / Production install build with `sh: 1: esbuild: not found` inside `scripts/bundle-dist.mjs`'s `npm exec --package esbuild` call — a pre-existing, repo-wide CI blocker unrelated to this diff (confirmed via a rerun that reproduced identically, and independently reproduced in an unrelated local sandbox). Filed as #2590 and fixed by #2594 (bundle-dist.mjs now resolves esbuild via `npm exec --prefix <fresh empty temp dir>` with `cwd: root`, avoiding the nested-esbuild collision). #2589 (a different, parallel fixer's PR for #2587) also landed, changing `package.json`'s `pi.skills` from `["../../skills"]` to `["./skills"]` and rewriting its `tests/packaging.test.ts` guard.

**Merge result:** `git fetch origin master && git merge origin/master --no-edit` — clean, zero conflicts. New head: `c81978ac61105ca102d17fa07c4303396712d788`. Verified semantically, not just textually:
- `tests/packaging.test.ts` carries BOTH test blocks side by side: `#2587`'s `"pi.skills resolves PACKAGE-ROOT-relative and never escapes the package (#2587)"` and this PR's `"pi-tui peer range covers every tested host version (#2586)"` — neither replaced the other, confirmed by grepping both describe names present after the merge.
- `package.json`: `pi.skills` is `["./skills"]` (master's #2589 change, untouched by me), `peerDependencies["@earendil-works/pi-tui"]` is still `"^0.84.1 || ^0.85.0"` and `devDependencies` still pin `^0.85.1` for both pi-coding-agent and pi-tui (this PR's changes, untouched by the merge).
- `scripts/bundle-dist.mjs` carries #2594's `--prefix`/`cwd: root` fix, unmodified by me (I didn't touch this file).

**`npm run build:dist` — the key regression check — now succeeds** (previously failed on this exact tree before #2594):
```
> pi-lens@4.1.3 build:dist
> node -e "require('fs').rmSync('dist',{recursive:true,force:true})" && npx --yes -p typescript@7.0.2 tsc --project tsconfig.dist.json --noCheck && npm run bundle:dist

> pi-lens@4.1.3 bundle:dist
> node scripts/bundle-dist.mjs

  dist/index.bundled.mjs  4.4mb ⚠️

⚡ Done in 48ms
[bundle] wrote self-contained dist/index.js
```
`dist/index.js` contains exactly 2 occurrences of `__pilensCreateRequire` (import alias + call), matching the `#335` bundle-shape test's expectation.

**Full re-verification after the merge, all green:**
- `npm run build` (tsc, tsconfig.build.json) — clean.
- `npm run lint` (`tsc --project tsconfig.json` + `oxlint --deny-warnings`) — clean.
- `tests/packaging.test.ts` — **37 passed** (was 32 passed | 4 skipped before the merge, since `dist/` now exists and the `#335` bundle-shape tests run instead of skipping).
- pi-tui-touching suites (`tui-fit`, `turn-summary-render`, `widget-state`, `widget-state-lsp-servers-bounded`, `deps-centralization`, `host-sdk-type-only`) plus `tests/scripts/check-lockfile-sync.test.ts` — **7 files, 107 tests, all passed**.
- Full governance/sweep/conformance/coverage/gate/silence/hermeticity/invariant/contract glob (mechanically re-derived post-merge — master added 3 new `tests/config/*.test.ts` files: `label-manifest-coverage`, `lsp-service-double-sweep`, `tracked-control-bytes`) plus every `tests/config/*.test.ts`, plus `tests/scripts/bundle-dist.test.ts` (new from #2594), plus `tests/packaging.test.ts` and `tests/scripts/check-lockfile-sync.test.ts` — **59 files, 798 tests, all passed** (`npm run test:targeted`).
- `npx oxfmt --check tests/packaging.test.ts` — already correctly formatted.
- `node scripts/check-changelog-fragments.mjs` — `changelog fragments OK (120 entries in .changelog/)`.

## Review round 1 (CI-caught coverage gap)

CI on head `c81978ac` (post-merge) came back Lint green but **Unit tests failed**: `tests/clients/host-edit-normalize-sync.test.ts > host still exports the line-ending + BOM primitives we vendored` — `expected src to contain 'startsWith("﻿")'`. This file is a drift guard (#257) that re-reads `node_modules/@earendil-works/pi-coding-agent/dist/core/tools/edit-diff.js` and asserts it still contains the literal BOM-detection primitive `clients/host-edit-normalize.ts` vendors a copy of. It is not in the pi-tui-touching list from my original brief, and doesn't match the mechanical governance-sweep keyword glob (sweep/ratchet/conformance/coverage/gate/governance/silence/hermeticity/invariant/contract) — so it was never run before push. **Class-sweep gap this PR is adding to the record:** a devDependency bump to a host SDK needs its own sweep for "reads the host's installed `dist/` source and asserts a literal" style guards (`hostPackageDir()`/`readFileSync` against `node_modules/@earendil-works/pi-coding-agent/dist` or `.../pi-tui/dist`), independent of the governance-suite keyword glob, since these guards pin runtime SOURCE, not test infrastructure. Swept `tests/` for this shape after the fact: only `tests/clients/host-edit-normalize-sync.test.ts` reads `pi-coding-agent`'s `dist/` this way; `tests/clients/console-capture-window-coverage.test.ts` reads the host's `.d.ts` structurally (member list derived dynamically, not a hand-pinned literal — unaffected by the bump, confirmed green) and `tests/packaging.test.ts`'s `pi-coding-agent` references are comments only. No `pi-tui` equivalent of this shape exists in `tests/`.

**Root cause (verified against the actual installed 0.85.1 source, not paraphrased):** `dist/core/tools/edit-diff.js` no longer inlines the BOM check — it now does `import { splitBom } from "../../utils/text.js";` and calls `splitBom(rawContent)`. The primitive itself is unchanged, just relocated to `dist/utils/text.js`:
```js
export function splitBom(content) {
    return content.startsWith("﻿") ? { bom: "﻿", text: content.slice(1) } : { bom: "", text: content };
}
```
`clients/host-edit-normalize.ts`'s own vendored logic already matches this (unchanged) runtime behavior — no production code fix needed, only the test's source-reader was stale. `detectLineEnding`/`normalizeToLF`/`restoreLineEndings`/`fuzzyFindText` and every other primitive this file guards are still in `edit-diff.js` unchanged (confirmed by grep against the installed source).

**Fix:** added `hostTextUtilsSource()` (mirrors `hostEditDiffSource()`) reading `dist/utils/text.js`, and split the single BOM assertion into two `it`s: `edit-diff.js` must still import+call `splitBom` (so a future host version that inlines something *different* instead of delegating still fails this guard — the guard's actual purpose, catching host drift, is preserved, not defanged) and the BOM code-point check is now asserted against `dist/utils/text.js`, wherever the primitive actually lives.

**Red proof** (this is the literal CI failure, reproduced locally against the same 0.85.1 tree pre-fix):
```
 ❯ tests/clients/host-edit-normalize-sync.test.ts:100:15
     98|   expect(src).toContain("export function normalizeToLF");
     99|   expect(src).toContain("export function restoreLineEndings");
    100|   expect(src).toContain(`startsWith("${esc(HOST_BOM_CODE_POINT)}")`);
       |               ^
    101|  });

 Test Files  1 failed (1)
      Tests  1 failed | 5 passed (6)
```

**Green after fix:**
```
 Test Files  1 passed (1)
      Tests  7 passed (7)
```

**Mutation-proof (both halves of the split guard, against the real installed `node_modules` files, reverted after):**
- Mutated `dist/utils/text.js`'s `startsWith("﻿")` → `startsWith("X")`: reds `"host still delegates BOM stripping to splitBom, and splitBom matches our vendored copy"` (`expected ... startsWith("﻿")` not found). Reverted; confirmed byte-identical to the installed copy via `diff`.
- Mutated `dist/core/tools/edit-diff.js` to drop the `import { splitBom } ...` line (simulating a future host inlining something else instead of delegating): reds the same test on the `toMatch(/import.../)` assertion. Reverted; confirmed byte-identical via `diff`.
- Re-ran green after each revert (7 passed).

**Full re-run after this fix**, `tests/clients/host-edit-normalize-sync.test.ts` plus the entire previous 59-file governance/targeted batch (66 files total) — **903 tests, all passed**. `npx oxfmt --check` clean on the touched file. `npm run build` (tsc) and `npm run lint` clean.

New head: `e32d814e` (one commit on top of the merge commit, no rebase — `git log --oneline` shows the merge commit followed by this fix commit).

## Review round 2

CI's `ci-verdict.mjs` (Unit tests + Lint only) read green on `e32d814e`, but that script does not cover the "Production install build" or "Install test" jobs — those were red. Reviewer read all jobs directly and found six findings; all reproduced and fixed on this branch, no new PR.

### F1 (blocking) — shell-splitting hazard in `--install-args`

**Reproduced exactly as reported.** `node scripts/supply-host-provided-deps.mjs --install-args` on the pre-fix code printed `typebox@^1.0.0 @earendil-works/pi-tui@^0.84.1 || ^0.85.0` (space-joined); `read -ra HOST_PKGS <<< "$(...)"` (all three `ci.yml` call sites, lines ~485/537/789) then split that into 4 tokens instead of 2:
```
Supplying: typebox@^1.0.0 @earendil-works/pi-tui@^0.84.1 || ^0.85.0
count: 4
npm ERR! code ETARGET
npm ERR! notarget No matching version found for undefined@^0.85.0.
EXIT: 1
```
(reproduced locally with the exact pre-fix output piped through `read -ra`, then confirmed a live `npm install` with those 4 tokens fails identically). `gh run rerun` on the original CI run reproduced the same failure a second time (deterministic, not flaky), plus a third job — "Production install build" — also red for the same reason.

**Fix:** `scripts/supply-host-provided-deps.mjs`'s `--install-args` mode now `join("\n")`s instead of `join(" ")`; all three `ci.yml` call sites switched from `read -ra HOST_PKGS <<< "$(node "$SUPPLY" --install-args)"` to `mapfile -t HOST_PKGS < <(node "$SUPPLY" --install-args)` (newline-only splitting). Confirmed no fourth caller exists (`grep -rn "install-args"` across the repo) and no other `read -ra`/`read -a` pattern exists anywhere in `.github/` or `scripts/` (the class is now fully closed, not just this one instance).

**Regression test:** `tests/scripts/supply-host-provided-deps.test.ts` — spawns the real script (real-process-spawn, admitted via header + `ADMITTED_AFTER_BASELINE` + `vitest.config.ts`'s `wallClockBudgetInclude`, per the flake-shape ratchet's two-part gate) and asserts the newline-split token count equals `HOST_PROVIDED_RUNTIME_PACKAGES.length`, plus that the space-containing pi-tui entry survives as ONE unsplit token. Red before the fix:
```
AssertionError: expected exactly 2 newline-delimited entries, got: ["typebox@^1.0.0 @earendil-works/pi-tui@^0.84.1 || ^0.85.0"]: expected 1 to be 2
AssertionError: expected [ Array(1) ] to include '@earendil-works/pi-tui@^0.84.1 || ^0.…'
Tests  2 failed | 1 passed (3)
```
Green after (`Tests 3 passed (3)`). Reverted the delimiter back to `" "` via a targeted edit (not checkout, to protect sibling uncommitted work), reran red, restored, reran green — `git diff` confirmed the fix intact.

**End-to-end shell-fragment proof** (not just the unit test — the review explicitly asked for the real `ci.yml` lines run locally): extracted and ran all three shapes verbatim (plain `npm install`, and the `--prefix "$INSTALL_DIR"` variant from the `install-test` job) against a scratch npm project:
```
Supplying: typebox@^1.0.0 @earendil-works/pi-tui@^0.84.1 || ^0.85.0

added 4 packages in 222ms
EXIT: 0
	"version": "0.85.1",
```
Both shapes install pi-tui 0.85.1 (satisfying the OR range as one token) with exit 0.

Class-sweep correction to the earlier "Class sweep" section above: the round-1 sweep only checked for hard-coded stale ranges and wrongly concluded `supply-host-provided-deps.mjs` needed no alignment — it dynamically reads the range (correct), but nothing downstream was safe against a range CONTAINING whitespace. See the amended bullet above; the shell-splitting class is now fully closed (verified: no other `read -ra`/`read -a` in the repo).

### F2 (high) — `package-lock.json` regenerated with a pre-11.11 npm, stripping `libc` metadata

**Reproduced:** `grep -c '"libc"' package-lock.json` on the branch was `0`; `origin/master`'s own lockfile (same package family, pre-#2586) has `36`. `.github/workflows/ci.yml` lines ~443-449 pin npm 11.18.0 specifically for this (#9025 — npm 11.11+ writes `libc` platform descriptors for glibc/musl optional-binary pairs; older npm silently drops them, and an install against a libc-blind lockfile installs BOTH variants).

**What I found doing this that the review didn't specify:** an in-place `npm install --package-lock-only` (or even a real `npm install`) against an EXISTING lockfile that already "satisfies" `package.json` is a no-op for metadata — it does not backfill missing `libc` fields, with or without existing `node_modules`. Only a full regeneration (delete `package-lock.json` AND `node_modules`, then `npx npm@11.18.0 install --package-lock-only`) actually re-fetches full registry metadata for every package and writes `libc`. That full regeneration also picked up ordinary patch/minor bumps for already-caret-ranged devDependencies (a fresh resolution reads "latest satisfying" as of today, not "whatever the old lock pinned") — I verified this honestly rather than assume the reviewer's "identical package set" held: `@ast-grep/cli`/`@ast-grep/napi` 0.45.2→0.45.3, `@biomejs/*` 2.5.11→2.5.12, several `@rolldown/binding-*` 1.2.5→1.2.7, `typebox` 1.3.20→1.3.28, `@babel/*` helpers, and a handful more (60 version diffs total, every one still satisfying its existing package.json range) plus a `std-env` nested-duplicate → hoisted-single dedup (4.1.0→4.2.0, a transitive vitest/coverage-v8 dep). **None of pi-lens's own runtime-critical dependencies changed** (`@earendil-works/pi-tui`/`pi-coding-agent` stayed 0.85.1; `minimatch`, `js-yaml`, `vscode-jsonrpc`, `pidusage` unchanged).

`grep -c '"libc"' package-lock.json` after regeneration: **36** (matches `origin/master`'s count for the same package family exactly).

Also caught a knock-on break from the ast-grep patch bump: `tests/packaging.test.ts`'s `allowScripts` pin (`"@ast-grep/cli@0.45.2": true`) no longer matched the newly-locked `0.45.3`, failing `expect(pkg.allowScripts?.[...]).toBe(true)`. Bumped the pin to match.

**Caution for local reproduction:** this repo's own installed npm (9.2.0 in my sandbox) silently re-strips the `libc` fields the NEXT time it touches `package-lock.json` (confirmed empirically) — so after regenerating with npm 11.18.0, do not run a further plain local `npm install` against it without re-verifying the `libc` count afterward.

**Superseded by review round 3 / N2 below:** the from-scratch regeneration described above was replaced with a minimal-delta regeneration once review flagged the ~60 unrelated floats as out of scope. See "Review round 3" for the corrected approach and its much smaller diff.

### F3 (medium) — tested-versions list didn't pin the low end of the range

**Reproduced:** mutating the peer range to `"^0.85.0"` (dropping 0.84.x) stayed green pre-fix, because `testedVersions = [lockVersion, "0.85.1"]` deduped to a single `0.85.1` entry once the devDependency matched the lockfile.

**Fix:** added `"0.84.1"` explicitly to `testedVersions` — the same version `scripts/install-selftest.mjs` (line ~256) already cites as verified against pi's package-manager resolver, so it's not an arbitrary floor.

**Red (mutated range to `^0.85.0`):**
```
FAIL tests/packaging.test.ts > ... > accepts tested host version 0.84.1
AssertionError: peerDependencies["@earendil-works/pi-tui"] (^0.85.0) must accept 0.84.1: expected false to be true
Tests  1 failed | 4 passed | 33 skipped (38)
```
**Green after revert:** `Tests 5 passed | 33 skipped (38)`. Reverted via targeted edit (package.json had no other uncommitted diff at the time — confirmed via `git diff` before reverting), re-verified clean.

### F4 (medium) — BOM assertion not scoped to `splitBom`'s own body

**Reproduced the MU3 attack exactly as specified:** built a fixture `dist/utils/text.js` with a `legacySplitBom` sibling keeping the old `startsWith("﻿")` literal while the REAL `splitBom` was rewritten to `charCodeAt(0) === 0xfeff`. Confirmed the LOOSE "anywhere in file" check (`textUtilsSrc.includes(...)`) stays green under this fixture (`loose anywhere-in-file check would pass: true`) — proving the round-1 guard wasn't actually scoped to `splitBom`.

**Fix:** added `sliceExportedFunction(source, name)` — slices one named top-level `export function` declaration's own source, bounded by the next top-level `export function`/`export const` or EOF — and assert the BOM literal against that slice instead of the whole file.

**Red under the MU3 fixture (tightened guard):**
```
FAIL ... > host still delegates BOM stripping to splitBom, and splitBom matches our vendored copy
AssertionError: expected 'export function splitBom(content) {\n…' to contain 'startsWith("﻿")'
Tests  1 failed | 6 passed (7)
```
Restored the real `node_modules` file afterward (`diff` confirmed byte-identical); re-ran green (`7 passed`).

### F5 (low) — SDK floor predates the BOM relocation

`HOST_EDIT_DIFF_SDK_FLOOR` bumped from `"0.79.9"` to `"0.85.1"` (the file/function `splitBom` lives in didn't exist before 0.85.1) so a host below the floor fails with the floor-check's own legible message rather than a confusing source-mismatch one from F4's tightened guard. Header comment updated to note the 0.85.1 relocation.

### F6 (low) — PR body corrections

Fixed the "Class sweep" and "Blast radius" sections above. Independently checked `tests/clients/pi-host-contract.test.ts` (not just trusted the reviewer's claim): it reads `core/extensions/runner.js`, `core/agent-session.js`, `core/tools/bash.js`, and `utils/paths.js` from the host build — none of which reference BOM, `splitBom`, `stripBom`, or `edit-diff.js` (grepped the file directly) — so it is genuinely unaffected by the 0.85.1 BOM relocation and needed no change. Confirmed still green (34 passed).

### Verification (full re-run against final committed state)

Clean rebuild (`rm -rf dist node_modules && npm install`, then restored `package-lock.json`'s libc metadata since the local npm re-stripped it on save — see F2 caution above; `npm run build`): 68 governance/sweep/conformance/coverage/gate/silence/hermeticity/invariant/contract + every `tests/config/*.test.ts` + the pi-tui/host-SDK-touching files + `tests/scripts/{audit-prod-deps,bundle-dist,check-lockfile-sync,supply-host-provided-deps}.test.ts` — **912 tests, all passed, 0 skipped** (dist/ now built, so the `#335` bundle-shape tests run instead of skipping). `npm run lint` (tsc + oxlint) clean. `npx oxfmt --check` clean on all touched files. `node scripts/check-changelog-fragments.mjs` — `changelog fragments OK (120 entries in .changelog/)`. `.github/workflows/ci.yml` parses as valid YAML (`js-yaml.load`).

New head: `9c638b4b` (5 commits on top of `e32d814e`: flake-shape admission, lockfile regen, testedVersions fix, BOM-guard scoping + floor bump, changelog update). This head then hit a real merge conflict against a concurrently-merged master PR (`tests/clients/flake-shape-ratchet.test.ts`'s `ADMITTED_AFTER_BASELINE` dict), resolved additively and re-verified (68 files / 912 tests, still green) at `fdd04be5`.

## Review round 3

CI on `fdd04be5` found a NEW regression from the round-2 F1 fix, plus review caught two more issues in an addendum. All reproduced and fixed on this branch, one push covering both.

### macOS regression from the F1 fix — `mapfile` doesn't exist on bash 3.2

**Reproduced from the job log verbatim** (`Install test (macos-latest)`, job 101527965573):
```
shell: /bin/bash --noprofile --norc -e -o pipefail {0}
/Users/runner/work/_temp/....sh: line 3: mapfile: command not found
##[error]Process completed with exit code 127.
```
`mapfile`/`readarray` are bash 4+ builtins; macOS has shipped bash 3.2 since Apple stopped updating it at the GPLv2→GPLv3 license switch, so the round-2 fix broke exactly the runner it hadn't been tested against.

**Fix:** replaced all three `.github/workflows/ci.yml` call sites with a portable loop bash 3.x supports:
```sh
HOST_PKGS=()
while IFS= read -r pkg; do
  [ -n "$pkg" ] && HOST_PKGS+=("$pkg")
done < <(node "$SUPPLY" --install-args)
```
`scripts/supply-host-provided-deps.mjs`'s newline-delimited output is unchanged (that part was correct); only the shell-side consumption changed. Updated the script's USAGE docstring to describe the portable loop and explicitly warn against `mapfile` (names why: macOS bash 3.2).

**Bashism sweep of `.github/workflows/ci.yml`** (the only workflow file this PR touches), checked against every construct review named:
- `mapfile`/`readarray` — 0 hits after the fix (was 3, now the portable loop everywhere).
- `${var,,}` / `${var^^}` (bash 4+ case-conversion) — **7 hits**, all in a duplicated "Validate merge-train dispatch payload" step (`${PAYLOAD_SHA,,}`). Mapped each to its job's `runs-on`:
  - Lines 71, 119, 165, 203, 281, 434 — jobs `validate-merge-train-dispatch`, `close-keyword-lint`, `changelog-fragment-fastfail`, `lint-and-typecheck`, `test`, `prod-install-build` — **all `runs-on: ubuntu-latest`** (confirmed via `awk` mapping job-start lines to `runs-on:` lines). Ubuntu ships bash 5.x; these are safe as-is, left unchanged, stated explicitly rather than waved off.
  - Line 630 — inside the `install-test` job, `runs-on: ${{ matrix.os }}` (`[ubuntu-latest, windows-latest, macos-latest]`), gated `if: github.event_name == 'repository_dispatch'` with no per-step `shell:` override. This one DOES reach macOS's bash 3.2 whenever that condition is true. **Fixed** — replaced with a portable `tr '[:upper:]' '[:lower:]'` lowercase:
    ```sh
    payload_sha_lower="$(printf '%s' "$PAYLOAD_SHA" | tr '[:upper:]' '[:lower:]')"
    if [[ "$payload_sha_lower" != "$checked_sha" ]]; then
    ```
  - `declare -A` — 0 hits.
  - `|&` — 0 hits.

`tests/config/github-workflow-permissions.test.ts` (covers this file) still passes; `.github/workflows/ci.yml` still parses as valid YAML.

**End-to-end proof of the new loop**, run through real bash piped from the real script (mirrors the F1 verification style):
```
count: 2
TOKEN: [typebox@^1.0.0]
TOKEN: [@earendil-works/pi-tui@^0.84.1 || ^0.85.0]

added 4 packages in 1s
EXIT: 0
	"version": "0.85.1",
```
Also re-ran the `--prefix "$INSTALL_DIR"` variant (the `install-test` job's exact shape) with the new loop — same result, exit 0, pi-tui resolves to 0.85.1.

### N2 (high, addendum) — the F2 lockfile regen floated ~60 unrelated packages

Round 2's F2 fix (delete `node_modules` + `package-lock.json`, regenerate from scratch) genuinely restored `libc` metadata, but a from-scratch resolution reads "latest satisfying" for every caret-ranged devDependency as of THAT moment — floating ast-grep, biome, rolldown bindings, typebox, babel helpers, etc. (~60 version diffs), none of them this PR's business, and forcing `package.json`'s `@ast-grep/cli` `allowScripts` pin to a version dependabot should own.

**Fix, minimal-delta regeneration:**
1. Merged `origin/master` again (moved to `a6019ad9` by this point).
2. `git checkout origin/master -- package-lock.json` — start from master's own current (already libc-aware, 36-entry) lockfile instead of nothing.
3. `npx npm@11.18.0 install --package-lock-only` against the already-bumped `package.json` — since the lockfile already satisfies every OTHER range, only the two actually-changed devDependencies (pi-coding-agent, pi-tui) and their own transitive closure get touched.
4. Reverted `package.json`'s `allowScripts` pin back to `@ast-grep/cli@0.45.2`, confirmed still master's current pin (it had not itself moved).

**Structural diff against master's lockfile** (verified, not assumed): **added: 32, removed: 6, versionChanged: 7** — exactly matching review's predicted shape. Every added/removed entry is under `node_modules/@earendil-works/pi-coding-agent/node_modules/**` (pi-coding-agent's own transitive deps moving with it to 0.85.1 — esbuild platform binaries, `@earendil-works/chord`, `@stablelib/base64`, etc. swapping in, `pi-client`/`pi-protocol`/`@opentelemetry/api`/`glob`/`minipass`/`path-scurry` dropping out) except one: `@types/semver`, which is this PR's OWN already-intentional devDependency (added for the peer-range regression test, not new here). All 7 version-changes are `pi-coding-agent`/`pi-tui` themselves plus their own `@earendil-works/pi-agent-core`/`pi-ai`/`pi-telemetry`/`pi-tui` nested copies and `@anthropic-ai/sdk`. Zero biome/rolldown/ast-grep/typebox entries in this diff. `grep -c '"libc"' package-lock.json` = **36** (unchanged, confirmed).

### F4-residual (low, addendum) — `sliceExportedFunction`'s boundary regex missed non-exported siblings

`/\n(?:export (?:function|const)\b)/` only recognized EXPORTED declarations as slice terminators. Built the exact attack: a fixture where the real `splitBom` is rewritten (no BOM literal) and an **un-exported** `function legacySplitBom` placed immediately after it keeps the old literal.

**Red (well, green when it should be red) under the CURRENT regex** — the false negative, reproduced:
```
Test Files  1 passed (1)
     Tests  7 passed (7)
```
All 7 pass even though `splitBom` itself no longer contains the BOM check — `legacySplitBom`'s body (no `export` prefix) never terminated the slice and got swallowed into "splitBom"'s captured source.

**Fix:** `/\n(?:(?:export )?(?:function|const)\b)/` — `export ` now optional in the boundary. Re-ran the IDENTICAL fixture:
```
FAIL ... > host still delegates BOM stripping to splitBom, and splitBom matches our vendored copy
AssertionError: expected 'export function splitBom(content) {\n…' to contain 'startsWith("﻿")'
Tests  1 failed | 6 passed (7)
```
Correctly reds now. Restored the real host source afterward (`diff` confirmed byte-identical); re-ran green (7 passed).

### Docstring fix

`scripts/supply-host-provided-deps.mjs`'s USAGE comment (previously prescribing `mapfile -t` by name) now shows the portable `while IFS= read -r` loop and explicitly names why `mapfile` is wrong here (macOS bash 3.2).

### Verification (full re-run against final committed state, post both fixes)

Merged `origin/master` again mid-round (clean, no conflicts this time — `git checkout origin/master -- package-lock.json` plus the minimal-delta regen happened as part of the same sequence). Clean rebuild (`rm -rf dist node_modules && npm install`, restored `package-lock.json`'s libc metadata per the standing local-npm caution, `npm run build`): the full governance/config/host-SDK-touching batch, now including master's new `tests/scripts/{build-dist-tsc,exec-isolation}.test.ts` (from master's own unrelated `#2593` bundle-dist refactor that merged in) — **70 files, 918 tests, all passed, 0 skipped**. `npm run lint` (tsc + oxlint) clean. `npx oxfmt --check` clean. `node scripts/check-changelog-fragments.mjs` — `changelog fragments OK (127 entries in .changelog/)`. `.github/workflows/ci.yml` still parses as valid YAML.

New head: `de7e728d` (4 commits on top of the round-2 head `fdd04be5`: macOS-portable loop + bashism sweep, F4-residual regex fix, merge + minimal-delta lockfile regen, changelog update).


apmantza:
CI's Unit tests / Lint & type-check / Production install build jobs are red on this PR's head (`d974f366`), but the failure is unrelated to this diff: `npm install`'s `prepare` script dies at `sh: 1: esbuild: not found` inside `scripts/bundle-dist.mjs`'s pre-existing `npm exec --package esbuild@0.28.1` invocation, before any of this PR's changes are exercised. Confirmed deterministic via a rerun (`gh run rerun 34036728661 --failed`, identical failure), and reproduced independently in an unrelated local sandbox. Filed as a repo-wide CI blocker: #2590.

Issue #2636, opened 2026-09-06
Bundled rule/query dirs (ast-grep, tree-sitter-queries) silently degrade to zero, same shape as #2626


apmantza:
## What

#2626's class sweep found the SAME silent-zero shape it fixed for skills in
two other `resolvePackagePath`-resolved bundled resource directories:

- `clients/ast-grep-client.ts:211` — `AstGrepClient`'s `ruleDir` falls back to
  `resolvePackagePath(import.meta.url, "rules")` when the project has none,
  with no existence check and no degradation if that bundled dir is also
  absent (entry copied out of the package tree, same as #2626's layout F).
- `clients/cache/rule-cache.ts:55` — `BUNDLED_RULES_ROOT` (`rules/tree-sitter-queries`)
  is used unconditionally by `RuleCache`, same absent-dir gap.
- `clients/tree-sitter-query-loader.ts:72,223` — `ruleFilesForLanguage`
  already does `if (!fs.existsSync(dir)) continue;` per candidate dir, so it
  degrades gracefully to "0 query files for this language" WITHOUT throwing,
  but records nothing either — a managed-cache relocation makes every
  language's bundled tree-sitter queries silently vanish with no signal,
  identical in effect to the pre-fix skills bug.

None of the three import `degradation-ledger.ts` or `user-notify.ts` today
(checked via `grep -l recordDegradation\|notifyUserDegradation`).

Excluded from this sweep (checked, NOT the same shape): the config-arg
builders in `clients/tool-policy.ts` (`markdownlintConfigArgs`,
`ruffConfigArgs`, `biomeConfigArgs`) and `clients/lsp/server.ts`'s typos
`configPath` all hand a `resolvePackagePath` result to an EXTERNAL spawned
tool as a `--config`/`--config-path` argument — a missing file there makes
the external tool itself error loudly (already surfaced through the existing
spawn-failure path), not silently.

## Suggested fix

Reuse `clients/skills-resolver.ts`'s pattern (or extract a shared
`resolveBundledResourceDir(importMetaUrl, subpath, kind)` helper) for the
three sites above: fail-closed on an absent/unreadable directory, one
`recordDegradationOnce` + `notifyUserDegradation` per kind, keyed on the
resolved path.

## Acceptance

- [ ] `AstGrepClient` records a bounded degradation when NEITHER the project
      `rules/` nor the bundled `rules/` exists.
- [ ] `RuleCache`'s bundled root records a bounded degradation when absent.
[...]

PR #2654, opened 2026-09-06
fix(rules): degrade-and-record bundled ast-grep/tree-sitter dirs (closes #2636)

apmantza:
## Summary

#2637 (still open at the time of this branch — see the base-branch note
below) fixed the silent-zero shape for `resources_discover`'s `skills/`
directory and filed #2636 for the other two `resolvePackagePath`-resolved
bundled resource dirs its class sweep found:

- `clients/ast-grep-client.ts:211` (now ~251) — `AstGrepClient`'s `ruleDir`
  falls back to `resolvePackagePath(import.meta.url, "rules")` with no
  existence check when the project has none of its own.
- `clients/cache/rule-cache.ts:55` — `BUNDLED_RULES_ROOT` is used
  unconditionally by `RuleCache` to classify a rule file as
  bundled-vs-project for cache fingerprinting.
- `clients/tree-sitter-query-loader.ts:72,223` — `ruleFilesForLanguage`
  already degrades gracefully to "0 query files" per candidate dir
  (`if (!fs.existsSync(dir)) continue;`) but records nothing, so a
  managed-cache relocation makes every language's bundled tree-sitter
  queries silently vanish with no signal.

Each now reports a bounded degradation, following #2637's shipped pattern
exactly (`clients/skills-resolver.ts`'s classify/report shape): purely
observational — never changes what path is returned or how the fallback
resolves — via a new shared pair, `clients/bundled-resource-health.ts`
(`classifyBundledResourceDir` + `reportBundledResourceDirHealth`), reused by
all three sites plus `clients/ast-grep-rule-manager.ts`'s
`checkAstGrepRulesHealth` (ast-grep needs a richer "does this hold a `.yml`
rule description" check mirroring `loadRuleDescriptions`'s own resolution,
so it wraps the shared pair rather than calling it directly).

**Two new `DegradationKind`s**, ONE per resource dir as the issue's
suggested-fix allows (`clients/degradation-ledger.ts`):
- `ast-grep-rules-dir-missing` — `clients/ast-grep-client.ts`.
- `tree-sitter-queries-dir-missing` — shared by `clients/cache/rule-cache.ts`
  and `clients/tree-sitter-query-loader.ts`, which read the IDENTICAL
  physical bundled directory (`rules/tree-sitter-queries`) independently.
  Using one kind + the same subject string means whichever call site
  observes the absence first is the ONE ledger row — no duplicate noise from
  two sites reporting the same missing directory.

**Acceptance criteria (all three; this PR is `closes #2636`):**
- [x] `AstGrepClient` records a bounded degradation when NEITHER the project
      `rules/` nor the bundled `rules/` exists (checked only when the
      bundled fallback is actually selected — a broken *project* override is
      a user-config concern, not this bug's shape).
- [x] `RuleCache`'s bundled root records a bounded degradation when absent.
- [x] `ruleFilesForLanguage` records ONE bounded degradation (not per
      language) when a language resolves zero files from EITHER candidate
      dir, investigated and DISTINGUISHED from "no queries authored for this
      language by design" (cobol, plsql — see `tree-sitter-shared.ts`'s
      disabled-by-design languages) by checking the shared bundled ROOT's
      health, never any one language's own subdirectory.

## Base branch note

#2637 was still `OPEN` (`gh pr view 2637 --json state` → `"OPEN"`) when this
branch was cut, per the task's instruction to branch from its head in that
case. This branch is `fix/2626-skills-dir-degradation` (#2637's head), and
this PR's base is set to that branch, not `master` — it should be retargeted
to `master` (or simply merged after #2637 lands, whichever the merge queue
prefers) once #2637 merges. The diff below is scoped to ONLY this issue's
files; `git diff --stat origin/master` will additionally show #2637's own
diff until it lands.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:observability
- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [ ] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds (changed `clients/`)
- [x] `package-lock.json` is in sync with `package.json` (no dependency changes)
- [ ] `AGENTS.md` is updated — not needed: no new convention/invariant, this follows the existing `incrementDegradationCount` + `notifyUserDegradation` seam #2637 established one PR earlier for the identical shape
- [x] `.changelog/fix-2636-bundled-resource-dir-degradation.md` — Fixed section, validated with `node scripts/check-changelog-fragments.mjs`
- [x] Commit subject includes the issue number: `(refs #2636)`

## Tests

- **NEW `tests/clients/bundled-resource-health.test.ts`** (8 cases) — drives
  the shared `classifyBundledResourceDir`/`reportBundledResourceDirHealth`
  pair directly with real temp dirs (absent/empty/healthy) plus one
  `node:fs` mock for the EACCES case; pins the once-per-(kind,subject)
  notify/tally split.
- **NEW `tests/clients/ast-grep-rule-manager.test.ts`** (7 cases) — drives
  `checkAstGrepRulesHealth` with real temp-dir layouts matching
  `loadRuleDescriptions`'s own candidate resolution exactly (including a
  cross-check against a REAL `AstGrepRuleManager` for the healthy case, and
  the "first existing candidate shadows a later populated one" edge case),
  plus two EACCES cases: one where the readdirSync(rulesPath) branch throws,
  one where NO candidate is even visible via `existsSync` and the
  `readdirSync(ruleDir)` probe itself throws (this second case was added
  after mutation-proofing surfaced it as untested — see Mutation proofs).
- **NEW `tests/clients/ast-grep-client-rules-health.test.ts`** (4 cases) —
  drives `AstGrepClient`'s real constructor: negative control (this repo's
  own `rules/` — no check, no record), the bundled-fallback branch firing
  (real bundled dir, healthy), an unhealthy bundled dir via
  `vi.spyOn(checkAstGrepRulesHealth)` returning `absent`, and the
  once-per-session notify gate across two constructions.
- **EDITED `tests/clients/cache/rule-cache.test.ts`** (+3 cases) — negative
  control against this repo's real bundled root (healthy), an absent case
  via `vi.spyOn(classifyBundledResourceDir)`, and the once-per-session gate
  across three constructions.
- **EDITED `tests/clients/tree-sitter-query-loader.test.ts`** (+4 cases) —
  a fully REAL (no mocking) proof that `cobol` (no bundled queries by
  design) records nothing against the real, healthy bundled root; a proof
  that the classify call never fires on the common non-empty path
  (`typescript`); an absent case via the same spy technique; and the
  "collapses to ONE row across languages" proof.

Why `vi.spyOn` on the module namespace instead of `vi.mock`: `BUNDLED_RULES_ROOT`/`BUNDLED_QUERIES_ROOT`
are module-load-time constants derived from EACH file's own fixed
`import.meta.url` (unlike `skills-resolver.ts`'s `resolveSkillPaths`, they
are not parameterized — changing that would risk changing resolution
behavior, which the issue explicitly forbids). Faking a different resolved
package root would need either an injectable `import.meta.url` (a
signature change to a production function used at several call sites) or
mocking `node:fs` against this repo's own REAL absolute bundled paths
(brittle, and risks affecting unrelated fs calls in the same process). The
pure classify/health functions (`classifyBundledResourceDir`,
`checkAstGrepRulesHealth`) get FULL, real-fs, no-mock coverage of every
status; the three call sites' WIRING (do they call it, with the right
subject, in the right branch) is proven by spying on that already-tested
function's return value — a stated, honest use of dependency substitution
for glue code, not a fixture-gamed test of the fix itself.

### Red-first proof (pre-fix behavior)

Committed the fix + tests first, then reverted ONLY the 5 pre-fix source
files (`clients/ast-grep-client.ts`, `clients/ast-grep-rule-manager.ts`,
`clients/cache/rule-cache.ts`, `clients/tree-sitter-query-loader.ts`,
`clients/degradation-ledger.ts`) via `git checkout HEAD~1 --` against the
committed tree, deleted the new `clients/bundled-resource-health.ts`,
rebuilt, and ran the five new/edited test files:

```
 Test Files  4 failed | 1 passed (5)
      Tests  13 failed | 31 passed (44)
```

(The 1 passing file, `bundled-resource-health.test.ts`, is brand-new
infrastructure with no pre-fix analog to regress against — see Mutation
proofs below for its coverage instead.) The 13 failures, one per new
behavior:

```
FAIL ast-grep-rule-manager.test.ts > reports absent when ruleDir itself does not exist
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-rule-manager.test.ts > reports empty when ruleDir exists but none of the three candidates hold a .yml file
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-rule-manager.test.ts > reports healthy via the nested ast-grep-rules/rules candidate, matching loadRuleDescriptions
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-rule-manager.test.ts > reports healthy via the flat rules/ candidate when ast-grep-rules/rules is absent
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-rule-manager.test.ts > reports empty (not absent) when the first EXISTING candidate holds no .yml, even though a later candidate would
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-rule-manager.test.ts > distinguishes an unreadable ruleDir (EACCES) from absent (ENOENT)
TypeError: checkAstGrepRulesHealth is not a function
FAIL ast-grep-client-rules-health.test.ts > checks bundled health and logs the phase record when the project has no rules/
FAIL ast-grep-client-rules-health.test.ts > records a bounded degradation + notify when the bundled rules/ is unhealthy
Error: The property "checkAstGrepRulesHealth" is not defined on the object.
FAIL ast-grep-client-rules-health.test.ts > does not re-notify on a second unhealthy construction in the same session
Error: The property "checkAstGrepRulesHealth" is not defined on the object.
FAIL cache/rule-cache.test.ts > records a bounded degradation + notify when the bundled root is absent
AssertionError: expected "classifyBundledResourceDir" to be called with arguments: [ undefined ]
Number of calls: 0
FAIL cache/rule-cache.test.ts > does not re-notify across repeated constructions in the same session
FAIL tree-sitter-query-loader.test.ts > records a bounded degradation + notify when the bundled root is actually gone
AssertionError: expected "classifyBundledResourceDir" to be called with arguments: [ undefined ]
Number of calls: 0
FAIL tree-sitter-query-loader.test.ts > collapses every zero-file language into ONE ledger row, not one per language
```

Restored via `git checkout HEAD -- <files>`; `git diff HEAD --stat` was
empty afterward (byte-identical restore).

### Mutation proofs (every new guard neutered, each reds)

1. **`bundled-resource-health.ts`: `if (health.status === "healthy") return;`
   → `if (true) return;`** (never records) — 3/8 red:
   ```
   FAIL carries the fsErrorCode into ledger metadata for an unreadable directory
   AssertionError: the given combination of arguments (undefined and string) is invalid…
   FAIL notifies exactly once per (kind, subject) across repeated calls
   ```
2. **`bundled-resource-health.ts`: notify-once gate `if (isFirstOccurrence)`
   → `if (true)`** — 1/8 red:
   ```
   FAIL notifies exactly once per (kind, subject) across repeated calls
   AssertionError: expected [ {…}, {…}, {…} ] to have a length of 1 but got 3
   ```
3. **`bundled-resource-health.ts`: ENOENT check `if (fsErrorCode === "ENOENT")`
   → `if (true)`** (collapses unreadable into absent) — 1/8 red:
   ```
   FAIL distinguishes an unreadable directory (EACCES) from absent (ENOENT)
   AssertionError: expected { status: 'absent' } to deeply equal { status: 'unreadable', fsErrorCode: 'EACCES' }
   ```
4. **`bundled-resource-health.ts`: `entries.length > 0` → `false`** (never
   healthy) — 1/8 red:
   ```
   FAIL reports healthy with the real entry count for a populated directory
   AssertionError: expected { status: 'empty' } to deeply equal { status: 'healthy', entryCount: 2 }
   ```
5. **`ast-grep-client.ts`: `if (usingBundledFallback)` → `if (false)`**
   (never checks) — 3/4 red:
   ```
   FAIL checks bundled health and logs the phase record when the project has no rules/
   FAIL records a bounded degradation + notify when the bundled rules/ is unhealthy
   AssertionError: expected "checkAstGrepRulesHealth" to be called 1 times, but got 0 times
   FAIL does not re-notify on a second unhealthy construction in the same session
   ```
6. **`ast-grep-client.ts`: `usingBundledFallback` computed as `!ruleDir`
   only** (drops the project-exists check, so it fires even when the
   project provides its own `rules/`) — the negative control reds:
   ```
   FAIL does NOT check bundled health when the project provides its own rules/
   AssertionError: expected [ { type: 'phase', … } ] to deeply equal []
   ```
7. **`ast-grep-rule-manager.ts`: `if (fs.existsSync(candidate))` → `if
   (false)`** (never finds a candidate) — 4/6 red:
   ```
   FAIL reports healthy via the nested ast-grep-rules/rules candidate
   AssertionError: expected { status: 'absent' } to deeply equal { status: 'healthy', entryCount: 2 }
   FAIL reports healthy via the flat rules/ candidate when ast-grep-rules/rules is absent
   AssertionError: expected { status: 'absent' } to deeply equal { status: 'healthy', entryCount: 1 }
   FAIL reports empty (not absent) when the first EXISTING candidate holds no .yml…
   AssertionError: expected { status: 'absent' } to deeply equal { status: 'empty' }
   ```
8. **`ast-grep-rule-manager.ts`: `entries.length > 0` → `false`** — 2/6 red:
   ```
   FAIL reports healthy via the nested ast-grep-rules/rules candidate
   AssertionError: expected { status: 'empty' } to deeply equal { status: 'healthy', entryCount: 2 }
   FAIL reports healthy via the flat rules/ candidate…
   AssertionError: expected { status: 'empty' } to deeply equal { status: 'healthy', entryCount: 1 }
   ```
9. **`ast-grep-rule-manager.ts`: `if (fsErrorCode && fsErrorCode !== "ENOENT")`
   → `if (false)`** — FIRST attempt was a FALSE GREEN (all 6 tests passed:
   the only existing EACCES test hit the OTHER readdirSync call site, never
   this branch). Added the "no candidate visible" EACCES test (see Tests
   above) specifically to close this gap; re-ran the same mutation, now 1/7
   red:
   ```
   FAIL distinguishes an unreadable ruleDir ITSELF (EACCES) from absent (ENOENT), when no candidate is even visible
   AssertionError: expected { status: 'absent' } to deeply equal { status: 'unreadable', fsErrorCode: 'EACCES' }
   ```
10. **`cache/rule-cache.ts`: `reportBundledRulesRootHealth();` call
    deleted** — 2/11 red:
    ```
    FAIL records a bounded degradation + notify when the bundled root is absent
    AssertionError: expected "classifyBundledResourceDir" to be called with arguments: [ Array(1) ]
    Number of calls: 0
    FAIL does not re-notify across repeated constructions in the same session
    ```
11. **`tree-sitter-query-loader.ts`: `if (files.size === 0)` → `if
    (false)`** (never checks, even when zero files found) — 2/15 red:
    ```
    FAIL records a bounded degradation + notify when the bundled root is actually gone
    AssertionError: expected "classifyBundledResourceDir" to be called with arguments: [ Array(1) ]
    Number of calls: 0
    FAIL collapses every zero-file language into ONE ledger row, not one per language
    ```
12. **`tree-sitter-query-loader.ts`: same gate → `if (true)`** (checks
    unconditionally, even on the common non-empty path) — 1/15 red:
    ```
    FAIL never classifies the bundled root on the common, non-empty path (typescript)
    AssertionError: expected "classifyBundledResourceDir" to not be called at all, but actually been called 1 times
    ```

All twelve mutations restored afterward; `git diff --stat HEAD` was empty
each time (confirmed via `cp`-based backup/restore, not `git checkout --`
against uncommitted work).

## Blast radius

- **`clients/bundled-resource-health.ts`** — new leaf module. Imports
  `degradation-ledger.js` and `user-notify.js`, both already
  process-lifetime leaves many `clients/` modules import; no new import
  cycle (`tests/config/dependency-boundaries.test.ts`'s cycle-baseline-count
  test passed unchanged).
- **`clients/ast-grep-client.ts`** — one call site changed (constructor).
  Only fires the extra check when `!ruleDir && !fs.existsSync(projectRuleDir)`
  — this repo's own real cwd has `rules/`, so this repo's own test/CI run
  never exercises the bundled branch except via the explicit `cwd` spy in
  the new test. `AstGrepClient()` is constructed once per host session
  (`index.ts`, `mcp/server.ts`, `clients/mcp/session.ts`), never per-edit —
  no hot path.
- **`clients/ast-grep-rule-manager.ts`** — `loadRuleDescriptions`'s own
  resolution logic is UNCHANGED (still reads `candidateAstGrepRulesPaths`
  the exact same way); only extracted the possible-paths list into a named
  function it now shares with the new `checkAstGrepRulesHealth`, and added
  the new export. Existing callers (`getRuleDescription`) unaffected.
- **`clients/cache/rule-cache.ts`** — `isBundledRuleFile`'s classification
  logic is UNCHANGED; the new call is purely additive in the constructor.
  `RuleCache` IS constructed per-dispatch (`dispatch/runners/tree-sitter.ts`),
  so the added work is bounded to: one cached-after-first-check status
  object property read when healthy (this repo's real state — no extra
  syscall after the first construction, same class as `package-root.ts`'s
  own `getPackageRoot` memo) — actually implemented as a plain classify
  call each time (see the module doc comment for why this is still bounded:
  it is the same order of magnitude as the `fs.existsSync` calls
  `ruleFilesForLanguage` already performs unconditionally on this path).
- **`clients/tree-sitter-query-loader.ts`** — `ruleFilesForLanguage`'s
  return value and per-language directory resolution are UNCHANGED; the new
  check runs ONLY inside the `files.size === 0` cold branch (proven never to
  fire on the common non-empty path — mutation proof 12), so the hot,
  healthy path (every real language with authored queries) pays nothing
  extra.
- **`clients/degradation-ledger.ts`** — additive: two new `DegradationKind`
  literals. No existing kind, exhaustiveness check, or
  `INFORMATIONAL_DEGRADATION_KINDS` entry touched.

## Observability

- `ast-grep-rules-dir-missing` degradation (`clients/degradation-ledger.ts`),
  recorded via `incrementDegradationCount`, visible in `pilens_health`'s
  `renderDegradationLines()`, plus one `notifyUserDegradation` warning
  (rising-edge gated). Subject: the resolved bundled `rules/` path.
- `tree-sitter-queries-dir-missing` degradation, same mechanism, shared by
  `RuleCache` and `ruleFilesForLanguage` (subject: the resolved bundled
  `rules/tree-sitter-queries` path — identical string from both sites, so
  the ledger naturally collapses duplicate reports of the same directory).
- `ast_grep_rules_resolved` phase/latency record (`clients/latency-logger.ts`,
  `latency.log`) — fires on every `AstGrepClient` construction that reaches
  the bundled-fallback branch, healthy or not, naming `filePath` (the
  resolved dir) + `metadata: { status, entryCount }`.
- `tree_sitter_queries_resolved` phase/latency record — fires inside
  `ruleFilesForLanguage`'s cold branch (zero files for that call), naming
  the bundled root path + `metadata: { languageId, status, entryCount }` —
  so a healthy-but-empty-for-this-language call (cobol) is distinguishable
  from a genuinely broken bundled root.

## Class sweep

Re-ran the issue's own `resolvePackagePath(` population sweep across the
whole tree (`grep -rn "resolvePackagePath(" clients/ tools/ mcp/ scripts/
index.ts`, excluding `.test.ts` and compiled `.js` twins): the population
and verdicts are UNCHANGED from #2637's table except the three rows this
issue targets, now fixed:

| Site | Verdict |
|---|---|
| `clients/ast-grep-client.ts` (`rules` fallback) | **fixed here** |
| `clients/cache/rule-cache.ts` (`BUNDLED_RULES_ROOT`) | **fixed here** |
| `clients/tree-sitter-query-loader.ts` (`rules/tree-sitter-queries/<lang>` + root) | **fixed here** |
| `clients/tree-sitter-client.ts` (`node_modules/web-tree-sitter/*`, `grammars`) | already safe (existsSync-guarded fallback chain + CDN fetch degradation) |
| `clients/tool-policy.ts` / `clients/lsp/server.ts` (external-tool `--config` args) | already safe (missing file errors loudly from the spawned tool) |
| `clients/sgconfig.ts` (bare `packageRoot`, no subpath) | not applicable (never read as a resource) |
| `scripts/benchmark-biome.ts` | not applicable (dev-only script) |

No new consolidation opportunity: the two `tree-sitter-queries`-root sites
(`rule-cache.ts`, `tree-sitter-query-loader.ts`) now share ONE
`DegradationKind` (see Summary), which is the fold the issue anticipated for
them; ast-grep's `rules/` dir is a physically different directory and keeps
its own kind.

## Test assessment

- `tests/clients/bundled-resource-health.test.ts` (new): pins the shared
  classify/report pair's every status + the once-per-session tally/notify
  split; no redundancy (new file).
- `tests/clients/ast-grep-rule-manager.test.ts` (new): pins
  `checkAstGrepRulesHealth`'s agreement with `loadRuleDescriptions`'s real
  resolution across every layout the mutation proofs above exercise; no
  redundancy (new file).
- `tests/clients/ast-grep-client-rules-health.test.ts` (new): pins the
  constructor's WIRING to the health check (branch gating, phase record,
  degradation, notify-once); no redundancy (new file).
- `tests/clients/cache/rule-cache.test.ts` (edited): the new describe block
  adds three cases with no overlap with the file's existing cache-roundtrip
  coverage.
- `tests/clients/tree-sitter-query-loader.test.ts` (edited): the new
  describe block adds four cases with no overlap with the file's existing
  metadata-parsing / `ruleFilesForLanguage` composition coverage.

## Verification

`npm run build` (clean); targeted run of the 17 files referencing the
touched symbols (`AstGrepClient`, `AstGrepRuleManager`,
`checkAstGrepRulesHealth`, `RuleCache`, `BUNDLED_RULES_ROOT`,
`ruleFilesForLanguage`, `BUNDLED_QUERIES_ROOT`,
`classifyBundledResourceDir`, `reportBundledResourceDirHealth`, the two new
degradation kinds) plus the mechanically-selected governance sweep
(`ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts`)
and every `tests/config/*.test.ts` — **75 files total**:

```
Test Files  4 failed | 71 passed (75)
     Tests  10 failed | 851 passed (861)
```

All 10 failures are `tests/clients/pi-host-contract.test.ts` (6) and
`tests/clients/console-capture-window-coverage.test.ts` (4) — both pinned
against a `node_modules/@earendil-works/pi-coding-agent` install this
worktree does not have (`ls node_modules/@earendil-works/` → no such
directory), a pre-existing environment gap unrelated to this diff
(`git diff --stat origin/master -- <those two files>` is empty). After
adding the new test case (mutation proof 9), re-ran the full 18-file symbol
set + `flake-shape-ratchet.test.ts`: **147/147 passed**.

`npm run lint` clean. `npx oxfmt --check` clean on every touched file (after
one auto-format pass). `node scripts/check-changelog-fragments.mjs` → OK
(134 entries). `npm run build:dist` succeeds.

## Fix round 1 (review verdict: NEEDS CHANGES)

Head at review: `fe4aef5b`. Reproduced every finding against that code before
fixing.

**Honest correction to my original report's Verification section:** I wrote
"4 failed files, 10 failed tests" and explained only the 2 files I could see
in a truncated `tail`. The other 2 were masked by an incomplete
`node_modules` in this worktree (this repo's own
`@earendil-works/pi-coding-agent` devDependency was entirely absent —
`ls node_modules/@earendil-works/` → no such directory — and a run without it
cascades into MORE than the 2 files whose names I quoted).
`tests/config/sweep-floor-coverage.test.ts` was one of the 2 I never named
(F1(b) below); fixed as part of this round. Per the reviewer's instruction,
this worktree's `node_modules` is now a symlink to the main checkout's
(`ln -s /home/akis/Desktop/pi-lens/node_modules node_modules`), so
`pi-host-contract.test.ts` and `console-capture-window-coverage.test.ts` now
run for real in every verification pass below.

**F1(b) — sweep-floor-coverage.test.ts red on this branch, green on
master.** Confirmed: `tests/clients/bundled-resource-health.test.ts`
matches the sweep-shape heuristic (a `node:fs` `readdirSync` mock +
`toHaveLength(0)`/similar emptiness assertions). Fix: added a
`DECLARED_EXCEPTIONS` entry, same reasoning #2637 used for
`skills-resolver.test.ts`. Re-running the full suite after the OTHER fixes
below ALSO flagged the edited `tree-sitter-query-loader.test.ts` (new
`readdirSync` mock added for F2/F6's real regression tests) — added a second
entry for it. Red-first (pre-fix):
```
FAIL sweep-floor meta-sweep: 1 flagged item(s) are neither registered nor exempted:
  tests/clients/bundled-resource-health.test.ts
```
Green after both entries: `tests/config/sweep-floor-coverage.test.ts` → 5/5.

**F2 — unbounded per-occurrence latency row on the warm path.** Confirmed:
`ruleFilesForLanguage`'s cold branch logged `tree_sitter_queries_resolved`
unconditionally, including on every dispatched file for one of the seven
REACHABLE by-design-empty grammars (bash, dart, elixir, lua, ocaml, swift,
zig — verified against `language-registry.ts`'s `EXTENSION_TO_GRAMMAR`; my
original comment named cobol/plsql, which are NOT in that registry at all —
fixed the comment, the changelog fragment, and the test language names to
the real seven). Fix: the phase record (and the degradation report) now
fire only when `health.status !== "healthy"`. Also confirmed and fixed the
mutation-vacuous claim: deleting the whole `logLatency` block left the
degradation-ledger assertions green because they never inspected
`latency.log`; added an explicit phase-entry assertion (mirroring
`ast-grep-client-rules-health.test.ts`'s pattern) that reds on that
deletion:
```
FAIL records a bounded degradation + notify + phase record when the bundled root is actually gone
AssertionError: expected [] to deeply equal [ ObjectContaining{…} ]
```
Mutation proof the NEW gate itself is load-bearing (`if (health.status !==
"healthy")` → `if (true)`):
```
FAIL records nothing for bash: zero files, but the REAL bundled root is healthy (no queries authored by design)
AssertionError: expected [ { … "status": "healthy" … } ] to deeply equal []
```

**F3 — `ast-grep-rules-dir-missing` destroyed at session start, never
re-recorded.** Confirmed: `AstGrepClient` is a per-process singleton
(`mcp/server.ts:570`, `index.ts:764`, `clients/mcp/session.ts:72`) and
`handleSessionStart` calls `resetDegradationLedger()` on every session
boundary, so a report that only ever fires at construction is invisible
after the first boundary — the skills row survives because
`resolveSkillPaths` runs per request, which `AstGrepClient` has no
equivalent of. Fix: `ensureRulesHealthReported` (new private method)
compares the ledger's `getDegradationLedgerGeneration()` against the last
generation it reported under — same pattern `tree-sitter-client.ts`'s
`refreshGrammarSessionLatches` already uses for its own per-instance session
latches — called at construction AND at the start of `ensureAvailable()`
(the real "start of a scan" entry point every `tools/ast-grep-*`/`ast-dump`
handler calls before doing any work). Red-first:
```
FAIL re-reports on the next scan after resetDegradationLedger() (session boundary), but not again within the same generation
```
(4 of 5 tests in the file reproduced red against the pre-fix constructor;
quoted the most direct one.) Mutation proof (the generation-compare guard
neutered, `if (this.rulesHealthReportedGeneration === generation) return;`
→ `if (true) return;`): 4/5 tests red, including the two PRE-EXISTING tests
("records a bounded degradation…", "does not re-notify…") — proving the
guard is load-bearing for the ORIGINAL behavior too, not just the new one.
Discovered a SEPARATE governance gap while wiring this: the compare tripped
`tests/clients/generation-guard-sweep.test.ts` (a hand-rolled generation
compare outside `clients/generation-guard.ts`). Reproduced it, confirmed the
shape is "clear-on-transition" (decides whether to RE-REPORT, never guards a
pending write against a stale generation) — the SAME exempted shape
`tree-sitter-client.ts`'s trust-notification set already carries in that
sweep's own exemption list — and added the matching exemption rather than
migrating to `createGenerationSource` (which models a different shape,
guarded writes racing an await).

**F4 — net-count: bundled-resource-health.ts duplicates skills-resolver.ts.**
Confirmed character-identical ENOENT/EACCES/absent/unreadable prose between
the two files. Fix: `classifyBundledResourceDir` now takes an optional
`countEntries` closure (default: bare directory-entry count) so a caller can
supply its OWN "what counts as found" predicate while sharing the
ENOENT/EACCES/empty classification shell; `describeBundledResourceHealth`
(now exported) takes an optional `emptyDescription` override;
`reportBundledResourceDirHealth` takes an optional `reasonOverride`.
`skills-resolver.ts` now calls `classifyBundledResourceDir(skillsDir, (dir)
=> scanEntriesForSkills(dir, readdirSync(dir, {withFileTypes:true}),
true).length)` and `reportBundledResourceDirHealth(...)` directly — its
LOCAL `SkillsHealth` type, `checkSkillsHealth`'s hand-rolled try/catch, and
`describeHealth` are all deleted. `tests/clients/skills-resolver.test.ts`
passes UNMODIFIED: **15/15**, `git diff --stat` on that file is empty.
Mutation proof the fold is real, not cosmetic (reverted `checkSkillsHealth`
to call `classifyBundledResourceDir(skillsDir)` with NO custom counter — the
generic default): 4/15 reproduced red —
```
FAIL H: skills/.hidden/SKILL.md is skipped (dot-dir), same as pi's own loader — records the degradation
AssertionError: expected undefined to be 1
```
(the other 3: a broken-symlink-at-root case, a subdirectory-loose-.md case, and
the node_modules-skip case — all cases where the generic bare-directory-count
predicate disagrees with `scanEntriesForSkills`'s real skip rules).

Also folded `BUNDLED_QUERIES_ROOT`/`BUNDLED_RULES_ROOT`/the third inline
`resolvePackagePath` spelling in `ruleFilesForLanguage`'s per-language loop
into ONE constant: `tree-sitter-query-loader.ts` is now the sole definition;
`rule-cache.ts` re-exports it as `BUNDLED_RULES_ROOT`
(`export { BUNDLED_QUERIES_ROOT as BUNDLED_RULES_ROOT } from
"../tree-sitter-query-loader.js";`), and `ruleFilesForLanguage`'s
per-language bundled path is now `path.join(BUNDLED_QUERIES_ROOT, lang)`.
The "one ledger row" claim now rests on reference equality, not two
independently-resolved strings that happen to match.

**F6 — unbounded per-occurrence classify call, claimed memo that didn't
exist.** Confirmed: `RuleCache`'s constructor called
`classifyBundledResourceDir(BUNDLED_RULES_ROOT)` fresh on every
construction (per dispatched file), contradicting my original PR body's
"same class as `package-root.ts`'s own `getPackageRoot` memo" claim — no
memo existed. Fix: `getBundledQueriesRootHealth()` (new export in
`tree-sitter-query-loader.ts`) memoizes the classification once
(`cachedBundledQueriesRootHealth`, a bare module-scope `let`, since the
underlying fact — does the package's own bundled directory exist — is a
process-lifetime install fact, not a session-scoped one); both
`RuleCache`'s constructor and `ruleFilesForLanguage`'s cold branch now call
it instead of `classifyBundledResourceDir` directly. Added
`_resetBundledQueriesRootHealthForTests()` for test isolation. Red-first/
mutation proof (memo condition neutered, `if
(!cachedBundledQueriesRootHealth)` → `if (true)`):
```
FAIL memoizes the bundled root's health across calls, even for different languages
AssertionError: expected [ …(3) ] to have a length of 1 but got 3
```
This new module-scope `let` tripped a SECOND governance sweep,
`tests/clients/session-state-conformance.test.ts` (the shared-state screen
the F3 finding itself pointed at) — reproduced it, confirmed the memo is
NOT session-scoped (same class as `project-scale.ts`/`slow-fs.ts`'s existing
exemptions: a host/install-derived fact a session boundary cannot move),
and added the matching `EXEMPT_SESSION_STATE_FILES` + `SESSION_STATE_SYMBOL_COUNTS`
entries rather than wiring a session-start reset that would defeat the
memo's whole purpose.

**F7 — `AstGrepClient` constructor's TOCTOU: `existsSync(projectRuleDir)`
called twice.** Confirmed by construction (reading the code): the ternary
re-evaluates `fs.existsSync(projectRuleDir)` independently of the
`usingBundledFallback` computation's own call. Fix: one
`const hasProjectRuleDir = fs.existsSync(projectRuleDir);`, reused for both.
Red-first proof (reverted to the two-call form, with a test that makes the
SECOND `existsSync(projectRuleDir)` call answer differently from the
first — simulating the directory appearing mid-construction):
```
FAIL checks the project rules/ directory exactly once (immune to it appearing mid-construction)
AssertionError: expected 3 to be 1
```
(3, not 2: the desync this finding describes ALSO fed the wrong directory
into `checkAstGrepRulesHealth`'s own candidate search, which then probed
`existsSync` on that same path a third time — direct evidence the race
produces exactly the misclassification F7 describes, not just an extra
syscall.)

### Round 1 verification

`npm run build` (clean); the 5 touched test files +
`tests/clients/skills-resolver*.test.ts` +
`tests/config/sweep-floor-coverage.test.ts` +
`tests/clients/bounded-telemetry-sweep.test.ts` +
`tests/clients/session-state-conformance.test.ts` +
`tests/clients/degradation-ledger*.test.ts` + every `tests/config/*.test.ts`
— **38 files, 474/474 passed**. Full symbol+governance batch (the same 75
files as the original PR, now with a real `node_modules`) — **75 files,
866/866 passed**, `pi-host-contract.test.ts` and
`console-capture-window-coverage.test.ts` included and green. `npm run
lint` clean. `npx oxfmt --check` clean on every touched file (one
auto-format pass). `node scripts/check-changelog-fragments.mjs` → OK (134
entries). `npm run build:dist` succeeds.

## Fix round 2 (review verdict: NEEDS CHANGES — new defect from round 1)

Head at review: `763d5900`. F1–F7 confirmed fixed and mutation-guarded; both
governance exemptions judged legitimate. Round 1's fold introduced three new
issues, addressed here.

**R2-F1 (user-visible regression) — the F4 fold reworded the shipped skills
notify sentence.** Confirmed: master notifies `pi-lens: registers zero
skills — <reason>.` (`git show origin/master:clients/skills-resolver.ts` —
line 136); round 1's fold silently switched it to the shared helper's
generic `pi-lens: skills unavailable — <reason>.` template. Confirmed M7:
reverting `checkSkillsHealth`'s custom `emptyDescription` to the generic
default left `skills-resolver.test.ts` 15/15 green — no test pinned the
skill-specific empty-reason text. Fix:
`reportBundledResourceDirHealth` (`clients/bundled-resource-health.ts`) now
takes an optional `notifyMessage` override; `skills-resolver.ts` passes
`` `pi-lens: registers zero skills — ${reason}.` `` explicitly. Added two
pins to `skills-resolver.test.ts`: `notified[0].message` must
`.toContain("registers zero skills")` (tests F and G), and the G (empty)
reason must `.toContain("no SKILL.md")`. Mutation proofs:
```
FAIL G: skills/ exists but is completely empty — records the degradation
AssertionError: expected '/tmp/…/skills exists but holds nothing (entry loaded from …) (count: 1)' to contain 'no SKILL.md'
```
(emptyDescription override removed — closes M7's exact gap)
```
FAIL F: entry copied out of the package … 
AssertionError: expected 'pi-lens: skills unavailable — no such directory: …' to contain 'registers zero skills'
FAIL G: skills/ exists but is completely empty …
AssertionError: expected 'pi-lens: skills unavailable — no SKILL.md …' to contain 'registers zero skills'
```
(notifyMessage override removed). No changelog change — the string does not
change from master.

**R2-F2 — the tree-sitter site's FAILURE path was the unbounded one.**
Confirmed: `ruleFilesForLanguage`'s cold branch logged
`tree_sitter_queries_resolved` once per DISPATCHED FILE while the bundled
root stayed broken — 200 `typescript`/`bash` touches would write 200 raw
`latency.log` rows beside the ledger's own 8 power-of-two-bounded rows
(`incrementDegradationCount`'s durable-write gate: `admitted &&
isPowerOfTwo(count)`, `clients/degradation-ledger.ts`). Fix: dropped the
phase/latency record entirely — the degradation ledger row already answers
"did this run" and "how many times" in a bounded way, and the record added
no field the ledger lacks once the (now-removed) `languageId` field is
weighed against 200x the log volume. `logLatency` import removed as
unused. Rewrote the test file's phase-record scaffolding (the `vi.mock`
wrapper and `resolvedPhaseEntries()` helper) away entirely, and replaced the
one test that exercised the OLD mutation-vacuous gap with a direct pin: 200
calls against a broken root still tally to `degradationGroup()?.count ===
200` (the ledger's exact in-memory total) while writing exactly one
grouped ledger entry, never a raw per-call record. Mutation proof (the
`reportBundledResourceDirHealth` call itself removed from the branch, to
confirm the new test actually exercises it):
```
FAIL records a bounded degradation + notify when the bundled root is actually gone
FAIL collapses every zero-file language into ONE ledger row, not one per language
FAIL tallies many occurrences into the ledger's exact count, never a raw per-call record
```

**R2-F3 — the F6 memo was keyed on `undefined`, probed once ever.**
Confirmed by reading the code: `cachedBundledQueriesRootHealth` was set once
on first access and NEVER invalidated, and the registry/comment claimed the
underlying fact "cannot change mid-process any more than the install
location can" — which overclaims: a managed extension cache relocating a
LIVE install mid-process is exactly the failure #2587/#2626 investigated, so
a permanently-cached "absent" (or "healthy") verdict from the first probe
would never notice the directory coming back (or disappearing) later in the
same process — the identical silent-zero shape this issue exists to end,
just latched at a different layer. Fix: `getBundledQueriesRootHealth` now
keys the memo on `getDegradationLedgerGeneration()` (bumped by
`resetDegradationLedger`, wired into `handleSessionStart`) — re-probing once
per SESSION, not once per call (still saves the 4.2 µs/call cost on the
per-dispatched-file hot path) and not once forever. Corrected the
overclaiming prose in both the source comment and the
`session-state-registry.ts` exemption reason. This is now the SAME
generation-keyed shape `ast-grep-client.ts`'s `ensureRulesHealthReported`
(round 1's F3 fix) already uses, as the finding predicted. Added a test and
mutation-proved it:
```
FAIL re-probes exactly once after a session boundary (resetDegradationLedger), not on every call within it
AssertionError: expected [ Array(1) ] to have a length of 2 but got 1
```
This also tripped `generation-guard-sweep.test.ts` a second time (the same
hand-rolled-generation-compare shape `ast-grep-client.ts` tripped in round
1) — added the matching "clear-on-transition, not a guarded write"
exemption there too.

### Rail deliverable — strings & records table

Every user-visible string and pushed record the three touched sites (plus
their shared skills sibling) emit. Grepped from the current head, not
recalled.

| Seam | master text/shape | head text/shape | Test line that pins it |
|---|---|---|---|
| skills notify (absent/unreadable/empty) | `pi-lens: registers zero skills — <reason>.` (`clients/skills-resolver.ts:136` on `origin/master`) | **identical** — `clients/skills-resolver.ts:123` | `tests/clients/skills-resolver.test.ts:274`, `:304` |
| skills empty reason | `no SKILL.md (or loadable .md) found under <dir>` | **identical** — `clients/skills-resolver.ts:99` | `tests/clients/skills-resolver.test.ts:302` |
| skills absent/unreadable reason | `no such directory: <dir>` / `cannot read <dir> (<code>)` | **identical** — shared via `clients/bundled-resource-health.ts:85`/`:87` | `tests/clients/skills-resolver.test.ts:264` (absent), `:432`/`:445` (EACCES/ENOENT split) |
| `skills_resolved` phase record | fires every call, `{status, entryCount}` | **identical** — `clients/skills-resolver.ts:85` | `tests/clients/skills-resolver.test.ts:141`, `:277` |
| ast-grep notify (new) | N/A — site did not exist on master | `pi-lens: ast-grep rule descriptions unavailable — <reason>.` (generic template, label `"ast-grep rule descriptions"`) | `tests/clients/ast-grep-client-rules-health.test.ts:160` (notify), `:157` (absent reason) |
| ast-grep empty/unreadable reason (new) | N/A | `<dir> exists but holds nothing` / `cannot read <dir> (<code>)` — generic, NOT overridden (`clients/ast-grep-client.ts:63-69` passes no 5th/6th arg) | pinned generically at `tests/clients/bundled-resource-health.test.ts:113-125` (`describeBundledResourceHealth` direct test); no ast-grep-specific literal-text pin for the empty/unreadable cases specifically — only `status` equality (`tests/clients/ast-grep-rule-manager.test.ts:69,107` for empty) |
| `ast_grep_rules_resolved` phase record (new) | N/A | fires on every bundled-fallback construction/scan, `{status, entryCount}` | `tests/clients/ast-grep-client-rules-health.test.ts:78` (helper), asserted at lines 135, 162 |
| tree-sitter notify (new) | N/A | `pi-lens: bundled tree-sitter query rules unavailable — <reason>.` (generic template, label `"bundled tree-sitter query rules"`) | `tests/clients/tree-sitter-query-loader.test.ts:490` |
| tree-sitter empty/unreadable reason (new) | N/A | generic, NOT overridden (`clients/tree-sitter-query-loader.ts:169-176`) | same generic pin as ast-grep's row above; no tree-sitter-specific literal pin for empty/unreadable (only the "absent" ENOENT path is exercised in `tree-sitter-query-loader.test.ts`, via the real fs-mock) |
| `tree_sitter_queries_resolved` phase record | N/A (didn't exist) | **DELETED this round (R2-F2)** — no phase record for this seam; the degradation-ledger row (`tree-sitter-queries-dir-missing`) is the only record | no positive pin (there is nothing to pin); `tests/clients/tree-sitter-query-loader.test.ts`'s "tallies many occurrences…" test (added this round) is the negative proof — 200 calls write no raw record, only one bounded ledger row |
| `ast-grep-rules-dir-missing` / `tree-sitter-queries-dir-missing` / `skills-dir-missing` ledger kinds | N/A / N/A / `skills-dir-missing` (identical) | as named, unchanged from round 1 | `tests/clients/ast-grep-client-rules-health.test.ts:70-73`, `tests/clients/tree-sitter-query-loader.test.ts` `degradationGroup()`, `tests/clients/skills-resolver.test.ts` `skillsDegradationGroup()` |

**Gap named, not hidden:** the ast-grep and tree-sitter sites' `empty` and
`unreadable` reason TEXT (as opposed to `status`) has no site-specific
literal pin — only the shared `describeBundledResourceHealth` function's
generic output is pinned (new this round, `bundled-resource-health.test.ts`),
and each call site is confirmed (by reading the source, not a test) to pass
no override for those two statuses. Given neither site ships a
site-specific empty/unreadable string (unlike skills), the generic pin is
the correct level — a site-specific literal pin here would just re-assert
the same generic template's output under a different file, which is not
what M7 asked for (M7 was specifically about an override being silently
DROPPABLE without any test noticing, and there IS no override to drop for
these two statuses on these two sites).

### Round 2 verification

Merged `origin/master` (`bbb4c5cf`) first — clean, no conflicts (`git
status` after merge showed nothing but the pending `node_modules` symlink).
`npm run build` clean. The reviewer's named test list (38 files) —
**477/477 passed**. Full symbol+governance batch (75 files, including
`pi-host-contract.test.ts`) — **869/869 passed**. `npm run lint` clean.
`npx oxfmt --check` clean on every touched file. `node
scripts/check-changelog-fragments.mjs` → OK (141 entries — master's own
merged fragments added 7). `npm run build:dist` succeeds.


apmantza:
Orchestrator note from the r3 verify (merge-ready; no fix round): (1) the strings-and-records table omits the fourth emission site, `clients/cache/rule-cache.ts:91` (`reportBundledResourceDirHealth` from the `RuleCache` constructor, pinned by `tests/clients/cache/rule-cache.test.ts`; strings identical to the loader's; no phase record by the R2-F2 reasoning). (2) The tree-sitter leg is now silent on a healthy root by design; a bounded success record on the memo-miss branch is filed separately.

PR #2680, opened 2026-09-06
fix(compat): locate relocated compat-contract sources independently (closes #2581)

### LINKED ISSUE #2581: compat-smoke: third-party contract drift detected
The nightly [subagent-extension compat smoke](https://github.com/apmantza/pi-lens/actions/runs/34035878832) detected drift in a pinned third-party contract used by pi-lens's subagent-compatibility features (#473/#474/#475).

- Layer A (pinned-contract verification): **failure**
- Layer B (real-pi behavioral smoke): **success**
- Installed versions this run: @earendil-works/pi-coding-agent@0.85.1 pi-subagents@0.66.0 avtc-pi-subagent@1.0.8 @tintinweb/pi-subagents@0.19.0

See the [run log](https://github.com/apmantza/pi-lens/actions/runs/34035878832) for failed checks, and `docs/subagent-compat.md` for the contracts and files each check verifies.

_This issue is auto-refreshed by the nightly compat-smoke workflow — do not close it while the check is failing. Once drift is resolved (upstream revert or pi-lens update), close it._

apmantza:
## Summary

The nightly compat-smoke Layer A (`scripts/compat-contracts.mjs`) was reporting `INFRA FAILURE` and getting logged into the tracking issue as generic "drift detected" for a plain `ENOENT`: `pi-subagents@0.65.0` rewrote subagents to run as native, in-process `AgentSession`s instead of spawning a separate `pi` CLI process per child, and relocated `src/runs/shared/pi-args.ts` (the file the `nicobailon.child-env` contract read) into two different files in the process. The old script read all five contract-backing files inside one shared `try`/`catch`, so that single `ENOENT` aborted the run before the other four files (`avtc`/`sdk`×3/`tintinweb`) — all fine — ever got checked, and exited 2 with no per-contract detail.

Root cause and fix, per the issue's four tasks:

1. **Where the contract moved.** Downloaded and extracted the real `pi-subagents@0.66.0` tarball (matching the failing run's installed versions exactly). The `SUBAGENT_CHILD_ENV` const moved to `src/runs/shared/child-runtime-config.ts`; the `env[SUBAGENT_CHILD_ENV] = "1"` assignment moved to `src/runs/background/subagent-runner.ts` — two files now, where `pi-args.ts` used to hold both. Separately, grep across the entire 0.66.0 source tree found `PI_SUBAGENT_RUN_ID` and `PI_SUBAGENT_CHILD_AGENT` (the contract's other two required consts) **gone entirely** — not relocated, removed. Child identity now travels through an in-process `ChildRuntimeConfig` object because foreground children no longer spawn a separate process at all.
2. **Robust locate + infra-vs-drift split.** New `scripts/lib/compat-contract-locator.mjs`: pure `locateContractSource`/`locateContractSources`, each contract's file(s) resolved through an ordered candidate-path list (oldest-observed-layout first, each candidate carrying the version it was observed at). `scripts/compat-contracts.mjs` now resolves and checks all seven contracts **independently** — one relocated file can no longer blind the other six — and reports a three-way `outcome` (`verified`/`drift`/`infra`) per contract and for the run, surfaced as a `GITHUB_OUTPUT` the workflow's alert step reads to word the tracking-issue body correctly instead of always saying "drift detected". `docs/subagent-compat.md` documents all three outcomes.
3. **Re-verified the actual contract.** Ran the (now-fixed) Layer A script against a real scratch install of the exact four versions from the failing run — see transcript below. All seven contracts **VERIFY cleanly**. The one genuine content drift (the removed identity consts) needed no pi-lens *behavior* fix: `getSubagentIdentity()` was already documented/tested as best-effort, degrading to `runId`/`agentName: undefined` when absent — that permanent-now state is exactly what it already handled. The fix on the pi-lens side is dropping the now-permanently-unsatisfiable requirement from `checkNicobailonChildEnv` (keeping only the `PI_SUBAGENT_CHILD='1'` check pi-lens's light-mode detection actually depends on) — updating the check, not the pin, since re-adding the removed vars is not something bumping the pin would ever fix.
4. **Red-first.** `tests/scripts/compat-contract-locator.test.ts` (new, 8 cases) exercises the locator against real on-disk fixture package trees: file at the original path, file moved to the new path, file absent everywhere, and the multi-file "co-located vs split across two files" cases the 0.65.0 rewrite introduced. `checkNicobailonChildEnv` proven red against the real (relocated, concatenated) 0.66.0 source using the **pre-fix** function (quoted below), green after.

Closes #2581 — every contract verifies clean after relocation.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:tests
- [ ] area:observability

## Checklist

- [x] `npm run lint` passes
- [ ] `npm run build:dist` succeeds if I changed code under `clients/`, `commands/`, `tools/`, or `index.ts` — N/A, no files under those paths touched
- [x] `package-lock.json` is in sync with `package.json` (no dependency changes)
- [ ] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there — its compat-smoke paragraph (~line 2745) still accurately describes the identity/marker behavior; the outcome-classification detail lives in `docs/subagent-compat.md`, which IS updated
- [x] `.changelog/2581-compat-contract-locator.md` added
- [x] Commit subjects carry `#2581`

## Tests

- **`tests/scripts/compat-contract-locator.test.ts` (new, 8 cases; now 9 after round 1's F1 fix, see below)** — `locateContractSource`: finds the file at its original pinned path, at a relocated path, originally asserted "prefers the FIRST existing candidate" — **flipped to newest-first in review round 1 (F1)**, reports `found:false` with every tried candidate when absent everywhere, and when the package dir itself doesn't exist. `locateContractSources`: concatenates parts resolved from two *different* files (the 0.65.0 split), resolves a single co-located file for both parts (the pre-0.65.0 layout), and reports `found:false` naming the specific unresolvable part rather than silently running a partial concatenation.
- **`tests/scripts/compat-contracts.test.ts` (edited)** — `checkNicobailonChildEnv`'s fixture dropped the two now-removed identity consts; replaced the "fails when run-id const missing" case with "passes even when the (now-removed upstream) identity consts are absent" and kept "fails when the const definition is missing" / "fails when the child-flag assignment is missing" as the real regression guards. Aggregation tests were originally against `runAllContractChecks` — **superseded in review round 1 (F3/F4), see that section below.**
- **`tests/clients/subagent-mode.test.ts` (edited, +1 case)** — added "still classifies as a subagent with marker set when PI_SUBAGENT_CHILD=1 is the ONLY var present (pi-subagents@0.65.0+ reality)". This is a **confirming test, not a red-first fix**: the code already passed it (best-effort degradation was already correct) — stated per the red-first exception policy, since forcing an artificial pre-fix regression here would just re-derive a test with no defect to catch. It closes the coverage gap the re-verification step (#2581 task 3) surfaced: the "child flag set, zero identity vars" shape is now the *permanent* nicobailon reality, not a hypothetical.

None of AGENTS.md's ten test-authoring screens apply adversely — the locator tests use real on-disk fixtures (no ambient-inspection double), real candidate lists (no loose bound), and each asserts a distinct outcome (no snapshot-as-behavior).

### Test assessment

- **`tests/scripts/compat-contract-locator.test.ts` (new; edited round 1 for F1)** — pins the candidate-path resolution behavior: original-path hit, relocated-path hit, multi-file concatenation, and the distinct "found:false, name which part" case. Round 1 flipped "prefers the FIRST existing candidate" to "prefers the NEWEST" (F1) and added the reviewer's exact stale-leftover-with-satisfying-content probe as its own case — the two newest-first cases have independent signatures (one checks which path/content wins on a plain conflict, the other additionally proves a check-passing stale file still loses). Nothing else in the suite covers this; no removal candidates.
- **`tests/scripts/compat-contracts.test.ts` (edited; edited again round 1 for F3/F4)** — still uniquely pins each pure regex check's pass/fail boundary. Round 1 deleted the `runAllContractChecks` describe block (F3 — the function it tested no longer exists) and replaced the near-duplicate "identity consts absent" case with a real 0.66.0-shaped two-file concatenation (F4) — distinct from the plain-`GOOD`-fixture case above it (disjoint fixture strings; the new one additionally proves cross-file concatenation, which the old duplicate never exercised). No other case in this file duplicates another's assertion.
- **`tests/scripts/compat-contract-resolution.test.ts` (new, round 1, F2)** — uniquely pins `resolveAndCheckContracts`'s per-contract independence (verified/drift/infra together in one run) and the structural elimination of the two-registry join F2 found — a case renaming one fixture contract's `package` field alone proves there is no second table left for that rename to desync from. Nothing else in the suite covers `resolveAndCheckContracts` directly; no removal candidates.
- **`tests/clients/subagent-mode.test.ts` (edited)** — the new case pins a specific env-shape (`PI_SUBAGENT_CHILD=1` alone, no identity vars) that no existing case in the file covered; the closest existing case ("carries marker 'pi-subagents' and agentName when PI_SUBAGENT_CHILD=1") always sets `PI_SUBAGENT_CHILD_AGENT` too, so it doesn't red on this shape (verified: reverting the new case's assertions to expect `agentName: "code-reviewer"` instead of `undefined` fails only the new test). No removal candidates.

### Red-first proof (checkNicobailonChildEnv against real pi-subagents@0.66.0 source)

Pre-fix function (`git show a85c6893~1:scripts/lib/compat-contracts.mjs`) against the real, concatenated `child-runtime-config.ts` + `subagent-runner.ts` from a live `pi-subagents@0.66.0` install:

```
PRE-FIX checkNicobailonChildEnv against real pi-subagents@0.66.0 source:
{
  "pass": false,
  "detail": "missing: PI_SUBAGENT_RUN_ID const, PI_SUBAGENT_CHILD_AGENT const"
}
```

Post-fix, same input:

```
POST-FIX checkNicobailonChildEnv against real pi-subagents@0.66.0 source:
{
  "pass": true,
  "detail": "PI_SUBAGENT_CHILD='1' set unconditionally (RUN_ID/CHILD_AGENT identity vars no longer required — removed upstream at pi-subagents@0.65.0, #2581; getSubagentIdentity() already degrades to unknown)"
}
```

### Reproducing the exact nightly failure (old orchestrator, real install, real versions)

Pre-fix `scripts/compat-contracts.mjs` (`git show a85c6893~1:scripts/compat-contracts.mjs`) run against a scratch install pinned to the exact reported versions:

```
versions installed:
  @earendil-works/pi-coding-agent@0.85.1
  pi-subagents@0.66.0
  avtc-pi-subagent@1.0.8
  @tintinweb/pi-subagents@0.19.0

INFRA FAILURE — expected source file not found (package layout changed?): ENOENT: no such file or directory, open '.../scratch-install/node_modules/pi-subagents/src/runs/shared/pi-args.ts'
EXIT CODE: 2
```

Matches the reported nightly log line verbatim.

### Post-fix, same real install, per-contract outcome (issue's task 4: "quote the outcome per contract")

```
versions installed:
  @earendil-works/pi-coding-agent@0.85.1
  pi-subagents@0.66.0
  avtc-pi-subagent@1.0.8
  @tintinweb/pi-subagents@0.19.0

contract checks:
  [PASS] nicobailon.child-env (pi-subagents) — PI_SUBAGENT_CHILD env var set on every child-hosting process
         PI_SUBAGENT_CHILD='1' set unconditionally (RUN_ID/CHILD_AGENT identity vars no longer required — removed upstream at pi-subagents@0.65.0, #2581; getSubagentIdentity() already degrades to unknown)
  [PASS] avtc.child-env (avtc-pi-subagent) — PI_SUBAGENT_CHILD_AGENT + PI_SUBAGENT_PARENT_PID pair set on every spawned child
         PI_SUBAGENT_CHILD_AGENT + PI_SUBAGENT_PARENT_PID both assigned on the per-spawn subagent env
  [PASS] sdk.extension-cache (@earendil-works/pi-coding-agent) — process-global extensionCache Map in the extension loader
         process-global `extensionCache = new Map()` present
  [PASS] sdk.bind-extensions-session-start (@earendil-works/pi-coding-agent) — bindExtensions() unconditionally emits a session_start-typed event
         bindExtensions() unconditionally emits a session_start-typed event
  [PASS] sdk.invalidate-called (@earendil-works/pi-coding-agent) — invalidate() called from the sequential session-replacement path
         _extensionRunner.invalidate(...) call site found
  [PASS] sdk.stale-ctx-message (@earendil-works/pi-coding-agent) — stale-ctx error message contains "stale after session replacement"
         stale-ctx message contains "stale after session replacement"
  [PASS] tintinweb.in-process-bind (@tintinweb/pi-subagents) — constructs DefaultResourceLoader + calls bindExtensions() in-process
         constructs DefaultResourceLoader + calls session.bindExtensions() in-process

ALL CONTRACT CHECKS VERIFIED
EXIT CODE: 0
```

### Mutation proof

`scripts/lib/compat-contract-locator.mjs`'s `if (fs.existsSync(filePath))` guard, both directions, against `tests/scripts/compat-contract-locator.test.ts`:

- Forced `true` (always "found"): **5 of 8 tests fail** — `fs.readFileSync` throws `ENOENT` on the absent-file cases, since the guard no longer actually checks existence.
- Forced `false` (never "found"): **6 of 8 tests fail** — every "should find it" case now reports `found:false`.

Both directions confirm the guard is load-bearing and each new locator test has its own signature.

**Orchestrator per-contract independence** (the actual defect this PR fixes) is proven by the before/after contrast above rather than a synthetic mutation of the fixed code — mutating my own new `.map()`-based loop back into the old shared-try/catch would just re-derive the pre-fix script I already quoted verbatim. Additionally verified with a mixed real-install fixture (one file deleted → `infra`, one file's content corrupted → `drift`, five untouched):

```
  [PASS] nicobailon.child-env ...
  [FAIL] avtc.child-env (avtc-pi-subagent) — ...
         missing: PI_SUBAGENT_CHILD_AGENT assignment, PI_SUBAGENT_PARENT_PID = String(process.pid) assignment
  [PASS] sdk.extension-cache ...
  [PASS] sdk.bind-extensions-session-start ...
  [PASS] sdk.invalidate-called ...
  [PASS] sdk.stale-ctx-message ...
  [INFRA] tintinweb.in-process-bind (@tintinweb/pi-subagents) — ...
         INFRA — expected source not found for "source" (package layout changed?): tried src/agent-runner.ts (observed at 0.13.0)

ONE OR MORE CONTRACT CHECKS FAILED (drift)
```
Exit code **1** (drift takes priority). With only the tintinweb file removed (no drift anywhere), exit code is **2** (infra-only) — confirmed separately. With neither mutation, exit code is **0**.

### Governance / targeted run

```
Test Files  61 passed (61)
     Tests  808 passed (808)
```
(3 directly-relevant files: `tests/scripts/compat-contracts.test.ts`, `tests/scripts/compat-contract-locator.test.ts`, `tests/clients/subagent-mode.test.ts`; plus the full mechanically-selected governance set — every `tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts` and every `tests/config/*.test.ts`, 58 files.)

## Blast radius

No `clients/`, `commands/`, `tools/`, `mcp/`, or `index.ts` files touched — this PR is confined to `scripts/`, `tests/scripts/`, `tests/clients/subagent-mode.test.ts` (test-only edit), `docs/`, `.github/workflows/`, and `.changelog/`. No production runtime module_report blast radius applies. The nightly workflow (`compat-smoke.yml`) is the only consumer of `scripts/compat-contracts.mjs`; its `continue-on-error: true` means even a bug in this PR degrades to "alert misfires" at worst, never a red gate on this or any other PR.

## Observability

The nightly's `outcome` (`verified`/`drift`/`infra`) is now a `GITHUB_OUTPUT` from `scripts/compat-contracts.mjs`, read by the workflow's alert step and printed into the tracking issue body — quoted end-to-end above (`ALL CONTRACT CHECKS VERIFIED`, exit 0, `outcome=verified` written to `GITHUB_OUTPUT` when set) and exercised for all three values via the mixed-fixture and infra-only runs above. No new production (`clients/`) log/ledger record — this is test/CI tooling only.

## Class sweep

**Shape:** a script hardcodes a single literal path into a third-party package's internal source tree for content verification, with no fallback, inside a `try`/`catch` shared across multiple unrelated reads — one relocation both silently degrades that ONE check from "verified" to "couldn't check" AND aborts every sibling check in the same `try` block.

Swept the whole tree for the same shape: `grep -rln "readFileSync" scripts/*.mjs | xargs grep -l "node_modules"` → 12 scripts touch `node_modules` paths at all (`compat-smoke-behavioral.mjs`, `bundle-dist.mjs`, `check-pr-body.mjs`, `scan-ast-grep-rules.mjs`, `smoke-tools.mjs`, `check-extensions.mjs`, `check-lockfile-sync.mjs`, `install-selftest.mjs`, `build-dist-tsc.mjs`, `release-qa.mjs`, `prune-agent-worktrees.mjs`, plus `compat-contracts.mjs` itself); none of the other 11 read a fixed *internal* third-party source-file path for pattern-matching against — they check package existence/version/`package.json`, which don't have this relocation exposure. `scripts/lib/warm-loader-cache.mjs` also references `dist/core/extensions/loader.js` in comments, but resolves it through Node's own module resolution (jiti/require), not a hand-built literal path — a different, already-resilient mechanism.

**Verdict: class of size 1.** `compat-contracts.mjs` was the only site with this exact shape; fixed at its only occurrence. No consolidation needed — `compat-contract-locator.mjs` is itself the reusable seam should a second site ever need it.

## Review round 1

Reviewer cleared the identity-var drop finding entirely (four tarballs 0.64.0–0.66.0 confirmed removal, `isSubagentSession()` only ever reads `PI_SUBAGENT_CHILD`, every `getSubagentIdentity()` consumer treats identity as optional — #2581 closes on that axis) and returned five findings. All five addressed on this branch.

**F1 (high, locator oldest-first + first-match-wins) — fixed.** `locateContractSource` now walks candidates NEWEST-observed first (`scripts/lib/compat-contract-locator.mjs`); candidate lists stay authored oldest-first (append the newest layout) — only the walk order changed. Reproduced the reviewer's exact probe as a new test (a stale leftover `pi-args.ts` whose content still satisfies the check, alongside a live `child-runtime-config.ts` that has dropped it) and a simpler "prefers newest when both exist" case.

Red-first, reverting to oldest-first iteration:
```
 ❯ prefers the NEWEST existing candidate when both exist (stale leftover must never outrank the live layout)
   AssertionError: expected 'src/runs/shared/pi-args.ts' to be 'src/runs/shared/child-runtime-config....'
 ❯ resolves the live (newest) file even when the stale leftover's content would satisfy a check and the live one wouldn't
   AssertionError: expected 'src/runs/shared/pi-args.ts' to be 'src/runs/shared/child-runtime-config....'
 Test Files  1 failed (1)
      Tests  2 failed | 7 passed (9)
```
`docs/subagent-compat.md` updated: candidates are walked newest-first; new entries are still appended (not prepended) since resolution order no longer depends on authored order.

**F2 (high, two parallel registries joined by a string id) — fixed at the root.** `parts` (candidate paths) and `package` (the exact npm install spec, not a lookup key) are now folded directly into each `CONTRACTS` entry in `scripts/lib/compat-contracts.mjs` — one table, no second `CONTRACT_SOURCE_LOCATIONS` map, no string join to desync. `resolveAndCheckContracts` moved to a new, directly-testable `scripts/lib/compat-contract-resolution.mjs` (no npm install, no process.exit) and is covered by `tests/scripts/compat-contract-resolution.test.ts` against a real on-disk 3-contract fixture exercising verified/drift/infra together, plus a case renaming one contract's `package` field alone (proving there's no second table left to desync from).

Reproduced the reviewer's exact failure mode against the PRE-fix orchestrator (a phantom 8th `CONTRACTS` entry with no matching `CONTRACT_SOURCE_LOCATIONS` entry, run against a real install):
```
EXIT: 2
compat-contracts.mjs crashed: TypeError: Cannot read properties of undefined (reading 'packageKey')
    at .../scripts/compat-contracts.mjs:214:41
    at Array.map (<anonymous>)
    at resolveAndCheckContracts (.../scripts/compat-contracts.mjs:212:19)
```
Same install, same phantom-contract shape, against the fixed (folded) design: the shape can no longer be expressed at all — there's nowhere to add a contract without its `parts` right there in the same object literal. Fixture test:
```
Test Files  1 passed (1)
     Tests  3 passed (3)
```

**F3 (medium, dead `runAllContractChecks` + `inputKey`) — deleted.** Removed `runAllContractChecks` and the `inputKey` field it existed only to feed from `scripts/lib/compat-contracts.mjs` and its `.d.mts`. Its two aggregation tests in `tests/scripts/compat-contracts.test.ts` are replaced by a comment pointing at `resolveAndCheckContracts` (the seam that ships) and its new fixture test file, which covers the same aggregation behavior against real per-contract resolution instead of a hand-built inputs object nothing in production built.

**F4 (medium, verbatim duplicate test) — given a distinct input.** `checkNicobailonChildEnv`'s "identity consts absent" case now uses a concatenated two-file fixture shaped like the REAL 0.66.0 layout (const in one file, assignment importing it in another) instead of the single-snippet `GOOD` fixture the first case already used — the input this whole round proved the check against, not a coincidental restatement. Confirmed the two cases now have independent signatures: they use disjoint fixture strings, and the second explicitly asserts the cross-file-concatenation shape the first never exercises.

**F5 (low, stale doc claims in `clients/subagent-mode.ts`) — fixed.** Module docstring and `getSubagentIdentity()`'s docstring updated to state the 0.65.0 removal as settled fact (not "a future extension that doesn't set them") and clarify identity is optional metadata that never gates light mode.

**F5's low note (alert issue title still says "drift detected" on an infra outcome) — left as-is, stated explicitly.** Adding a per-kind label (`outcome:infra`/`outcome:drift`) would require the workflow to also ensure those labels exist in the repo first (`gh issue create --label` fails on a label that doesn't exist; none of `infra`/`drift`/`outcome:*` currently exist — checked via `gh label list`) — repo-administration surface a contract-verification workflow doesn't otherwise touch. The title-search dedupe itself is unaffected either way (it matches on the literal title string, not labels), and the alert BODY (fixed in the original PR push) already states the real per-layer outcome distinctly from the title. Leaving the title generic and letting the body carry the specific outcome.

### Re-verification after all fixes

Real scratch install, exact reported versions, after every fix:
```
versions installed:
  @earendil-works/pi-coding-agent@0.85.1
  pi-subagents@0.66.0
  avtc-pi-subagent@1.0.8
  @tintinweb/pi-subagents@0.19.0
...
ALL CONTRACT CHECKS VERIFIED
EXIT: 0
```

### Tests run this round

Three original targeted files + the new orchestrator test + `tests/config/*` + `flake-shape-ratchet`, per the coordinator's instruction:
```
Test Files  32 passed (32)
     Tests  332 passed (332)
```
Full governance sweep re-run (58 files, mechanically selected — same list as the original push):
```
Test Files  58 passed (58)
     Tests  753 passed (753)
```
`npm run lint` and `npx oxfmt --check` both clean on every touched file.

New/changed files this round: `scripts/lib/compat-contract-locator.mjs` (F1), `scripts/lib/compat-contracts.mjs` + `.d.mts` (F2/F3), `scripts/lib/compat-contract-resolution.mjs` + `.d.mts` (new, F2), `scripts/compat-contracts.mjs` (F2), `tests/scripts/compat-contract-locator.test.ts` (F1), `tests/scripts/compat-contracts.test.ts` (F3/F4), `tests/scripts/compat-contract-resolution.test.ts` (new, F2), `docs/subagent-compat.md` (F1/F2 doc updates), `.github/workflows/compat-smoke.yml` (F2 doc reference), `clients/subagent-mode.ts` (F5).

### DIFF
--- a/.changelog/2581-compat-contract-locator.md
+++ b/.changelog/2581-compat-contract-locator.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **compat-smoke Layer A resolves each pinned contract's source file independently (refs #2581)** — `pi-subagents@0.65.0` relocated the file behind the `PI_SUBAGENT_CHILD` env-var contract and removed the `PI_SUBAGENT_RUN_ID`/`PI_SUBAGENT_CHILD_AGENT` identity vars entirely, so the nightly's Layer A read a single hardcoded path inside one shared try/catch and reported an `INFRA FAILURE` (a plain `ENOENT`) as generic contract drift for the whole run, blinding verification of the other six contracts too. `scripts/compat-contracts.mjs` now resolves each contract's source through an ordered candidate-path list (`scripts/lib/compat-contract-locator.mjs`) and reports one of three outcomes per contract — verified / drift / infra — instead of a single pass/fail; `docs/subagent-compat.md` documents all three. The identity-var requirement was dropped from `checkNicobailonChildEnv` since pi-lens's `getSubagentIdentity()` already degrades gracefully when they're absent — no runtime behavior changed.
--- a/.github/workflows/compat-smoke.yml
+++ b/.github/workflows/compat-smoke.yml
@@ -137,9 +137,10 @@ jobs:
               LAYER_A_LINE+=" moved/renamed the file, or the install itself"
               LAYER_A_LINE+=" failed). NOT a confirmed contract regression —"
               LAYER_A_LINE+=" see the run log for which candidates were"
-              LAYER_A_LINE+=" tried, then add the new location to"
-              LAYER_A_LINE+=" \`scripts/compat-contracts.mjs\`'s"
-              LAYER_A_LINE+=" \`CONTRACT_SOURCE_LOCATIONS\`."
+              LAYER_A_LINE+=" tried, then append the new location to that"
+              LAYER_A_LINE+=" contract's \`parts\` in"
+              LAYER_A_LINE+=" \`scripts/lib/compat-contracts.mjs\`'s"
+              LAYER_A_LINE+=" \`CONTRACTS\` list."
               ;;
             verified)
               LAYER_A_LINE="verified — passed (this alert fired because of"
--- a/.github/workflows/compat-smoke.yml
+++ b/.github/workflows/compat-smoke.yml
@@ -93,17 +93,28 @@ jobs:
           ANTHROPIC_API_KEY: sk-ant-dummy-ci-no-credits-needed
         run: node scripts/compat-smoke-behavioral.mjs
 
-      # Low-noise alert: only fires when a layer failed for a REAL reason
-      # (exit 1 = contract/assertion drift), not an infra error (exit 2) —
-      # an infra failure means this run couldn't test anything meaningful,
-      # so alerting on it would be noise, not signal. Searches for the
-      # existing open alert issue by title before creating a new one so
-      # repeated nightly failures refresh one issue instead of piling up.
-      - name: Alert on contract drift (create-or-update tracking issue)
+      # Alert whenever either layer's step outcome is non-success — this
+      # covers BOTH real content drift and an infra failure (a dependency's
+      # file layout moved, npm install failed, ...): an infra failure means
+      # Layer A couldn't actually re-check anything, which is itself worth a
+      # maintainer's attention, just a DIFFERENT one than drift. #2581: the
+      # body used to print the raw GH step outcome ("failure") for both
+      # cases under a title that always says "drift detected" — a
+      # relocated file (ENOENT) got reported and re-confirmed as drift for
+      # two nightly runs before anyone opened the log and found it was
+      # never a contract regression. Layer A's own `outcome` output
+      # (verified/drift/infra from compat-contracts.mjs, distinct from the
+      # GH step outcome above) drives the wording below so the two are never
+      # conflated again. Searches for the existing open alert issue by title
+      # before creating a new one so repeated nightly failures refresh one
+      # issue instead of piling up; the title itself stays constant so that
+      # search keeps finding it regardless of which outcome kind is current.
+      - name: Alert on contract drift or infra failure (create-or-update tracking issue)
         if: >-
           steps.layer_a.outcome == 'failure' || steps.layer_b.outcome == 'failure'
         env:
-          LAYER_A_OUTCOME: ${{ steps.layer_a.outcome }}
+          LAYER_A_STEP_OUTCOME: ${{ steps.layer_a.outcome }}
+          LAYER_A_KIND: ${{ steps.layer_a.outputs.outcome || 'unknown' }}
           LAYER_B_OUTCOME: ${{ steps.layer_b.outcome }}
           LAYER_A_VERSIONS: >-
             ${{ steps.layer_a.outputs.versions ||
@@ -112,24 +123,53 @@ jobs:
           set -uo pipefail
           TITLE="compat-smoke: third-party contract drift detected"
           BODY_FILE="$RUNNER_TEMP/compat-smoke-alert-body.md"
+
+          case "$LAYER_A_KIND" in
+            drift)
+              LAYER_A_LINE="**DRIFT** — a pinned contract's installed source"
+              LAYER_A_LINE+=" no longer matches the shape pi-lens depends on."
+              LAYER_A_LINE+=" This is a real regression to investigate."
+              ;;
+            infra)
+              LAYER_A_LINE="**INFRA FAILURE** — Layer A could not locate one"
+              LAYER_A_LINE+=" or more contracts' source files at any known"
+              LAYER_A_LINE+=" candidate path (a dependency's layout likely"
+              LAYER_A_LINE+=" moved/renamed the file, or the install itself"
+              LAYER_A_LINE+=" failed). NOT a confirmed contract regression —"
+              LAYER_A_LINE+=" see the run log for which candidates were"
+              LAYER_A_LINE+=" tried, then add the new location to"
+              LAYER_A_LINE+=" \`scripts/compat-contracts.mjs\`'s"
+              LAYER_A_LINE+=" \`CONTRACT_SOURCE_LOCATIONS\`."
+              ;;
+            verified)
+              LAYER_A_LINE="verified — passed (this alert fired because of"
+              LAYER_A_LINE+=" Layer B, see below)."
+              ;;
+            *)
+              LAYER_A_LINE="unknown (step outcome: ${LAYER_A_STEP_OUTCOME};"
+              LAYER_A_LINE+=" the script did not report an outcome — check"
+              LAYER_A_LINE+=" the run log directly)."
+              ;;
+          esac
+
           {
             echo "The nightly [subagent-extension compat smoke](https://github.com/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID})" \
-              "detected drift in a pinned third-party contract used by" \
-              "pi-lens's subagent-compatibility features (#473/#474/#475)."
+              "flagged pi-lens's subagent-compatibility contracts" \
+              "(#473/#474/#475)."
             echo ""
-            echo "- Layer A (pinned-contract verification):" \
-              "**${LAYER_A_OUTCOME}**"
+            echo "- Layer A (pinned-contract verification): ${LAYER_A_LINE}"
             echo "- Layer B (real-pi behavioral smoke): **${LAYER_B_OUTCOME}**"
             echo "- Installed versions this run: ${LAYER_A_VERSIONS}"
             echo ""
             echo "See the [run log](https://github.com/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID})" \
               "for failed checks, and \`docs/subagent-compat.md\` for the" \
-              "contracts and files each check verifies."
+              "contracts and files each check verifies, including the" \
+              "verified / drift / infra outcome definitions."
             echo ""
             echo "_This issue is auto-refreshed by the nightly compat-smoke" \
               "workflow — do not close it while the check is failing. Once" \
-              "drift is resolved (upstream revert or pi-lens update)," \
-              "close it._"
+              "resolved (upstream revert, pi-lens update, or the file" \
+              "location added to the locator), close it._"
           } > "$BODY_FILE"
 
           EXISTING_ISSUE="$(
@@ -141,7 +181,7 @@ jobs:
             echo "updating existing issue #$EXISTING_ISSUE"
             gh issue edit "$EXISTING_ISSUE" \
               --repo "$GITHUB_REPOSITORY" --body-file "$BODY_FILE"
-            COMMENT="Still failing: Layer A=${LAYER_A_OUTCOME},"
+            COMMENT="Still flagged: Layer A=${LAYER_A_KIND},"
             COMMENT+=" Layer B=${LAYER_B_OUTCOME}"
             COMMENT+=" (versions: ${LAYER_A_VERSIONS})."
             gh issue comment "$EXISTING_ISSUE" --repo "$GITHUB_REPOSITORY" \
--- a/clients/subagent-mode.ts
+++ b/clients/subagent-mode.ts
@@ -4,8 +4,16 @@
  * Two known subagent-spawning ecosystems set different env vocabularies in
  * every spawned child's environment:
  *
- * - nicobailon/pi-subagents sets `PI_SUBAGENT_CHILD=1` unconditionally (plus
- *   `PI_SUBAGENT_RUN_ID` / `PI_SUBAGENT_CHILD_AGENT` for best-effort identity).
+ * - nicobailon/pi-subagents sets `PI_SUBAGENT_CHILD=1` unconditionally on
+ *   every process that hosts a child session — the only var light mode keys
+ *   on for this vocabulary. It ALSO used to set `PI_SUBAGENT_RUN_ID` /
+ *   `PI_SUBAGENT_CHILD_AGENT` alongside it for best-effort identity, but
+ *   removed both entirely in `pi-subagents@0.65.0`'s native-AgentSession
+ *   rewrite (grep-verified absent from the whole 0.66.0 source tree, #2581/
+ *   #2680) — child identity now travels through an in-process object, not
+ *   env vars. `getSubagentIdentity()` below already treats identity as
+ *   optional best-effort metadata, so this permanent absence for the
+ *   nicobailon vocabulary needed no behavior change, only the doc update.
  * - avtc-pi-subagent (the spawn engine under avtc-pi-feature-flow) is the same
  *   execution model — real child-process `pi --mode rpc` / `--mode json -p`
  *   spawns, full env inheritance — but never sets `PI_SUBAGENT_CHILD`. It sets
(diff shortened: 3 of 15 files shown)


apmantza:
Orchestrator note from the r2 verify (merge-ready; no fix round): a `CONTRACTS` entry missing `parts` still throws (`parts is not iterable`, exit 2) rather than becoming that contract's own `infra` outcome; it cannot ship because `tests/scripts/compat-contract-resolution.test.ts:155` runs the real registry, but "every contract resolves independently" has one remaining throw. One-line hardening for whoever touches the resolver next.

Issue #2726, opened 2026-09-07
tokenizeShellCommand is heredoc-unaware: body words leak into the token stream


apmantza:
`clients/bash-file-access.ts`'s `tokenizeShellCommand` is heredoc-unaware: it
walks straight through a heredoc body and flattens the body's words into the
segment's token stream. Its consumers (`clients/git-guard.ts`,
`clients/runtime-tool-result.ts`) then analyse those tokens as if they were a
command the user is running.

Surfaced by the class sweep on #2699 / PR #2702, where the *same* defect shape
was fixed in the new `PreToolUse` hook's own lexer. The hook is a separate
bounded context and was deliberately not folded into this tokenizer (see PR
#2702's file header), so this is filed rather than fixed there.

## What is verified

Driving the REAL built tokenizer (`clients/bash-file-access.js`,
`PI_LENS_HOME` pinned):

```
$ node -e '…tokenizeShellCommand("cat > NOTES.md <<\x27EOF\x27\ngit reset --hard HEAD\nEOF")'
[["cat","NOTES.md","EOF","git","reset","--hard","HEAD","EOF"]]

$ node -e '…tokenizeShellCommand("git commit -F - <<\x27EOF\x27\nfix: do not git reset --hard\nEOF")'
[["git","commit","-F","-","EOF","fix:","do","not","git","reset","--hard","EOF"]]
```

Both bodies are inert to bash (a quoted delimiter means no expansion, and body
lines are data fed to `cat`/`git commit`, never commands), yet their words are
in the token stream. A `$( )`-wrapped heredoc feeding a CLI flag
(`gh issue comment … --body "$(cat <<'EOF' … EOF)"`) instead lands as ONE
double-quoted token, so the two shapes are inconsistent with each other as
well.

## What is NOT verified (the premise this issue needs)

Whether any consumer reaches a WRONG DECISION from those tokens. Driving
`collectGitInvocations` / `isGitCommitOrPushAttempt` from a bare `node -e`
returned `false`/`[]` even for a plain `git commit -m x` control, so those
predicates need session/config state the probe did not supply. **Reproduce from
the production call path before writing any fix** — if the guard's own segment
predicates only ever look at a segment's command position, the flattened body
words may be unreachable and the correct deliverable is an enforced invariant
plus a report, not a heredoc parser.

## Suggested scope

1. Reproduce (or refute) a false git-guard decision through the real dispatch
   path with a heredoc-carrying Bash tool call.
[...]

Issue #2705, opened 2026-09-07
hooks: guard-bash should subtract inert regions (heredoc bodies, single-quoted spans, comments) before classifying, instead of tokenizing the whole command (refs #2699, #2697)


apmantza:
Named output from the #2702 round-1 review (2026-09-07). The hook's lexer tokenizes the entire command and recurses into every backtick pair, which is how heredoc bodies that merely mention `git stash` were denied (round-1 F1). The inverse shape is smaller and makes that class structurally impossible: first strip the regions the shell never executes as commands — heredoc bodies, single-quoted spans, `#` comments — then tokenize and classify only what remains.

## Acceptance
- `scripts/hooks/guard-bash.mjs` classifies on an "inert-subtracted" view; the heredoc-awareness added in #2702 round 2 becomes a consequence of that view rather than a special case.
- The round-2 survey harness (every Bash command from a session transcript fed through the hook) is kept as a test fixture with the 2026-09-07 corpus; the pin is "zero non-rule denies".
- Lexer size and the header's blind-spot list both shrink; say by how much.

Not for #2702's fix round.

apmantza:
Remainder after PR #2732 (`refs #2705`), for the record: acceptance (a) — the inert-subtracted view — is on master via PR #2702 round 3; acceptance (c) — the measurements — is stated in #2732's body from the reproducible numbers (973 → 988 → 1000 lines; the four remaining limitations named in the header). Acceptance (b) is NOT met: the survey fixture in `tests/scripts/guard-bash-hook.test.ts` is a checked-in synthetic corpus (43 must-deny / 31 must-allow rows, one real hook spawn per row, pin "zero non-rule denies"), not the 2026-09-07 session-transcript corpus, which was not available to the fixer offline. Closing this issue needs that transcript exported into the fixture (or a decision that the synthetic corpus is the durable form).

apmantza:
Acceptance note for item (b): the retained fixture (PR #2748) is the 2026-09-06..08 transcript corpus (1,122 distinct commands), which contains the 2026-09-07 survey commands the item names; accepting the broader corpus as superseding. Pin: two expected denies by hash, zero non-rule denies, two-sided.

PR #2730, opened 2026-09-07
fix(guard): heredoc-aware tokenizeShellCommand (closes #2726)

apmantza:
## Summary

Round 2 fixes heredoc delimiter parsing in `tokenizeShellCommand`. Delimiters
now stop at shell metacharacters, including `;`, `&`, `|`, `<`, `>`, and
whitespace. Quoted and escaped delimiter spellings keep the same behavior.

## Tests

Round 1 transcripts carried forward:

- `npm run build` passed before the original heredoc tests.
- The original heredoc tokenizer tests passed after the round 1 body-consumption
  change.
- The original `handleToolCall` heredoc case passed through the real
  `lens-guard` path and did not block prose containing `git push`.

Round 2 red-first and verification:

- Before the source fix, the new `;`, `&`, `|`, `<`, and `>` × quoting cases
  failed: 20 cases total. The built tokenizer merged `echo git push` into the
  heredoc segment. The four whitespace cases were already green because the
  old parser already stopped at whitespace. The new real-path case also
  returned `{ block: true, ... }`.
- After the fix, `npm run build` passed.
- After the fix, `node_modules/.bin/vitest run tests/clients/bash-file-access.test.ts tests/clients/runtime-tool-call.test.ts --configLoader runner` passed: 2 files, 130 tests.
- The matrix covers six delimiter terminators (`;`, `&`, `|`, `<`, `>`, and
  space) across bare, single-quoted, double-quoted, and backslash-escaped
  delimiters: 24 token-stream cases.
- The real `handleToolCall` case with `cat <<EOF; echo git push` returned
  `undefined` and did not block.
- The real `handleToolCall` control with `cat <<EOF; git push` returned a
  blocking result.
- Mutation proof: restoring the old whitespace-only delimiter loop made 21
  targeted tests fail, including all 20 operator cases and the real non-blocking
  `handleToolCall` case.
- The six environment-blocked governance files remain for the orchestrator to
  run outside this sandbox.

## Blast radius

The changed parser is consumed by `commandSegments` in
`clients/bash-file-access.ts`, by `clients/git-guard.ts` for commit and push
classification, and by `clients/runtime-tool-result.ts` for command-result
analysis. The changed tool surface is `handleToolCall` in
`clients/runtime-tool-call.ts`, which invokes the git guard for Bash calls.
No durable storage shape or telemetry record changes.

## Class sweep

The repository sweep found one delimiter parser and one independent sibling
[...]

PR #2732, opened 2026-09-07
fix(hooks): guard-bash here-string and unclosed-substitution residuals (refs #2705, refs #2726)

apmantza:
## Summary

Fixes the round-2 heredoc substitution loss in `scripts/hooks/guard-bash.mjs`.
The scanner now retains substitutions found before a later malformed `$()` or
backtick span. It discards substitutions nested inside the malformed span,
matching bash behavior for the mirror case.

This round changes guard behavior and references #2705 and #2726. The PR does
not close either issue. The named remainder of #2705 is the requested
2026-09-07 session-transcript corpus survey. The checked-in 44-deny/32-allow
fixture is synthetic regression coverage, not that session-transcript survey.
The orchestrator must post that remainder on #2705.

The production hook measured 973 lines at `origin/master`, 988 lines at the
PR head before this round, and 1000 lines after this round. The documented
blind-spot list has four entries at both `origin/master` and this head, so no
blind spots were removed by the prior or current change.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:security
- [x] area:tests

## Checklist

- [x] I have read `CONTRIBUTING.md` and `AGENTS.md`
- [x] The change has a regression test for the bug
- [ ] Targeted Vitest passes locally; the offline grammar prefetch hung before results
- [x] The regression case has a pre-fix red proof
- [x] The new flush branches are mutation-proven
- [x] The PR title carries the issue references
- [ ] `npm run lint` was not run in this sandbox
- [x] `npm run build` succeeds
- [x] The changelog fragment is valid
- [x] The commit message references #2705 and #2726

## Tests

The edited `tests/scripts/guard-bash-hook.test.ts` drives the real hook CLI
through its PreToolUse JSON stdin and exit-code/stderr contract. The new case
pins a valid forbidden substitution before an unclosed substitution. It is
also added to the real deny corpus. The mirror case pins that an unclosed
outer substitution before a nested valid substitution remains allowed.

Red-first reproduction on the pre-fix hook:

```text
reviewer-probe-status=0
stderr: <empty>
```

The command was:

```text
cat <<EOF
$(git stash)
$(echo harmless
EOF
git diff
```

Post-fix direct hook probe:

```text
reviewer-probe-status=2
stderr: git stash is forbidden (CLAUDE.md non-negotiable) -- ...
backtick-probe-status=2
stderr: git stash is forbidden (CLAUDE.md non-negotiable) -- ...
mirror-status=0
stderr: <empty>
```

Real bash mirror probe:

```text
bash-mirror-status=0
bash-mirror-stdout=AFTER
bash-mirror-stderr=bash: command substitution: line 4: unexpected EOF while looking for matching `)'
```

The malformed outer substitution runs neither nested substitution. The hook
therefore allows the mirror case.

Mutation proof removed both `out.push(...found)` flushes. The same valid-before-
unclosed probe then produced:

```text
mutation-valid-before-unclosed-status=0
stderr: <empty>
```

This is red against the regression's required exit status 2. The flushes were
restored afterward.

Carried-over W1/W2 probes remain green:

```text
W1 status=2 stderr=git stash is forbidden (CLAUDE.md non-negotiable) -- ...
W2-dollar status=0 stderr=<empty>
W2-backtick status=0 stderr=<empty>
```

`npm run build` passed. `node scripts/check-changelog-fragments.mjs` passed
with `changelog fragments OK (10 entries in .changelog/)`. The instructed
Vitest command was attempted once with `--configLoader runner`; it printed the
offline tree-sitter grammar prefetch failures and hung before test results. It
was stopped after 36 seconds. The orchestrator must rerun it outside this
sandbox.

### Test assessment

- `tests/scripts/guard-bash-hook.test.ts`: uniquely pins the real hook's
  deny/allow CLI contract, including the new valid-before-malformed heredoc
  ordering and the real-bash mirror ordering. No existing test became
  redundant.

## Blast radius

The production module is the standalone `PreToolUse` Bash hook. Its entry
point reads JSON from stdin and writes one bounded rule message to stderr with
exit status 2 on denial; malformed input still exits 0. The changed helper is
called only by `consumeHeredocBody`, which is reached from `lexRegions` for
unquoted heredocs. No callers outside this scanner were found.

Call-tree diff:

```text
run -> classifyPayload -> scannableRegions -> lexRegions
                                      -> consumeHeredocBody
                                         -> scanHeredocBodyForSubstitutions
```

The helper runs once per unquoted heredoc body. The change adds a small
per-substitution array and does not add a process, filesystem, or network
operation. A repository `module_report` was unavailable in this sandbox, so
the caller sweep above is the available evidence.

## Observability

No new failure path or telemetry record exists. Denials continue to use the
existing bounded stderr rule message and exit status 2. Allows remain silent
with exit status 0. The regression observes both the discriminator message
and status through the real CLI.

## Class sweep

For the malformed-substitution class, the whole-tree search covered
`clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts`, `tests/support/`, and the
guard test. It found one production scanner, two `!span.closed` branches in
that scanner, and the one real-CLI test family. No duplicate production seam
exists. The population sweep covers both substitution forms: `$()` and
backticks each flush previously found substitutions and each has a direct
post-fix probe. The family stays on one scanner seam.

## Round 2

1. Finding 1: flushed `found` before returning from both malformed `$()` and
   backtick paths; isolated nested discoveries so the mirror case matches
   bash; added the deny regression and the mirror allow case; recorded red,
   post-fix, and mutation probes.
2. Finding 2: stated plainly that the 44/32 checked-in fixture is synthetic
   and that the 2026-09-07 session-transcript survey remains a `refs #2705`
   issue remainder.
3. Finding 3: recomputed measurements from `origin/master` and this head:
   973 lines at origin, 988 before this round, 1000 after this round, and
   four documented blind spots at both endpoints.

https://claude.ai/code/session_013gEp548mWcf8nWL9xivZWV

### DIFF
--- a/.changelog/fix-2705-guard-bash-inert-regions.md
+++ b/.changelog/fix-2705-guard-bash-inert-regions.md
@@ -1,5 +1,4 @@
 ---
 section: Fixed
 ---
-- **Guard-bash preserves here-string boundaries and rejects malformed heredoc substitutions (refs #2705, #2726)** — `<<<` is consumed as a here-string operator instead of being re-read as a phantom heredoc delimiter, and an unclosed `$(` or backtick in an unquoted heredoc body is treated as non-executable by the scanner while later live commands remain classified.
-- **Guard-bash retains valid heredoc substitutions before malformed ones (refs #2705, #2726)** — a valid `$()` or backtick substitution discovered before a later unclosed substitution remains classified, while substitutions nested inside the malformed span remain inert like bash.
+- **Guard-bash preserves here-string boundaries and classifies heredoc substitutions the way bash expands them (refs #2705, #2726)** — `<<<` is consumed as a here-string operator instead of being re-read as a phantom heredoc delimiter; an unclosed `$(` or backtick in an unquoted heredoc body is treated as non-executable while later live commands remain classified; and a valid `$()` or backtick substitution discovered before a later unclosed one stays classified, with substitutions nested inside the malformed span kept inert like bash.
--- a/.changelog/fix-2705-guard-bash-inert-regions.md
+++ b/.changelog/fix-2705-guard-bash-inert-regions.md
@@ -2,3 +2,4 @@
 section: Fixed
 ---
 - **Guard-bash preserves here-string boundaries and rejects malformed heredoc substitutions (refs #2705, #2726)** — `<<<` is consumed as a here-string operator instead of being re-read as a phantom heredoc delimiter, and an unclosed `$(` or backtick in an unquoted heredoc body is treated as non-executable by the scanner while later live commands remain classified.
+- **Guard-bash retains valid heredoc substitutions before malformed ones (refs #2705, #2726)** — a valid `$()` or backtick substitution discovered before a later unclosed substitution remains classified, while substitutions nested inside the malformed span remain inert like bash.
--- a/.changelog/fix-2705-guard-bash-inert-regions.md
+++ b/.changelog/fix-2705-guard-bash-inert-regions.md
@@ -1,0 +1,4 @@
+---
+section: Fixed
+---
+- **Guard-bash preserves here-string boundaries and rejects malformed heredoc substitutions (refs #2705, #2726)** — `<<<` is consumed as a here-string operator instead of being re-read as a phantom heredoc delimiter, and an unclosed `$(` or backtick in an unquoted heredoc body is treated as non-executable by the scanner while later live commands remain classified.
--- a/scripts/hooks/guard-bash.mjs
+++ b/scripts/hooks/guard-bash.mjs
@@ -224,6 +224,8 @@ function parseHeredocMarker(text, start) {
  * @param {string[]} out flat sink for every substitution body found
  */
 function scanHeredocBodyForSubstitutions(body, out) {
+	/** @type {string[]} */
+	const found = [];
 	let i = 0;
 	while (i < body.length) {
 		const ch = body[i];
@@ -232,19 +234,22 @@ function scanHeredocBodyForSubstitutions(body, out) {
 			continue;
 		}
 		if (ch === "$" && body[i + 1] === "(") {
-			const span = lexRegions(body, i + 2, ")", out);
-			out.push(span.retained);
+			const span = lexRegions(body, i + 2, ")", found);
+			if (!span.closed) return;
+			found.push(span.retained);
 			i = span.end;
 			continue;
 		}
 		if (ch === "`") {
-			const span = lexRegions(body, i + 1, "`", out);
-			out.push(span.retained);
+			const span = lexRegions(body, i + 1, "`", found);
+			if (!span.closed) return;
+			found.push(span.retained);
 			i = span.end;
 			continue;
 		}
 		i++;
 	}
+	out.push(...found);
 }
 
 /**
@@ -308,7 +313,7 @@ function consumeHeredocBody(text, start, heredoc, out) {
  * @param {number} start
  * @param {")"|"`"|null} closer
  * @param {string[]} out
- * @returns {{ end: number; retained: string }}
+ * @returns {{ end: number; retained: string; closed: boolean }}
  */
 function lexRegions(text, start, closer, out) {
 	let retained = "";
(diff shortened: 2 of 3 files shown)


apmantza:
## Review round 1 (Luna reviewer via plegma, head 6086cf0c) — verdict: needs changes

### Spec-compliance
- **HIGH — `scripts/hooks/guard-bash.mjs:236-246`: a valid substitution is lost when a LATER substitution in the same heredoc body is unclosed.** `return` discards `found` before it reaches `out`. Real bash executes the earlier valid substitution: for `cat <<EOF\n$(echo FIRST >&2)\n$(echo SECOND >&2\nEOF\necho AFTER` bash prints `FIRST`, `AFTER`, exit 0 (with the unexpected-EOF warning). The hook returns allow for `cat <<EOF\n$(git stash)\n$(echo harmless\nEOF\ngit diff` — a false allow. Remedy: flush already-discovered substitutions before abandoning the malformed one (`if (!span.closed) { out.push(...found); return; }`), plus a regression case with a valid forbidden substitution before an unclosed one.
- **HIGH — `tests/scripts/guard-bash-hook.test.ts:204-211`: the survey corpus is a checked-in synthetic fixture (43 deny / 31 allow — not vacuous), not "every Bash command from a session transcript" as #2705(b) asks.** The PR already says `refs #2705` for this; the remainder must be named on the issue rather than presented as the survey result.
- **MEDIUM — PR body measurements do not reproduce.** merge-base `guard-bash.mjs` is 973 lines, not 804, so the delta is 973 → 988; "3 blind spots removed" does not appear in the header nor derive from the 15-line source delta.

### Standards-compliance
None.

### Red-run / mutation verification
Removing the `<<<` branch: W1 `stash` → `null`. Removing both `!span.closed` guards: W2 `null` → `stash`. Source restored, `node --check` ok, tree clean. W1 variants cleared: `<<<`, `<<<"$(git stash)"`, `$'x\n'`, `a<<<b<<EOF`. A later live `git stash` after malformed heredoc syntax is executed by bash and correctly denied. Test wiring uses the real hook process, exact PreToolUse JSON, `cwd` set, file admitted to the `wall-clock-budget` project. Vitest could not complete in the sandbox (offline grammar prefetch hung); orchestrator's 246-passed run stands as the local result, CI green on this head.

### #2730 interaction
`clients/bash-file-access.ts` is a separate conservative tokenizer with no deny/allow classification; no textual conflict, independent merge, intentionally different bounded contexts.

Fix round dispatched (behaviour finding → same-reviewer verify afterwards).

apmantza:
## Verify round (round 2, Luna reviewer via plegma, head efb36b00) — verdict: needs changes, contract-only (the behaviour fix itself passes)

### Round-1 findings
- HIGH valid substitution lost before a malformed `$(` — **FIXED.** Both `$()` and backtick forms now exit 2 with stderr naming `git stash`.
- HIGH synthetic survey corpus — **PARTIAL by design**: remainder recorded on #2705 (reviewer could not read it offline).
- MEDIUM measurements — **FIXED**: 973 (origin/master) → 988 (6086cf0c) → 1000 (HEAD); the header's four-item limitation list is unchanged.

### New finding
- **MEDIUM (standards) — incomplete mutation coverage.** `tests/scripts/guard-bash-hook.test.ts:128-131` covers only the `$()` form; the production fix has a separate backtick flush at `scripts/hooks/guard-bash.mjs:249-255`. Deleting only the backtick flush made this real-hook probe ALLOW (`hook_status=0`): ``cat <<EOF\n`git stash`\n`echo harmless\nEOF\ngit diff``. Add the equivalent backtick regression case.

### Behaviour probes (bash oracle vs hook)
Mirror case (malformed outer substitution): bash runs neither nested substitution and continues to `echo AFTER`; hook allows — matches. `git stash` after `echo AFTER`: bash executes it; hook denies. Two valid substitutions before a malformed one: denied, names `git stash`. Valid substitution after the malformed span: not classified, matching bash's swallowed span. Valid substitution containing a heredoc with an unclosed nested substitution: allowed; bash runs neither.

### Mutations
Removing the `<<<` branch → all here-string variants allow. `!span.closed` guards → false: `$()` W2 case allow→deny, backtick unchanged. Removing both new flushes → both regression probes allow.

### Test quality / changelog
Real hook process, exact PreToolUse JSON, repository `cwd`, asserts exit + stderr (not tautological). One fragment, valid front matter, one entry describing all three behaviours. Commits conventional. Tree clean. Vitest hung on the offline grammar prefetch in the sandbox; direct probes supplied the evidence (orchestrator: 250 passed outside the sandbox).

Backtick regression added as a trailing test commit (contract-only → merge on green).

PR #2748, opened 2026-09-08
test(hooks): pin the guard-bash hook against the 2026-09-08 transcript corpus (closes #2705)

apmantza:
## Summary

Add the real transcript survey required by #2705. The harness loads the
1,122-row corpus extracted from this project's Claude Code Bash transcripts
from 2026-09-06 through 2026-09-08. Secrets and the maintainer's email are
scrubbed in the checked-in fixture.

The harness feeds every command to the real hook with the full PreToolUse
payload and the repository worktree as `cwd`. It pins zero non-rule denies.
The two expected denies are exact SHA-256 hashes for:

- a double-force `git worktree remove` command;
- an unpinned `node dist/cli.js` runtime probe.

The test also fails when either expected deny becomes allowed.

## Tests

- Added one serialized corpus test with a 180,000 ms total budget. It reports
  every offending command instead of stopping at the first false positive.
- The real corpus test was not completed in this sandbox. Vitest hung during
  repeated nested Node child spawns after the offline grammar prefetch. The
  orchestrator must run this file outside the sandbox and record the emitted
  row count, deny count, and elapsed time.
- Direct hook probe for the W1 mutation: replacing the `<<<` branch with a
  compile-valid `false &&` guard changed `grep x <<< foo\ngit stash` from
  deny to exit 0. Restoring the branch restored the deny.
- Direct synthetic false-positive probe: a quoted heredoc body containing
  `git stash` exits 0, proving inert body text is not classified.
- `vitest run tests/clients/flake-shape-ratchet.test.ts --configLoader runner`
  passes: 39 tests.
- `npm run build` passes.
- `npm run lint` passes.
- `npx oxfmt --check tests/scripts/guard-bash-hook.test.ts` passes.

## Blast radius

Test-only. The production hook is unchanged. The existing real-process-spawn
admission remains at eight static spawn-shape hits. The new corpus budget adds
one elapsed-time assertion, pinned in the flake-shape baseline. The existing
wall-clock lane already contains this test file.

## Class sweep

The corpus fixture is under `tests/fixtures`, while the flake-shape and test
file sweeps enumerate `*.test.ts` files. The 1.1 MB JSON is therefore data,
not a test file, and is not walked as executable test source.

## Observability

The corpus test logs its row count, observed deny count, and elapsed time. A
failure includes every non-rule-denied command and every expected-deny command
that was allowed.

## Test assessment

This adds a regression fixture and one real-runner contract test. The real
child process is required because the acceptance criterion is the hook's
stdin, exit-code, stderr, and environment contract, not only its exported
classifier.

The changelog fragment changes the #2705 reference to `closes #2705`.

## Orchestrator measurement

Outside the sandbox (this machine, 32 threads, serialized lane): `tests/scripts/guard-bash-hook.test.ts` 253 tests pass, wall 95.4 s for the whole file including the 1,122-row corpus, under the 180 s budget.


apmantza:
## Review round 1 (Luna reviewer via plegma, head bcb68a72) — verdict: needs changes (MEDIUM), both findings resolved by the orchestrator

### The 112-vs-2 question
The fixer's early "112 denies" probe is not reproducible against this head. Shipped hook over the corpus: `PI_LENS_HOME` unset → 1,122 rows, 2 denies (double-force worktree remove, unpinned `node dist/cli.js` probe); `PI_LENS_HOME` set → 1 deny (the probe rule reads `process.env.PI_LENS_HOME`, so a pinned session allows that row). `cwd` does not affect classification. The test deliberately unsets the variable, i.e. it pins the stricter unpinned-environment model; both pinned hashes match those two rows exactly.

### Findings
- **MEDIUM** `tests/fixtures/guard-bash/transcript-corpus-2026-09-08.json` — 593 rows carried the maintainer's home path (`/home/akis/…`, scratchpad paths). → Scrubbed to a placeholder user segment in 81286a10; the one expected-deny row whose text changed is re-pinned; corpus test + ratchet re-run outside the sandbox: 292 tests pass, 96.4 s wall.
- **MEDIUM** `closes #2705` — the issue names "the 2026-09-07 corpus", the fixture spans 2026-09-06..08. → Accepted as superseding on the issue (comment below).

### Cleared
Build; fixture excluded from Vitest collection and from the package (`files` excludes tests/); baseline elapsed-time count 1, real-process-spawn 8; the test is in `wallClockBudgetInclude`, the right lane for a 180 s wall-clock assertion; no `ghp_`/`sk-`/`Bearer`; one changelog entry, valid front matter.

### Could not verify (sandbox)
The corpus test hung at grammar prefetch/nested spawn in the reviewer sandbox (orchestrator ran it: 253 → 292 tests pass after the scrub); source mutations not run by the reviewer (read-only checkout) — the fixer's W1 and synthetic-false-positive reds stand as claimed; only four broad `<<` rows with `git stash` in bodies exist in the corpus, several being quoted transcript text.

Merge on green (data + hash re-pin only after the review).

PR #2755, opened 2026-09-08
fix(scripts): liveness-gated scratch-dir sweeps for smoke-tools and the harness scratch home (closes #2688, closes #2687)

apmantza:
## Summary

Fixes the shared scratch-directory race in #2688 and the unfindable, unswept
LSP harness homes in #2687.

`scripts/lib/scratch-dir.mjs` is now the single ownership seam. It claims
stable-named per-run directories under `/tmp/pi-lens-scratch`, writes
`owner.pid`, preserves live owners, removes dead owners, and age-gates
directories without a pid file. `withScratchHome()` and smoke fixture
workspaces use it. Scratch-home paths are printed at start and completion.

Option (a), extending `PILENS_PROBE=1`, would require moving the installer tool
tree and instances registry into the probe-home contract. That would expand a
log redirection contract into machine-state ownership and could reintroduce
cross-run contamination. Option (b) keeps the fresh-per-run isolation requested
by #2670, gives each directory a stable prefix plus pid and timestamp, and
sweeps siblings through the shared owner gate. This PR implements option (b).

Closes #2688 and closes #2687.

## Type of change

- [x] Bug fix
- [ ] New feature (net-new capability)
- [ ] Enhancement (improvement to existing capability)
- [ ] Documentation

## Area

- [x] area:tests

## Checklist

- [x] I read `CLAUDE.md`, `AGENTS.md`, `docs/pi-lens-subagent.md`, and `docs/pi-lens-fixer.md`.
- [x] The change has regression and edge-case tests.
- [x] The new regression test was red on the pre-fix smoke sweep.
- [x] New liveness and age guards were mutation-tested with compile-valid edits.
- [x] `npm run lint` passes.
- [x] The changelog fragment passes `node scripts/check-changelog-fragments.mjs`.
- [x] No commit, push, or `CHANGELOG.md` edit was made.

## Tests

The red-first and mutation transcripts are below. Vitest also reports the
expected offline grammar prefetch degradation; those grammars are unrelated to
this seam.

- `tests/scripts/scratch-dir.test.ts`: pins live-owner preservation, dead-owner
  removal, smoke-tools consumer wiring, and the young/old pid-less age fallback.
- `tests/scripts/lsp-fixture-workspace.test.ts`: edits the shared-home fixture
  to use the shared root and pins start/completion announcements. Twenty of 21
  tests pass; the git-init test cannot spawn `git` in this sandbox.
- `tests/config/lsp-fixture-home-workflow-pin.test.ts`: pins the honest option
[...]

Issue #2687, opened 2026-09-07
scripts: harness scratch home should reuse the PILENS_PROBE stable-path contract rather than a random /tmp dir (refs #2670)


apmantza:
Design note from the #2686 review (F3 named output). `withScratchHome()` pins `PI_LENS_HOME` to `os.tmpdir()/lsp-fixture-home-<random>`, printed nowhere and swept by nothing; it also pre-empts `getGlobalPiLensLogDir()`'s `global-dir-probe-redirect` degradation row. The repo already has a contract for "a harness run's logs go somewhere stable and announced": `PILENS_PROBE=1` → `<cwd>/.pi-lens-probe-home` (`clients/probe-home-state.ts:computeProbeHomeDir`), gitignored, with a degradation row on every fire — but it redirects logs only and leaves `tools`/`bin`/`instances.json` in the real home (`clients/degradation-ledger.ts:~1157`). Weigh: (a) extend the probe-home contract to optionally relocate the tool tree too, and have the five scripts set `PILENS_PROBE=1` instead of a random dir; or (b) keep the fresh-per-run dir #2670 asked for but make it stable-named and swept. Either removes the unfindable-logs problem and the /tmp growth. Part of #2608.

apmantza:
Follow-up: option (b)'s removal of the job-level `PI_LENS_HOME` pins from `tool-smoke.yml`/`parser-smoke.yml` (#2755) broke the nightly Format layer (#2762) — the six `--install` steps must share one tool tree in CI (#2670 review F1). Restored in a follow-up PR; the per-run scratch isolation stays the local default because `withScratchHome()` respects an explicit `PI_LENS_HOME`.

Issue #2670, opened 2026-09-06
scripts: one fixture-workspace bootstrap helper for the five LSP scripts, pinning PI_LENS_HOME (refs #2655, #2369, #2658)


apmantza:
Named output from the #2656 review (closes #2655, refs #2369).

**1. The preamble #2655 asked to fold is still five copies.** #2656 extracted the 5-line guard but grew the duplicated "copy fixture → `initLSPConfig` → optional `disableServers` → optional `gitInit`" block by 4 lines per script; `scripts/characterize-lsp.mjs:48-75` and `scripts/server-capabilities.mjs:67-94` are now a 23-line byte-identical block, `probe-clean-signal.mjs:230-251` the same with two flags swapped, `bench-lsp.mjs` (#2658) and `smoke-tools.mjs` variants. One helper in `scripts/lib/` (`bootstrapFixtureWorkspace(fx, {disableServers, gitInit})`) returning the workspace path, used by all five.

**2. None of the four scripts pins the home (#2506 shape, ~12×).** The now-unconditional `initLSPConfig` emits a `config_resolution_pending` + `config_resolved` pair per fixture (49 per full `--lsp` run, up from 3), and a local run writes them, rooted at `/tmp/char-lsp-*`, into the maintainer's real `~/.pi-lens/latency.log` and `sessionstart.log`, which `scripts/analyze-pi-lens-logs.mjs` joins on. The helper pins `PI_LENS_HOME`/`PILENS_DATA_DIR` to a per-run temp dir unless the caller opts out.

**3. Docblock drift.** `scripts/lib/lsp-fixture-session-guard.mjs` says the guard "throws loudly", but in `probe-clean-signal.mjs` the throw lands in a per-fixture try/catch (exit 0) and the three non-smoke steps in `tool-smoke.yml` are `continue-on-error`; only `smoke-tools.mjs --lsp` reds the nightly. State that.

Sequencing: after #2661 and #2658 so it folds all five at once. Part of the test-harness umbrella #2608.

apmantza:
[...]

PR #2766, opened 2026-09-08
fix(ci): restore the job-level shared tool tree in the smoke workflows (refs #2762, refs #2687)

apmantza:
## Summary

#2755 removed the job-level `PI_LENS_HOME`/`PILENS_DATA_DIR` pin from `tool-smoke.yml` and `parser-smoke.yml` (the #2670 review F1 measure that lets the six `--install` steps share one tool tree) and rewrote the pin test to assert its absence. The first nightly after it (run 34215458928) started the Format layer from an empty scratch home; seven managed formatters were selected but "ran clean and left the file unchanged" (#2762). This restores the two workflow hunks with their rationale and the pin test, and records the recurrence in the test header.

`withScratchHome()` still returns early on an explicit `PI_LENS_HOME` (`scripts/lib/lsp-fixture-workspace.mjs:153-156`), so the per-run scratch isolation #2687 introduced remains the local default; CI is the one place a job-level shared tree is intended.

## Tests

`tests/config/lsp-fixture-home-workflow-pin.test.ts` restored: it asserts both workflows carry the job-level env, that the two values are equal, and that neither uses `${{ runner.temp }}`. Red-first is the current master (the test as restored fails against master's workflows); green on this branch. lsp-fixture-workspace and scratch-dir suites unchanged and green.

## Blast radius

Two workflow files (job-level `env:` only) and one test. No runtime code.

## Class sweep

No other workflow runs more than one `--install` step; `parser-smoke.yml` is pinned for symmetry as before.

## Observability

Nightly `Tool smoke` Format layer; the drift tracker #2762 closes itself on the next green run.

### Test assessment

Restored workflow-contract test; the nightly run is the only executable check for the shared-tree behaviour.

Open question for the reviewer: why the seven rows reported `success` + unchanged instead of the harness's `tool not installed → skip` — the formatter's binary resolution in an empty home deserves its own row.

https://claude.ai/code/session_013gEp548mWcf8nWL9xivZWV

### DIFF
--- a/.changelog/fix-2762-smoke-shared-home.md
+++ b/.changelog/fix-2762-smoke-shared-home.md
@@ -1,0 +1,4 @@
+---
+section: Fixed
+---
+- **Nightly tool smoke: the six `--install` steps share one tool tree again (refs #2762, refs #2687)** — #2755 dropped the job-level `PI_LENS_HOME`/`PILENS_DATA_DIR` pin from `tool-smoke.yml` and `parser-smoke.yml`, so the Format layer started from an empty scratch home and seven managed formatters ran against nothing; the pin and its workflow test are restored, per-run scratch isolation stays the local default.
--- a/.github/workflows/parser-smoke.yml
+++ b/.github/workflows/parser-smoke.yml
@@ -36,8 +36,19 @@ jobs:
     # The installer resolves GitHub release assets through the REST API.
     # Unauthenticated is 60 requests/hour per IP and shared runners exhaust it.
     #
+    # PI_LENS_HOME/PILENS_DATA_DIR (#2670 review F1, for symmetry with
+    # tool-smoke.yml): this job only runs one `--install` step today, so there
+    # is no per-job cache to preserve here yet, but pinning keeps this
+    # workflow behaving identically to its sibling rather than depending on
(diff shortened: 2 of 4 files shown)


apmantza:
## Review (Luna reviewer via plegma, Tier A, head 737fe88d) — verdict: merge as-is

Trace confirms the diagnosis and the fix: with no job-level pin, each of the six `--install` invocations calls `withScratchHome()` and mints a fresh empty home; the installer's `TOOLS_DIR`, GitHub-binary dir and probe cache derive from that home at module load, so tools installed by earlier layers are invisible to the Format layer (install → home A, LSP → home B, Format → empty home C). The job-level pin turns that into one shared tree; `withScratchHome()`'s pinned branch returns before both `sweepScratchDirs` and `claimScratchDir`, and the shared path sits outside the scratch prefix, so the sweep cannot touch it. Build, restored pin test 24/24; the untouched `lsp-fixture-workspace` suite needs `npm run build:dist` (environmental, same for the orchestrator).

Open second defect, filed separately: a config-selected formatter whose managed binary is absent reports `success` + unchanged through its fallback resolution instead of the harness's `tool not installed → skip`.

Merge on green.

Issue #2767, opened 2026-09-08
tool-smoke Format layer misreads formatFile's typed 'unavailable' outcome as 'ran clean but unchanged', and --install skips the managed formatters (refs #2762)


apmantza:
## Context

Nightly run 34215458928 (#2762): the Format layer's step started from an empty `PI_LENS_HOME` (regression fixed in #2766), so black, cmake-format, stylua, cljfmt, php-cs-fixer, google-java-format and oxfmt had no managed binary. The smoke harness reported each as `ran clean but left the mis-formatted file unchanged` — i.e. `formatFile` returned `success: true, changed: false` — instead of the harness's `tool not installed → skip` path (which keys on `ENOENT|not found` in the formatter error).

## Defect

For those seven formatters, `resolveCommand` resolves to something that exits 0 without formatting when the managed install is absent (a PATH/launcher fallback, `python -m`, `npx`-style shim, or a wrapper swallowing the spawn error). A formatter that reports success without running is a silent no-op in production too: a user with a config-gated formatter and a broken install sees no error and no formatting.

## Acceptance

- `formatFile` for each of the seven reports a failure naming the missing binary (`ENOENT`/`not found`) when the managed tool is absent, so the smoke harness classifies it `skip` and the runtime records a bounded degradation.
- One test per formatter through the real `resolveCommand` with an empty `PI_LENS_HOME` and a scrubbed PATH (red-first on master).
- Class sweep across all formatters in `clients/formatters.ts`: which resolve to a fallback that can exit 0 without the tool; fold onto one resolution helper if the pattern repeats.

Found by the #2766 review (2026-09-08).

apmantza:
Re-scoped after reproduction (manual run 34264080916 with the shared home restored still shows the seven rows; local run reproduces 12). `formatFile` is honest: a missing binary yields `{ success: true, changed: false, outcome: "unavailable", error: "Cannot spawn black: tool not found (spawn black ENOENT)" }` (#2413's typed outcome). The harness at `scripts/smoke-tools.mjs` (Format layer verdict) checks `target.success` before `outcome`, so the typed `unavailable` is reported as "ran clean but left the mis-formatted file unchanged" instead of `skip (tool not installed)`.

## Acceptance
1. The Format-layer verdict keys on `outcome === "unavailable"` first → `skip` with the error text; red-first through the harness's verdict function with a fake `FormatterResult` of that shape.
[...]

Issue #2514, opened 2026-09-02
findInNodeModules walks ancestors past HOME: a stray ~/node_modules/.bin/oxfmt.cmd is picked up as the project's formatter (refs #2472)


apmantza:
Found during #2503 round 2 (2026-09-03). `tests/clients/formatter-unavailable-outcome.test.ts`'s "real oxfmt" case fails on the maintainer's box on both pre- and post-fix heads: `findInNodeModules`'s ancestor walk has no HOME ceiling, so it climbs out of the temp project and finds `C:\Users\R3LiC\node_modules\.bin\oxfmt.cmd` (the home-level manifest for pi extensions installs its own bins). Same walk-confinement class as #2503 F2 (`findLocalToolConfig`, fixed there with a default-on `isAtOrAboveHomeDir` ceiling), #250/#253, #622/#625.

Acceptance: `findInNodeModules` (and any sibling `node_modules/.bin` resolver — grep) stops at/above HOME via the shared ceiling; a test with a bin planted at HOME and a project under tmp proves it is not resolved (red on current master, cross-form path); the formatter-unavailable test passes on a box with a home-level node_modules. Enumerate the behaviour change in the fragment (a formatter installed only at HOME is no longer used).

Refs #2472, #2503.

PR #2775, opened 2026-09-09
fix(smoke): the Format layer installs the managed formatters it selects (refs #2767)

apmantza:
### Root cause and fix

The Unit tests job has no `PI_LENS_HOME`, `XDG_*`, or global install step. It
runs `npm install`, then `npm test`; npm prepends the repository's
`node_modules/.bin` to the child PATH. `package.json` declares `oxfmt`, so that
PATH contains a real `oxfmt` shim even though each test's mocked `which()`
returns `null`.

The temporary `getToolPath("oxfmt")` trace identified the returning leg:

```text
[getToolPath oxfmt] PATH candidate oxfmt: true
[getToolPath oxfmt] returning PATH candidate oxfmt
```

The PATH leg is legitimate production discovery and is independent of the
formatter module's mocked `which` seam. The fix pins that independent input in
both absence tests to an empty test-owned PATH directory and restores PATH in
`finally`. It does not change production PATH discovery or the deliberate
POSIX HOME-level local-bin case.

### Red and mutation evidence

With the temporary trace enabled in built output, a CI-shaped run used a fresh
HOME, `USERPROFILE`, `XDG_CONFIG_HOME`, and `XDG_CACHE_HOME`, and ran through
`npm test` so npm supplied its script-local PATH. Before the test pin, both
cases failed:

```text
Tests  2 failed | 9 passed (11)
AssertionError: expected 'failed' to be 'unavailable'
```

The trace above proves the PATH leg was re-enabled and returned the bare
command. This is the mutation red for the surviving CI path.

### Verification

- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/formatter-unavailable-outcome.test.ts --configLoader runner`
  — 11 passed.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/formatters.test.ts tests/clients/installer/ --configLoader runner`
  — 540 passed, 56 skipped.
- `PI_LENS_HOME=$PWD/.probe-home npm run lint` — passed.
- `PI_LENS_HOME=$PWD/.probe-home npm run build` — passed.

The pattern sweep found the shared installer PATH leg and no other formatter
absence test that mocks `which` while leaving npm's PATH prefix unpinned. The
population sweep covered all oxfmt absence cases in
`formatter-unavailable-outcome.test.ts`; the HOME drive-case test retains its
intentional POSIX PATH-independent local-bin assertion. The blast radius is
limited to those two test inputs; production callers of `getToolPath` are
unchanged.

The single changelog fragment is `.changelog/2767-format-smoke-managed-formatters.md`.
`CHANGELOG.md` is untouched. Changes remain uncommitted.

### Orchestrator note (round 5)

Root cause of the CI-only red: `npm test` prefixes PATH with `node_modules/.bin`, where the repo's dev-dependency `oxfmt` lives, so the installer's legitimate PATH leg returned it past the formatter module's mocked `which`. Fix is test-side pinning of PATH; production discovery is unchanged. The changelog fragment was folded to a single top-level entry (the fragment gate failed on ea52d38f with three bullets).

### DIFF
--- a/.changelog/2767-format-smoke-managed-formatters.md
+++ b/.changelog/2767-format-smoke-managed-formatters.md
@@ -5,3 +5,6 @@ section: Fixed
 - **Install managed formatters in the format smoke lane (closes #2767)** — The
   `--install` path now prefetches configured formatters through `ensureTool`
   and the formatter pipeline resolves those managed binaries.
+- **Record unsupported formatter platforms as unavailable (refs #2767)** —
+  GitHub and archive installs preserve the typed unavailable outcome when no
+  asset exists for the host platform and architecture.
--- a/.changelog/2767-format-smoke-managed-formatters.md
+++ b/.changelog/2767-format-smoke-managed-formatters.md
@@ -1,0 +1,7 @@
+---
+section: Fixed
+---
+
+- **Install managed formatters in the format smoke lane (closes #2767)** — The
+  `--install` path now prefetches configured formatters through `ensureTool`
+  and the formatter pipeline resolves those managed binaries.
--- a/.changelog/2767-format-smoke-managed-formatters.md
+++ b/.changelog/2767-format-smoke-managed-formatters.md
@@ -2,9 +2,12 @@
 section: Fixed
 ---
 
-- **Install managed formatters in the format smoke lane (closes #2767)** — The
-  `--install` path now prefetches configured formatters through `ensureTool`
-  and the formatter pipeline resolves those managed binaries.
-- **Record unsupported formatter platforms as unavailable (refs #2767)** —
-  GitHub and archive installs preserve the typed unavailable outcome when no
-  asset exists for the host platform and architecture.
+- **The format smoke lane installs the managed formatters it selects (refs #2767)** —
+  The `--install` path prefetches configured formatters through `ensureTool`
+  and the formatter pipeline resolves those managed binaries, with venv/local
+  → PATH → managed precedence; every managed resolver returns the typed
+  unavailable outcome instead of falling through to a bare command; GitHub and
+  archive installs record a typed unavailable outcome when no asset exists for
+  the host platform and architecture; the formatter-absence tests pin the
+  installer's independent PATH lookup so npm's `node_modules/.bin` prefix on
+  CI cannot resolve a dev-dependency binary.
--- a/clients/dispatch/runners/utils/availability-policy.ts
+++ b/clients/dispatch/runners/utils/availability-policy.ts
@@ -97,7 +97,7 @@ export interface ProbeEvidence {
 	 * same empty result as one that tried and failed, and writing `failed` for
 	 * both fabricates an attempt that never happened.
 	 */
-	install?: "succeeded" | "failed" | "not-attempted";
+	install?: "succeeded" | "failed" | "unavailable" | "not-attempted";
 	/**
 	 * Bounded (200 char) reason the installer gave, verbatim.
 	 *
@@ -225,7 +225,7 @@ export function describeProbeEvidence(
 
 /** The installer's own record of what its last attempt did. */
 export interface InstallAttemptFact {
-	outcome: "succeeded" | "failed" | "declined" | "skipped";
+	outcome: "succeeded" | "failed" | "unavailable" | "declined" | "skipped";
 	reason?: string;
 }
 
@@ -265,6 +265,11 @@ export function describeInstallAttempt(
 			return { install: "succeeded", ...(reason && { installReason: reason }) };
 		case "failed":
 			return { install: "failed", ...(reason && { installReason: reason }) };
+		case "unavailable":
+			return {
+				install: "unavailable",
+				...(reason && { installReason: reason }),
+			};
 		default:
 			return {
 				install: "not-attempted",
--- a/clients/formatters.ts
+++ b/clients/formatters.ts
@@ -48,6 +48,7 @@ import {
 import { safeSpawnAsync } from "./safe-spawn.js";
 import { assertInstallAllowed } from "./project-trust.js";
 import { tryLazyInstallForFormatter } from "./dispatch/runners/utils/lazy-installer.js";
+import { getToolPath } from "./installer/index.js";
 import {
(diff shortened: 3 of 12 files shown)


apmantza:
## Review round 1 (Luna reviewer via plegma, Tier A, head 2ac09042) — verdict: needs changes (HIGH)

- **HIGH** precedence: `stylua` (project-local → managed → PATH, `clients/formatters.ts:1629`), `cljfmt` (managed → PATH, `:1694`) and `google-java-format` (managed → PATH, `:1676`) now prefer pi-lens's managed copy over a user's PATH binary; before, they fell through to the bare PATH command, and the formatter policy elsewhere is project-local → PATH → managed. Restore PATH precedence (or document and test an intentional exception).
- **HIGH** registry packaging is not cross-platform: `php-cs-fixer` always matches `php-cs-fixer.phar` (`clients/installer/index.ts:924`) and on Windows `installGitHubTool` writes the PHAR bytes to `php-cs-fixer.exe` with no PHP launcher; `cljfmt` maps macOS and linux-arm64 to `standalone.jar` on the bare-binary path (`:950`), written as an executable with no `java -jar` launcher. Use the launcher-aware Maven/archive strategy or platform wrappers.
- **MEDIUM** `tests/scripts/smoke-tools-format-install.test.ts` calls `ensureFixtureTools` directly with fixture data; removing the production call at `scripts/smoke-tools.mjs:2163` leaves it green. Test through the real format-layer entry point.
- **MEDIUM** `scripts/smoke-tools.mjs:262` — Go, PowerShell, Rust, C#, Zig, Java and Dart TOOL-layer fixtures now declare formatter tools, so `--install` performs seven extra installs in the Tool layer; keep the declarations on `FORMAT_FIXTURES` only.

### Cleared
Hook path: the new resolver calls `getToolPath` (discovery only, `formatters.ts:702`), never `ensureTool`, so no download inside turn-end formatting; pip uses the shared `pip3 → pip → python3 → python` ladder (`installer/index.ts:5118`); Maven entries create a `java -jar` launcher gated on Java (`:4522`); the ubuntu workflow has setup-python, setup-java and setup-php; `.d.mts` parity; changelog; build, lint, 8 files / 209 tests (orchestrator: 10 / 369 outside).

Routing: fix round 2 on the full-access sandbox (real installs), then the same reviewer verifies (runtime precedence change).

apmantza:
CI on 44a4e81a: Unit tests red, two cases in `tests/clients/formatter-unavailable-outcome.test.ts` — "real oxfmt with no executable resolves to unavailable" and the #2514 HOME-bin sibling — both `expected 'failed' to be 'unavailable'`. Green locally in the retained worktree (11/11) and master f7a98cfa is green on the same job, so this is branch-specific and environment-sensitive: the new `resolveManagedFormatterCommand` path (`local ?? which ?? getToolPath`) reaches `getToolPath("oxfmt")` past the test's `which` mock, and on the runner something resolves to a spawnable path. Verify r2 is running on this head; the r3 brief carries this red as its first item.

apmantza:
Verify r2 (Luna, head 44a4e81a — settlement was missed for ~90 min by the fleet watch, plegma#339): **STILL NEEDS CHANGES — HIGH**; round 3 dispatched (full-access fixer), then re-verify.

- HIGH: five of seven managed resolvers (`black`, `cmake-format`, `stylua`, `cljfmt`, `google-java-format`) return `null` when nothing resolves, so the generic bare-command fallback runs and spawns ENOENT; `php-cs-fixer` and `oxfmt` correctly return `formatter-unavailable`. Probe with empty PATH and no managed copy: `black null / cmake-format null / stylua null / cljfmt null / php-cs-fixer "formatter-unavailable" / google-java-format null / oxfmt "formatter-unavailable"`. Remedy: `resolveManagedFormatterCommand` (clients/formatters.ts:704) returns `FORMATTER_UNAVAILABLE` when every candidate fails.
- MEDIUM: unsupported platforms return `undefined` from asset selection and log it, but the install-attempt record is a generic `failed`; preserve a typed unsupported reason.
- CI red on this head (`formatter-unavailable-outcome.test.ts`, oxfmt `'failed'` vs `'unavailable'`) is the same seam: on the runner the `getToolPath("oxfmt")` leg resolves something spawnable past the test's `which` mock — round 3 must make that file green on CI, not locally only.
- Cleared: PATH precedence table (venv/local → PATH → managed) at the shared resolver; launcher matrix (php shim, java shim, direct binaries) with the runtime guard at installer/index.ts:3645; `_arch` mapping cases; three fixture-population mutations red; 262 + 65 targeted tests; dependency-cruiser and jscpd environment-blocked in the sandbox (CI is the evidence).

apmantza:
Verify r3 (Luna, full-access, head 36e344d3): **NEEDS CHANGES** — one HIGH, one MEDIUM, both prescribed; round 4 dispatched and merges on green with quoted reds.

- HIGH: unsupported-platform GitHub installs still record `failed` — `clients/installer/index.ts:5556-5559` always finishes GitHub attempts as failed; `installGitHubTool` detects the unsupported platform at `:3556-3561` but returns `undefined`. Affects `stylua`, `php-cs-fixer`, `cljfmt`; the new typed branch (`:5574-5578`) only covers `archive` tools. Remedy: propagate `unavailable` when `assetMatch` has no asset; real installer regression test.
- MEDIUM: that archive branch is not mutation-proven (`&& Boolean(false)` → 96 tests still green).
- Cleared: built seven-resolver probe with empty PATH and isolated `PI_LENS_HOME` returns `formatter-unavailable` for all seven; sentinel mutation → 7 failed / 180 passed; php-cs-fixer sentinel-indexing fix at `formatters.ts:1564-1565`, no other misuse; 8 files / 330 tests; workflow runtimes and launcher guards present; jscpd no new clones; `.d.mts`, single fragment, no `CHANGELOG.md` edit.
- Could not verify: the historical CI-only red exactly (controlled HOME/PATH reproduction passes 11/11); CI on the round 4 head is the evidence. Title retitled to `refs` for the close-keyword gate (item 2 of #2767 only).

Issue #2697, opened 2026-09-07
Umbrella: mechanize the workflow — hooks for the non-negotiables, ast-grep sweeps, tool-smoke cwd row, incremental mutation, model routing, knip (refs #2691, #2693, #2694, #1844)


apmantza:
## Why

The 2026-09-07 lane for #2691 (an external report of a defect the catalog had already fixed by symbol in #1731) took a docs PR, a fix PR at three rounds, and two contract batches. The round generator was a hand-rolled source scanner; the process slips were rules that live only in prose. Each item below replaces a rule that agents read every round with a mechanism that runs for free.

Ranked by evidence. Items 1–3 are quick wins; 4 and 5 are lanes; 6 is a chore.

## 1. `PreToolUse` hooks for the non-negotiables

Today `.claude/settings.json` has only the worktree-prune hooks (SessionStart, SubagentStop). Add a Bash-matcher `PreToolUse` hook that DENIES, with a one-line teaching message:
- `git stash` (any form) — CLAUDE.md non-negotiable; a fixer ran it on 2026-09-07 (#2686 lane) despite the playbook.
- `git reset --soft origin/master` and `git reset --hard` outside a worktree.
- `git worktree remove -f -f` / `--force --force`.
- `node -e` / `node <script>.mjs` against `clients/*.js` with no `PI_LENS_HOME` in the command's environment (probe hygiene, #2506).

Acceptance: hook script under `scripts/hooks/`, unit-tested on the command strings above (deny) and on their safe neighbours (`git stash list` is still a stash — deny; `git diff > fix.patch` — allow); documented in AGENTS.md "Contributing"; the prose rules stay but cite the hook.

## 2. Governance sweeps as ast-grep rules

`@ast-grep/napi` is already a dependency. Relational rules (`has` / `not` / `inside`) express "a `safeSpawnAsync` call whose last argument is an object literal with no `cwd` key" without a bracket scanner. #2693 spent three rounds on exactly that scanner (whole-call token → raw-vs-blanked slice → positional wrapper params); shapes 33 and 34 are the same failure class.

Acceptance: #2694 is delivered as the first migration (a `sweep-kit` seam that runs an ast-grep rule over a file set and returns `file:line` rows); the runner-spawn-cwd sweep consumes it; a follow-up lists the other `stripSource`/balanced-scan consumers in `tests/support/sweep-kit.ts` with a migrate/keep verdict each.

[...]

PR #2778, opened 2026-09-09
ci(complexity): advisory complexity report over clients/, tools/ and mcp/ from pi-lens's own client (refs #2697)

apmantza:
### Summary

Keep the JS/TS complexity client’s cyclomatic contribution zero-based while
the shared dispatch helper remains 1-based McCabe complexity (refs #2697).

### Tests

- Added `tests/clients/complexity-client.test.ts` coverage for no-decision and
  one-`if` JS/TS functions through the real `ComplexityClient` seam.
- Pre-fix red with core grammars available: `expected 1 to be +0`; the existing
  complexity case also reported `expected 6 to be 5`.
- Mutation red after removing `- 1` from built output: `expected 1 to be +0`.
- Green: `npm run build`; targeted complexity suites, 24 tests passed.

### Blast radius

`collectFunctionMetrics` → `subtreeCyclomatic` → `calcCyclomaticComplexity`.
Only JS/TS per-function complexity changes; Python, Go, Rust, dispatch facts,
and aggregate baseline behavior remain unchanged.

### Observability

The existing per-function metrics and complexity report expose the corrected
value. No new failure path or telemetry record is introduced.

Grammar download was attempted with `node scripts/download-grammars.js --core
--dest grammars` but the sandbox had no network access. The test was completed
with the pinned core grammar cache already present in a sibling checkout.

### DIFF
--- a/.changelog/ci-2697-complexity-advisory.md
+++ b/.changelog/ci-2697-complexity-advisory.md
@@ -1,0 +1,4 @@
+---
+section: Added
+---
+- **`complexity (advisory)` CI job dogfoods pi-lens's own complexity client over `clients/`, `tools/` and `mcp/` (refs #2697)** — `npm run complexity` writes a Markdown report (top functions by cyclomatic/cognitive complexity, files over 1,000 lines, split candidates at the dispatch threshold of 15) to the step summary and an artifact; `ComplexityClient` now exposes the per-function metrics it already computed.
--- a/.changelog/ci-2697-complexity-advisory.md
+++ b/.changelog/ci-2697-complexity-advisory.md
@@ -1,4 +1,4 @@
 ---
 section: Added
 ---
-- **`complexity (advisory)` CI job dogfoods pi-lens's own complexity client over `clients/`, `tools/` and `mcp/` (refs #2697)** — `npm run complexity` writes a Markdown report (top functions by cyclomatic/cognitive complexity, files over 1,000 lines, split candidates at the dispatch threshold of 15) to the step summary and an artifact; `ComplexityClient` now exposes the per-function metrics it already computed.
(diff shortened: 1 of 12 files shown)


apmantza:
Review r1 (Luna, worktree at 4e0d03ed): **NEEDS CHANGES** — fix round r2 dispatched, merge on green after it.

- HIGH: the report accepts zero analyzed files and exits 0; `lint.yml` runs `npm ci --ignore-scripts`, so the grammar download never runs and the advisory lane always reports "None identified" (reproduced locally). Remedy: grammar download step + non-zero exit on zero files / analysis failure.
- MEDIUM: `complexity-client.ts` omits `do_statement`/`catch_clause` that `function-facts.ts` counts, so the report's threshold is not the dispatch metric. Remedy: reuse the dispatch seam or document the difference.
- MEDIUM: boundary test covers 16 only; `>=`→`>` mutation survives. Remedy: exact-15 case with the mutation red quoted.
- Cleared: build, 105 targeted tests, SHA pins/permissions, .d.mts parity, consumer sweep of `FileComplexity` (no strict parser found).

apmantza:
Verify r2 (Luna, head cf8d60b5): **NEEDS CHANGES** — one HIGH introduced by r2, round 3 dispatched with the prescribed remedy (merge on green).

- HIGH: `clients/complexity-client.ts:268` delegates JS/TS cyclomatic to `calcCyclomaticComplexity` (base `cc = 1`) while every other language path returns a 0-based contribution. Built probe: no-decision JS function now 1 (master 0), one `if` now 2 (master and Python/Go 1); the existing suite's `if + && + for-of + if + ternary = 5` case returns 6. User-visible `lens_complexity` regression. Remedy: the client's JS/TS branch subtracts the base so one convention holds per consumer; test at 0 and 1 with the `- 1` mutation red.
- Cleared: zero-file exit 1 probe, `>=`→`>` and `requireAnalyzedFiles` mutations red, 104 workflow/verdict tests, grammar step before build with `continue-on-error`, `.d.mts` parity, single fragment, no import cycle.
- Could not verify: the ComplexityClient suite (grammars + CDN unavailable in the sandbox); CI on the r3 head is the evidence for it.

Issue #2790, opened 2026-09-09
preflight: one command that runs every mechanical PR gate locally, and the fixer contract requires its transcript in the handoff


apmantza:
Wave-2 umbrella #2784. Record from 2026-09-09: of the CI reds across #2775, #2780, #2781, #2782, #2783, #2785, #2787, ELEVEN were repo-shape gates a fixer could have run locally in one command, not defects: close-keyword title/body mismatch (#2775), changelog fragment with three top-level entries and a fragment without front matter (#2775 r4, #2787), `vi.mock` doubles missing exports (#2782 r1), flake-shape ratchet on a raw `vi.waitFor` (#2781), lsp-spawn-heavy lane admission (#2783), `wallClockBudgetInclude` sort order and the ci-verdict advisory classifier (#2785), the generation-guard ratchet (#2782 r4), the install-step count pin (#2780), the format-smoke style contract (#2782), and oxfmt on merged test files (#2778/#2783). Each cost a fix round or a trailing commit and a CI cycle.

Deliverable: `npm run preflight` → `scripts/pr-preflight.mjs` (+ `.d.mts`) that runs, in order, with one summary table and a non-zero exit on any red: `npm run build`; `npm run lint`; `npm run fmt:check`; `npm run changelog:check` (+ `check-changelog-fragments.mjs`); `check:lockfile`; `tests/config/` (all admission/registry/workflow pins); the governance sweeps that red most often (`tests/clients/generation-guard-sweep.test.ts`, `tests/clients/flake-shape-ratchet.test.ts`, `tests/config/lsp-spawn-heavy-coverage.test.ts`, `tests/config/vi-mock-export-sweep.test.ts` once #2787 lands, `tests/scripts/ci-verdict.test.ts`); and, when `COMMIT_MSG.txt`/`PR_BODY.md` exist at the root, `check-pr-title.mjs` on the first line, `check-close-keywords.mjs` on title+body, and `check-pr-body.mjs` on the body (template headings, Observability literal per #2543 when that lands). Vitest calls use `--configLoader runner` and `PI_LENS_HOME=$PWD/.probe-home`; total runtime target under 3 minutes; the table names the gate, the CI job it mirrors, and pass/fail with the first red line.

[...]

Issue #2784, opened 2026-09-09
Umbrella: mechanize the workflow, wave 2 — CI verdict fidelity, mock-double drift, admission registries, worktree hygiene


apmantza:
Second wave of #2697 (mechanize the workflow). Every item below is an open issue that surfaced or recurred in the 2026-09-09 lanes and is mechanical (a check, a sweep, a hook, a lane), not a runtime fix. Order is by how often it bit this week.

| # | item | why it is first-class now |
|---|---|---|
| #2281 | whole-module mock factories silently drop new exports (22 known) — a governance sweep that reds when a `vi.mock` factory of a production module omits an export the module gained | cost #2782 a full round: 30 reds in 6 files from two missing pass-throughs |
| #2616 | the merge-train warden's red-ci label checks only Unit tests + Lint, not every non-advisory check | same class as ci-verdict's absent-is-not-green; the changelog-fragment gate and close-keyword gate both redded PRs this week without the label |
| #2671 | admission registries (flake-shape baseline, wallClockBudgetInclude, lane membership) are merge-conflict magnets under parallel PRs | #2782 conflicted on a registry file when #2775 merged first; #2783 moved the spawn-heavy baseline the same day |
| #2103 | CI failure classifier: label infra kills (exit 137, SARIF, codeload) vs assertion failures; auto-rerun infra once | a #2783 run died with exit 137 mid-job |
| #2543 | check-pr-body: the Observability section must name a record literal present in the diff | three fixer bodies this week passed with prose-only observability sections |
| #2563 | flake-shape ratchet walks `tests/support/*.ts` helpers too | the governed-wait rule was applied by hand in #2781 r3; a helper clone would evade the detector |
| #2614 | label-sync post-merge validation fails on any `deleted` line | recurrence #2553, twice |
| #2625 | workflow-shell portability guard (no bash-4-only builtins in `shell: bash` steps that can run on mac) | — |
| #2631, #2538 | worktree pruning: any worktree whose branch is merged into master, per-run cap, unregistered dirs | the merge-train prunes by hand today; the reaper parked on hand-removed trees |
| #2663 | release-qa baseline includes the tool-smoke install lane | #2767 shipped because the install lane was not in the release gate |

Acceptance for the umbrella: each row is closed by its own PR (refs this issue), or explicitly parked here with the reason. Same brief rules as #2697: detector and sweep authoring goes to the strongest model from round 1; every new check is proven red on the recurrence it names; advisory until clean, then gating.

[...]

PR #2796, opened 2026-09-09
ci(preflight): one command for every mechanical PR gate (refs #2790, #2784)

apmantza:
## Summary

Fixes the round 3 preflight findings for selector validation, local Test assessment parity, spawn failures, test locking, and CI job labels.

## Tests

- `npm run build` passes.
- `npm run lint` passes.
- `npm run fmt:check` passes.
- `npm run changelog:check` passes.
- `npm run check:lockfile` passes.
- Targeted Vitest passes: 2 files, 97 tests, 0.32 seconds.
- Governance siblings pass: 2 files, 58 tests, 2.90 seconds.
- `npm run preflight` is run last below.

## Test assessment

`tests/scripts/pr-preflight.test.ts` uniquely pins unknown-selector rejection, spawn-failure rows, shared Vitest locking, and CI label vocabulary drift. `tests/scripts/check-pr-body.test.ts` uniquely pins local Test assessment parity for a diff touching `tests/`. No existing test became redundant.

## Blast radius

`runPreflight` calls `parseArgs`, builds the effective gate list, invokes `runChild`, and formats every gate row. Local `check-pr-body` execution reads the Git diff before calling `lintPrBody`. The CI job vocabulary is consumed by preflight and its declaration mirror.

## Class sweep

Searched all preflight selectors, direct Vitest gate commands, local `lintPrBody` calls, and preflight CI labels. The touched seams are the only members requiring this behavior.

## Observability

Spawn exceptions produce a failed row whose first red line is the original error. The runner continues in declared gate order, so every attempted gate remains visible in the summary.

## Round 3

Red-first and mutation transcripts:

- Unknown selector before the guard: `Test Files 1 failed (8 tests | 1 failed)` and `expected [Function] to throw an error`.
- Spawn catch neutered: `Test Files 1 failed (8 tests | 1 failed)` with the injected error escaping from `runChild`.
- Shared lock prefix removed: `Test Files 1 failed (8 tests | 1 failed)` because the command lacked `scripts/with-test-lock.mjs` and `--shared`.
- Local Test assessment option neutered: `Test Files 1 failed (88 tests | 1 failed)` with `Received: 0` and the expected status `1`.
- CI changelog label reverted: `Test Files 1 failed (8 tests | 1 failed)` with `Received: "Validate changelog fragments"`.
- Reviewer red reproduced: `--only does-not-exist` previously printed an empty table and exited `0`; an injected `ENOENT` previously threw before a summary row.

The targeted command that previously encountered the stale-source setup works after the fresh build:

`PI_LENS_HOME=$PWD/.probe-home node scripts/with-test-lock.mjs --shared -- node_modules/.bin/vitest run tests/scripts/pr-preflight.test.ts tests/scripts/check-pr-body.test.ts --configLoader runner`

It passes 2 files and 97 tests in 0.32 seconds. The sibling governance run passes 2 files and 58 tests in 2.90 seconds. The preflight command below is the final verification.

The first preflight attempt exposed two governance reds caused by the new local Git probe and its subprocess test. The probe now uses `scripts/lib/git-fixture-env.mjs`, and the test uses the injectable production seam, so both governance suites pass.

### Orchestrator run (round 3, 2026-09-09)

```text
build ok, lint ok, oxfmt clean
pr-preflight + check-* + tests/config/ — 42 files, 506 passed
npm run preflight -- --only does-not-exist → usage error naming the valid gates, exit 2
```


apmantza:
Review r1 (Luna, full-access, head f36bc329): **NEEDS CHANGES** — two HIGH, three MEDIUM, all prescribed; round 3 dispatched and merges on green with quoted reds.

- HIGH: `scripts/pr-preflight.mjs:49-64, 203-211` — unknown `--only`/`--skip` names are accepted; `--only does-not-exist` prints an empty table and exits 0.
- HIGH: `scripts/check-pr-body.mjs:501-505` — local mode omits `requireTestAssessment`, which CI sets for test-touching PRs (`:489-491`); preflight can pass where CI fails.
- MEDIUM: `:80-97` a spawn `ENOENT` throws before any FAIL row; `:154-164` vitest gates bypass `with-test-lock.mjs`; `:12-19` two "mirrored CI job" labels are step names, not job names (`Changelog fragment (fast-fail)`, `Lint & type-check`).
- Cleared: real reds for fmt:check, a broken tests/config pin and a `build:` title each name the right gate and exit 1; `--skip fmt:check`/`--skip build` refused; `--only fmt:check` works; `PI_LENS_HOME` default/preserve; two full runs 16.3 s and 17.0 s; CI paths of the three checkers unchanged without the local flag; `.d.mts` parity; single fragment; `CHANGELOG.md` untouched.
- Could not verify: the targeted Vitest command (global freshness setup reported 463 stale sources after build) — carried into round 3 to reproduce.

Issue #2563, opened 2026-09-03
flake-shape ratchet: walk tests/support/*.ts helpers too, so an unsanctioned delay() clone cannot evade the raw-timer detector (refs #2547)


apmantza:
#2559 review: the three detectors only walk `*.test.ts` (964/964 scanned files). A local helper `export const delay = (ms) => new Promise(r => setTimeout(r, ms))` in tests/support/ and its call sites `await delay(50)` evade all three detectors; only `interleaving-kit.ts` is the sanctioned exemption. Extend the walk to `tests/support/**/*.ts` (exempting interleaving-kit by name) and add a call-site detector for exported helpers that wrap setTimeout. Shape 34 (spelling enumeration) applies — prove red with a planted clone.

apmantza:
Tracked under the test-harness umbrella #2608.

apmantza:
Closed by #2814 (merged 82b79167): the flake-shape ratchet walks tests/support/**/*.{ts,mts} with the three time detectors, resolves timer aliases on the AST (local bindings, timers/promises imports, destructured globalThis), requires the two-part admission (header + entry) for support-helper rows, keeps the spawn detector test-file-only, and parses only timer-bearing files once per run (CI 38 s → ~9 s locally).

PR #2807, opened 2026-09-09
ci(pr-body): the Observability section names a record literal from the diff (refs #2543, #2784)

apmantza:
## Summary

Round 3 merges `origin/master`, preserves the preflight local-input contract, and restores the shared `gitExecFileSync` seam so the observability diff is acquired.

## Round 3

- F1: fixed in round 2. CI uses a full checkout and reports `diff unavailable:` when the ref is absent.
- F2: fixed in round 2. Test, `__tests__`, and declaration files are excluded from runtime marker checks.
- F3: fixed in round 2. Comment and template-literal laundering cases fail, including the blanking mutation.
- F4: fixed. The merge keeps master's `--lint-local PR_BODY.md` invocation from `scripts/pr-preflight.mjs` and this PR's `--body <body> --title <title>` title-aware variant.
- F5: fixed. `localDiff()` uses the master's `gitExecFileSync` wrapper, which calls `execFileSync("git", args, options)`.

The pre-fix red was `The "file" argument must be of type string. Received an instance of Array`.
The pre-fix local live-path probe returned `{"valid":true,"errors":[]}` for a runtime-shaped body with no record.

## Tests

- `tests/scripts/check-pr-body.test.ts`: added full-checkout diff acquisition and runtime missing-record regressions, and updated the local fallback matrix for the merged seam.
- `tests/clients/flake-shape-ratchet.test.ts`: ran unchanged to verify the existing admission population.
- `tests/scripts/check-pr-body.test.ts` and `tests/clients/flake-shape-ratchet.test.ts`: 148 tests passed.
- `npm run build`: passed.
- Full-checkout probe: acquired 17,952 bytes from `origin/master...HEAD`.
- Shallow-clone probe: reports `diff unavailable:` only when `origin/master` is absent in CI mode.
- `npm run preflight`: run below; its table is the final verification record.

Final preflight table:

| gate | mirrored CI job | pass/fail | first red line |
| --- | --- | --- | --- |
| build | Lint & type-check | pass | |
| lint | Lint & type-check | pass | |
| fmt:check | oxfmt format check (advisory) | FAIL | `tests/clients/dispatch/runners/eslint.test.ts` |
| changelog:check | Unit tests | pass | |
| check-changelog-fragments | Changelog fragment (fast-fail) | pass | |
| check:lockfile | Lint & type-check | pass | |
| tests/config | Unit tests | pass | |
| generation-guard | Unit tests | pass | |
| flake-shape-ratchet | Unit tests | pass | |
[...]

PR #2814, opened 2026-09-09
test(governance): flake-shape ratchet walks tests/support helpers (refs #2563, #2784)

apmantza:
Summary

Resolve destructured `globalThis`, `window`, and `self` timer aliases in the
flake-shape support scan. Add regression coverage for aliased, shorthand, and
non-timer object patterns. F2 is fixed.

Tests

- `tests/clients/flake-shape-ratchet.test.ts`: add aliased and shorthand
  global timer fixtures and a non-timer destructure negative case.
- `tests/support/flake-shape-scan.ts`: resolve object-pattern timer bindings
  in the existing `timerBindings` seam.
- Red-first: before the fix, the prescribed aliased fixture failed with:
  `AssertionError: expected [] to have a length of 1 but got +0`.
- Mutation: after removing the object-pattern branch, the same fixture failed
  with `AssertionError: expected [] to have a length of 1 but got +0`.
- Green: build plus ratchet and `tests/config/`: 39 files, 414 tests passed.
- `npm run lint`: passed.
- `git diff --check`: passed.

Blast radius

The change is confined to `timerBindings`, called by `scanRawTimerWait`, and
its raw-timer ratchet fixtures. The call visitor is unchanged. The new branch
adds only bindings whose initializer is a known timer global and whose object
property is `setTimeout` or `setInterval`; non-timer objects remain unbound.

Class sweep

The existing timer-binding population covers direct globals, local aliases,
namespace imports, named `timers/promises` imports, and now object patterns
from `globalThis`, `window`, and `self`. The support walk and comment/string
blanking paths remain unchanged. Consolidation stays on the shared
`timerBindings` seam.

Observability

This detector change adds no failure path, durable record, or runtime logging.
The ratchet remains the independent observable and mutation proof.

Test assessment

The edited ratchet test adds three cases for the prescribed recurrence: two
positive destructured global timer aliases and one negative non-timer object.
The tests exercise the real AST scanner and preserve comment/string safety
coverage in the existing suite.

## Round 3

F2 fixed. The remaining destructured `globalThis` timer-alias shape now
resolves both `const { setTimeout: t } = globalThis` and the shorthand form.
The required non-timer destructure remains `[]`.

Quoted red proofs:

- Pre-fix prescribed probe: `scanRawTimerWait(...)` returned `[]` for the
  destructured alias; Vitest reported `expected [] to have a length of 1 but got +0`.
- Mutation with the object-pattern branch neutered: the same test reported
  `expected [] to have a length of 1 but got +0`.

Could not run `npm run preflight` because this checkout has no `preflight`
script. No CI verdict or OS matrix was run.


apmantza:
Verdict: merge after fixes.

## HIGH findings

### F1 — The new support baseline row bypasses the two-part admission gate

1. Probe: with the tests retained and `tests/support/flake-shape-scan.ts` reverted to `origin/master`, `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/flake-shape-ratchet.test.ts --configLoader runner` failed 3 tests / passed 48. The first failure was:

   `raw-timer-wait:support/fault-injection.ts`

   The head baseline contains `"support/fault-injection.ts": 3` under `raw-timer-wait`, but `tests/support/fault-injection.ts` has no `// flake-shape: raw-timer-wait — ...` header, no `ADMITTED_AFTER_BASELINE` entry, and no `wallClockBudgetInclude` membership.

2. The support helper was admitted by a bare baseline edit, not by the claimed reason-plus-header admission.

3. `auditAgainstBaseline` only applies the header and serialized-lane checks to `ADMITTED_AFTER_BASELINE` entries; a newly surfaced support row can therefore be frozen into the baseline without either proof. This turns the new population scan into an allowlist bypass at exactly the boundary this ratchet is meant to govern. The relevant code is `tests/clients/flake-shape-ratchet.test.ts:94-360, 652-704`, while the added row is `tests/support/flake-shape-baseline.json:221`.

4. Fix A: give the support helper an explicit, detector-specific admission path with a reason and a support-helper equivalent of the required execution-lane proof, then make the ratchet validate it. Fix B: do not baseline this helper; route its real delay/teardown timing through the sanctioned primitive or add a narrowly named exemption that still records the reason. Do not solve this by adding only a comment to the JSON.

### F2 — Aliased delay clones pass silently despite the named delay-clone shape

1. Probe results from `scanRawTimerWait` (expected one hit for each real timer helper):

   `globalThis.setTimeout(...)` → 1 hit (caught).

   `export function pause(ms) { const t = setTimeout; t(() => {}, ms); }` → `[]` (missed).

   `export function pause(ms) { return new Promise(r => setTimeout(r, ms)); }` → 1 hit (caught).

   `import { setTimeout as timer } from "node:timers/promises"; export function pause(ms) { return timer(ms); }` → `[]` (missed).

   The direct detector catches a `.mts` source when called manually, but the whole-tree walk returns no such file because both walkers pass `extensions: [".ts"]`; a `.js` helper is likewise outside the walked population.

2. A support helper can hide the raw wait behind a local alias or a `timers/promises` alias and avoid both `RAW_TIMER_CALL` and the delay/sleep-name regex.

3. `tests/support/flake-shape-scan.ts:331-347` only recognizes a literal `setTimeout(`/`setInterval(` call and a binding whose declaration name starts with `delay` or `sleep`. A helper named `pause`, `tick`, or `waitMs` using `const t = setTimeout` or `timers/promises` therefore reaches callers with no count. This is the delay-clone reuse vector named by #2563, so it is a HIGH false negative. The extension gap is a separate scope decision: the issue says `tests/support/**/*.ts`, so it is not independently actionable unless the intended population includes `.mts`/JavaScript.

4. Fix A: use the TypeScript AST to resolve local aliases and imports from `node:timers/promises`, then flag the awaited timer call or its enclosing helper. Fix B: define and test a bounded support-helper policy that recognizes timer-bearing exported functions regardless of their names, while retaining the current name check as a cheap first pass. Add mutation probes for both alias forms.

## Cleared

- Red-first: reproduced 3 failures / 48 passes on the pre-fix scanner; the live head passes 51/51.
- Comment and string safety: `// setTimeout(() => {}, 1)` and `const x = "setTimeout(() => {}, 1)"` both returned `[]`.
- `globalThis.setTimeout`, direct Promise-arrow `setTimeout`, and direct `.mts` detector calls are caught.
- Delay-definition mutation: replacing `DELAY_SLEEP_HELPER_DEFINITION` with `/(?!)/` failed 2 tests, including the ratchet count falling from 3 to 2 and the hidden-delay fixture.
- Spawn exclusion is intentional and supported by existing governance: `fault-injection.ts:25-28` imports `node:child_process` and defines `spawnWedgedChild`; its test carries the real-process header at `tests/support/fault-injection.test.ts:316`, is in `ADMITTED_AFTER_BASELINE` at `tests/clients/flake-shape-ratchet.test.ts:345-349`, and is in `vitest.config.ts:429`. The same pattern exists for `git-fixture-env.test.ts` at `tests/support/git-fixture-env.test.ts:146` and `vitest.config.ts:431`.
- Runtime and memoization: build passed; under `--configLoader runner`, the ratchet passed 51/51 in 2.98s test time / 3.30s wall time. `countsCache` is populated once at `tests/support/flake-shape-scan.ts:467-501`, and subsequent detector reads return cached maps.

## Could not verify

- I did not run the full suite, CI verdict, or an OS matrix.
- I did not extend the issue's explicitly `*.ts` population to `.mts` or `.js`; those extensions are currently outside the stated scan contract.

## Named output

- Strong: the scan has one whole-tree cache and a single support-population detector policy, so the right consolidation point for alias handling is `scanRawTimerWait`/the support walker, not per-helper exceptions.
- Worth exploring: a shared AST binding resolver could close local aliases and `timers/promises` imports without broadening the spawn boundary.

apmantza:
Round 2 (3aee5702, Luna): F1 support-helper two-part admission (header + entry, both mutation-red), F2 alias-resolved timer detection on the AST incl. timers/promises, .mts walked. Detector round → fresh verify follows.

apmantza:
Verdict: merge after fixes.

## Disposition

| Finding | Disposition |
| --- | --- |
| F1 | fixed |
| F2 | not fixed |

## HIGH findings

### F2 — destructured `globalThis` timer aliases still pass silently

1. Probe: `scanRawTimerWait("support/destructure.ts", "const { setTimeout: t } = globalThis; export function pause(ms: number) { t(() => {}, ms); }")` returned `[]`; expected one raw-timer hit. The same probe returned one hit for the namespace import, one for the `timers/promises` named alias, one for the class method, and one for the `.mts` named alias. Comment and string probes returned `[]`.
2. The claimed AST alias resolution does not resolve a destructured `globalThis.setTimeout` binding, so a `pause` delay clone remains invisible.
3. `timerBindings` only records identifier-to-identifier declarations and imports at `tests/support/flake-shape-scan.ts:342-395`; the call visitor consumes those sets at `:467-482`. This is the named delay-clone shape: a support helper can provide a real timer through an arbitrary name, and the new support population is intended to catch it. The false negative is HIGH because callers can reuse the ungoverned delay without a ratchet count.
4. Resolve object-pattern declarations whose initializer is `globalThis` and whose property is `setTimeout` or `setInterval`, or add an equivalent AST test and binding path. Keep the comment/string blanking invariant.

## Cleared

- F1 admission is two-part and mutation-red. Removing only the helper header failed with `raw-timer-wait:support/fault-injection.ts: missing "// flake-shape: <detector> — <reason>" header`. Removing only the `ADMITTED_AFTER_BASELINE` entry failed with `raw-timer-wait:support/fault-injection.ts: missing ADMITTED_AFTER_BASELINE entry`.
- Namespace `timers/promises` import: one hit. Named `timers/promises` alias in `.mts`: one hit. Class method with direct `setTimeout`: one hit. Re-export outer file: `[]`, while its inner implementation returned one hit; forwarding is not itself a delay clone and the walked inner file remains governed.
- Alias text inside a comment and string: both `[]`.
- Spawn exclusion remains unchanged. `SUPPORT_POPULATION_DETECTORS` excludes `real-process-spawn` at `tests/support/flake-shape-scan.ts:604-607`; support files are the sanctioned child-process boundary, and existing real-spawn test admissions remain in `tests/clients/flake-shape-ratchet.test.ts:352-364` and `vitest.config.ts:429-431`.
- The AGENTS.md line is accurate: support-helper rows inherit lane proof through importing tests, matching `supportHelperHasLaneProof` and the validation path at `tests/clients/flake-shape-ratchet.test.ts:397-423, 714-724`.
- Build passed. The ratchet passed 54/54. The ratchet plus `tests/config/` passed 411/411 across 39 files. The scan cache remains memoized by the single `countsCache` initialization at `tests/support/flake-shape-scan.ts:605-638`.
- `git merge-tree` reports no textual conflict merging open PR #2807 (`67c3436`) into this head. The overlapping insertion regions are `tests/clients/flake-shape-ratchet.test.ts:131-135` here versus #2807's `real-process-spawn` entry at its diff lines 338-344, and `tests/support/flake-shape-baseline.json:221` here versus #2807's baseline insertion at its diff lines 536-540. Its third overlap is `vitest.config.ts:429-431` here versus #2807's insertion at diff lines 548-554. These are adjacent/same-list edits, but Git resolves them automatically; review the combined sorted lists after merge.
- `npm run lint` passed, and `git diff --check origin/master...HEAD` passed.

## Could not verify

- No CI verdict or OS matrix was run. Full CI remains to verify the exact head.

## Named output

- Strong: the scan has one cached support-population walk and one AST binding seam. The destructuring fix belongs in `timerBindings`, not in per-helper exceptions.

apmantza:
Round 3 (8070bbb9): F2 destructured globalThis aliases resolved in timerBindings, red-first and mutation quoted. Prescribed round → merges on green.

apmantza:
Round 4 (14abedd5): the whole-tree AST pass now runs only on timer-bearing files after blanking, one parse shared across detectors — local 27.1 s → 8.9 s, pre-pass mutation back to 25.7 s (CI was 38 s). Second changelog fragment folded by the orchestrator. Prescribed round → merges on green.

Issue #2817, opened 2026-09-09
Stale LSP snapshot: server reports TS2305 for an export present on disk, surviving mode=full rescans


apmantza:
[...]

Issue #1783, opened 2026-08-19
Untracked disk edits leave the LSP view stale indefinitely (no drift backstop)


apmantza:
An edit made outside the tracked write/edit path (a bash-tool bulk edit) leaves the primary LSP server's in-memory view of open documents stale indefinitely. The server keeps publishing pre-edit diagnostics; nothing detects that disk mtime is newer than the last synchronized content.

Proven live in the 2026-08-20 plegma dogfood session:

- `src/backends/model-meta.ts` and `src/mcp/stdio.ts` both have disk mtime 22:48:19.177Z (identical — one bulk operation). sessionstart.log 22:48:20.093 shows the only activity: `tool_result: skipped turn tracking - toolName="bash" (not write/edit)` — by design, no didChange was sent.
- The typescript server has published `diagnosticCount:2` for `src/mcp/stdio.ts` on every diagnostic sequence from 22:27 through 23:07+ (publicationIndex 77-82) — pre-edit content — while `tsc` on disk reports 0.
- The resync-on-read path exists (the 3000ms resync budget seen at 22:13) but fires only on tracked reads; nothing sweeps open documents for disk drift.
- Same family, second witness: `tests/unit/pi-codec.test.ts` produced repeated past-EOF demotions (42 then 54 cumulative) 12 and 29 minutes after its last edit — the server answering workspace rounds from a longer, pre-edit version. The past-EOF backstop demoted every arrival, so no wrong render escaped, but the upstream view never healed.

## Acceptance criteria

1. Detect drift: for documents the server holds open, a check (on workspace rounds, on a bounded cadence, or on the resync path) compares disk mtime/size against the last-synchronized state and resynchronizes when they diverge. State the trigger choice and its cost on hot paths (measured).
2. The bash/untracked-edit case heals within a bounded window without requiring a tracked read of the file. State the bound.
3. Regression test: open a document through the normal sync path, modify it on disk behind the server's back, and prove the pre-fix behavior (stale diagnostics keep serving) red, post-fix the view resynchronizes.
4. Observability: a bounded record naming the resynchronized file and the drift age.

Screens: freshness keys pair mtime with size (catalog shape 6); the empty/changed distinction preserves identity.

Coordination: the resync seam (`clients/pipeline.ts` resyncLspFile, `clients/lsp/index.ts`) is being touched by PR #1770 (in verify round). Merge its branch or land after it; do not race the seam.

[...]

PR #2823, opened 2026-09-09
fix(lsp): re-sync git-changed files and their open importers; scoped mode=full touches open dependencies (refs #2817)

apmantza:
Operating rule: derive language capability rows from the production import-facts predicate.

Kept: the test still uses one real fake LSP server and preserves the existing cache and resync assertions.

## Summary

- Fixed F8 by exposing `importFactsApplyTo` from the production import-facts provider.
- Derived one facts row and one no-facts row from `LANGUAGES` through that predicate.
- Printed selected ids: `facts=javascript; no-facts=ada`.

## Tests

- Edited `tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts` to pin the production capability matrix and real-server behavior.
- Added `clients/dispatch/facts/import-facts.ts` predicate reuse so provider selection and matrix selection share one seam.
- Red-first: the hard-coded Python row failed with `AssertionError: expected false to be true`.
- Mutation: changing the predicate to claim facts for every language failed with `Error: language registry lost a facts matrix class`.
- Targeted test: 1 file, 1 test passed.
- Ratchets and `tests/config/`: 44 files, 464 tests passed.

## Blast radius

`importFactsApplyTo` is called by `importFactProvider.appliesTo` and the language-neutral LSP test. The export adds no runtime state, durable record, or new failure path.

## Class sweep

Searched all `isJstsFactFile`, `importFactProvider`, and language-registry capability references. The provider is the single import-facts capability seam; the test is the only hard-coded `.py`/`.lua` matrix member.

## Observability

The test prints the selected registry ids. Existing `lsp_dependency_touch_capped` and `lsp_external_change_resync` records remain the bounded runtime evidence for the surrounding resync path; this predicate adds no new failure path.

## Test assessment

The existing language-neutral test is edited to derive facts and fallback classes from the production predicate. Its real fake-server path remains the regression proof for cache exclusion and dependency touch behavior.

## Round 5

F8 fixed. Prescribed option A applied. Handoff is uncommitted with no Git authority.

### Preflight

| gate | mirrored CI job | pass/fail |
| --- | --- | --- |
| build | Lint & type-check | pass |
| lint | Lint & type-check | pass |
| fmt:check | oxfmt format check (advisory) | pass |
| changelog:check | Unit tests | pass |
| check-changelog-fragments | Changelog fragment (fast-fail) | pass |
| check:lockfile | Lint & type-check | pass |
| tests/config | Unit tests | pass |
| generation-guard | Unit tests | pass |
| flake-shape-ratchet | Unit tests | pass |
| lsp-spawn-heavy-coverage | Unit tests | pass |
| ci-verdict | Unit tests | pass |
| check-pr-title | PR title | pass |
| check-close-keywords | Close-keyword syntax | pass |
| check-pr-body | PR body (advisory) | pass |


apmantza:
Verdict: redesign before merge.

PR #2823 repairs the direct same-size/same-mtime document path, and its
all-server touch scope preserves the existing coverage rule. It is not
merge-ready because Git recovery is unbounded, the dependency cap can serve
stale cached diagnostics, and required governance suites are red.

## HIGH

### F1: Git recovery bypasses drift pacing and has no batch bound

Probe: a real `LSPService` with 500 open TypeScript files and one recovered Git
path set produced `review-probe git-resync touches=500`; the expected bound was
the existing drift batch of four. The calls originate at
`clients/lsp/index.ts:3532-3555`.

The new path runs every open target in one `Promise.all`. It does not use the
four-file drift pacing or another replacement bound. A checkout changing 500
files can therefore enqueue 500 `touchFile` operations at once, multiplied by
every live server holder, and overload the TypeScript process and notify queues.

A: route recovered targets through the existing paced drift/resync scheduler.
B: add an explicit bounded batch with deferred targets and an honest record of
the deferred count. Keep one `lsp_external_change_resync` row for the event.

### F2: The 32-import cap still serves the cached result as confirmed

Probe: with 40 open imports, the real `runWorkspaceDiagnostics` seam touched
32 dependencies, emitted `cap-record-count=1` across two scans, returned one
result, and returned the independent payload `stale cached diagnostic` unchanged.
The behavior is at `clients/lsp/index.ts:8714-8737` and the cache hit is pushed
at `clients/lsp/index.ts:8760-8772`.

The cap records a degradation but does not change the cache result or its
confirmation state. If the changed dependency is item 33 or later, the server
view can remain stale while `mode=full` reports the requested file from cache.
That violates the operating rule and the claim that the scan remains honest
when capped.

A: when the cap trips, bypass the cache for the requested file and perform a
fresh bounded touch. B: return the cache result as unconfirmed and exclude it
from clean or authoritative reporting until the remaining dependencies are
repaired.

## MEDIUM

### F3: Uncovered files silently skip dependency repair

Probe: `WorkspaceDiagnosticsCacheContext.importsFor()` reads the persisted
snapshot’s forward `imports[filePath]` entry. The cache contract returns
`undefined` when that file was never covered. The new code returns immediately
at `clients/lsp/index.ts:8709-8710`, so an importer whose module facts were not
built receives no dependency touch and can still be served from cache.

This is not a review-graph lookup. A dependency edited before the module graph
was built, or a file never module-reported, remains outside the repair set.
The state-space table documents the limitation, but the operating rule says
every affected open TypeScript view is resynchronized before reporting.

A: make an uncovered TypeScript cache entry ineligible for a scoped cache hit
and touch the requested file. B: build or refresh module facts before admitting
the cache entry, with an explicit bounded failure result.

### F4: The new export is missing from the vi-mock export ratchet

Probe: `vitest run tests/config` failed `vi-mock-export-sweep.test.ts` with the
exact message that `resyncGitChangedFiles` changed the missing export set in
many whole-module mock baselines. The new export is at
`clients/lsp/index.ts:10095-10099`.

The PR’s preflight table says `tests/config` failed only on
`hook-await-bounds.test.ts`, but the full suite also fails the vi-mock sweep
because the production export baseline was not updated. This is a required
Unit-tests governance failure, not a pre-existing formatting issue.

Update the checked-in vi-mock export baseline through its documented workflow,
then rerun the full `tests/config` population.

### F5: The hook-await ratchet also rejects the new LSP awaits

Probe: `vitest run tests/config` reports
`clients/lsp/index.ts@161` as an unregistered unbounded await in
`hook-await-bounds.test.ts`.

The new asynchronous resync seam adds awaits to a module in the direct hook
helper population, but the PR adds no registration, exemption, or bounded
handoff. The preflight claim is incomplete and the required suite is red.

Register the intentional helper with its measured reason, or move the work
behind the project’s bounded hook handoff. Do not raise the ratchet without
explaining the new failure path and its bound.

## Cleared categories

- The submitted drift regression reaches the real `LSPService.touchFile` seam.
- The existing all-server coverage logic excludes holders that do not have the
  document open, so the new call does not intentionally spawn cold servers.
- `lsp_external_change_resync` is logged once after the batch at
  `clients/lsp/index.ts:3556-3565`, not once per file.
- The cap record is once per `(kind, subject)` per session. The probe observed
  one record across two capped scans.
- Build passed, and the targeted drift, cache, spawn-coverage, and flake-ratchet
  suites passed: 4 files, 121 tests.
- `git diff --check` passed.

## Could not verify

- The requested master-versus-head production reproduction used a real
  temporary Git repository, the real `lens_diagnostics mode=full` handler, and
  the real TypeScript launch path. Both scans returned zero confirmed LSP files;
  the second reported two unconfirmed files. The probe produced no valid
  stale/fresh diagnostic quote, so I do not treat it as green evidence.
- I did not run the requested per-file observability mutation. Source inspection
  shows the production row is outside the target loop, so the current path is
  one row per Git event.
- CI checks were queued during this review. `mergeStateStatus=BLOCKED` is not a
  pass signal.

## Named output

- Strong: `resyncGitChangedFiles` is a second fan-out scheduler beside the
  existing paced drift scheduler. Removing the shared drift scheduler would
  concentrate an unbounded event path in this new method, so the fix should
  fold onto the existing scheduler rather than add another pacing policy.
- Worth exploring: the persisted reverse-dependency snapshot is the admission
  boundary for both freshness and repair. A single explicit dependency-coverage
  value could prevent undefined, empty, and capped states from being confused.

## Test assessment

The edited `tests/clients/lsp/service-document-drift.test.ts` uniquely pins the
same-size/same-mtime Git recovery path through `LSPService.touchFile`. It does
not pin importer expansion, the 32-item cap, once-only cap telemetry, or the
all-server Git recovery path. Mutation evidence: removing the importer enqueue
and removing the scoped dependency-touch call both left the targeted suite
green at 16/16.

apmantza:
Round 2 (230f8d93): F1 folded onto the paced drift scheduler (500-file probe bounded at four per pass, deferred retained), F2 cap trip → fresh bounded touch, F3 uncovered/no-facts → ineligible for a scoped hit for every language, F4/F5 ratchets green. Language-neutral scope per #2817's clarification. Release-QA on the r1 head: SHIP, every row witnessed. Fresh verify follows.

apmantza:
VERDICT: merge after fixes

## Disposition table

| ID | Disposition |
| --- | --- |
| F1 | fixed |
| F2 | fixed |
| F3 | not fixed |
| F4 | fixed |
| F5 | fixed |

## New findings

### MEDIUM — F6: The F3 regression does not detect removal of the uncovered guard

1. Probe: after mutating `clients/lsp/index.ts:8685` from `if (!imports) return false` to `if (!imports) return true`, this command still passed:

   `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/lsp/workspace-diagnostics-cache.test.ts -t 'freshly scans a cached TypeScript file absent from the importer snapshot' --configLoader runner`

   Output: `Test Files 1 passed (1)\nTests 1 passed | 57 skipped (58)`. Expected: the F3 test turns red.

2. The uncovered-language test does not reach the new guard under mutation.

3. Its cache fixture sets `depIndexAtScan: false` at `tests/clients/lsp/workspace-diagnostics-cache.test.ts:1218`, so cache lookup rejects the entry before `importsFor(filePath)` runs. The test therefore proves the older cache rejection, not the new no-import-facts rule. Removing the guard can silently regress the exact acceptance criterion while this test remains green.

4. Option A: make the fixture cache-admissible, with missing `imports[filePath]`, then assert the fresh touch and stale payload exclusion. Option B: add a direct cache-admission case that proves the `importsFor` branch is reached, while retaining the production-path test.

### MEDIUM — F7: Language-neutral behavior is only mocked, not verified through the fake server

1. Probe: the claimed Python/Go coverage at `tests/clients/lsp/service-document-drift.test.ts:202-239` uses `makeClient(..., "python")`, a mocked `getReverseDepsFromIndex`, and a mocked `createLSPClient`. It does not launch `tests/fixtures/fake-lsp-server.mjs`, exercise registry selection, or derive import facts from a Python/Go fixture. Expected: a real fake-server Python or Go importer produces the same repair outcome as TypeScript, plus a no-facts language takes the fallback.

2. The non-TypeScript test changes only identifiers and the mocked return value; it can pass even if language-registry routing, the real facts loader, or the fallback admission is broken.

3. This leaves #2817's every-LSP-backed-language scope unproved. The PR body names broad families, but it does not enumerate the exact registry entries with import facts and the entries using the no-facts fallback. The missing real-server matrix also leaves the requested language-neutral acceptance unpinned.

4. Option A: add a real fake-server matrix through the production LSP path for one Python or Go importer with facts and one registry language without facts, and list the exact registry IDs in the PR body. Option B: split a real import-facts adapter test from a real fake-server resync test, but still enumerate the registry population and fallback rows.

## Cleared categories

- F1: 500 queued targets are paced by the existing four-file scheduler. The expanded probe reported `first-pass touches=4`; the next pass reported `resynced=4` and `deferred=556` for the combined queued and ordinary-drift work.
- F2: 40 open dependencies with the changed dependency beyond position 32 did not return the stale cached diagnostic. The cap mutation returned the stale payload and failed the test.
- F4: the vi-mock export ratchet passed in the targeted run and in preflight.
- F5: the hook-await ratchet passed in the targeted run and in preflight.
- Build passed on the merged tree.
- Targeted suites passed: 5 files, 153 tests.
- `npm run preflight` passed every listed gate.

## Could not verify

- The F1 probe ran under Vitest test mode, which suppresses `logLatency` writes. I could quote the scheduler result, but not a persisted `lsp_external_change_resync` row from that run. The source emits one row with `enqueued` and `deferred`.
- I did not obtain a valid end-to-end real TypeScript server quote for the reporter's `lens_diagnostics mode=full paths=[importer]` scenario on both master and this merged tree. The PR's tests use mocked clients, and the existing r1 attempt recorded zero confirmed LSP files. This remains unverified, not green evidence.
- CI status was read from PR metadata as clean with successful required checks, but this review's authoritative local evidence is the merged-tree build, targeted suites, mutations, and preflight.

## Named output

- Strong: the shared drift scheduler is the correct single pacing seam. The 500-file probe reached four touches per pass, while the existing stat-drift population also contributes to the pass's deferred count.
- Worth exploring: `depIndexAtScan` and `imports[filePath]` are separate admission facts. The F3 test conflates them, which allowed the guard mutation to pass.

apmantza:
Round 3 (2c70d06d): uncovered-import guard now proven by a cache-admissible fixture (mutation red), language-neutral resync verified through the real fake-server path with the registry ids named in the body. Second fragment folded by the orchestrator. Verify follows.

apmantza:
VERDICT: merge after fixes

Round 3 verification for PR #2823 at `2c70d06d`, after merging `origin/master`
(already up to date).

## MEDIUM

### F6: The uncovered-import guard remains mutation-unproven

1. Probe: after the compile-valid mutation at `clients/lsp/index.ts:8685`
   from `if (!imports) return false` to `if (!imports) return true`, this
   command still passed:

   `PI_LENS_HOME=$PWD/.probe-home npm run build && PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/lsp/workspace-diagnostics-cache.test.ts -t 'freshly scans an admissible cached TypeScript file absent from importer facts' --configLoader runner`

   Output: `Test Files 1 passed (1)` and `Tests 1 passed | 57 skipped (58)`.
   Expected: the test turns red.
2. The new cache-admissible fixture reaches `importsFor(filePath)` and checks
   a fresh touch and stale-payload exclusion, but the focused test does not
   observe the mutated production module instance.
3. The guard can therefore be deleted while the claimed regression remains
   green. The PR body correctly discloses this gap, so the round-3 claim is
   not evidence of mutation resistance.
4. Make the test load the same production module instance as the guard, or add
   an independent production-path assertion whose mutation turns red. Rebuild
   between the source mutation and the test.

### F7: The language-neutral matrix does not prove the claimed facts path

1. Probe: the test contains the real seams
   `await service.touchFile(...)` at
   `tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts:98-112`
   and the real handler call through
   `createLensDiagnosticsTool(...).execute(... mode: "full" ...)` at
   `:151-167`. It launches `tests/fixtures/fake-lsp-server.mjs` through the
   configured child command. After forcing the cache facts lookup to
   TypeScript-only, the command still passed:

   `PI_LENS_HOME=$PWD/.probe-home npm run build && PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts --configLoader runner`

   Output: `Test Files 1 passed (1)` and `Tests 1 passed (1)`.
   Expected: the Python row turns red.
2. The Python row does not prove that Python facts drive repair. The test
   explicitly calls `resyncGitChangedFiles([pythonDependency, luaImporter])`,
   so the Python importer can be repaired from the persisted reverse-deps edge
   independently of the scoped `importsFor` lookup.
3. The requested language-neutral acceptance remains unproved at the facts
   admission boundary. The test uses a hand-seeded snapshot edge and a custom
   server id, so it cannot detect a facts-loader or registry-routing regression.
4. Keep the real fake-server path, but make the Python/Go row depend on the
   scoped facts lookup rather than direct Git-resync inclusion. Add a separate
   registry-backed fallback row and assert both stale payloads are excluded.

### F8: The claimed registry population is not exercised by the test

1. Probe: `tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts:44`
   configures the custom id `"fake-language-neutral"`, while
   `clients/dispatch/facts/import-facts.ts:99-105` registers a provider whose
   `appliesTo` is only `isJstsFactFile`. `clients/file-kinds.ts:146-159` limits
   that provider to ordinary JS/TS extensions. Expected: the test exercises a
   real registry id and verifies the provider/fallback classification.
2. The matrix does not select `python`, `go`, or `lua` from
   `clients/language-registry.ts`; it selects one ad hoc server for `.py` and
   `.lua`.
3. The PR body calls `typescript`, `typescriptreact`, `javascript`,
   `javascriptreact`, `python`, and `go` “facts-bearing,” but the import-facts
   provider is JSTS-only. Python and Go are registry entries at
   `clients/language-registry.ts:343` and `:413`, but they are not covered by
   this provider. The claim conflates persisted reverse-deps snapshot data with
   the import-facts provider, leaving the population contract ambiguous.
4. A: use actual registry server ids and document that Python/Go facts come
   from a different provider, then test that provider. B: narrow the body claim
   to the actual `isJstsFactFile` population and add explicit Python/Go fallback
   rows.

## Disposition table

| ID | Disposition |
| --- | --- |
| F3 | fixed |
| F6 | not fixed |
| F7 | not fixed |

## Cleared categories

- F1 and F2 remain fixed by the round-2 implementation.
- The F3 fixture is cache-admissible and asserts fresh touch plus stale-payload
  exclusion, but its guard mutation is separately reported as F6.
- The real fake-server test reaches `LSPService.touchFile` and the real
  `lens_diagnostics` full handler. It does not use `getReverseDepsFromIndex` or
  `createLSPClient` mocks.
- `vitest.config.ts:277` admits the new test to the serialized
  `lsp-spawn-heavy` lane. `lsp-spawn-heavy-coverage` and
  `flake-shape-ratchet` pass: 2 files, 61 tests.
- Build passes. The touched LSP suites pass: 2 files, 59 tests.
- `npm run preflight` passes every listed gate.
- The registry ids named in the body all exist in `clients/language-registry.ts`.
  The facts-provider claim is still inaccurate or underspecified for Python and
  Go, as described in F8.

## Could not verify

- I could not produce a valid master-versus-head reporter quote for the full
  end-to-end temporary-Git scenario. The required evidence is “stale on master,
  fresh here”; this round has no such quote and does not treat the scenario as
  verified.
- Exact-head `node scripts/ci-verdict.mjs 2823` reports `Lint & type-check`
  success, but `Unit tests` remains in progress. It also reports a failing
  `PR body (advisory)` check and in-progress advisory prose checks. The local
  preflight table is green; CI is not fully witnessed.

## Named output

- Strong: the test currently combines two repair mechanisms, direct Git
  resync and scoped cache dependency repair. Removing the shared facts seam
  would leave the direct Git path carrying the test, so the matrix should split
  those mechanisms before it can certify language coverage.

## Test assessment

The edited cache test adds the cache-admissible uncovered-import case and the
dependency-cap case. The new language test adds one real child-process case, but
its snapshot and custom server setup do not certify registry/provider coverage.
The two mutation transcripts above are the deciding evidence.

apmantza:
Round 4 (948ffba5): guard proven through the production module instance, language matrix selected from the registry by capability with the ids named from the test's own output, drift pacing count settled against #1783. Verify follows.

apmantza:
VERDICT: merge after fixes

Round 4 verification for PR #2823 at 948ffba5, on a checkout after merging
origin/master locally.

## Disposition table

| ID | Disposition |
| --- | --- |
| F6 | fixed |
| F7 | fixed |
| F8 | not fixed |
| CI-red | fixed locally; exact-head Unit tests still in progress |

## Findings

### MEDIUM — F8: The matrix still does not derive capability from the facts loader

1. Probe: `LANGUAGES.filter(entry => entry.extensions.some(... .py/.lua ...))`
   produces `python,lua`, matching the body's assertion at
   `tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts:226`.
   The actual import-facts provider at
   `clients/dispatch/facts/import-facts.ts:103-104` applies only when
   `isJstsFactFile(ctx.filePath)` is true. Expected: the facts-selected
   population comes from the provider capability, not only from extensions.
2. The test selects Python and Lua by hard-coded extension predicates, then
   manually seeds a persisted project snapshot. It does not recompute a
   registry population from the facts provider capability.
3. Python is not in the `isJstsFactFile` population, so the test's Python row
   proves the snapshot reverse-dependency loader and fake server, but not the
   claimed registry-plus-facts capability selection. The body therefore
   overstates what F8 covers.
4. Option A: expose or reuse the facts capability and derive the selected
   registry ids from it, then assert the generated ids. Option B: narrow the
   claim to registry extension selection plus persisted-snapshot coverage and
   document that `python,lua` are not import-facts-provider capability rows.

## Cleared categories

- F6 is fixed. The guard mutation `if (!imports) return false` to `true`,
  rebuilt with `npm run build`, failed at
  `workspace-diagnostics-cache.test.ts:1252` with
  `AssertionError: expected false to be true`; the test reached the production
  `LSPService` module. `tests/config/module-instance-coverage.test.ts` passed.
- The runtime and test resolve the same module path:
  `file:///home/akis/.plegma/work/sub-mtue4dsw-20/clients/lsp/index.js`.
  The test import is `tests/clients/lsp/workspace-diagnostics-cache.test.ts:981`.
- F7 is fixed. Forcing the lookup to TypeScript-only caused the real fake-server
  test to fail because the trace lacked
  `textDocument/didChange file:///home/akis/.plegma/work/sub-mtue4dsw-20/.probe-lsp-language-*/workspace/dependency.py`.
  Removing the Python importer edge then produced no dependency touch on the
  unmutated code.
- The registry recomputation from the test's `LANGUAGES` extension predicate
  is `python,lua`, and the body names `python,lua` from its own output. The
  remaining issue is that this is extension selection, not facts capability
  selection.
- The #1783 pacing code changed. It adds `pendingResync` to
  `clients/lsp/document-drift.ts`, applies the existing four-resync budget,
  records completed keys, and skips those keys in the ordinary drift list.
  The code, rather than the pin, implements the no-double-schedule rule.
- Independent probe: eight queued files completed in exactly two passes of four;
  output assertions were `first.resynced=4`, `second.resynced=4`, `calls=8`,
  and `unique calls=8`. The PR pacing test also passed with four first-pass
  touches and one second-pass resync for five files.
- Merged-tree targeted suites passed: 4 files, 81 tests. `tests/config` passed:
  42 files, 385 tests. `git diff --check` passed.
- `npm run preflight` passed every gate: build, lint, fmt:check,
  changelog:check, check-changelog-fragments, check:lockfile, tests/config,
  generation-guard, flake-shape-ratchet, lsp-spawn-heavy-coverage, and
  ci-verdict.
- Exact-head `node scripts/ci-verdict.mjs 2823` reports `Lint & type-check`
  success, all completed non-advisory checks success, and `Unit tests` in
  progress. The earlier drift pacing CI red is not reproduced locally or by
  completed exact-head checks, but the gating run is not finished.

## Could not verify

- I could not produce the requested reporter end-to-end quote showing “master
  stale, head fresh.” The merged-tree real fake-server test proves stale cache
  exclusion and dependency `didChange`, but it does not run the same temporary
  Git scenario once on `origin/master` and once on this head. I do not treat
  that scenario as green evidence.
- CI `Unit tests` remained in progress at the final exact-head verdict read.

## Named output

- Strong: `DocumentDriftTracker` is the single pacing seam. Removing the shared
  queue would move the four-file bound back into `resyncGitChangedFiles` and
  recreate the competing scheduler shape.
- Worth exploring: the persisted snapshot loader and the dispatch import-facts
  provider expose different meanings of “facts.” Naming the coverage source in
  the cache context would prevent provider capability and snapshot presence from
  being conflated.

## Test assessment

- `tests/clients/lsp/workspace-diagnostics-cache.test.ts` uniquely pins an
  admissible cache entry with missing importer facts and reaches the compiled
  production module instance.
- `tests/clients/lsp/workspace-diagnostics-language-neutral.test.ts` uniquely
  pins real fake-server cache exclusion, Python snapshot-edge dependency touch,
  facts-edge removal, and Lua fallback. It does not prove import-facts-provider
  capability enumeration.
- `tests/clients/lsp/service-document-drift.test.ts` pins same-size/same-mtime
  Git recovery, language-neutral importer recovery, and shared four-file pacing.
- `tests/fixtures/fake-lsp-server.mjs` adds opt-in synchronous URI tracing for
  real `didChange` observation without changing protocol behavior.

apmantza:
Round 5 (e81e1a02): the language matrix now derives capability from the production import-facts predicate and names the selected registry ids from its own output (red-first and mutation quoted). Prescribed round → merges on green.

Issue #2625, opened 2026-09-06
Workflow-shell portability guard: no bash-4-only builtins (mapfile/readarray/${var,,}) in `shell: bash` steps that can run on macOS (recurrence #2588 r2)


apmantza:
## Recurrence

PR #2588 round 2 replaced `read -ra` with `mapfile -t` at three `shell: bash` call sites in `.github/workflows/ci.yml`; the `install-test` matrix leg on `macos-latest` runs `/bin/bash` 3.2.57, which has no `mapfile`, and failed with exit 127 (job 101527965573). It was the repo's first use of the builtin, so no prior art proved it portable, and the JS-level regression test models mapfile's semantics, not its availability. The gate script (ci-verdict.mjs, #2609) did not see the red job.

## Guard

A `tests/config/` sweep over `.github/workflows/*.yml`: for every `run:` block whose job or matrix can execute on a macOS runner, reject bash-4-only constructs (`mapfile`, `readarray`, `${var,,}`, `${var^^}`, `declare -A`, `|&`, `;;&`). Steps pinned to ubuntu/windows are exempt (labels.yml's `${PAYLOAD_SHA,,}` runs on ubuntu). Alternative: run `actionlint` with shellcheck's `-s bash` and a target-version note. Red-first against ci.yml at fdd04be5.

Filed from the #2588 verify round; tracked under #2608 cluster (d).

apmantza:
From the #2588 verify: the same class is already latent on master, pre-dating #2588 — `${PAYLOAD_SHA,,}` (bash 4.0 case conversion) appears in ~20 workflow steps, several on macOS legs: .github/workflows/ci.yml:624 (install-test) and install-smoke.yml:159/276/419 (three jobs with os: [ubuntu-latest, macos-latest]). All sit behind `if: github.event_name == 'repository_dispatch'` (the merge-train path), so they have never executed on a pull_request; the macOS log on #2588 proves `shell: bash` there is /bin/bash 3.2, so a merge-train dispatch would fail those steps. That is the recurrence this guard prevents, and it argues for the class guard over a one-line fix. Include those sites in the red-first proof.

apmantza:
Also fold in when fixing: tests/scripts/supply-host-provided-deps.test.ts:15/48/50 still describe the removed `mapfile -t` idiom in prose and in the test name (the assertion itself is correct); update the wording when the guard lands.

apmantza:
[...]

PR #2830, opened 2026-09-09
test(governance): workflow-shell portability sweep — no bash-4-only builtins on steps that can run on macOS (refs #2625, #2784)

apmantza:
## Summary

Extend the workflow shell portability scanner so command substitutions inside double quotes re-enter executable shell lexing with nested depth tracking. Revert the unnecessary Ubuntu-only `host-range-smoke` rewrite and pin job-local runner reachability with a sibling-job fixture.

## Tests

- `tests/config/workflow-shell-portability.test.ts`: add double-quoted `$()` fixtures for `mapfile`, a no-needle substitution, a single-quoted literal, nested `declare -A`, and job-local runner reachability.
- `tests/support/workflow-shell-portability.ts`: recursively lex `$()` bodies and treat `$(` as a command-position boundary for `declare -A`.
- Red-first F6: before the re-entry branch, the targeted suite failed with `expected [] to deeply equal ["mapfile"]`.
- Mutation F7: broadening reachability to any job matrix failed the sibling-job and matrix-neutering tests.
- `npm run build`: pass.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/config --configLoader runner`: 43 files, 407 tests passed.
- `npm run preflight`: run last; see the table below.

## Blast radius

The changed seam is `findBash4PortabilityFindings` → `canRunOnMacOS` / `lexShell` → `needlePattern`. Callers are the workflow portability config tests and the workflow population sweep. No runtime production module or durable record changes.

## Class sweep

The shell scanner population remains `.github/workflows/*.{yml,yaml}`. The reachability matrix covers literal runners, matrix dimensions, matrix `include`, reusable inputs, explicit step exclusions, and a fixed Ubuntu job beside a macOS matrix job. The shell matrix covers comments, single and double quotes, parameter expansions, nested command substitutions, quoted and unquoted heredocs, continuations, and non-Bash shells.

## Observability

This governance scanner reports bounded finding rows containing workflow, job, step, and needle identity. No new failure path or telemetry surface is introduced.

## Test assessment

The edited test file uses the real YAML loader and scanner seam. It does not replace an in-process implementation with a double. The red-first probe reaches the original false-negative path, and the compile-valid mutations make the re-entry and job-local reachability guards red.

## Round 3

- F1–F5: fixed in round 2 and retained.
- F6: fixed by recursively re-entering executable `$()` bodies inside double quotes, including nested substitutions. The prescribed fixtures pass.
- F7: fixed by reverting the `host-range-smoke` rewrite. Before: the review reported the job's reachability as `[]` while the rewrite remained at the Ubuntu-only site. After: the job remains `[]`, and the current-tree sweep is clean without that rewrite. The cause was scope drift between reachability and rewrite selection: the rewrite treated every `${PAYLOAD_SHA,,}` occurrence as a target instead of using the job-local `runs-on` result.

| job `runs-on` | job-local matrix | expected macOS reachability |
| --- | --- | --- |
| `matrix.os` | `os: [macos-latest]` | yes |
| `ubuntu-latest` | `os: [macos-latest]` on the same job | no |
| `ubuntu-latest` | `os: [macos-latest]` on a sibling job | no |

Preflight

The final `npm run preflight` table is appended below.

| gate | mirrored CI job | pass/fail | first red line |
| --- | --- | --- | --- |
| build | Lint & type-check | pass | |
| lint | Lint & type-check | pass | |
| fmt:check | oxfmt format check (advisory) | pass | |
| changelog:check | Unit tests | pass | |
| check-changelog-fragments | Changelog fragment (fast-fail) | pass | |
| check:lockfile | Lint & type-check | pass | |
| tests/config | Unit tests | pass | |
| generation-guard | Unit tests | pass | |
| flake-shape-ratchet | Unit tests | pass | |
| lsp-spawn-heavy-coverage | Unit tests | pass | |
| ci-verdict | Unit tests | pass | |
| check-pr-title | PR title | pass | |
| check-close-keywords | Close-keyword syntax | pass | |
| check-pr-body | PR body (advisory) | pass | |


apmantza:
Verdict: redesign.

The detector is not merge-ready. It passes the authored fixtures and the clean
tree sweep, but the adversarial probes find core false negatives and a false
positive in the claimed shell-code boundary.

## Findings

### F1 — HIGH — double-quoted Bash expansions are missed

1. Probe: `findBash4PortabilityFindings(loadWorkflow(install-smoke.yml), ...)`
   returned `[]`; expected at least the `${PAYLOAD_SHA,,}` findings in the
   macOS legs of `smoke`, `pi-load`, and `mise-repro`. The same construct is in
   `.github/workflows/install-smoke.yml` inside `"${PAYLOAD_SHA,,}"`.
2. `blankShellProse` blanks the complete double-quoted span at
   `tests/support/workflow-shell-portability.ts:124-133`, so it removes real
   parameter expansion code along with quoted prose.
3. This is a direct miss of the issue's bash-4 portability shape. A forbidden
   construct already exists in a macOS matrix workflow, while the governance
   sweep reports it clean.
4. Parse shell structure enough to preserve parameter expansions and operators
   inside double quotes, or use a Bash-aware tokenizer. Keep the separate
   string-literal policy explicit, then add a regression fixture for
   `"${VALUE,,}"` and `"${VALUE^^}"`.

### F2 — HIGH — matrix `include` runners are not followed

1. Probe: a job with `strategy.matrix.include: [{ os: macos-latest }]`,
   `runs-on: ${{ matrix.os }}`, and `run: mapfile -t x` returned `[]`; expected
   `{ job: "j", step: "x", needle: "mapfile" }`.
2. `canRunOnMacOS` only reads `job.strategy.matrix[matrixName]` at
   `tests/support/workflow-shell-portability.ts:79-81`; it never evaluates
   `include` rows.
3. GitHub Actions treats the include row as a valid matrix expansion, so this
   is an evasion of the issue's matrix-runner acceptance, not an exotic input.
4. Expand the matrix runner model to inspect direct dimensions and `include`
   rows, and add a fixture whose only macOS value comes from `include`.

### F3 — MEDIUM — quoted heredoc text is falsely treated as shell code

1. Probe: a macOS step containing `cat <<'EOF'`, a body line `mapfile -t x`,
   and `EOF` returned a `mapfile` finding; expected `[]`. An unquoted heredoc
   with `$(mapfile -t x)` also returned a finding, which is the expected hit.
2. `blankShellProse` has no heredoc state and scans the quoted heredoc body as
   ordinary shell source at `tests/support/workflow-shell-portability.ts:113-145`.
3. This violates the stated comments-and-strings boundary and can block a
   workflow that only emits literal documentation. The mutation probe confirms
   the blanking guard is live, but not that its shell model is correct.
4. Track quoted and unquoted heredoc delimiters. Blank quoted bodies, while
   retaining command substitutions in unquoted bodies, and add both fixtures.

### F4 — MEDIUM — lexical Bash-4 forms evade the needle matcher

1. Probes returned `[]` for `map\\\nfile -t x` and for `${VALUE,}`; expected
   `mapfile` and the Bash-4 lowercase-expansion family respectively.
2. The detector applies regular expressions directly to physical source lines
   at `tests/support/workflow-shell-portability.ts:103-110`; it does not join
   shell line continuations and only lists the double-comma expansion form.
3. The line-continuation case is covered by the issue's named `mapfile` shape.
   The single-comma form is not named in the acceptance list, but it is part of
   the broader claim to reject Bash-4-only constructs. Both are bypasses of the
   detector's stated purpose.
4. Normalize shell continuations before matching and enumerate the complete
   Bash-4 parameter-expansion family, or document and test a deliberately
   narrower policy that the issue explicitly accepts.

### F5 — MEDIUM — step conditions are ignored

1. Probe: a macOS job with `if: runner.os != 'macOS'` and `run: mapfile`
   returned a `mapfile` finding; expected `[]` because the step cannot execute
   on macOS.
2. The step type omits `if`, and `findBash4PortabilityFindings` filters only
   by job runner and shell at `tests/support/workflow-shell-portability.ts:39-43`.
3. This creates false blockers for explicitly non-macOS steps. The issue says
   to inspect steps that can run on macOS, so the condition is part of the
   acceptance shape.
4. Model the supported runner predicates conservatively: recognize the
   explicit `runner.os != 'macOS'` exclusion, test it, and document unsupported
   expressions as intentional findings rather than silently claiming coverage.

## Cleared categories

- Literal macOS runners, direct matrix dimensions, default bash, `shell: bash`,
  and `shell: bash -e {0}` produced expected hits.
- Comments, ordinary quoted strings, `pwsh`, `sh`, Ubuntu-only jobs, and
  Windows-only jobs produced `[]` in the authored and adversarial fixtures.
- A reusable workflow input with `default: macos-latest` produced the expected
  hit. An input with no default is conservatively treated as potentially macOS.
  The detector does not resolve a caller workflow's `with:` value; a caller
  passing macOS to an input whose reusable-workflow default is Ubuntu remains
  unverified and can be missed.
- The existing YAML seam is used: `clients/deps/js-yaml.js`; no second loader
  was introduced.
- The planted CI probe produced the required red:
  `.github/workflows/ci.yml`, job `install-test`, step `Review probe bash 4
  builtin`, needle `mapfile`.
- Mutation proof: changing `return stripSource(chars.join(""))` to
  `return stripSource(source)` made the quoted-comment tests fail, including
  `Comment only` and `Comment and quote`.

## Verification

- `npm run build`: pass.
- `vitest run tests/config/workflow-shell-portability.test.ts`: pass after
  restoring probes, 7 tests.
- `vitest run tests/config`: pass, 43 files and 392 tests.
- `npm run preflight`: every reported gate passed, including build, lint,
  format, changelog checks, `tests/config`, generation, flake-shape,
  lsp-spawn-heavy coverage, and ci-verdict.
- Current scratch tree: clean after removing all probes.

## Could not verify

- Full CI was not rerun locally. The PR's reported required checks are green on
  `d39c08d9`, but that does not close the adversarial detector gaps above.
- Cross-file reusable-workflow caller resolution was not available through the
  detector's single-workflow API; the caller-passes-macOS/default-Ubuntu case
  needs an explicit policy and fixture.

## Named output

- Strong: the detector should have one shell-structure seam that distinguishes
  executable code, quoted literals, heredocs, and expansions. Deleting the
  current blanker would expose the false-positive tests, but extending its
  ad-hoc state machine will continue to duplicate shell parsing unless the
  replacement owns these boundaries centrally.

apmantza:
Round 2 (3e64536b): the sweep now runs on a bash lexer (quoted contexts, double-quoted expansions kept visible, heredocs, continuations), follows matrix include and reusable inputs, honours the two macOS-excluding step conditions; one fixture per state-space cell, the five review cases red-first, both mutations quoted. Detector redesign → fresh verify follows.

apmantza:
Verdict: redesign.

The round-2 redesign fixes F1–F5, but the lexer still misses executable command substitutions inside double quotes. The workflow rewrite also changes an Ubuntu-only job unnecessarily. The exact-head required Unit tests check is still in progress.

## Findings

### F6 — MEDIUM — command substitutions inside double quotes are missed

1. Probe: `findBash4PortabilityFindings` on a macOS step with `run: echo "$(mapfile -t x)"` returned `[]`; expected one `mapfile` finding. The probe failed in `tests/config/_review2830-probe.test.ts` with `expected [] to have a length of 1`.
2. `lexShell` blanks every double-quoted character except `${...}` at `tests/support/workflow-shell-portability.ts:186-193`. It does not preserve or recursively lex `$()`.
3. Bash executes the command substitution on macOS, so this is a false negative in the same shell-code boundary that F1–F5 claim to cover. The existing unquoted heredoc substitution fixture passes because its `$()` remains visible, which does not cover the double-quoted form.
4. Option A: add nested command-substitution state to the lexer and preserve its executable contents while retaining the surrounding double-quoted literal policy. Option B: use a shell tokenizer that exposes command substitutions and parameter expansions as executable subtrees. Add a red-first fixture for `echo "$(mapfile -t x)"`.

### F7 — LOW — one workflow rewrite targets an Ubuntu-only job

1. Probe: `.github/workflows/install-smoke.yml:630` is inside `host-range-smoke`, whose job declares `runs-on: ubuntu-latest` at line 603 and has no macOS matrix. The detector correctly excludes that job; the current-tree sweep returns `[]`, while the diff still rewrites its `${PAYLOAD_SHA,,}` expression.
2. The round-2 change applies the portable `tr` rewrite to a site that cannot execute on macOS.
3. For a valid 40-hex SHA, `printf '%s' "$PAYLOAD_SHA" | tr '[:upper:]' '[:lower:]'` is byte-equivalent to `${PAYLOAD_SHA,,}` for the accepted input, so the runtime change is low risk. It is nevertheless unnecessary scope and shows the workflow rewrite was not derived from the reachability result. The other three sites are in `smoke`, `pi-load`, and `mise-repro`, each with `runs-on: ${{ matrix.os }}` and `os: [ubuntu-latest, macos-latest]`.
4. Remove the Ubuntu-only rewrite, or document why all occurrences are intentionally normalized. Keep the reachability-driven sweep as the source of the replacement list.

## Cleared categories

- F1 fixed: quoted `${VALUE,,}` and `${VALUE^^}` are detected.
- F2 fixed: a macOS value supplied only through `matrix.include` is detected.
- F3 fixed: quoted heredoc bodies are blanked, while an unquoted heredoc `$(mapfile ...)` is detected.
- F4 fixed: continuations and single-character parameter operators are detected.
- F5 fixed: `runner.os != 'macOS'` and `matrix.os != 'macos-latest'` are excluded; `if: ${{ !startsWith(matrix.os, 'macos') }}` remains conservatively reachable.
- `shell: sh` is excluded. A step without `shell:` in a macOS job is treated as Bash, including a job whose `defaults.run.shell` is Bash.
- `${!var}` and `${var:0:2}` are Bash-3 forms. `${var@L}` is Bash-5 syntax. Neither belongs in the Bash-4 needle set, and `@L` is not listed.
- The current `.github/workflows` population produces no findings.
- Shape 43 in `AGENTS.md:470` accurately describes the implemented F1–F5 boundaries, matrix reachability, and mutation requirements. It does not mention the newly proven `$()`-inside-double-quotes boundary and should be extended when F6 is fixed.

## Mutation evidence

- F1 mutation: replacing `index = preserveExpansion(chars, index)` with a compile-valid blanking assignment made `flags 'double-quoted parameter expansions'` fail; 16 other tests passed.
- F2 mutation: replacing the matrix-`include` condition with compile-valid `false &&` made `follows a macOS runner introduced by matrix include` fail; the same run also showed the F1 mutation before restoration. After restoring both mutations, the targeted suite passed 17/17.

## Verification

- `PI_LENS_HOME=$PWD/.probe-home npm run build`: pass.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/config/workflow-shell-portability.test.ts --configLoader runner`: pass, 17 tests.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/config --configLoader runner`: pass, 43 files and 402 tests, 53.73 seconds.
- `PI_LENS_HOME=$PWD/.probe-home npm run preflight`: every reported gate passed.
- Current workflow sweep: `[]`.
- Exact-head CI verdict for `3e64536b`: `Lint & type-check` passed; `Unit tests` was still in progress. Other non-advisory checks reported by `ci-verdict.mjs` passed or were explicitly skipped.
- The scratch checkout merged `origin/master` first, producing local merge commit `d87588a`; temporary probe files and source mutations were removed. No Git commit or branch operation was performed by this review.

## Could not verify

- The exact-head required `Unit tests` check had not completed when this review was written. Local `tests/config` and preflight passed, but that does not replace the remote check.
- No macOS runner was available locally. Reachability was verified from the workflow YAML and the detector's matrix model.

## Named output

- Strong: `lexShell` is now the single lexical seam for this detector, but its state model is still incomplete for nested executable regions. The deletion test is already satisfied by the quoted-string and heredoc fixtures; extending this seam is preferable to adding needle-specific exceptions.

## Previous-round disposition

| Finding | Disposition |
| --- | --- |
| F1 | fixed |
| F2 | fixed |
| F3 | fixed |
| F4 | fixed |
| F5 | fixed |

apmantza:
Round 3 (bc3e6941): `$(…)` inside double quotes re-enters code with nesting tracked (mutation red), the ubuntu-only rewrite reverted and the reachability cause fixed. Prescribed round → merges on green.

Issue #2536, opened 2026-09-03
No CI lane runs vitest on Windows: audit every skipIf(platform !== win32) in tests/ against the 'name the lane that satisfies it' rule


apmantza:
Named output from the #2533 review (2026-09-03). The Unit tests job is `ubuntu-latest` only; `windows-latest` appears in `install-test` (tarball smoke, no vitest) and `lifecycle-smoke.yml` (nightly PowerShell). So every `it.skipIf(process.platform !== "win32")` in `tests/` is a dev-box-only guard that the authoritative gate cannot see — e.g. `tests/scripts/prune-agent-worktrees.test.ts:1221` ("reaps under the registered argv when the listing itself fails"), and the #2501 case before #2533 round 2 made it cross-platform. AGENTS.md:3351 ("for every environment gate, name the lane that satisfies it") is unmet for all of them.

Acceptance: (1) inventory every win32-gated test (grep-guarded going forward: a `skipIf(process.platform` must carry a `// lane:` annotation naming the CI lane or `dev-box-only` with a reason); (2) decide: add a targeted `windows-latest` unit lane running exactly the win32-gated files (cheap — a dozen files) OR document the dev-box-only stance in AGENTS.md; (3) convert gates that are technique artifacts (not real platform divergences) to cross-platform variants via the seams the tests already own, per the #2533 round-2 pattern.

Refs #2533, #2501, #2519.

apmantza:
Tracked under the test-harness umbrella #2608.

apmantza:
#2793 merged (dca72a33): the Windows vitest lane exists in CI now, so every `skipIf(process.platform !== "win32")` in `tests/` has a lane that satisfies it.

Remainder before this closes: the audit half — a governance assertion (or a one-time sweep pasted here) that each win32 `skipIf` names the Windows lane per the "name the lane that satisfies it" rule, and that the lane actually executes those tests (a count of win32-only tests run on the Windows job > 0, read from the job log). If #2793's body already covers the audit, close this crediting it.

apmantza:
Closed by #2793 (the Windows vitest lane) and #2833 (merged 711483dd): every win32 gate names the Windows lane with an adjacent header, the lane's file population is computed by one shared script for both the job and the governance sweep, and the job prints the executed win32-gate count each run.

PR #2833, opened 2026-09-09
test(governance): every win32 gate names the Windows lane and the lane is proven to run them (refs #2536, #2784)

apmantza:
Operating rule: The Windows Vitest population has one executable source, and governance independently proves every detected Windows-only gate enters that population.

Kept: The detector still blanks comments and strings and requires the immediately preceding `// lane: windows-vitest` header.

## Summary

- HIGH-1 fixed: `scripts/lib/win32-gate-population.mjs` computes the source-derived gate population. The Windows job invokes it, and the governance sweep compares its output with an independent detector.
- MEDIUM-1 fixed: The Windows job records `win32-gate population: N files, M tests executed` in the step summary from the same script. This count is run evidence, not a test assertion.
- Updated `knip.jsonc` and the existing `.changelog/fix-2536-win32-gate-lane.md` fragment.

## Round 2

HIGH-1 — fixed.

The reviewer’s pre-fix mutation changed the workflow’s per-file filter to a constant workflow-file filter. Before the fix, the focused suite stayed green:

```text
Test Files  1 passed (1)
Tests  2 passed (2)
```

After the shared population filter was mutated with the compile-valid constant `TEST_FILE.test("ci.yml")`, the governance suite failed:

```text
FAIL tests/config/win32-gate-lane.test.ts
AssertionError: expected 0 to be greater than 0
```

MEDIUM-1 — fixed.

The job-side summary now computes the executed file count and gate count from the shared script. The latest completed Windows job available at handoff was run `34389484675`, job `102594091065`, and quoted `Windows Vitest subset: 77 files`. That older log is execution evidence only, not a durable test assertion. The next Windows run will additionally quote the new `win32-gate population: N files, M tests executed` summary.

## Tests

- `tests/config/win32-gate-lane.test.ts`: independently detects gate-bearing files, checks immediate headers, and proves the shared population includes every detected file.
- `tests/config/windows-vitest-workflow.test.ts`: checks the exact shared-script invocation, absence of inline `git grep`, and summary wiring.
- `tests/config/knip-entry-coverage.test.ts`: confirms the directly invoked script is registered.
- `tests/config/`: 45 files, 410 tests passed.
- Focused governance suite: 2 files, 5 tests passed; after the mutation, 1 test failed as expected.
- `npm run build`: passed.
- `npm run lint`: passed.
- `npx oxfmt --check` on touched source/config files: passed after formatting.

## Blast radius

The changed seam is the new `getWin32LaneFiles` / `findWin32Gates` script API. Callers are the Windows CI job and `tests/config/win32-gate-lane.test.ts`; `windows-vitest-workflow.test.ts` observes the job text. No production runtime, durable record, callback, or entry point changes.

Call-tree diff:

```text
- ci.yml → inline git grep/filter → Windows Vitest file list
+ ci.yml → win32-gate-population.mjs → Windows Vitest file list
+ win32-gate-lane.test.ts → win32-gate-population.mjs
```

## Class sweep

Searched all `tests/**/*.test.ts` files for the exact `it`/`describe` `process.platform` win32 gate forms. The independent detector found 16 gates across 16 files. Searched both workflow governance tests for the old inline `git grep` population; the old seam is removed. Consolidation verdict: fold onto the shared script because deleting it would recreate the duplicated population logic in the workflow and governance test.

## Observability

The Windows job summary preserves the bounded file and gate counts for each run. It does not claim that Linux governance tests prove Windows execution. The latest Windows evidence remains the quoted run-level file count until a post-fix Windows job reports the new summary.

## Test assessment

The edited governance tests use the real filesystem and workflow parser. The independent detector is retained so a source-population filter mutation reds the test. No mocks or fake in-process stores were added.


apmantza:
Verdict: merge after fixes.

The detector and headers pass the focused checks, and the header and source-blanking mutations both red. The lane-membership claim is not enforced, and the execution-count evidence is stale.

## Findings

### HIGH-1 — The governance test does not bind gates to the lane population

Probe: I changed `.github/workflows/ci.yml` from `git grep -qi 'win32' -- "$file"` to `git grep -qi 'win32' -- .github/workflows/ci.yml`. The command `PI_LENS_HOME="$PWD/.probe-home" node_modules/.bin/vitest run tests/config/win32-gate-lane.test.ts --configLoader runner` still passed: `Test Files 1 passed (1); Tests 2 passed (2)`. Expected: the test must fail when the enumerator no longer selects each gate-bearing file.

The test at `tests/config/win32-gate-lane.test.ts:79-98` checks workflow fragments with `toContain`, but it never computes the source-derived file set or compares it with the files containing detected gates. A broken per-file filter can therefore omit every Windows gate while the governance test remains green, which defeats the acceptance criterion that the lane include set is verified.

Remedy: A) parse the workflow command and execute its source-derived enumeration against the checkout, then assert every detected gate file is in the resulting set; or B) move the population computation into a shared testable script and assert both the workflow and governance test use that result. Add the mutation above as a regression probe.

### MEDIUM-1 — The quoted execution proof does not match the latest Windows job

Probe: `gh run view 34384920407 --job 102578780270 --log | rg -c 'Run Windows Vitest subset.*✓.*win32'` returned `7`. The PR body reports run `34384820209` and count `6`. The latest completed Windows job also reports `Windows Vitest subset: 75 files`, `7 failed`, `65 passed`, and `3 skipped`; the cited run reports 74 files and 6 matching pass lines. Expected: the PR evidence must identify the latest completed job used for the review and reproduce its count.

The body’s execution proof is a one-time log sample from an older run, while the new test only asserts job-definition text. Logs are useful evidence but are not a durable test assertion, so the claimed number can drift without a red test.

Remedy: update the one-time evidence to the latest completed job and state its limitations, or add a deterministic job-side summary/assertion that counts the gate-bearing population and records the executed population. Do not present the log count as a test assertion unless the workflow itself enforces it.

## Cleared categories

- Exact detector forms: `skipIf(process.platform !== "win32")` and `runIf(process.platform === "win32")` are detected. The accepted single-quote forms also match because the regex accepts both quote types.
- Opposite gate: `skipIf(process.platform === "win32")` is not classified as Windows-only.
- Comment and string laundering: the blanking mutation produced `win32 gate detection: scanned/matched 0`, and the test failed. The detector does not count a quoted or commented call.
- Header proximity: removing the immediately preceding header produced the expected file and line, `tests/clients/read-guard-path-normalization.test.ts:85`.
- Variant scope: `skipIf(isWindows)` and `skipIf(platform() !== "win32")` are not detected. No such production test gates were found in the scanned `tests/` tree, so this is a documented detector boundary, not a finding against the stated exact pattern.

## Verification

- Build: `npm run build` passed.
- Focused suite: 2 files, 5 tests passed in 2.65 seconds.
- Config suite: 45 files, 410 tests passed in 41.21 seconds. This differs from the PR body’s 44 files and 409 tests.
- Preflight: `npm run preflight` passed. Its complete local table reports all 13 gates as pass, including `tests/config`, `lockfile:complete`, and `ci-verdict`.
- PR state: GitHub reports `BLOCKED`; required checks were queued or in progress at review time. I did not treat absent or pending checks as green.

## Could not verify

- I could not verify a Windows run for PR head `5507342c`; the PR’s required checks had not completed.
- The local Linux checkout cannot execute the Windows-only tests. The Windows log evidence above is from completed CI jobs, not a local execution.
- No production blast radius exists in this test-only change. A structural `module_report` map was not available in this reviewer context.

## Named output

- Strong: the Windows job has one dynamic population source and one Vitest runner filter, but the governance test currently proves only that their source text exists. The shared population computation is the natural seam for both workflow execution and test enforcement.

apmantza:
Round 2 (2ff9bb6d): one shared population script (scripts/lib/win32-gate-population.mjs) drives both the Windows job and the sweep, the reviewer's constant-filter mutation reds, and the execution count is a job-side summary from the same script. Prescribed round → merges on green.

Issue #2815, opened 2026-09-09
observability: stamp a turnId on every latency.log, extension.log and degradation-ledger line so one id reconstructs a turn across the three sinks


apmantza:
Today the only field shared by `latency.log`, `extension.log` and the degradation ledger is `pid` (process-level). An investigator reconstructing one turn (the 2026-09-09 pi-free read: which edit paid the aux wait, which agent_end drain formatted which files) has to correlate by timestamp. There is no turn identity in `runtime-turn.ts` / `session-lifecycle.ts`.

Ask:
- A monotonically increasing `turnId` (session id + turn counter) minted at `turn_start`, carried by the existing emit context, and stamped as one field on every line the three sinks already write. Field-only: no new logging call site, no stdout (pi owns the terminal), no per-occurrence records — the bounded rules are untouched because nothing new is emitted.
- Hook-phase lines outside a turn (`session_start`, `session_shutdown`, deferred `agent_end` work) carry the id of the turn that scheduled them, or `turn:0`.
- Test through the real logger: one scripted turn → every line of the three sinks carries the same `turnId`; a second turn increments it; the strict-key baseline pins (`BASELINE_KEYS`-style) are EXTENDED with the key, never loosened.
- The investigator contract gains one line: grep the turn id first.

Evidence: the session read on #2809/#2810 needed per-phase timestamp joins to attribute the 1.5 s aux wait to the tool_result path.

apmantza:
Closed by #2836 (merged): one turnId stamped across latency.log, extension.log and the degradation ledger, minted per session (secondary sessions keep their own), detached writers carry turn:0, explicit queued ids stay authoritative, the reset runs once from the session_start closure before the prehandler row, and turn-context is a registered session-state module. Seven rounds with a state-space table from round 3 on.

PR #2836, opened 2026-09-09
feat(observability): one turnId stamped across latency.log, extension.log and the degradation ledger (refs #2815)

apmantza:
## Summary

Refs #2815. Register `clients/turn-context.ts` as session-state-shaped and wire its counter reset directly into the primary `session_start` closure. The closure performs the only session reset; `RuntimeCoordinator` only creates and adopts session identities.

## Round 5

The lifecycle invariant is one reset per primary session start, with no reset during a turn. The session-start closure decides whether the event is a primary continuation before touching process-shared state.

- Primary start: `handleSessionStart` calls `runtime.resetForSession`, which creates the runtime's random telemetry id. The session-start closure then calls `resetTurnContext(stableSessionId)` once, and `runtime.setSessionLifecycle` adopts the stable id without resetting the counter.
- Secondary start: the concurrent-secondary decision returns before the primary reset block, so it does not reset shared turn state or update the coordinator's `_telemetrySessionId`.
- Resume/fork restore: both are primary session starts. The direct reset runs after `handleSessionStart` and before the lifecycle pin, so the first later turn is `stableSessionId:1`.

This chooses rule (a): `resetTurnContext` is directly reachable from the session-start closure, and coordinator reset calls are removed to avoid a double reset.

## Type of change

- [x] Bug fix

## Area

- [x] area:observability
- [x] area:session

## Tests

The edited `tests/clients/turn-id-observability.test.ts` pins reset behavior through the direct turn-context seam while preserving per-session prefixes, detached `turn:0`, and explicit queued ids. The edited `tests/index-integration.test.ts` pins one direct reset invocation through the real `index.ts` session-start callback. The edited registry pins coverage, closure wiring, and the sweep's zero module-scope-symbol count. `tests/config/` pins the eager static-import allowlist.

Red-first on the merged round-4 head:

```text
clients/turn-context.ts is session-state-shaped but is neither registered nor exempted
turn-context.ts: session-state symbol count is missing
turn-context:perSessionCounters resetTurnContext must be directly called inside index.ts's session_start closure
```

Mutation proof after deleting the direct call from `index.ts`:

```text
turn-context:perSessionCounters: resetTurnContext is not reachable from index.ts's session_start closure
```

## Test assessment

The conformance sweep uniquely pins registry completeness, symbol counts, and direct closure reachability. The turn-id suite uniquely pins the real counter and identity semantics. The integration test uniquely pins the host callback ordering. The dependency-boundary test uniquely pins the static import admission.

## Blast radius

The production blast radius is the gated primary `index.ts` `session_start` callback, `clients/runtime-session.ts`'s `handleSessionStart` caller, `clients/runtime-coordinator.ts`'s identity lifecycle, and `clients/turn-context.ts`'s process-singleton-backed counters. The order is session classification → `handleSessionStart`/`resetForSession` → direct `resetTurnContext` → `setSessionLifecycle` → later `turn_start`/`beginTurnContext`.

## Observability

Turn-bearing rows continue to expose the session reset through their `turnId` field. Stable-session identity fallback remains bounded by the existing `turn-context-identity-fallback` record in the degradation ledger.

## Class sweep

The whole-tree session-state sweep found `turn-context.ts` as the only unregistered member. Its live symbol-count pin is zero because the state is process-singleton-backed rather than a module-scope container. The family remains distributed because each module owns a distinct lifecycle seam.

## Verification

The targeted matrix passed with 51 test files and 601 tests. It included `tests/clients/session-state-conformance.test.ts`, `tests/clients/turn-id-observability.test.ts`, `tests/index-integration.test.ts`, and `tests/config/`. `npm run build` and the mutation build both passed. Round 5's stale patch was not applied directly because it omitted this round-4 branch's current coordinator and test seams; the corrected six-file implementation compiles.

## Round 6

The round-5 decision is implemented on the merged round-4 base. The direct reset remains inside the primary `session_start` closure. Both coordinator reset calls are removed, so the closure is the single reset site. The registry entry uses `sessionStartClosureReset: true`, and `SESSION_STATE_SYMBOL_COUNTS` records the sweep's own count of `0` for `turn-context.ts`.

| gate | mirrored CI job | pass/fail |
| --- | --- | --- |
| build | Lint & type-check | pass |
| lint | Lint & type-check | pass |
| fmt:check | oxfmt format check | pass |
| changelog:check | Unit tests | pass |
| check-changelog-fragments | Changelog fragment (fast-fail) | pass |
| check:lockfile | Lint & type-check | pass |
| lockfile:complete | Lint & type-check | pass |
| tests/config | Unit tests | pass |
| generation-guard | Unit tests | pass |
| flake-shape-ratchet | Unit tests | pass |
| lsp-spawn-heavy-coverage | Unit tests | pass |
| ci-verdict | Unit tests | pass |
| knip | knip (advisory) | pass |

## Test assessment

No existing test was removed. The turn-id test was edited to model the production session-start reset seam explicitly after coordinator lifecycle methods stopped owning that reset.

## Blast radius

No additional production consumers are affected. The static dependency allowlist now admits the direct `index.ts` import, and the registry sweep owns the lifecycle contract.

## Observability

The existing turn-bearing sinks remain the proof of the reset identity. The existing `turn-context-identity-fallback` degradation record remains bounded and preserves the fallback identity.

## Round 7

R6-1 is fixed by moving the single `resetTurnContext(stableSessionId)` call into the gated primary path immediately before `handleSessionStart`. The prehandler row therefore carries `primary-prehandler:0`, including resumed or replaced stable sessions. The secondary early return and coordinator no-reset rule remain unchanged.

| Check | Result |
| --- | --- |
| Red-first prior-counter row | `primary-prehandler:2` |
| Turn-id, integration, session-state suites | 3 files, 186 tests passed |
| `tests/config/` | 49 files, 418 tests passed |
| `npm run fmt:check` | passed, 1795 files |
| Inverse mutation | red with `primary-prehandler:2` |
| `npm run preflight` | pending |
| PR body lint | pending |


apmantza:
Verdict: merge after fixes.

## Findings

### HIGH

#### F1. Concurrent secondary turns inherit the primary session id

Probe:

```text
PI_LENS_HOME=$PWD/.probe-home node --input-type=module -e '...'
primary primary:1
secondary without session bind primary:2
```

Expected: the secondary turn carries its own stable session id, such as
`secondary:1`. Actual: it is stamped `primary:2`.

`index.ts:onTurnStart` calls `runtime.beginTurn()` without the firing context's
session id, while the concurrent-secondary `session_start` branch returns before
`runtime.setSessionLifecycle`. The process-singleton context therefore retains
the primary identity and misattributes every secondary turn's direct sink row.
This defeats the issue's #473 isolation requirement and makes cross-session
investigation unreliable.

Option A: bind the current event session to a turn-scoped context before
`beginTurn`, with a context object selected by activation/session identity.
Option B: if the host guarantees serialized turns, document and enforce that
guarantee at the secondary gate, then add a regression test proving it.

### MEDIUM

#### F2. Session-reset mutation is not detected by the new test

Mutation: remove `resetTurnContext(this._telemetrySessionId)` from
`clients/runtime-coordinator.ts:395`, rebuild, and run
`tests/clients/turn-id-observability.test.ts`.

Result:

```text
Test Files  1 passed (1)
Tests       1 passed (1)
```

Expected: the shape-17 reset assertion reds. Actual: it remains green because
the test starts in a fresh worker and never performs two session resets in one
process. The PR claims the counter resets at session start, but no test pins the
second-session counter back to one. Add a real two-session case, then repeat the
mutation and quote its red.

## Cleared categories

- Real writers: a hermetic built-code probe found `latency.log` 3 rows,
  `extension.log` 1, `review-graph.log` 1, and `read-guard.log` 1. The rows
  carried `probe-session:1` except the second latency row, which carried
  `probe-session:2`. Deferred work carried `probe-session:1` after turn two.
- The PR test and the built targeted suites passed. Its extension-stamp mutation
  red at `tests/clients/turn-id-observability.test.ts:101` with
  `expected undefined to be 'session-2815:1'`.
- The module-instance sweep has one existing reviewed `vitest.config.ts`
  exception and no new turn-context `.ts` import. `turn-context.ts` uses the
  process singleton, so duplicate module evaluation does not diverge ids.
- No `BASELINE_KEYS`-style test changed. The only matching baseline is
  `tests/clients/security-scan-probe-telemetry.test.ts:138`; it concerns probe
  metadata, not these sink rows, and remains strict. No strict sink-row consumer
  was found. Pre-turn writers use the shared `turn:0` value.
- The diff adds no `logExtension` or `recordDegradationOnce` call site and no
  per-occurrence ledger record. It only adds fields and passes queued identity.
- The reset mutation above is the sole unclosed mutation category; it is F2.

## Could not verify

- I did not run the complete repository test suite. The requested `npm run
  preflight` completed with all 13 reported gates passing, and the targeted
  `tests/config` plus touched logger/runtime suites passed 50 files and 537
  tests.
- I did not execute a full host `agent_end` event with two live activations.
  The direct runtime probe and the unconditional `onTurnStart` call establish
  the secondary identity failure; the host-level scenario should be added with
  F1's regression test.

## Named output

- Strong: `turn-context.ts` is correctly process-singleton-backed for module
  duality, but that same process-wide current value cannot represent concurrent
  activation identity. The seam should be redesigned around activation/session
  context before adding more sink stamps.

## Verification table

| Check | Result |
| --- | --- |
| Build | pass |
| Targeted turn-id, runtime, and module-instance suites | 35 tests pass |
| `tests/config` plus touched logger/runtime suites | 50 files, 537 tests pass |
| `npm run preflight` | all 13 reported gates pass |
| Extension-stamp mutation | red, expected undefined at line 101 |
| Session-reset mutation | incorrectly green, F2 |

apmantza:
Verdict: merge after fixes.

## Findings

### MEDIUM

#### F3. Rows outside a host-event scope inherit the process-global latest turn

Probe on the merged head:

```text
primary-probe_turn primary-probe:1
secondary-probe_turn secondary-probe:1
outside secondary-probe:1
```

The probe ran `getTurnId()` after both real `RuntimeCoordinator` instances had
interleaved wrapped turns, but outside either `wrapSessionEventHandler` scope.
The PR body promises “Rows outside a turn use `turn:0`.” Instead,
`clients/turn-context.ts:71-74` falls back from `AsyncLocalStorage` to the
process-global `fallbackSessionId`, and `beginTurnContext` updates that fallback
at `:54`. A background drift batch or deferred task that starts outside a host
event can therefore be stamped with the last session's counter, not the writer's
identity and not `turn:0`.

This is an observability misattribution across concurrent sessions. It does not
reopen F1 for work scheduled inside a wrapped event, because async resources
created by that handler inherit the bound context. Fix by making truly detached
writers pass an explicit session/turn id, or make the no-context fallback
unambiguously `turn:0`; then add a real detached-writer regression case.

## Cleared categories

- F1 fixed. The two-real-coordinator probe passed with exact prefixes:
  `primary-probe_turn primary-probe:1` and
  `secondary-probe_turn secondary-probe:1`.
- F2 fixed. Removing only `resetTurnContext(this._telemetrySessionId)` made the
  reset test fail: `expected 'reset-2815:1'`, received `reset-2815:3`.
- Per-session keying is mutation-protected. Replacing the session map key with a
  single shared key made the interleaving test fail at line 180 and the reset
  test fail with `reset-2815:7` instead of `:1`.
- The `tool_result` path reads `ctx` after `await loadBootstrapClients()` at
  `index.ts:2427-2452`. The wrapper binds the session before invoking the async
  handler at `clients/session-event-guard.ts:182-194`; the existing interleaving
  test also reads the context after an await and passed.
- Resume and fork semantics match the contract: `setSessionLifecycle` resets
  the supplied stable id only, so a resumed same id and a new fork id each start
  at `:1`, while a concurrent different session does not reset the other map.
- Strict-key coverage was not loosened. The diff does not touch the
  `BASELINE_KEYS` assertions in `tests/clients/security-scan-probe-telemetry.test.ts`;
  degradation-ledger and extension-log rows still use their existing writers,
  with only additive `turnId` stamping.
- The AGENTS.md paragraph is accurate for host-event-scoped writes, but its
  unqualified “Sinks therefore read the writer's session id” claim is too broad
  in light of F3's detached fallback.

## Verification

- Build: pass.
- Targeted runtime/logger suites: 7 files, 115 tests passed.
- `tests/config/strictness-ratchet.test.ts` and
  `tests/config/timing-sensitive-coverage.test.ts`: 2 files, 6 tests passed
  under the configured default project. The lsp-heavy coverage file is
  lane-filtered and was not a member of that project.
- `npm run preflight`: exit 0.
- PR head checked: `0d46b46931a1dd492df45d9ce69292f33d9a14bd`; local verify head
  after the requested `origin/master` merge: `dd6051000`.

## Could not verify

- I did not run the complete repository test suite.
- I did not drive a live pi host's actual `agent_end` event with two live
  activations. The real wrapper probe and the real `tool_result` after-await
  path passed, but the host-level dual-activation scenario remains unexecuted.
- The PR body contains no literal table headed “Writers-by-axis”; I checked the
  body’s equivalent summary, operating rule, and blast-radius table instead.

## Named output

- Strong: `turn-context.ts` now solves event-scoped identity with
  `AsyncLocalStorage`, but its fallback remains a second identity mechanism.
  The deletion test says the per-session map is load-bearing; the next design
  question is whether detached work should carry explicit provenance or be
  deliberately anonymous as `turn:0`.

## Disposition table

| Finding | Disposition |
| --- | --- |
| F1 | fixed |
| F2 | fixed |
| F3 | new defect |

apmantza:
Verdict: merge after fixes.

Round 3 fixes F3's behavior, but the exact head is not merge-ready. The
explicit queued-id invariant is not mutation-pinned, and the required gates
are red.

## Findings

### HIGH

#### F4. Required preflight gates are red on the exact head

Probe: `npm run preflight` after merging `origin/master`.

```text
lint FAIL: clients/turn-context.ts(41,8): error TS6133: 'current' is declared but its value is never read.
fmt:check FAIL
tests/config FAIL: strictness-ratchet.test.ts
AssertionError: expected [ 'clients: regression (468 -> 470)' ] to deeply equal []
```

The PR cannot merge with the lint and format gates red. Remove the unused
local, format the changed files, and resolve or explicitly rebase the
strictness baseline regression before rerunning preflight.

### MEDIUM

#### F5. Explicit queued identity is not mutation-protected

Probe: mutate `clients/latency-logger.ts:819` from
`entry.turnId ?? getTurnId()` to `getTurnId()`, rebuild, and run
`tests/clients/turn-id-observability.test.ts`.

```text
Test Files  1 passed (1)
Tests       4 passed (4)
```

The test suite stays green when the sink overwrites an explicit queued or
completed `turnId`, although the Round 3 invariant says that identity is
authoritative. The existing interleaving test supplies explicit ids, but its
active context supplies the same prefixes, so it cannot detect this mutation.
Add a context-free latency row with `turnId: "owner-session:7"` and assert
that exact value reaches `latency.log`; rerun the mutation and quote its red.

### LOW

#### F6. The PR body fails runtime observability lint

Probe: `node scripts/check-pr-body.mjs --lint-local <(gh pr view 2836 --json body --jq .body)`.

```text
PR body Observability must name a record literal from the runtime diff;
"No new failure path; no record added." is not valid when the added lines
contain a failure path.
```

The diff adds the `stableSessionId` catch in `clients/session-event-guard.ts`,
which is a new failure path that degrades to an unbound context. The diff adds
no bounded record literal for that catch, so the body cannot honestly name one.
Either add and name a bounded identity-resolution record, or remove the
failure-path claim from the diff/body contract if this catch is intentionally
not observable. `turnId: getTurnId()` is not a record discriminator accepted
by the checker.

## Cleared categories

- F3 is fixed. The direct built probe now prints:

  ```text
  primary-probe_turn primary-probe:1
  secondary-probe_turn secondary-probe:1
  outside turn:0
  ```

  The prior F3 probe's failing output was:

  ```text
  primary-probe_turn primary-probe:1
  secondary-probe_turn secondary-probe:1
  outside secondary-probe:1
  ```

- Restoring the process-global fallback is detected. The Round 3 test reds at
  `tests/clients/turn-id-observability.test.ts:243` with `Expected:
  "turn:0"` and `Received: "secondary-probe:1"`.
- `clients/turn-context.ts` contains no `fallbackSessionId`, latest-session
  pointer, or equivalent process-global fallback. It stores a per-session map
  and uses `AsyncLocalStorage`; the context-free branch is exactly
  `return "turn:0"`.
- Host-event handlers bind the stable event session in
  `clients/session-event-guard.ts:182-185`; async work inherits it.
- Deferred-format and requeue latency rows in `clients/runtime-agent-end.ts`
  pass `queuedTurnId` explicitly. The generic drift and reaper/timer sinks have
  no owner id at their call sites and therefore land on `turn:0` when detached.
  The latency flush only drains its queue and does not select identity.
- A direct built probe confirmed an explicit context-free latency row preserves
  `owner-session:7`. F5 shows the regression test does not protect that fact.
- The per-turn row, degradation-ledger row, extension log, review graph log,
  and read-guard log remain stamped. The strict security metadata baseline in
  `tests/clients/security-scan-probe-telemetry.test.ts:138` is untouched.
- The AGENTS.md turn-identifier contract matches the implementation: sinks use
  bound context, explicit queued ids, or `turn:0`, never a process-global latest
  session.

## Verification

| Check | Result |
| --- | --- |
| `origin/master` merge | pass, clean merge |
| Build | pass |
| Round 3 targeted test | pass, 4 tests |
| Built F3 probe | pass, outside `turn:0` |
| Process-global fallback mutation | red, F3 assertion |
| Explicit-id overwrite mutation | incorrectly green, 4 tests pass |
| `tests/config/` | fail, strictness ratchet 468 -> 470 |
| `npm run preflight` | fail, lint, fmt, and tests/config |
| PR body lint | fail, observability record requirement |

## Could not verify

- I did not run the complete repository suite. `npm run preflight` already
  failed its required gates, so a full run would not establish merge readiness.
- I could not prove that every asynchronous drift producer has owner provenance;
  several detached drift paths call `logLatency` without an explicit id. They
  correctly land on `turn:0` when no context exists, but the source does not
  expose an owner contract for those callbacks.

## Named output

- Strong: `getTurnId()` is now a single anonymous fallback seam, while
  `logLatency()` is the only sink that preserves caller-owned identity. The
  deletion test says the fallback seam is correctly centralized, but the
  explicit-id test seam is missing.

## Previous-round disposition

| Finding | Disposition |
| --- | --- |
| F1 | fixed |
| F2 | fixed |
| F3 | fixed |

apmantza:
Verdict: merge after fixes.

## Findings

### MEDIUM

#### R6-1. Session-start telemetry is emitted before the claimed reset

Evidence: `index.ts:2255-2286` awaits `handleSessionStart` and only then calls
`resetTurnContext(stableSessionId)`. The real handler emits
`session_start_prehandler` at `clients/runtime-session.ts:1927-1937` through
`logLatency`. The session-event wrapper binds the stable session before this
await at `clients/session-event-guard.ts:190-192`.

Expected: every row after a primary session-start boundary uses the reset
counter. Actual: a resumed or replaced stable session can emit the prehandler
row with its prior `sessionId:turnIndex` before the reset. A new stable id can
emit `stable-id:0` before the reset. This is a bounded observability
misattribution, not a turn counter mutation during `turn_start`.

Remedy: move the single `resetTurnContext(stableSessionId)` into the gated
primary path before `handleSessionStart`, or explicitly classify the
prehandler row as pre-reset and add a test that proves that contract. Preserve
the existing secondary early return and the coordinator's no-reset rule.

## Cleared categories

- F1: fixed. `turn-id-observability.test.ts` passed the two-real-coordinator
  interleaving case; the independent probe printed `p1 p:1`, `s1 s:1`,
  `p2 p:2`, and `detached turn:0`.
- F2: fixed. The required suites passed with 52 files and 603 tests. Removing
  the closure call produced 2 failures and 185 tests: the registry reachability
  pin and the integration call-count pin.
- F3: fixed. The detached-writer test passed and the independent probe returned
  `turn:0` outside both AsyncLocalStorage scopes.
- F4: fixed. Build, lint, format, config, generation, flake, LSP coverage, and
  CI-verdict gates passed in clean preflight.
- F5: fixed. The explicit queued-id test passed; replacing
  `entry.turnId ?? getTurnId()` with `getTurnId()` produced 1 failure and 4
  passing tests, with expected `owner-session:7` and received `turn:0`.
- F6: fixed. Body lint exited 0. The existing bounded identity-fallback record
  is named in the body.

The inverse mutation was also attempted. Adding `resetTurnContext` to
`RuntimeCoordinator.setSessionLifecycle` left 2 files and 56 tests green, so
no test pins that the coordinator cannot perform a duplicate reset. The
integration mock pins one direct closure invocation, but it does not detect a
second real coordinator reset.

Strict-key review: the turn-id suite passed its per-turn, degradation-ledger,
and extension-log assertions. The security scan exact-key test is not in the
default Vitest population or the wall-clock lane, so it was not a valid direct
target in this configuration; its source remains untouched.

## Verification

| Check | Result |
| --- | --- |
| Required build and targeted command | 52 files, 603 tests passed |
| Additional runtime/index tests | 3 files, 73 tests passed |
| F1/F3/F5 probes | passed |
| Closure-removal mutation | 2 failed, 185 tests total |
| Explicit-id overwrite mutation | 1 failed, 4 passed |
| Coordinator duplicate-reset mutation | 2 files, 56 tests passed |
| `npm run fmt:check` | passed, 1795 files |
| PR body lint | exited 0 |
| `npm run preflight` | required gates passed; advisory knip failed after purging 476 compiled siblings |

## Could not verify

- I did not run the complete repository test suite.
- I did not drive a live pi host with two concurrent activations. The real
  wrapper and two-coordinator probes cover the session identity seam, but not
  the host's event scheduler.
- The repository has no direct test that observes the production
  `session_start_prehandler` row across a same-id resume boundary.
- The strict-key security scan test could not be selected by the configured
  Vitest projects, so its exact-key assertions were not rerun.

## Named output

- Strong: the reset's placement is load-bearing, but the current placement is
  after an awaited handler that already writes latency and degradation rows.
  Moving the reset earlier should simplify the lifecycle rule and make the
  “before a new session can publish rows” comment in `turn-context.ts` true.
- Worth exploring: add a single-reset spy around the full primary session-start
  path. The current test proves direct closure wiring, not absence of another
  writer.

Issue #2838, opened 2026-09-09
Context injection budget by model tier: record injected bytes per source per turn, compact digest for non-frontier models (gated), measured through dogfood runs


apmantza:
Maintainer observation (2026-09-09): non-frontier models hallucinate more with pi-lens's inline context than without. Plausible causes, in order of suspicion: quoted tool/linter output injected verbatim (treated as instructions), the same finding re-injected every turn until fixed, imperative framing on the advisory tail, and standing description cost (already reduced by #2808). None of it is measurable today: the injection paths record bytes only for cache attribution (`clients/cache-observability.ts`), not per source.

Slices, in order:
1. **Observability first (no behaviour change).** One field on the existing per-turn latency row: `injectedBytes: { sessionGuidance, turnFindings, testFindings, agentNudge, turnEndAdvisory }` plus `injectedFindingsRepeated` (findings re-injected from a previous turn). No new record, no per-occurrence lines. The monitor readout (`docs/pi-lens-monitor.md`) gains the per-turn injected-bytes distribution.
2. **Model-tier digest (gated).** `context.digest: "auto" | "full" | "compact"` (default `auto`: compact when the session model id is not on a frontier allowlist; the allowlist lives in one place and is documented). Compact = counts per source, the top three findings with file:line and one-line message, no quoted bodies, deltas only after the turn a finding first appears, advisory tail in declarative wording; blockers keep their imperative form. Full = today's text. Byte-identical output when `full`. Marked experimental in `docs/configuration.md` until slice 3 reports.
3. **Measurement.** Through #2826 (autonomous dogfood): the same task set on one small model with `full` vs `compact`, comparing turn count, tool-call error rate, repeated-finding count and the monitor readout; the default for non-frontier models is decided from those numbers and recorded here.

Catalog: shapes 10, 13, 17, 18 screened; the digest is language-neutral (findings from every runner, not TypeScript diagnostics only).

apmantza:
Slice 1 merged in #2841: per-turn `injectedBytes` by source and `injectedFindingsRepeated` (bounded identity set, reset on session_start after review round 2), monitor readout line. Remaining slices: 2 (compact digest for repeated findings) and 3 (budget by model tier), plus the per-call tool-output figures now tracked as #2800 item 7.

Issue #2839, opened 2026-09-09
ci-failure-classifier: a test timeout in a job that also logged registry ECONNRESET/ETIMEDOUT is infra-net, not real


apmantza:
Record (2026-09-09, PR #2834): the Unit tests job logged three `infra: registry unreachable — npm … ECONNRESET` lines and one `Test timed out in 5000ms` on a whole-tree sweep test under that load; the classifier labelled the PR `ci:real` because a timeout is a real-failure shape, and no rerun was armed. When a job carries network-unreachable evidence AND the only test-level failure is a timeout, the classification should be `infra-net` (rerun once); an assertion failure still wins as real. Fixture: the #2834 log excerpt. Refs #2103.

apmantza:
Second instance within the hour: PR #2835's Unit tests job logged `npm error code ECONNRESET` through all npm-retry attempts and was labelled `ci:real`; re-ran by hand. Two instances → this is the catalogued recurrence for the fix.

apmantza:
Closed by #2844 (merged): a test or hook timeout beside registry network evidence classifies infra (rerun once) — decided per FAIL block on evidence lines only, prose and echoed needles cannot arm it, kill evidence outranks it, Windows annotation prefixes and CRLF are normalized once; every demotion predicate has a distinguishing fixture, mutations re-run by the orchestrator before merge.

PR #2841, opened 2026-09-09
feat(observability): injected bytes per source and repeated-finding count on the per-turn row (refs #2838)

apmantza:
## Summary

Reset repeated-finding identities and the identity-cap once latch at `session_start` for the owned session role. Preserve injected-byte counters, summaries, and concurrent-secondary isolation.

## Tests

- `tests/clients/cache-observability.test.ts`: Added a regression test for the reused-session-id probe. It asserts turn 1 reports `0`, the post-boundary turn reports `0`, and byte attribution remains unchanged.
- `tests/index-integration.test.ts`: Added an index wiring test that observes the real `session_start` handler and requires the primary owned-role reset.
- `tests/clients/session-state-conformance.test.ts`: Re-ran the session-state lifecycle sweep after the new reset seam was wired.

Red-first proof:

```text
AssertionError: expected 1 to be +0
tests/clients/cache-observability.test.ts:1675
```

Mutation proof, with the new `index.ts` reset call removed:

```text
AssertionError: expected "vi.fn()" to be called with arguments: [ 'reset-cache', 'primary' ]
Number of calls: 0
tests/index-integration.test.ts:2767
```

## Blast radius

`index.ts` `session_start` → `resetCacheFindingIdentitiesSession` → `attributionBySession[key].findingIdentities` and `findingIdentityCapRecorded`. The reset runs after the existing concurrent-secondary decision and before `handleSessionStart`. It does not delete the attribution state or touch byte counters, summaries, or the per-turn repeated count.

## Class sweep

Searched all `session_start` reset seams, `findingIdentities` references, and cache-observability lifecycle callers. The only production writer is `observeCacheContext`; shutdown still uses `clearCachePrefixSession`, while `session_start` now uses the role-scoped identity reset. No sibling identity set or cap latch exists.

## Observability

No new failure path; no record added. The existing `cache_usage` row remains the independent observable for repeated findings and injected bytes. The existing `injected_finding_identity_cap` record remains once-latched per session identity state.

## Test assessment

The new cache test drives the real cache-observability state and observes the existing latency sink mock for the row fields. The index test uses the established host boundary mock only to verify lifecycle wiring and preserves the real module exports with `importActual`.

## Verification

- `npm run build` — pass.
- `PI_LENS_HOME=$PWD/.probe-home-round2 node_modules/.bin/vitest run tests/clients/cache-observability.test.ts tests/index-integration.test.ts tests/clients/session-state-conformance.test.ts --configLoader runner` — 3 files, 255 tests passed.
- Mutation run with the reset call removed — failed at the lifecycle assertion as quoted above.

## Round 2

Fixed the prescribed MEDIUM finding from PR #2841. Repeated-finding identities and the cap latch reset on `session_start` through the existing owned-role closure seam. Byte counters remain untouched.


apmantza:
CI: Unit tests on c2afc89b died in a registry ECONNRESET storm (`npm ci` failed 3 times, one ratchet test timed out at 5 s while the runner had no network) — same class as #2839; failed jobs re-run by hand, no code change.

apmantza:
Verdict: merge after fixes.

## Findings

### MEDIUM — finding identities do not reset on `session_start`

Probe:

```text
session "reset", turn 1: injectedFindingsRepeated=0
session "reset", turn 2 after the session_start boundary: injectedFindingsRepeated=1
```

The new `findingIdentities` set is created in `clients/cache-observability.ts:325-331` and only removed by `clearCachePrefixSession`, which `index.ts:3298` and `index.ts:3345` call on `session_shutdown`. The `session_start` handler at `index.ts:1838-1842` has no corresponding reset. Reusing a stable session id across reload, fork, or replacement therefore carries identities across the required session boundary and reports false repeats. Add a session-start reset for the owned role, preserving the concurrent-secondary rule, and add a lifecycle test that is red when that reset is skipped.

The mutation is proven by the probe above: skipping the required reset leaves the second row at `1`, while the contract requires `0`.

## Cleared categories

- Behavior neutrality: the changed `index.ts` path only calls `recordTurnEndAdvisoryBytes` after the existing successful `sendMessage`; the context message construction is unchanged. The scripted host-text SHA-256 is `bc4575a75f44a04557620b6bc2a3020305485cf49ff5d822453c24c475f2801e` on the reviewed tree. The diff leaves the pre-change return path unchanged, so the before value is the same hash.
- UTF-8 bytes: `é` records `turnEndAdvisory: 2`, not 1. The non-ASCII finding `src/例.py:7 rule=E501` records its UTF-8 byte length.
- Repeated identity: the same `file:line:rule` in turn 2 counts 1. Moving the finding from line 3 to line 4 counts as new; that is intended by the issue's explicit `file:line:rule` identity.
- Cap: 2,001 YAML identities produced exactly one record:

  ```text
  phase=injected_finding_identity_cap metadata={"cap":2000}
  ```

- Disabled context: `no-lens-context` produces all six fixed keys with zero values. The strict exact-key consumer is `tests/clients/cache-observability.test.ts:53-99`; the targeted suite passes.
- Logging: the diff adds no logging call in `index.ts`. The only new `logLatency` call is the once-only cap record at `clients/cache-observability.ts:1078-1084`.
- Language neutrality: Python and YAML findings both participate in the same identity parser. A repeated Python finding counted 1, while the moved-line YAML finding counted 0.
- Monitor contract: `docs/pi-lens-monitor.md` reports per-source p50/p95 bytes and the repeated-findings ratio.

## Verification

- `npm run build` — pass.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/cache-observability.test.ts tests/index-integration.test.ts --configLoader runner` — 2 files, 125 tests passed.
- `PI_LENS_HOME=$PWD/.probe-home-config node_modules/.bin/vitest run tests/config --configLoader runner` — 45 files, 411 tests passed.
- `PI_LENS_HOME=$PWD/.probe-home-preflight npm run preflight` — all reported gates passed.
- `git diff --check` — pass.

## Could not verify

- A second isolated pre-change checkout was blocked by the repository's one-worktree guard. The before/after host-text conclusion is therefore based on the exact diff: no context return construction changed, and the current scripted host-text hash is recorded above.

## Named output

- Strong: `findingIdentities` is a shared language-neutral parser, and the issue's line-sensitive identity means line movement intentionally appears as a new finding. The follow-up should preserve that contract while adding the missing lifecycle reset.

PR #2842, opened 2026-09-09
fix(scripts): entry-module gates compare against pathToFileURL so the Windows lane enumerates its files (refs #2536)

apmantza:
## Summary

Master run 34400281631 (head 711483dd4): `Unit tests Windows (advisory)` errored with `enumeration empty` because `node scripts/lib/win32-gate-population.mjs --files` printed nothing on `windows-latest`. The script gated its main block on `import.meta.url === \`file://${process.argv[1]}\``; on Windows `process.argv[1]` is `D:\a\...\script.mjs` and `import.meta.url` is `file:///D:/a/...`, so the gate never fires. Six scripts carried the shape (sweep recorded below).

- Every gate compares against `pathToFileURL(process.argv[1]).href` (the idiom `scripts/lint-js-advisory.mjs` already used).
- `win32-gate-population.mjs` excludes fixtures by `join(...) + sep` instead of a hand-appended `/`.
- `tests/config/script-entry-portability.test.ts` rejects the hand-built comparison across `scripts/` (comment-blanked, template literals kept because the needle's evidence is a string; census floor 60 of 120 files).
- AGENTS.md: catalog shape 44; orchestrator rule that a PR changing what an advisory lane runs is gated on that lane's row (#2833 merged with the row red on its own head).

## Tests

Red first, scripts reverted to master, test kept:

```
× no script compares import.meta.url with a hand-built file:// string
+   "scripts/complexity-report.mjs",
+   "scripts/lib/lockfile-completeness.mjs",
+   "scripts/lib/win32-gate-population.mjs",
+   "scripts/strictness-report.mjs",
+   "scripts/strip-dev-deps-for-pack.mjs",
+   "scripts/upsert-tracking-issue.mjs",
      Tests  1 failed | 1 passed (2)
```

Green with the fix: that file plus `win32-gate-lane`, `sweep-floor-coverage`, `lockfile-completeness`, `complexity-report`, `upsert-tracking-issue` tests. `npm run preflight` on the exact tree: 13 rows pass, exit 0. Linux enumeration unchanged (57 files, 0 fixtures). The Windows proof is this PR's own `Unit tests Windows (advisory)` row, which must be green before merge.

## Sweep

`grep -rnE 'file://\$\{process\.argv\[1\]\}' scripts tests tools clients`: six gates (all fixed); the remaining `file://${…}` hits are LSP URIs and a clone URL, not entry gates. Consolidation verdict: stay distributed (one expression per script; a shared helper would relocate one line).

## Blast radius

Six scripts' entry condition only; exported functions untouched. `strip-dev-deps-for-pack.mjs` keeps its `endsWith` fallback.

## Observability

## Detection retrospective

Caught by: the Windows advisory lane on master. Should have been caught by: the same row on #2833's head, which was red and ignored because the verdict skips advisory rows. Cost of the layer: reading one row on PRs that change what the lane runs (now an AGENTS.md rule).


apmantza:
Verdict: merge after fixes.

The six Windows entry gates are correct for normal CLI execution. The exact
head has green required checks, a green `Unit tests Windows (advisory)` row,
and the targeted build, lint, and portability sweep pass. Two defects remain
in the new import safety and governance coverage.

## Findings

### MEDIUM

#### F1 — The governance regex has two evasion boundaries

Probe:

```text
canonical: true
notEqual: true
rightHandImport: false
operatorSplit: true
memberSplit: false
templateSplit: true
uriOnly: false
```

The pattern at `tests/config/script-entry-portability.test.ts:26` catches the
canonical gate, `!==`, and whitespace or line breaks around the operator and
template expression. It misses a reversed comparison such as `` `file://${process.argv[1]}` === import.meta.url `` and a member split such as
`import.meta.\nurl === \`file://${process.argv[1]}\``.

The sweep therefore does not enforce its stated whole-tree prohibition: either
miss can reintroduce the Windows-empty-enumeration defect while the test stays
green. Extend the scanner to recognize both operand orders and use an AST or a
small normalized lexical form for member access. Add compile-valid fixtures for
both misses and mutate each new branch.

#### F2 — Importing any of the six scripts without `argv[1]` now throws

Probe:

```text
node -e 'import("./scripts/lib/lockfile-completeness.mjs")'
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```

The same probe exits 1 for all six changed scripts. Supplying one dummy argv
value makes all six imports exit 0, which is why the Vitest imports pass. The
new unguarded `pathToFileURL(process.argv[1])` call executes during module
evaluation before the entry comparison. The old hand-built template did not
throw for an absent argument.

This breaks a valid import/test invocation that has no script argument and
turns a safe entry guard into an import-time exception. Guard the conversion
with an argv presence check before calling `pathToFileURL`; preserve the
optional fallback in `strip-dev-deps-for-pack.mjs`.

## Cleared categories

- LSP URI false positives: no exact `file://${process.argv[1]}` hit exists in
  `tests/clients/lsp`; the sampled hits are ordinary URI fixtures such as
  `tests/clients/lsp/watch-queue.test.ts:20` and
  `tests/clients/lsp/refresh-and-sync-kind.test.ts:48`.
- Linux behavior parity: old and new `win32-gate-population.mjs --files`
  outputs are byte-for-byte identical, 57 paths each.
- `strip-dev-deps-for-pack.mjs:79-80`: the `endsWith` fallback is not dead;
  importing with `process.argv[1] = "strip-dev-deps-for-pack.mjs"` reaches it
  and exits 64 with the usage message.
- AGENTS.md advisory wording: its manual-review rule does not contradict
  `scripts/ci-verdict.mjs:266-268`. The former requires reading the affected
  advisory row; the latter excludes advisory rows from the automated verdict.

## Could not verify

- I could not run the Windows job locally. I relied on the exact-head green
  `Unit tests Windows (advisory)` check and the local Linux probes.
- The advisory mutation job is red. Its log shows the initial related-test run
  failed on a strictness-ratchet assertion and timed out in the Windows gate
  test, so it does not provide clean mutation evidence for this change.

## Named output

- Strong: the distributed one-line gate remains the smallest structure. A
  shared helper would relocate, rather than remove, one entry check per script.

apmantza:
Round 2 (both findings prescribed, applied by the orchestrator):

**F2** — every gate is now `process.argv[1] && import.meta.url === pathToFileURL(process.argv[1]).href`. Red-first (this head before the change), the reviewer's probe:
```
node -e 'import("./scripts/lib/lockfile-completeness.mjs")'
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
```
After: all six imports exit 0 with no argv.

**F1** — the sweep regex matches both operand orders and whitespace/line breaks inside `import.meta.url`. Red-first with the previous regex:
```
old regex reversed false
old regex memberSplit false
```
Both spellings are fixtures now (plus `!==`, a fully split multi-line form, and three clean sources); `tests/config/script-entry-portability.test.ts` green.

Merged origin/master (now carries #2843, which untracks `.probe-home`). Merge on green with the Windows advisory row read.

Issue #2800, opened 2026-09-09
Umbrella: tool-roster economics — description budget with a governance pin, fold lsp_diagnostics into lens_diagnostics, situational gate + per-tool enabled


apmantza:
Every registered tool's description and schema is paid on every turn of every session, whether or not the tool is used, and adjacent tools compete for the same intent (#2792 was the cost of `lsp_diagnostics` next to `lens_diagnostics`).

## Measured today (description text only, before parameter schemas)

| tool | bytes |
|---|---:|
| module_report / read_symbol / read_enclosing (one file, three tools) | 5,630 |
| lsp_navigation (situational) | 2,135 |
| lens_diagnostics | 1,931 |
| ast_grep_search (situational) | 1,483 |
| project_report | 1,323 |
| lens_diagnostic_mark | 981 |
| ast_grep_outline / symbol_search / ast_grep_replace / effective_config / pi_lens_activate_tools / ast_grep_dump | 253–852 each |
| lsp_diagnostics | 243 |

≈17 KB of description ≈ 4k tokens per turn plus schemas; 13 tools always on, the situational set behind `pi_lens_activate_tools`. The MCP mirror reuses descriptions by reference since #2795, so every change below lands on both surfaces.

## Items, in order

1. **Description budget + governance pin.** Cut every description to its contract sentence plus one example; operational detail moves into the RESULT text (paid only on use — with #2799's bounded truncation). A `tests/config/` pin on total roster bytes per surface (pi and MCP) so the budget cannot regrow silently; baseline = the post-trim number. Target: ≤ 8 KB total description.
2. **Fold `lsp_diagnostics` into `lens_diagnostics`.** One diagnostics tool with a `source` modifier (`session` cache | `lsp` probe | `analyzers`) and `scope` (`delta` | `paths` | `workspace`), one result contract; `lsp_diagnostics` retired with a compatibility shim for one release (the host name mapping in the MCP mirror follows). #2792's guidance sentences become the single description. Design-first: the state-space table over (source × scope × cache state) before edits; #2418 stability policy governs the deprecation.
3. **Do not fold** search, navigation, structural search/replace, and reports: different intents and result shapes; a polymorphic `analyze` tool would lengthen the description past the sum of the parts and hide affordances the model finds by name (a known `Read` collapse worked because those tools shared one intent).
4. **Situational gate stays the mechanism** for everything else; land #2774 (`tools.<name>.enabled`) so users can drop tools they never want.
[...]

Issue #2799, opened 2026-09-09
tool results: middle-out truncation with an omitted-count and a log-file path; annotate benign nonzero exits; cursor reads for paginated reports


apmantza:
1. **Truncation tells the model what it lost.** "tells the model exactly how many characters were omitted, and writes the full output to a log file whose path is in the result." Audit `tools/render-compact.ts` and every tool that clips output (module_report, project_report, ast_grep_search, lens_diagnostics mode=full): truncate middle-out (head + tail), state `N characters omitted`, and name the file under the session data dir that holds the full text. A hard clip with no count is the "retry the same call" trap.
2. **Benign nonzero is annotated.** "an unannotated grep 1 is a retry loop." Any tool whose child exits nonzero for a benign reason (no matches, no diagnostics, tool unavailable) says so in the result text — most already do via the typed `unavailable` outcome; sweep for the rest.
3. **Cursor reads, never the same bytes twice.** Paginated results (symbol search, module/project reports) carry a `from_offset`/cursor so a follow-up call returns only new bytes.

Ties to the tool-economics thread: result-side text is paid only on use, so moving operational detail out of descriptions (paid every turn) into results is net positive once results are bounded this way.

apmantza:
Status comment — this issue has had **no comment at all** since filing, while PR #2848 merged work against it as `refs`. Recording the disposition of each of the three asks so the remainder is visible rather than implied.

**Ask 1 — middle-out truncation, `N characters omitted`, name the file under the session data dir.** *Seam delivered; the per-tool audit is not.* `boundToolText` (`tools/render-compact.ts:97`) does head+tail with `[incomplete: N bytes omitted, budget …]` and `[Full output: <path>]`, and it is wired into both host-adapter delivery paths through `boundResultPayload` (`:284`) and `finalizeToolResultWithDelivery` (`:378`). What the issue asked for beyond the seam was an **audit across `module_report`, `project_report`, `ast_grep_search` and `lens_diagnostics mode=full`** — confirming each of those actually routes through it. PR #2848's own body limits its claim to the seam, so that audit is the open half.

[...]

PR #2795, opened 2026-09-09
fix: clarify diagnostic verification (closes #2792)

Don-Yin:
## Summary

Closes #2792. `mode=all` reports cached findings; it does not actively verify changed files. Agent guidance now directs absent or stale findings to `mode=full` with `paths` and asks callers to check coverage. Omitting `paths` still means a whole-project scan.

The MCP entry reuses the diagnostic tool description, translating the LSP tool reference to its registered MCP name. No scan mode, cache policy, runtime decision, dependency, or configuration is added. All acceptance criteria are covered. The existing `paths` contract in AGENTS.md already describes this behavior correctly.

## Type of change

- [x] Bug fix
- [x] Documentation

## Area

- [x] area:diagnostics
- [x] area:session
- [x] area:tests

## Tests

- `npm run build`, `npm run lint`, `npm run fmt:check`, and `npm run build:dist` pass.
- `npm run test:targeted -- tests/tools/lens-diagnostics.test.ts tests/clients/runtime-turn-session.test.ts tests/tools/lens-diagnostics-mode-all-freshness.test.ts tests/mcp/server.smoke.test.ts`: 184 tests pass on the updated master base.
- `npm run changelog:check` and `git diff --check` pass.
- The new tool-metadata test failed on pre-fix code: `expected ... to match /(?:mode=)?full[^.\n;]*paths/`.
- The extended session-orientation test failed on pre-fix code: `expected ... to match /mode=all[^.\n]*cache-only/`.
- Mutating the MCP description back to independent, whole-project-only advice makes the real stdio `tools/list` test fail its cache-only assertion. Reusing the Pi description without translating the tool name fails `toContain("unlike pilens_lsp_diagnostics")`.

Tests inspect returned metadata, injected context, and a real MCP transport, not source text. Existing execution tests continue to cover cache-only queries, empty scoped caches, and targeted active scans. No new runtime guard or branch needs separate mutation coverage. The full suite remains CI's job.

### Test assessment

- `tests/tools/lens-diagnostics.test.ts`: adds one case for the tool description, completion hint, and parameter guidance, which callers consume independently.
- `tests/clients/runtime-turn-session.test.ts`: extends the existing orientation test to guard cached/active wording while retaining its size budget.
- `tests/mcp/server.smoke.test.ts`: extends the existing tool-list assertion to catch MCP description drift and invalid cross-tool names over real stdio.

No tests are made redundant or removed. Existing fixtures and transport are reused; no new mocks, timers, or processes are added.

## Blast radius

[...]

PR #2848, opened 2026-09-09
feat(tools): roster-bytes pin on both surfaces; descriptions cut to contract sentence plus one example (refs #2800)

apmantza:
## Summary

Round 3 makes the MCP truncation seam own the AST replacement preview cap.
The MCP tool renders all replacements once, then `boundToolText` writes that
complete payload to the session log and bounds the model-facing response.
Pi callers retain their historical 50-match content cap.

## Tests

- `tests/mcp/server.smoke.test.ts`: verifies the bounded MCP response and the
  complete `value319` replacement in the persisted session log.
- `tests/config/tool-roster-budget.test.ts`: re-runs the duplicate-name roster
  pin from review F2.
- `tests/config/`: verifies the merged master tree and configuration contracts.

Red-first proof on the pre-fix tree:

```text
AssertionError: expected '[DRY-RUN] 320 replacement(s): … Found 320 matches (showing first 50)' to contain 'value319'
```

Mutation proof after removing the MCP complete-render request:

```text
FAIL tests/mcp/server.smoke.test.ts > pi-lens MCP result bounds > bounds a large AST replacement and keeps the full result in the session log
AssertionError: expected '[DRY-RUN] 320 replacement(s): …' to contain 'value319'
```

## Blast radius

Call-tree diff:

```text
callTool
  astGrepReplaceTool.execute
-   formatMatches(matches, dryRun, showModeIndicator)
+   formatMatches(matches, dryRun, showModeIndicator, resultMaxItems)
  boundToolText
```

Only the MCP AST replacement call passes `Number.POSITIVE_INFINITY`; pi callers
still use the default 50-item formatter limit. `boundToolText` is also changed
to use binary search when shrinking large UTF-8 results, avoiding quadratic
work on complete MCP payloads.

## Class sweep

The formatter cap population includes AST replace and AST search. AST replace
is the only changed MCP result path; AST search keeps its explicit paging
contract. The duplicate roster-name mutation remains fixed and the roster pin
was re-run.

## Observability

(The existing `boundToolText` session-log file is the durable record: it stores the complete MCP result before the model-facing bound.)

## Test assessment

The existing smoke test uses the real stdio MCP server, real AST replacement
tool, real session-log writer, and a real temporary workspace. The mutation
proves the assertion is not tautological.

## Round 3

Option (a) is implemented. This does not change pi's caller-visible result
contract: only the MCP invocation requests the complete formatter output.
The session-log contract now receives the tool's complete result before
truncation.

## Round 4

Implemented the maintainer-prescribed hard input budget. The dated constant is
`COMPLETE_MCP_RESULT_INPUT_BUDGET_BYTES` in `tools/render-compact.ts` and is
documented in `docs/settings.md`. Results above it are explicitly incomplete;
results below it are unchanged. The existing fragment in
`.changelog/tool-roster-budget-2800.md` was extended; no new fragment was added.

## Round 5 (merge with master)

Resolved the one conflict hunk in `tests/config/tool-roster-budget.baseline.json` by retaining the branch's per-tool budget schema, removing `ast_grep_dump`, and regenerating the merged-tree values. No other conflict hunks occurred. Preserved the master ast-grep fold, registry removal, and out-of-registry MCP compatibility shim together with this branch's trims, roster pin, truncation seam, 8 MiB budget, and once-only record. No new changelog fragment was added.


apmantza:
Verdict: merge after fixes.

## Critical

None.

## High

None.

## Medium

### F1 — Result guidance is not bounded by #2799

Probe:

```text
tools/call pilens_ast_grep_replace {pattern: "const $X", rewrite: "let $X", apply: false}
=> 177 replacement(s), output length 44,745 bytes
=> "Found 177 matches (showing first 50):"
```

`mcp/server.ts:490-499` uses `toolText` to append the complete JSON payload, and
`mcp/server.ts:323-344` returns the formatted replacement output without a
middle-out character bound, omitted-count line, or session log path.

The changed description moves preview guidance into this result, but the real
MCP handler can still emit tens of thousands of bytes. This does not satisfy
#2800 item 1's explicit #2799 condition: the model must learn what was omitted,
and the complete output must be available through a bounded session log.

Remedy: apply the #2799 result truncation seam to this MCP result, preserving
the existing `truncated` detail and adding the omitted count and log path.

### F2 — Duplicate MCP names disappear from the measurement

Mutation probe: add a second `pilens_health` entry to `ALL_TOOLS` in
`mcp/server.ts`. The roster test still passes:

```text
Test Files  1 passed (1)
Tests       2 passed (2)
```

`tests/config/tool-roster-budget.test.ts:58-68` builds `measured` with
`Object.fromEntries(tools.map(...))`. A duplicate name overwrites the earlier
row, and both totals are then reduced from the overwritten object. The real
`tools/list` response contains the duplicate, but the pin silently measures
only one row.

Remedy: reject duplicate names before converting the list to a keyed baseline,
or measure totals from the original array and assert unique names explicitly.

## Low

None.

## Nitpick

None.

## Cleared categories

- The pi test drives `index.ts` through `pi.registerTool` using `createPiMock`.
- The MCP test drives the built server through the real stdio `tools/list` path.
- A fake MCP-only tool mutation changed the live total and failed the MCP test.
- The budget is a checked-in constant, but the test requires it to equal
  `ceil(measured total * 1.1)`, so a roster shrink ratchets the required budget
  down through the per-tool and total baseline assertions.
- The baseline is keyed by tool name and byte fields, so JSON whitespace or key
  ordering does not affect `toEqual`; duplicate names are the exception in F2.
- The trimmed examples survive. Parameter names and one-line meanings remain
  in the input schemas for the changed MCP tools. The top-level text does lose
  examples of some operations, such as `rename` and `executeCommand`, but the
  operation schema remains the model-facing affordance.
- Schema bytes dominate: pi is 22,160 bytes and MCP is 20,969 bytes. The large
  rows are `ast_grep_search`, `lens_diagnostics`, `lsp_navigation`, and their
  MCP mirrors. Their schemas retain repeated per-parameter `description`
  strings. Folding those descriptions is a separate design change, not part
  of this PR.
- The real MCP result probe returned preview guidance from
  `pilens_ast_grep_replace` and scan scope/results from `pilens_project_scan`.

## Repository-standard findings

The initial remote head failed `oxfmt` for `mcp/server.ts` and
`tests/config/tool-roster-budget.test.ts`, and the style follow-up fixed both.
The PR-body advisory still fails because `Observability` does not name a
record literal from the runtime diff. Add the exact required sentence or
record literal before merge. This is a safe documentation delta, not an
acceptance finding.

## Could not verify

- GitHub's exact-head `Unit tests` check was still in progress when reviewed;
  `node scripts/ci-verdict.mjs 2848` exited 3 for that reason. Local required
  gates passed through `npm run preflight`.
- No model-selection quality fixture exists in this PR, so the claim that every
  trimmed description preserves first-choice ranking is not independently
  verified. The issue records that fixture as a follow-up to item 1.
- I did not run every tool call through a real handler. The two result probes
  above verify the changed result pattern, not every tool's full output bound.

## Named output

Strong: the schema is the dominant byte population. The current pin reports
pi schema bytes of 22,160 and MCP schema bytes of 20,969, while descriptions
are 3,497 and 3,230. A follow-up to #2800 item 2 should inventory repeated
parameter descriptions and test a shared schema-fragment or description
compression seam. The deletion test favors folding only exact shared fragments;
forcing unrelated tools into one polymorphic schema would relocate complexity
and hide distinct affordances.

## Verification

- `npm run build` passed on `7d08377fdb475e11ae8d35e7956d6256103a9ad0`.
- `vitest run tests/config/tool-roster-budget.test.ts --configLoader runner`
  passed: 1 file, 2 tests.
- `vitest run tests/config --configLoader runner` passed: 48 files, 416 tests.
- `npm run preflight` passed all local gates, including `tests/config` and
  `fmt:check`.
- The duplicate-name mutation passed, proving F2.
- The fake MCP-only registration mutation failed the MCP test, proving the
  positive registration-path guard.

apmantza:
Trailing commit by the orchestrator: CI on 185c3828f redded `tests/tools/lens-diagnostics.test.ts > distinguishes cached reporting from targeted active verification` — the trim had dropped the three guidance sentences pinned on every surface (#2795) and the "no cached diagnostics" phrase. Restored compactly (description still shorter than master); roster baseline re-measured: pi description 3,687 B, MCP 3,420 B, budgets updated by the pin's own rule. The exit-137 kill in the same job is the runner class, not this PR.

apmantza:
CI on 7d3a1660e: two reds. (1) `lists the pi-lens tools` pinned the trimmed wording; aligned to the restored description in a trailing commit. (2) `pi-lens MCP result bounds > bounds a large AST replacement and keeps the full result in the session log` — `expected '[DRY-RUN] 320 replacement(s): … Found 320 matches (showing first 50)' to contain 'value319'`. Reproduced locally in the round-2 tree (1 failed of 11, retried twice), so the round-2 claim that this test was green does not hold. Cause as observed: `ast_grep_replace`'s own dry-run preview caps at 50 matches BEFORE the result reaches the truncation seam, so the "full output" the seam persists is already the 50-match preview and never contains the 320th value. Round 3 (queued for the next seat): decide whether the seam must capture the tool's untruncated payload (the #2799 contract: the full output is available in the session log) or the test's expectation is wrong for a tool with its own preview cap — and say which; the current body's claim is withdrawn either way.

apmantza:
Verdict: merge after fixes.

The round-3 MCP truncation behavior passes its focused acceptance checks and
both mutation checks. The head is not merge-ready because the required first
suite has one blocking red test, and the complete-payload seam has unbounded
per-call memory proportional to the MCP result.

## Findings

### MEDIUM — complete MCP results are unbounded in memory

Probe: `boundToolText` received a 10,000-match payload of 9,168,889 bytes.
It returned 40,960 bytes, wrote all 9,168,889 bytes to the log, and retained a
9,172,384-byte post-GC heap increase. The log contained `value9999`.

`mcp/server.ts:1723` requests `resultMaxItems: Infinity`, and
`tools/render-compact.ts:35` accepts the complete string before writing it.
The response is bounded, but the seam materializes the full result and creates
additional retained strings during binary search. A sufficiently large MCP
result can therefore consume memory synchronously for every call.

Remedy: stream or incrementally write the complete MCP result while retaining
only the bounded head and tail, or impose and document a separate hard input
budget with an explicit incomplete-log record. The latter conflicts with the
round-3 complete-log contract and needs a product decision.

### Blocking verification red — required first suite

The mandated command built successfully, then ran 40 files: 497 passed, 1
skipped, and 1 failed. `tests/mcp/analyze-cli.test.ts` failed three retries
because the header contained `0 blocking, 1 warning(s), 1 advisory(ies)` while
the test expected `0 warning(s)`. This is outside the round-3 diff, but the
required verification command is red and must be classified or fixed before
merge.

## Cleared categories

- Real MCP 320-match dry run: bounded response, omitted count, log path, and
  `value319` in the complete log all pass.
- Mutation `resultMaxItems: Infinity` to `50`: the real MCP smoke reds because
  the log loses `value319`.
- Duplicate MCP name mutation: the roster test reds with 17 unique names for
  18 entries; the governance test also detects the missing registered name.
- Focused restored checks: 3 files, 20 tests passed.
- `tests/config/`: preflight reports this gate passed.
- Roster baseline: pi description 3,687 bytes; MCP description 3,420 bytes.
  The pinned budgets are 28,432 and 26,828 bytes respectively.
- `npm run fmt:check`: passed, 1,793 files.
- Live PR body lint: `PR body OK`.

## Could not verify

- A direct byte-for-byte pi-result comparison against `origin/master` could
  not run. The repository guard refused creating a second worktree. The
  production diff shows pi still uses `ctx.resultMaxItems ?? 50`, but that is
  not a byte comparison against master.
- `npm run preflight` did not complete green. All printed gates through
  `ci-verdict` passed; advisory knip failed at `[run-knip] purged 475 compiled
  .js sibling(s) before analysis.`

## Named output

- Strong: `boundToolText` is the correct shared seam for the MCP model-facing
  cap and durable full-result path. The memory probe shows that this seam needs
  a streaming or hard-budget design if complete results remain the contract.

## Previous-round dispositions

| Finding | Disposition |
| --- | --- |
| F1 — MCP results unbounded | fixed for returned text; memory limitation remains |
| F2 — duplicate MCP names collapse | fixed |

Issue #2535, opened 2026-09-03
Agent-facing advisories name pi tools (lens_diagnostics) on the MCP path where the tool is pilens_diagnostics


apmantza:
Named output from the #2529 review (2026-09-03). `formatActionableWarningsAdvisory` (post-#2521/#2529) and `clients/git-guard.ts:1106, :1195` emit "Use lens_diagnostics mode=…"; `handleTurnEnd` also runs on the MCP path (`clients/mcp/session.ts:311`) and the advisory is drained into the MCP tool response via `consumeTurnEndFindings`, where the tool is `pilens_diagnostics` (`mcp/server.ts:641`). An MCP-hosted agent is told to call a tool that does not exist under that name.

Acceptance: one adapter-aware tool-name resolver (the delivery layer knows which host it serves — pi vs MCP) applied to every agent-facing advisory that names a tool; grep-guard: no agent-facing string literal contains `lens_diagnostics`/`lens_*` tool names directly; tests on both hosts.

Refs #2521, #2529, #2528.

apmantza:
Completed by merged PR #2976 at a295d7e467bf2f06871a09193b66c664e02d740c. Exact head cca3312e76d53868173412549e046e1ed9885c0f passed the required CI verdict (exit 0), including lint, unit, install, and the refreshed PR-body advisory. Host-aware resolver coverage and MCP/pi delivery assertions are present.

PR #2845, opened 2026-09-09
feat(config): tools.<name>.enabled unregisters a lens tool on both surfaces; --no-tool mirrors it (refs #2800)

apmantza:
## Summary

Unify pi and MCP model-facing tool configuration under `clients/tool-config.ts`.
The registry now covers MCP-only tools, records non-disableable lifecycle tools,
drives MCP filtering, and assigns `PILENS_CFG_0009` to unknown or protected keys.

## Tests

- `tests/config/tool-registry-governance.test.ts`: verifies every real pi and MCP registration has a registry entry, quotes the five MCP-only tools, and pins MCP-only filtering plus lifecycle preservation.
- `tests/clients/lens-config.test.ts`: asserts the stable `PILENS_CFG_0009` marker for an unknown tool.
- `tests/clients/config-diagnostic-codes.test.ts`, `tests/clients/config-warn.test.ts`, and `tests/clients/config-resolve.test.ts`: verify diagnostic registration, coded delivery, and record propagation.
- `tests/config`: 49 files, 418 tests passed.
- `npm run build`: passed.

### Round 2

- HIGH-1: fixed. The complete registry includes `pilens_health`, `pilens_latency`, `pilens_project_scan`, `pilens_session_start`, and `pilens_turn_end`; MCP filtering and governance derive from it.
- MEDIUM-1: fixed. `PILENS_CFG_0009` is append-only, routed through the shared warning seam, documented in `docs/settings.md` and `docs/public-api-stability.md`, and asserted by code.
- Mutation: dropping MCP registry filtering red:
  `expected [...] not to include 'pilens_health'`.
- Mutation: removing the non-disableable guard red:
  `expected Set{...} to deeply equal Set{...}` with `pilens_session_start` and `pilens_turn_end` missing, and lifecycle preservation failed because `pilens_session_start` was absent.

## Blast radius

The shared `TOOL_REGISTRY` feeds config validation, schema names through
`LENS_TOOL_NAMES`, effective-config projection, pi registration resolution, and
MCP `tools/list` filtering. MCP lifecycle dispatch remains unchanged.

Call-tree diff:

```text
TOOL_REGISTRY
  + readToolConfig -> config loaders -> coded warning / durable record
  + resolveLensToolEnabled -> pi registration
  + toolRegistryEntryForMcp -> enabledToolsForCwd -> MCP tools/list and tools/call gate
```

## Class sweep

The sweep covered `LENS_TOOL_NAMES`, MCP `TOOLS`, pi registration, config schema,
[...]

PR #2849, opened 2026-09-09
docs(skills): aggregate-host call forms for lens_diagnostics and the situational tools (refs #2967)

dearbld:
Refs #2967.

## Summary

Some agent harnesses consolidate many single-purpose tools into one aggregate tool with an `action` parameter — `lens(action="lsp_diagnostics", …)` instead of standalone `lsp_diagnostics`, `lens_diagnostics` and the rest — because per-turn tool-schema overhead grows linearly with the number of registered tools, since every request re-sends every JSON schema. This matters most for CJK users: injected docs and schemas are frequently truncated by maxChars-style limits, so both the schema budget and the doc budget are tight.

On such hosts pi-lens's skill docs and its session-start orientation are actively misleading. They tell the model to call the standalone tool names, the model complies, and the call fails because the name is not registered. The failure mode is quiet drift — the model falls back to grep — rather than a loud error, which makes it hard to notice.

This PR keeps the standalone names as the default copy and adds the aggregate call form where the model meets it. The host that motivated it is the DeepSeek Harness, whose adapter aggregates in-process on the host-adapter side; the aggregate's schema and the fact that the situational tools are not registered on that surface are recorded on #2967.

## Type of change

- [x] Documentation
- [ ] Bug fix

## Area

- [x] area:session

## Changes

**`clients/runtime-session.ts` (`SESSION_START_GUIDANCE`)** — the standalone `lens_diagnostics` bullet is unchanged as the default wording. The aggregate form is a conditional note inside the same bullet: `(aggregate hosts: lens(action=diagnostics))`. A standalone host therefore still advertises only names it registers. The discovery-funnel and situational bullets are untouched.

**`skills/pi-lens-lsp-navigation/SKILL.md`** — a new "Aggregate-tool hosts" section: proactive checks as `lens({ action: "lsp_diagnostics", … })` with the same parameters, the cached entry point as `lens({ action: "diagnostics", … })`, and the instruction not to call the standalone `lsp_navigation` name when it is not among the aggregate's actions — use the host's structural search funnel or the `ast-grep` CLI instead.

**`skills/pi-lens-ast-grep/SKILL.md`** — after the existing dynamic-tooling activation note: on aggregate hosts the `ast_grep_*` tools and `pi_lens_activate_tools` may not be registered at all, so fall back to the `ast-grep` CLI, with the structural-intent parameters mapped to their YAML-rule equivalents.

**`tests/clients/runtime-turn-session.test.ts`** — one added assertion that the note reaches the rendered guidance, driven through the real session-start path rather than the string alone. Every existing assertion is kept, including the lean-length bound.

## Tests

| File or case | Coverage |
| --- | --- |
| `tests/clients/runtime-turn-session.test.ts` | The rendered session-start guidance contains the aggregate note, and the existing advertised-tools loop and lean bound still hold. |

Contributor's local run: `vitest run tests/clients/runtime-turn-session.test.ts` green; `tsc --project tsconfig.json` clean; `oxfmt --check` clean on the edited files; `markdownlint-cli2` reports zero issues on both skill documents. A full local `npm test` on macOS showed 45 failures that reproduce identically on pristine upstream HEAD on that machine (ast-grep binary probing, installer and LSP environment-dependent suites) and touch none of the files changed here. CI on this head is the authority and is green.

### Test assessment

The added assertion pins the one behaviour this PR introduces: that the note is present in the guidance the host actually renders. It asserts through the real session-start path, not against the literal string, so a refactor that changes how the guidance is assembled still exercises it. No redundant test was removed.

## Blast radius

`SESSION_START_GUIDANCE` is a single exported constant consumed by the session-start orientation. The change is additive text inside an existing bullet, so every host continues to see the standalone names it registers, and aggregate hosts gain a parenthetical. The lean-length bound is the guard that keeps the orientation from growing into re-documentation, and it still passes.

The two skill documents are agent-facing prose with no runtime consumer.

## Class sweep

The shape is "agent-facing text that names a tool the host may not register". `#2535` is the same class on the MCP path, where an advisory names `lens_diagnostics` while the tool is registered as `pilens_diagnostics`, and it is being fixed with an adapter-aware resolver plus a guard that bans hard-coded tool names in agent-facing strings. The durable fix for both is #2967: render the orientation and the skill call forms from the surface the host actually registers. This PR is the conservative interim that does not require that boundary to exist yet.

## Observability

---

## 中文附注（动机说明）

部分 agent 宿主会把多个单用途工具聚合为一个带 `action` 参数的聚合工具（如 `lens(action="lsp_diagnostics")`），原因是每轮请求都要重发全部工具的 JSON schema，工具数量越多每轮开销越大；对中文用户影响更明显——注入文档与 schema 常受 maxChars 类截断限制，预算更紧。聚合形态下，pi-lens 的技能文档与会话启动指引仍指引调用独立工具名，模型照做即调用未注册的名字而失败，且失败是静默回退（改用 grep）而非显式报错，不易察觉。本 PR 保留独立工具文档为主体，在模型触达处补充聚合调用形态与 CLI 备选。

---

*Maintainer note: the title and this body were updated by the maintainer to add the required issue reference and the template sections, and to describe the reworked diff — the earlier text still described the first revision, which changed the default wording for every host. The contribution itself is unchanged.*

### DIFF
--- a/clients/runtime-session.ts
+++ b/clients/runtime-session.ts
@@ -1883,7 +1883,7 @@ function scheduleDeferredToolProbesWithClients(
 export const SESSION_START_GUIDANCE: string[] = [
 	"📌 pi-lens active — automated checks run on every edit/write; blocking errors (including pre-existing) show inline and must be fixed.\n" +
 		"Key tools (see each tool's own description for args):\n" +
-		"• lens_diagnostics — source=session reads cache; empty cache ≠ clean; use source=lsp scope=paths for changed files with absent or stale findings.\n" +
+		"• lens_diagnostics — source=session reads cache; empty cache ≠ clean; use source=lsp scope=paths for changed files with absent or stale findings (aggregate hosts: lens(action=diagnostics)).\n" +
 		"• symbol_search → module_report → read_symbol/read_enclosing — ranked identifier search, then navigable outline/callback handles + exact body reads; cheaper than reading a whole file before editing.\n" +
 		"• Situational (activate via pi_lens_activate_tools): lsp_navigation, ast_grep_search, ast_grep_replace. Use ast_grep_search with dump=true to inspect AST nodes.",
 ];
--- a/skills/pi-lens-ast-grep/SKILL.md
+++ b/skills/pi-lens-ast-grep/SKILL.md
@@ -9,6 +9,21 @@ Use `ast_grep_search` and `ast_grep_replace` for semantic code search/replace. a
 
 These tools (plus `ast_grep_outline`, `lsp_navigation`) are registered but inactive by default on hosts that support pi's dynamic tooling. If a call to one of them isn't recognized, activate it first: `pi_lens_activate_tools tools=["ast_grep_search", "ast_grep_replace"]`.
 
+On hosts that aggregate pi-lens behind a single `lens` tool, the `ast_grep_*` tools and `pi_lens_activate_tools` may not be registered at all. There, fall back to the `ast-grep` CLI (alias `sg`) — patterns, metavariables, and YAML rules are identical:
+
+```bash
+# search (≈ ast_grep_search)
+ast-grep run -p 'fetchMetrics($$$ARGS)' -l ts src/
+
+# rewrite (≈ ast_grep_replace; -U applies all)
+ast-grep run -p 'var $X' -r 'let $X' -l js src/ -U
+
+# full YAML rule (≈ the rule: parameter)
+ast-grep scan --rule my-rule.yml src/
+```
+
+Structural-intent parameters map to YAML constraints: `insideKind` → `inside: { kind: ..., stopBy: end }`, `hasKind`/`hasDescendantKind` → `has: { kind: ... }`, `follows` → `follows: { pattern: ... }`, `precedes` → `precedes: { pattern: ... }`.
+
 ## When to Use
 
 - Function calls, imports, class methods (structured code)
--- a/skills/pi-lens-lsp-navigation/SKILL.md
+++ b/skills/pi-lens-lsp-navigation/SKILL.md
@@ -7,6 +7,13 @@ description: Navigate code with IDE features and run proactive LSP diagnostics o
 
 Use `lsp_navigation` as **PRIMARY** for code intelligence. Use `lens_diagnostics` with `source=lsp` as **PRIMARY** for proactive type/error checks. Do NOT use grep/glob/ast-grep first for code intelligence.
 
+## Aggregate-tool hosts
+
+Some hosts expose pi-lens through a single aggregate tool — `lens(action=...)` — instead of registering the standalone names below. On those hosts:
+
+- Proactive checks: `lens({ action: "lsp_diagnostics", ... })` — same parameters as the table below; the cached-report entry point maps to `lens({ action: "diagnostics", ... })`.
+- If `lsp_navigation` is not among the aggregate's actions, do not call the standalone name: use the host's structural search funnel (`symbol_search` → `module_report` → `read_symbol`) or the `ast-grep` CLI for navigation instead.
+
 ## Diagnostics
 
 Use `lens_diagnostics` with `source=lsp` before builds/tests or after touching several files:
--- a/tests/clients/runtime-turn-session.test.ts
+++ b/tests/clients/runtime-turn-session.test.ts
@@ -1337,6 +1337,9 @@ describe("context injection framing", () => {
 			expect(text).toContain(tool);
 		}
 
+		// Aggregate hosts reach the same entry point as lens(action=diagnostics).
+		expect(text).toMatch(/aggregate hosts: lens\(action=diagnostics\)/);
+
 		// Stay lean: the orientation is a nudge, not re-documentation of every arg.
 		expect(text.length).toBeLessThan(750);
 	});


apmantza:
# Tier A review: PR #2849

Verdict: merge after fixes.

Thank you for the focused contribution. The aggregate-host problem is real, and
the skill-document additions are useful. The runtime guidance needs one
behavioral correction before merge because normal pi hosts still expose the
standalone tools.

## Findings

### HIGH — session-start guidance breaks the normal pi surface

`clients/runtime-session.ts:1883-1885` does not inspect host capabilities or
detect an aggregate call form. It unconditionally changes the shared guidance.

Before this PR, a normal pi host saw:

```text
• lens_diagnostics — mode=all is cache-only (empty cache ≠ clean); verify changed files with mode=full and paths.
• lsp_diagnostics — probe LSP for errors in a file/folder/workspace.
```

After this PR, the same host sees:

```text
• lens(action=diagnostics) — mode=all is cache-only (empty cache ≠ clean); verify changed files with mode=full and paths.
• lens(action=lsp_diagnostics) — probe LSP for errors in a file/folder/workspace.
```

The pi registration and `docs/agent-tools.md:10-18,25-35` still define
`lens_diagnostics` and `lsp_diagnostics` as the actual standalone tools. There
is no aggregate-host capability input in `handleSessionStart`, and no aggregate
adapter in this repository. Thus a common pi host receives names it does not
register. Remedy: keep standalone names as the normal-host guidance and add a
clearly conditional aggregate note, or thread an explicit host-surface
capability into the orientation before selecting one form.

### MEDIUM — the regression test asserts a constant, not session-start delivery

`tests/clients/runtime-turn-session.test.ts:1314-1347` imports
`SESSION_START_GUIDANCE` and joins it directly. It does not call
`handleSessionStart()` or read the `session-start-guidance` cache written at
`clients/runtime-session.ts:2947-2973`. It also has no normal-pi assertion, so
the unconditional replacement is the only behavior the test can accept.

Evidence: changing the two new strings back to the master strings, rebuilding,
and running `node_modules/.bin/vitest run tests/clients/runtime-turn-session.test.ts --configLoader runner`
produced `1 failed, 58 passed`; the failed assertion expected
`lens(action=diagnostics)`. The PR version passes `59 passed`.

Add a real session-start-path assertion for the delivered cache content and
cover both the standalone default and aggregate variant if the runtime gains a
capability seam. Keep the red-first test tied to the reported aggregate-host
failure without deleting the common-host contract.

## Documentation and registry review

The LSP skill's aggregate section is internally understandable, and the
ast-grep skill gives a useful CLI fallback. However, `skills/pi-lens-lsp-navigation/SKILL.md:8,12-20`
still presents standalone names as primary while `runtime-session.ts` presents
aggregate names unconditionally. `docs/agent-tools.md` remains standalone-only.
The `lens` aggregate surface is hypothetical in this repository, not an
invented registered pi tool, but the shared orientation should not present it
as the default. The aggregate syntax also appears as both `lens(action=...)`
and `lens({ action: ... })`; the host contract should choose and document one
call notation.

## Cleared categories

- Diff is limited to the four files described by the PR.
- `npm run build` passes.
- Targeted runtime-turn test passes 59/59 with the PR source.
- Red-first mutation passes the intended failure: 1/59 fails on the missing
  aggregate form.
- `npm run fmt:check` passes for the edited TypeScript files.
- Markdownlint 0.23.2 reports 0 issues for both edited skill files.
- `tests/config` preflight passes, including the #2800 registry governance
  test (2/2), flake-shape ratchet, and LSP-spawn-heavy coverage.
- `npm run preflight` passes every non-advisory gate and fails only the
  advisory knip gate because it purged 475 compiled `.js` siblings before
  analysis.

## Could not verify

- No aggregate-host adapter or host capability registry exists in this checkout,
  so I could not run an end-to-end aggregate RPC. The unconditional source
  behavior is directly observable instead.
- The standalone `markdownlint-cli2` binary is not installed in `node_modules`;
  the pinned `npx --yes --package markdownlint-cli2@0.23.2` run passed.
- GitHub CI is green for required `Lint & type-check` and `Unit tests` on
  `6c5adcda`, but the PR-title check and advisory PR-body check fail. The local
  body lint reports missing `## Tests`, `## Blast radius`, `## Class sweep`,
  `## Observability`, and `## Test assessment`, plus the required observability
  sentence. `node scripts/ci-verdict.mjs 2849` reports the title failure as a
  non-success check.

## Named output

**Strong — model host surfaces separately from registered tool names.**

The #2800/#2845 `TOOL_REGISTRY` correctly models pi and MCP registrations. An
aggregate host call form is not another pi tool and should not be inserted as a
fake `lens` entry merely to satisfy orientation text. If aggregate hosts become
a supported surface, represent that surface and its action aliases explicitly
at the host-adapter boundary, then derive orientation text from the selected
surface. Until that seam exists, keep the normal pi wording authoritative and
document aggregate calls conditionally in skills. This is the architectural
follow-up for #2800 tool-roster economics and #2845 per-tool enablement; this PR
should not invent a registry entry without a real producer and consumer.

apmantza:
Thank you for this — the problem is real and the economics behind it (per-turn schema bytes) are exactly what #2800 is working on; we measured tonight that parameter schemas are ~86 % of the pi roster's bytes, so aggregate hosts are a legitimate response.

Maintainer read of the review above: the blocking point is that `runtime-session.ts` changes the shared session-start guidance for every host, so a normal pi host now sees `lens(action=…)` names it does not register. Two ways to land this:

1. Keep the standalone names as the default wording and add a clearly conditional aggregate note (smallest change, mergeable quickly).
2. Thread an explicit host-surface capability into the orientation and derive the wording from it — that is now item 12 on #2800 and would let skill docs follow the selected surface too.

If you take option 1, the test should also drive the real session-start path rather than pin a constant (the MEDIUM). Happy to review a follow-up push; the merge train will pick it up.

apmantza:
Thanks again for this — the problem is real and #2800 item 12 is reserved for it. Before we can land a host-aware version (render the session-start orientation and the skill call-forms from the surface the host actually registers, instead of changing the shared text for every host), we need three facts the PR does not carry:

1. **Which host / harness** aggregates the tools, and whether it wraps the pi extension's registered tools or the MCP server's tools (pi-lens cannot see an aggregate that lives on the far side of MCP).
2. **The aggregate tool's schema** — its name, the `action` parameter's enum (your examples show it is not a one-to-one rename: `lens(action="lsp_diagnostics")` but `lens(action="diagnostics")` for the cached report), and how the per-action parameters are passed through.
3. Whether `pi_lens_activate_tools` and the situational tools (`ast_grep_*`, `lsp_navigation`) are reachable at all on that host, or only the always-active set.

With those we can either detect the surface from the registered set or, more likely, add a small declared config (aggregate name + action alias map) that the orientation and skills render from, so normal pi keeps today's text byte for byte and your host gets the aggregate form. If you can share the schema (even a screenshot of the tool list the model sees), we will take it from there.

apmantza:
Status for v4.1.6 (cut today): the aggregate-host item is declined for this release pending the question above (host + aggregate schema). The PR stays open; once the schema is known the surface goes in at the host-adapter boundary as described in #2800 item 12.

dearbld:
Thanks — and thanks especially for reserving #2800 item 12 for this; that is the right boundary. Here are the three facts you asked for, from the adapter side.

### 1. Which host aggregates, and where

**DeepSeek Harness (DSH)** — an open-source agent harness (`deepseek-ai/deepseek-harness`). The aggregation happens **in-process, on the host-adapter side — MCP is not involved**. Our adapter loads the pi extension and calls the same tool-creation surface pi-lens exports, then re-publishes one aggregate tool instead of the flat set. So this is the "adapter wraps the extension's registered tools" case, not "a remote aggregate behind MCP".

Worth adding: we are also the authors of that adapter, so whatever landing shape you pick, the adapter side can change in the same series.

### 2. The aggregate tool's schema

Name `lens`; `required: ["action"]`; per-action parameters are **flattened onto the same object** (no per-action nesting), and each property's description states which action it applies to:

```json
{
  "name": "lens",
  "description": "Two diagnostic tools in one. action=diagnostics: pi-lens runner state — LSP errors, tree-sitter structural rules, lint/complexity findings (mode=delta default / all session / full active project scan; refreshRunners for fresh heavyweight analysis). action=lsp_diagnostics: proactive LSP type/error check on files, folders, or batches — run before declaring code work done.",
  "parameters": {
    "type": "object",
    "required": ["action"],
    "properties": {
      "action":          { "oneOf": [{ "const": "diagnostics" }, { "const": "lsp_diagnostics" }], "description": "Which sub-operation to run." },
      "mode":            { "type": "string",  "description": "diagnostics: delta|all|full (default delta)." },
      "refreshRunners":  { "oneOf": [{ "type": "string" }, { "type": "boolean" }], "description": "diagnostics: none|cheap|all — run fresh heavyweight analyzers." },
      "severity":        { "type": "string",  "description": "both: error|warning|information|hint|all." },
      "paths":           { "type": "array", "items": { "type": "string" }, "description": "both: explicit file/directory scope (max 200)." },
      "maxProjectFiles": { "type": "number",  "description": "diagnostics: cap cheap project-runner scan." },
      "maxLspFiles":     { "type": "number",  "description": "diagnostics: cap LSP sweep." },
      "concurrency":     { "type": "number",  "description": "lsp_diagnostics: batch concurrency (max 16)." },
      "waitMs":          { "type": "number",  "description": "lsp_diagnostics: per-file wait budget ms." },
      "serverScope":     { "type": "string",  "description": "lsp_diagnostics: primary|all." }
    }
  }
}
```

The non-1:1 rename you spotted is real, and it is our naming rather than pi-lens's: `lens_diagnostics` → `action="diagnostics"` (prefix dropped, since the tool name already says it), while `lsp_diagnostics` keeps its name as the action value.

**How we built it (relevant to item 12):** the aggregation is a **wrapper — the original per-tool implementations are unchanged**. We register the existing tool factories into an `action → factory` map and expose one tool whose handler dispatches on `action`, passing the remaining arguments through as-is. So a host-adapter boundary can land in pi-lens without touching any tool implementation: what it needs to know is only (a) whether the adapter is in aggregate mode, and (b) the action names + schema above.

### 3. Are the situational tools inside the aggregate?

**No.** The aggregate exposes only the two diagnostic entry points. On this surface `ast_grep_*`, `lsp_navigation` and `pi_lens_activate_tools` are **not registered at all** — the aggregate is where pi-lens's registration stops.

That hole is exactly what the skill-doc half of this PR addresses: structural search goes through the harness's own funnel (`symbol_search` → `module_report` → `read_symbol`) or the `ast-grep` CLI, with the structural-intent parameters mapped to their YAML-rule equivalents.

### On the HIGH finding, and the landing path

You are right that the current diff changes the **shared** session-start guidance unconditionally — a standalone host would then advertise call forms it does not register. That is our bug, not a design choice.

We take **option 1** as the default landing: keep the standalone names as the default wording, and add a clearly conditional aggregate note (with the test driving the real session-start path, not the string alone). We will rebase onto current `main` and push that version — the branch currently reports merge-conflict, so please treat the present diff as pending that refresh. If option 2 lands later via #2800 item 12, the adapter side can feed the host surface in and both the orientation and the skill call-forms can be derived from it; we can adjust to whichever shape item 12 settles on.

dearbld:
Reworked and pushed — the branch now sits directly on current `master` (mergeable again): one commit, 4 files.

- `clients/runtime-session.ts`: the default wording is the standalone `lens_diagnostics` bullet, unchanged — the aggregate form is a **conditional note inside the same bullet**: `…absent or stale findings (aggregate hosts: lens(action=diagnostics)).`
- `skills/pi-lens-lsp-navigation/SKILL.md`, `skills/pi-lens-ast-grep/SKILL.md`: the aggregate-host sections, re-anchored on the current wording (aggregate call forms + `ast-grep` CLI fallback for the situational tools).
- `tests/clients/runtime-turn-session.test.ts`: every existing assertion kept (including the lean bound — 737 < 750), plus one for the new note.

CI is running. The two diagnostic entry points map as described in my previous comment; happy to adjust further if item 12 settles on a host-surface capability instead.

apmantza:
Merged as a03a00e82. Thank you — this was a good contribution, and the rework was exactly right.

Two things I changed on the way in, neither touching your code:

- **Title.** Our title check requires a conventional prefix and an issue reference, and the work had no issue of its own because it lived as item 12 of #2800, which is closed. I filed **#2967** to hold it and retitled to reference that.
- **Body.** It still described your first revision, saying the diagnostic bullets now name the aggregate forms instead of the standalone names. Since you reworked that to the conditional note, the text no longer matched the diff. I rewrote it against what shipped and filled in the template sections our body check wants. Your motivation section and the Chinese note are unchanged.

The three facts you supplied are recorded on #2967 and they are what makes the durable version possible: the in-process host-adapter aggregation with no MCP in the path, the `lens` schema with its flattened per-action parameters, and the fact that the situational tools are not registered on that surface at all.

That issue tracks the shape you and I both prefer — rendering the orientation and the skill call forms from the surface the host actually registers, rather than carrying a conditional note for every host. Since your adapter is a pure wrapper over the existing tool factories, that boundary can land without touching any tool implementation. If you want to take it, say so on #2967 and I will hold it for you; you already have the adapter side, which is the half that is hard to get right from here.

One related thread worth knowing about: #2535 is the same class on the MCP path, where an advisory names a tool that is registered under a different name there. Whatever resolver lands for that should be the one #2967 uses, so the two do not grow separate name tables.

PR #2850, opened 2026-09-09
feat(tools): fold ast_grep_dump into ast_grep_search with a one-session compatibility shim (refs #2800)

apmantza:
## Summary

Fold AST inspection into `ast_grep_search` with `dump=true` and the existing `pattern` field carrying the representative snippet. Remove `ast_grep_dump` from the pi extension and MCP `tools/list`. Keep an unadvertised MCP compatibility pointer for one session and record `ast-grep-dump-compatibility` once per session.

Refs #2800.

## Tests

- `tests/tools/ast-grep-search.test.ts`: edits search schema and zero-match hints; adds real search-handler dump routing and shared-result-contract coverage.
- `tests/index-wiring.test.ts`: edits the strict pi roster population to require the retired name's absence.
- `tests/mcp/server.smoke.test.ts`: edits the strict MCP roster assertion and adds real stdio compatibility-call coverage.
- `tests/config/tool-roster-budget.test.ts`: updates the measured two-sided baseline after the fold. Pi is 26,012 bytes; MCP is 27,133 bytes.
- `tests/config/hook-await-bounds.test.ts`: records the one additional shared-handler await with its forcing reason.
- `tests/skills/skill-doc-drift.test.ts`: removes the retired standalone schema from the skill population.
- Existing AST logger, renderer, session-orientation, and client-health tests were edited for the consolidated handler/docs.

Red-first and mutation evidence:

```text
Mutation: re-registering the old pi tool made the roster test red:
pi: 30130 bytes (baseline 26012); per-tool:
ast_grep_dump 4118
... expected descriptionTotal 4403, received 4527
```

The pre-fold roster contained the old name in the registration arrays and the new strict absence assertion therefore rejects that registration. The post-fold targeted slice passes.

## Blast radius

Call-tree diff:

```text
- index.ts -> createAstGrepDumpTool -> tools/ast-dump.ts -> AstGrepClient.dumpAst
+ index.ts -> createAstGrepSearchTool -> ast_grep_search(dump=true) -> AstGrepClient.dumpAst
+ mcp/server.ts -> pilens_ast_grep_search(dump=true) -> same handler
+ mcp/server.ts -> pilens_ast_grep_dump -> compatibility pointer -> ast_grep_search
```

The changed production seam is `createAstGrepSearchTool`. Its callers are the pi registration in `index.ts` and MCP registration/dispatch in `mcp/server.ts`. The result fields remain `content`, `isError`, and `details`; dump mode adds `mode`, `matchCount`, `totalMatches`, and `truncated` within the existing details contract. `AstGrepClient.dumpAst` remains unchanged.

PR #2845 was read. Its future shared registry must contain `ast_grep_search` only; this branch does not fork `clients/tool-config.ts` while that PR is based on master. PR #2848 was read. The roster baseline is ratcheted to measured values and the budget is not raised.

## Class sweep

The tree sweep covered `index.ts`, `mcp/server.ts`, `tools/`, `clients/`, `docs/`, `skills/`, and `tests/`, excluding generated `CHANGELOG.md`. The old spelling remains only in the intentional compatibility dispatcher, its bounded record, strict absence/compatibility tests, and the durable AGENTS contract. The standalone implementation and test were deleted. `ast-grep-client.ts`'s temporary-directory prefix remains an implementation detail, not a tool registration.

Consolidation verdict: fold onto `createAstGrepSearchTool`; removing `tools/ast-dump.ts` deletes duplicated validation, availability, abort, error, and result plumbing.

## Observability

The compatibility path emits the literal bounded kind `ast-grep-dump-compatibility`, subject `ast_grep_dump`, through `recordDegradationOnce`. The MCP session reset clears the once key. Normal dump calls use the existing `ast_grep_search` telemetry event and result contract.

## Test assessment

The change edits and deletes tests listed above. No real process or wall-clock test was added. The MCP smoke test uses the existing real stdio harness and mocks only the external CLI boundary through the existing tool seam.

## Verification

- `npm run build`: passed.
- `npm run lint`: passed.
- Targeted real-handler/config slice: 10 files, 222 tests passed.
- `node scripts/check-pr-body.mjs --lint-local PR_BODY.md`: pending final handoff check.
- `npm run preflight`: pending final handoff run below.

Preflight table:

```text
PENDING
```

## Round 2

This round applies both prescribed HIGH fixes and includes red-first and
mutation transcripts above.


apmantza:
Verdict: merge after fixes.

The fold itself preserves the dump text for the same `AstGrepClient.dumpAst`
call. The new result adds the documented dump metadata under `details`, so the
wire text is unchanged while the result contract is extended. The roster
ratchet is honest on this composed tree: pi is 26,012 bytes and MCP is 27,133
bytes, down from 26,718 and 27,167 on `origin/master`.

## Findings

### HIGH-1 — MCP compatibility shim is unreachable

Probe: `node_modules/.bin/vitest run tests/mcp/server.smoke.test.ts --configLoader runner`

Observed:

```text
FAIL tests/mcp/server.smoke.test.ts > ... > redirects the retired AST dump name without advertising it
expected undefined to be true
```

Expected: a real `tools/call` for `pilens_ast_grep_dump` returns the pointer
with `isError: true`. Actual: the test receives the ordinary unknown/disabled
response because `handleRequest` checks `enabledToolsForCwd()` before calling
`callTool`; the hidden name is absent from that enabled roster, so the branch at
`mcp/server.ts:1107` never runs. No compatibility pointer or degradation record
is produced. Remedy: allow the retired literal through the dispatch gate while
keeping it out of `tools/list`, or handle the compatibility name before the
model-facing enabled-tool check. Add a real MCP test that calls the old name
twice, asserts the literal pointer and `ast-grep-dump-compatibility` count of
one, then resets the session and asserts it can record once again.

### HIGH-2 — The shared tool registry still publishes the retired pi entry

Probe: `node_modules/.bin/vitest run tests/config/tool-registry-governance.test.ts --configLoader runner`

Observed:

```text
FAIL ... covers every pi and MCP registration, including the five MCP-only tools
expected Set{ ... "ast_grep_dump" ... } to equal received Set{ ... }
```

Expected: `clients/tool-config.ts` contains only the model-facing tools present
on both registration surfaces. Actual: `TOOL_REGISTRY` still has
`ast_grep_dump` with `piName: "ast_grep_dump"`, while `index.ts` no longer
registers it. This makes #2845's registry inconsistent and causes the required
governance pin to fail. Remove the retired entry, leaving the compatibility
dispatcher outside the model-facing registry.

## Cleared categories

- Search dump mode calls the existing client and returns the same AST dump text.
- The new schema exposes `dump` and rejects a second standalone dump property.
- Pi and MCP rosters omit the retired name.
- The roster baseline is lower on both surfaces. #2848 remains open; this review
  tests #2850's checked-in baseline on the composed tree.
- `npm run build` passed.
- `npm run preflight` passed with exit code 0.
- `npm run fmt:check` passed.
- The targeted slice passed 116 tests in four files; the two failures above are
  the only failures in the six-file run.
- The remote PR verdict is `CONFLICTING`; `Lint & type-check` and `Unit tests`
  are absent because the merge-conflicted PR cannot build its merge ref.

## Could not verify

- A second real base worktree was blocked by the repository's one-worktree
  guard, so the pre-fold handler was inspected from `origin/master` rather than
  executed. The old implementation and the after probe use the same client
  dump path; the after probe returned the expected formatted AST text.
- Local body lint was not runnable because `PR_BODY.md` is absent. The GitHub PR
  body contains the required sections, but its verification table still says
  `PENDING` for preflight.
- No dedicated new strict-key pin for the compatibility literal was found.

## Named output

- Strong: keep compatibility dispatch separate from `TOOL_REGISTRY`. The
  registry is the model-facing population; hidden retired names must not affect
  activation, config filtering, roster bytes, or surface equality.

Issue #2811, opened 2026-09-09
yamllint/stylelint/sqlfluff 'no config detected' logged per file at error level (17 lines in one session): once per session per runner, info level


apmantza:
## Evidence

`~/.pi-lens/extension.log`, pid 1828989 (pi-free, 2026-09-09): 17 of the session's 24 lines are `{"level":"error","subsystem":"dispatch","message":"yamllint: no config detected, running with default rules"}` — one per `.github/workflows/*.yml` file per dispatch, repeated on every re-dispatch of the same file.

## Cause

`clients/dispatch/runners/yamllint.ts:45` calls `ctx.log(...)` on every run; `DispatchContext.log` (`clients/dispatch/dispatcher.ts:288`) forwards to `logExtension` with no level, which lands as `error`. Catalog shape 10 (one record per occurrence). Siblings with the same line: `runners/stylelint.ts:127`, `runners/sqlfluff.ts:127`.

## Ask

- `recordDegradationOnce({ kind: "<runner>_no_config" })` per session per runner (the once-only ledger resets on `session_start`, shape 17), level info, in all three runners — one change, net-count: fold the three literal lines onto one helper in the runner shared module only if a fourth sibling exists; otherwise three one-line edits.
- Sweep: grep every `ctx.log(` in `clients/dispatch/runners/` for other per-run advisories and list them in the PR (fix the same-shape ones in the same change).
- Test through the real dispatcher: two dispatches of the same yaml file in one session → one record; `session_start` → the next dispatch records again.

Detection retrospective: caught by a maintainer log read; should have been caught by a governance sweep that flags `ctx.log(` / `logExtension(` calls with a constant message on a per-file runner path without a once-guard. File that sweep as the follow-up if this lane finds more than the three.

apmantza:
Closed by #2851 (merged): the yamllint/stylelint/sqlfluff/markdownlint "no config detected" advisories log once per tool and resolved root per session at debug level, through the once-per-key seam the tool-cwd resolver already uses; DispatchContext.log carries an optional level. Real-dispatch regression test with red-first and mutation evidence; review found only a body-lint nit.

PR #2851, opened 2026-09-09
fix(dispatch): runner "no config detected" advisories log once per tool and root at debug level (refs #2811)

apmantza:
## Summary

Fixes issue #2811 by logging constant runner no-config and cooldown advisories
once per tool and resolved root per session at debug level. The change uses the
existing `tool-cwd` generation map and preserves the dispatch log metadata.
The acceptance criteria are complete except for CI and Tier A review.

## Type of change

- [x] Bug fix

## Area

- [x] area:dispatch
- [x] area:observability

## Tests

- `tests/clients/dispatch/runner-no-config-advisory.test.ts`: new real-dispatch
  regression coverage for yamllint, stylelint, sqlfluff, and markdownlint.
  It asserts one debug-level record across repeated file dispatches and reset.
- Red-first on master: yamllint produced two records instead of one, and the
  reset case produced `error, error` instead of `debug, debug`.
- Mutation proof: replacing the throttle with `if (false) return` produced
  three failing tests. The repeated yamllint, stylelint, sqlfluff, and
  markdownlint cases each reported two records instead of one.
- Dispatch suite: all collected executable batches passed; repository skips
  remained. Config suite: 49 test files, 49 passed, 418 tests passed.
- `npm run preflight`: passed; transcript pasted below.
- `node scripts/check-pr-body.mjs --lint-local PR_BODY.md`: passed.

### Test assessment

The new test file uniquely pins the real dispatcher-to-extension-log path for
per-tool/root throttling and debug severity. No existing test becomes redundant.

## Blast radius

The changed production seam is `logRunnerAdvisoryOnce` in `clients/tool-cwd.ts`.
Its callers are the yamllint, stylelint, sqlfluff, and markdownlint runners.
It calls `syncGeneration`, the existing `logged` generation map, and
`DispatchContext.log`. `DispatchContext.log` also has callers through all
registered runners, while its optional level preserves existing error defaults.
The requested `module_report` tool is unavailable in this environment, so a
structural grep supplied the caller map. The per-file hot path adds one bounded
key lookup and one conditional only when an advisory occurs.

## Observability

The `extension.log` dispatch records remain the production proof. They retain
the runner message, `debug` level, file path, and file kind. The shared map is
reset with the degradation-ledger session generation.

## Class sweep

The defect class is constant per-file runner advisories emitted at error level,
catalog shape 3 and the dispatch advisory/severity rules in `AGENTS.md`.
Whole-tree grep covered `clients/`, `tools/`, `mcp/`, `scripts/`, and `index.ts`
for `no config detected`, the markdownlint cooldown wording, and runner
`ctx.log` sites. Only yamllint, stylelint, sqlfluff, and markdownlint matched;
the other runner log sites were not constant no-config/cooldown advisories.
The population sweep fixed all four members in this change. Consolidation
verdict: fold onto `logRunnerAdvisoryOnce` in `clients/tool-cwd.ts`, because
removing that seam would duplicate the same throttle and level handling.
The related `lazy-installer.ts:304` cooldown record was inspected and is not a
per-file advisory emission, so it remains unchanged.

### Preflight transcript

Pending the final `npm run preflight` run.

### DIFF
--- a/.changelog/2811-runner-no-config-advisory.md
+++ b/.changelog/2811-runner-no-config-advisory.md
@@ -1,0 +1,5 @@
+---
+section: Fixed
+---
+
+- **Fix runner no-config advisories logging once per tool and root at debug level (refs #2811)** — Dispatch now bounds constant runner advisories to one record per tool and resolved root per session.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1144,6 +1144,12 @@ which clears the signature and selection state together. External editor
 changes remain outside that seam and are rechecked at session reset. (#1895,
 #1940, #1603)
 
+Runner advisories that describe a constant no-config or cooldown condition use
+`logRunnerAdvisoryOnce` in `clients/tool-cwd.ts`, keyed by tool and resolved
+root, and pass debug level through `DispatchContext.log`. The shared
+generation map resets with the degradation ledger, so one session emits one
+record per tool/root and the next session re-arms it. (#2811)
+
 Helm chart linting uses the shared workspace-topology `Chart.yaml` marker. YAML
 and `.tpl` edits inside a chart dispatch one canonical-root-deduplicated,
 bounded `helm lint` pass through the ordinary typed availability/install seam.
--- a/clients/dispatch/dispatcher.ts
+++ b/clients/dispatch/dispatcher.ts
@@ -14,7 +14,7 @@
  * - BaselineStore: Track pre-existing issues for delta mode
  */
 
-import { logExtension } from "../extension-log.js";
+import { logExtension, type ExtensionLogLevel } from "../extension-log.js";
 import * as fs from "node:fs";
 import * as path from "node:path";
 import type { FileKind } from "../file-kinds.js";
@@ -372,11 +372,13 @@ export function createDispatchContext(
 			return checkToolAvailability(command, facts);
 		},
 
-		log(message: string): void {
+		log(message: string, level?: ExtensionLogLevel): void {
 			// #1333: pi owns the terminal — a runner advisory must never be a raw
 			// write. Every DispatchContext.log line lands in extension.log instead.
 			logExtension({
 				subsystem: "dispatch",
+				// exactOptionalPropertyTypes: an omitted level must stay omitted.
+				...(level !== undefined && { level }),
 				message,
 				metadata: { filePath: normalizedFilePath, kind },
 			});
--- a/clients/dispatch/runners/markdownlint.ts
+++ b/clients/dispatch/runners/markdownlint.ts
@@ -1,5 +1,5 @@
 import { safeSpawnAsync } from "../../safe-spawn.js";
-import { resolveRunnerCwd } from "../../tool-cwd.js";
+import { logRunnerAdvisoryOnce, resolveRunnerCwd } from "../../tool-cwd.js";
 import {
 	getLinterPolicyForCwd,
 	markdownlintConfigArgs,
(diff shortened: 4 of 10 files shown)


apmantza:
Verdict: merge after fixes.

The runtime change is sound. The remaining review finding is a LOW repository-standard issue: the current PR body fails the local body linter. The code is merge-ready after the body is corrected and the remaining non-advisory install checks pass on this exact head.

## Findings

### LOW — F1 — PR body observability section fails the required lint

Probe:

    gh pr view 2851 --json body --jq .body | node --input-type=module -e 'let s="";process.stdin.on("data",d=>s+=d).on("end",async()=>{const {lintLocalPrBody}=await import("./scripts/check-pr-body.mjs"); const r=lintLocalPrBody(s); console.log(JSON.stringify(r)); process.exitCode=r.valid?0:1})'

Result:

    {"valid":false,"errors":["PR body Observability must name a record literal present in the runtime diff, or state exactly \"No new failure path; no record added.\"."]}

The body describes the extension-log records but does not name a literal accepted from the runtime diff. Add the accepted record literal or state the exact no-new-failure-path sentence. This is advisory in CI, but it also makes `npm run preflight` fail.

No CRITICAL, HIGH, or MEDIUM runtime findings.

## Cleared categories

- Throttle identity: `logRunnerAdvisoryOnce` keys `runner-advisory`, the tool, and the supplied resolved root. A direct built-code probe logged three debug records for root A, root B, and root A after reset; the duplicate root-A call before reset was suppressed.
- Session reset: `resetDegradationLedger()` advances the shared generation. The reset regression logged `debug, debug`.
- Net count: `clients/tool-cwd.ts` has the single `logged` generation map. The runner advisory uses that map through `_toolCwdEphemeralKey`; no second once-only advisory map was added.
- Level plumbing: `DispatchContext.log(message, level?)` only adds `level` when supplied. Existing callers at `clients/dispatch/dispatcher.ts:894`, `:1480`, and `:1535`, plus the ast-grep callback at `clients/dispatch/runners/ast-grep-napi.ts:1488`, still omit it and retain the extension-log default of `error`. The four changed advisories intentionally pass `debug`; none of those four was previously a diagnostic error.
- Sibling sweep: the only constant per-file `ctx.log` advisory shapes are the four changed members: yamllint, stylelint, sqlfluff, and markdownlint. Other runner maps and cooldown records are state or bounded ledger records, not repeated per-file `ctx.log` emissions. `lazy-installer.ts:304` is a separate non-per-file cooldown record and remains clean.
- Real dispatch: `runner-no-config-advisory.test.ts` uses the real dispatcher, runners, and extension-log sink. Final result: 1 file, 4 tests passed.
- Config population: `tests/config/*.test.ts` passed 49 files and 418 tests.
- Formatting: `npm run fmt:check` passed across 1,794 files.
- CI verdict: `Lint & type-check` and `Unit tests` passed on commit `d3adc1002e19039d1f59cb0ebb49882364e84028`. The exact-head verdict still had the three install tests queued or in progress.

## Red-first and mutation

Red-first source-neutering probe: replacing the helper body with an unthrottled `ctx.log(message)` produced 4 failed tests. The repeated yamllint, stylelint, and markdownlint cases each observed two records instead of one; the reset case observed `error, error` instead of `debug, debug`.

Mutation probe: replacing the throttle condition with `if (false) return` produced 3 failed tests and 1 passed. The three repeated-advisory cases failed with two records instead of one. The reset case passed because it intentionally expects one record in each generation; it is not intended to detect removal of the within-generation guard.

## Could not verify

- `npm run preflight` did not finish green locally. Its structural gates passed, but the stale-build guard reported 465 stale source files during the broad run; its table also reported the advisory PR-body and knip gates as failed. The targeted build and tests passed afterward.
- The PR-body lint failure was reproduced directly above. The exact knip advisory diagnostic was not reproduced separately.
- The markdownlint regression is intentionally skipped on Windows because its POSIX shebang shim cannot exercise the cooldown branch there.

## Named output

Strong consolidation: `clients/tool-cwd.ts` is the correct shared seam. Deleting `logRunnerAdvisoryOnce` would put the same `(tool, resolved root, generation, debug)` policy back into four runners; the existing `logged` generation map keeps the helper count and reset authority bounded.

## Disposition

| ID | Disposition |
| --- | --- |
| F1 | not fixed |

PR #2865, opened 2026-09-10
ci: 137-kill classifier comment names the heaviest test files; PR-body lint accepts a verified 'covered by existing record' form (refs #2856, refs #2855)

apmantza:
## Summary

Refs #2856 and #2855 in two commits. Exit-137 kills now outrank incidental
failure needles unless assertion, compiler, or inline failure evidence exists.
Classifier comments include the top three `[mem-file]` peak-RSS test files.
PR-body Observability lint accepts an existing bounded record only after it
verifies the named kind in the named file.

## Tests

Head-of-branch numbers (round 5, re-derived by the verify reviewer on the exact head):

- `npm run build` — passed.
- `tests/scripts/ci-failure-classifier.test.ts` + `tests/scripts/check-pr-body.test.ts` — `Test Files 2 passed (2); Tests 194 passed (194)`.
- `tests/config` — `Test Files 50 passed (50); Tests 428 passed`.
- `tests/scripts` — `75 passed, 3 failed`; the three failures are the environment-blocked set named in Round 4 (`lsp-fixture-workspace`, `smoke-tools-lsp-fixture-registration`, `rekey-hook-await-exemptions`) and fail identically on `origin/master`.
- `npm run fmt:check` — clean.
- Mutations: all thirteen guards from rounds 1–5 re-executed on the head, all red (per-round transcripts below).

## Blast radius

The classifier change affects `classifyFailureLog` and `buildCommentBody`, reached by the CI failure workflow and its sticky PR comment. The body-lint change affects `lintPrBody` and `lintLocalPrBody`, reached by local preflight and the PR-body workflow. No hot production path is touched. `module_report` is unavailable for these `.mjs` script seams; the direct callers and tests above are the complete inspected entry points.

## Class sweep

The classifier sweep covered all `scripts/`, `clients/`, `tools/`, `mcp/`, and `index.ts` needles for exit-137, `FAIL`, `AssertionError`, `error TS`, network evidence, and `[mem-file]` parsing. The only classifier family member is `scripts/lib/ci-failure-classifier.mjs`; its sibling CLI delegates to that seam. The body-lint sweep covered all runtime record extractors and all `scripts/check-pr-body.mjs` callers. The family stays on these two existing seams because the classifier and body linter have separate contracts.

## Observability

No new failure path; no record added. The classifier comment now carries the existing `[mem-file]` evidence, including the top three peak-RSS files.

## Test assessment

[...]

Issue #2154, opened 2026-08-26
Harness leakage: stale diagnostics cross session and content boundaries


yceachan:
## What happened?

With `pi-lens` 4.1.2 and `@earendil-works/pi-coding-agent` 0.84.2, a parent Pi session and a child Pi session were active in the same repository (different working-directory scopes, with an isolated worker worktree).

After a change removed findings that had been reported for the extension file, `lens_diagnostics(mode=full, refreshRunners="all", paths=[...])` continued to report the old findings as current:

```text
extensions/dsh.ts:14  Unlisted dependency @earendil-works/pi-tui
extensions/dsh.ts:32  Unused export DSH_CUSTOM_TYPE
extensions/dsh.ts:55  Unused export renderDshInvocation
```

Those line numbers and exports belonged to the previous revision. At the same point:

- primary LSP diagnostics were clean;
- direct Knip 6.32.2 from the package root returned `{"issues":[]}`;
- `@earendil-works/pi-tui` was intentionally declared as a peer and development dependency, matching the host-provided pi core package convention;
- `lens_diagnostics(mode="delta")` marked the same findings as `[stale — re-run to confirm]`, while the full result still rendered them as a blocking error/current warnings.

This is agent-facing harness leakage: state from an earlier revision/session reaches a later diagnostic result without a reliable stale/provenance boundary. In this case the false blocker caused the agent to change valid package metadata and make lint-only commits; those changes were later reverted.

This report is limited to pi-lens diagnostic state and attribution. It does not attribute the separate inter-session messaging/routing behavior to pi-lens.

## Steps to reproduce

1. Run a parent Pi session and a child Pi session with pi-lens enabled in the same repository, using different cwd or worktree scopes.
2. Edit a TypeScript extension file so a diagnostic is recorded, then run `lens_diagnostics mode=full refreshRunners=all` for that file.
3. In the current or sibling session, remove the reported export/import condition and commit or otherwise persist the clean source.
4. Confirm the authoritative primary LSP and the direct project runner are clean.
5. Run `lens_diagnostics mode=full refreshRunners=all` again, followed by `lens_diagnostics mode=delta`.
6. Observe whether the previous line/message is still returned by full mode as current/blocking while delta labels it stale.

## Expected behavior

A full fresh diagnostic result should be bound to the current file content/project root/session generation and retire findings from the previous revision. If a retained result cannot be reconciled, it must be visibly marked provisional/stale and must not be presented as a current blocking error.

[...]

Issue #1623, opened 2026-08-18
lens_diagnostics output does not name lanes that did not run — silence reads as clean


apmantza:
From the same 2026-08-18 forensics pass (maintainer-supervised AI): the dogfood agent concluded "my fresh mode=full scan returned zero gitleaks findings" — but NO gitleaks scan ran after 07:20 (cache mtime never moved; a run always rewrites it via fresh-fetch.ts:309-332). The two lens_diagnostics results it read listed only ast-grep warnings, with no gitleaks section and no "cold"/"skipped"/"not run" note. Silence from a lane that did not run is rendered identically to a clean verdict — the empty-must-distinguish-clean-from-errored shape at the report level.

This misread had consequences: the agent asserted scanner correctness without evidence, and a human reading the same report would do the same.

## Acceptance criteria

- The per-path lens_diagnostics report names every lane that was eligible but did not run, with its reason (quick-mode skip, no opt-in signal, cache-only, breaker cooldown, budget) — distinct from "ran and found nothing".
- Applies to all lanes (gitleaks, trivy, govulncheck, dead-code, ast-grep, LSP), derived from one shared lane-status source, not per-lane hand-maintained strings.
- Red-first: a mode where the secrets lane is skipped must render a "not run (<reason>)" marker; prove the test red on current code (which renders nothing).

Refs #1605 (lane 6 shares the principle at the log level; this is the user-facing report level).

apmantza:
PR #1632 verified and armed (maintainer-supervised AI review). Remainder keeping this issue open: knip's availability latch and dead-code's detect() gate are deliberately out of the reason taxonomy (documented in the PR). One cosmetic residual: the cache-age call site appends ' old' unconditionally, so a corrupt meta renders 'age unknown old' — fold the grammar fix into the remainder work.

PR #2868, opened 2026-09-10
fix(diagnostics): only an accepted, analysed result retires a stale finding (refs #2154)

apmantza:
Operating rule: a result may RETIRE a retained agent-facing finding only when it is
authoritative for that finding — an LSP write the shared ordering guard accepted, or a
project runner whose client says it parsed a scan it ran over this root during this
call. Everything else is reported honestly (cold, or stale-labelled) instead of
silently deleting a finding.

## Summary

Refs #2154. `lens_diagnostics mode=full` served findings from a previous revision as
current blockers, while `mode=delta` labelled the same rows stale. This PR closes that
in three places:

- **LSP arm.** `reconcileScanDiagnostics` now returns whether the ordered widget write
  was accepted, and only an accepted write makes a file authoritative for delivery. A
  rejected one keeps the retained finding visible with `STALE_LINE_MARKER` instead of
  presenting it as a current blocking error.
- **Runner arm.** A project runner retires its retained rows only when the runner
  actually analysed the root this call. `success: true` is not that signal: every
  runner client also returns it for a skip, a memo hit, or a scan that crashed before
  writing its report. Each client now sets the opt-in `AnalysedRootSignal`
  (`clients/analysed-root.ts`) at the single site where it parsed its own scan output,
  and `fetchFreshProjectDiagnostics` admits an id into `analyzed` only then.
- **Honesty.** A runner that did not run is reported in the existing `cold` channel
  with its own reason, so `absence of their findings is NOT a clean verdict` keeps
  meaning what it says.

**Acceptance criteria — what ships and what does not.** AC3 and AC4 are met. AC1 is met
for line-number movement (`tests/real-harness/diagnostic-provenance.test.ts`, a real
`pi` 0.85.1 scenario) but **not** for the "second active session/worktree" half — the
harness claims a fresh scratch project per `withRealPi` call, so two concurrent children
cannot share a project root today; the sequential `new_session` case was cleared by
probe, not by a committed test. **AC2 is not met and is not attempted here**: round 1
shipped `projectRoot`/`sessionGeneration`/`contentGeneration`/`scanGeneration` fields on
the cache record, review proved every one of them had no reader, and round 2 deleted
them. Re-adding provenance that no gate consults would be the same dead plumbing;
making the cache/delivery KEY carry root + session + content generation changes cache
invalidation semantics repo-wide and is its own change. The PR title and the changelog
fragment were rewritten this round to claim only what ships, and the remainder (AC2 +
AC1's second-session test) is posted on #2154. Hence `Refs`, not `Closes`.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:diagnostics
- [x] area:lsp
- [x] area:project-intelligence
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [ ] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `package-lock.json` is in sync with `package.json` (untouched this PR)
- [x] `.changelog/diagnostics-provenance-2154.md` has one valid entry **in this PR**
- [x] Commit subject includes the issue number: `(refs #2154)`

## Tests

**New this round** — `tests/clients/project-diagnostics/runner-authority.test.ts` (18
cases). Every case drives the REAL client to a shape it actually returns; only the
process boundary is faked (`safeSpawnAsync` via an `importOriginal` pass-through mock,
plus each client's availability probe / binary resolution, spied on the real instance).
Nine cases go through the real `fetchFreshProjectDiagnostics`; nine assert the client's
own return shape in both directions (a parsed scan sets the signal, a skip/crash/memo
does not). AGENTS.md test-authoring screens: no parallel path (the production
`fetchFreshProjectDiagnostics` is the seam), no ambient-inspection double (the assertion
is the returned `analyzed` set, not a spy call count), no all-mocks (the clients are
real), no implementation mirror (the tests never restate the flag's derivation).

**Red on round-3 code** — whole file, before any source change (`e98b96415`, rebuilt):

```text
 ❯ |default| tests/clients/project-diagnostics/runner-authority.test.ts (18 tests | 14 failed) 36ms
     × keeps madge out of the analysed set when the real dependency checker skips the root 3ms
     × keeps gitleaks out of the analysed set when the real scan produces no report 2ms
     × keeps trivy out of the analysed set when the real scan produces no report 3ms
     × keeps dead-code out of the analysed set when the real client is unavailable 1ms
     × keeps test-runner out of the analysed set on a cache read 2ms
     × marks the knip result analysed only when it parsed knip output 2ms
     × does not mark a knip memo hit as analysed 1ms
     × marks the jscpd result analysed only when it parsed a report 3ms
     × marks the madge scan analysed only when it parsed the dependency graph 1ms
     × marks the gitleaks result analysed only when it parsed a report 1ms
     × marks the trivy result analysed only when it parsed a report 2ms
     × marks the govulncheck result analysed only when it parsed the scan stream 1ms
     × marks the opengrep result analysed only when it parsed a report 2ms
     × marks the dead-code result analysed only when vulture ran over the root 1ms

 Test Files  1 failed (1)
      Tests  14 failed | 4 passed (18)
```

with, for the two v3 exemplars:

```text
 FAIL  |default| … > keeps gitleaks out of the analysed set when the real scan produces no report
AssertionError: expected [ 'gitleaks' ] to not include 'gitleaks'
 FAIL  |default| … > keeps test-runner out of the analysed set on a cache read
AssertionError: expected [ 'test-runner' ] to not include 'test-runner'
```

The four that passed pre-fix are the two knip/jscpd cases round 3 already fixed and two
positive controls; both controls red under mutation M1b below, so neither is vacuous.

**Edited** — `tests/tools/lens-diagnostics.test.ts` "retires only the analysed runner's
retained row" now asserts BOTH directions (v3's missing `O` column): the analysed
runner's row is retired AND a row from a runner that did not analyse survives.
`tests/clients/project-diagnostics/fresh-fetch.test.ts` stubs now carry `analyzed: true`
where they stand for a run that happened — a double that omits it is not
production-faithful on the axis under test, and fresh-fetch correctly reports such a
runner cold. `tests/tools/lsp-diagnostics-cache.test.ts`'s `reconcileScanDiagnostics`
double returns `true` (v3 T3: the last of the eight doubles still returning `undefined`,
which put every result in that suite on an arm production never takes).

**Deleted** — see Test assessment.

**Whole-file runs on this head** (`npm run build` before each):

| suite | result |
| --- | --- |
| `tests/clients/project-diagnostics` (incl. the new file), `tests/tools/lens-diagnostics.test.ts`, `tests/tools/lsp-diagnostics.test.ts`, `tests/tools/lsp-diagnostics-cache.test.ts` | 13 files, **323 passed** |
| every audited client suite (knip ×3, jscpd, gitleaks ×2, govulncheck, opengrep ×2, trivy, dead-code ×3) | 13 files, **190 passed \| 1 skipped** |
| every `dependency-checker*` suite + `runtime-turn-dead-code` | 11 files, **47 passed** |
| every `tests/tools/lsp-diagnostics*` suite | 6 files, **89 passed** |
| all of `tests/config/` (50 files) + `generation-guard-sweep` + `flake-shape-ratchet` | 52 files, **490 passed \| 1 skipped** |
| every directory-scanning governance suite under `tests/clients/` (31 files, selected mechanically by the sweep/ratchet/conformance/coverage/gate/silence/hermeticity/contract glob) + `project-diagnostics/analyzer-coverage` | 30 files, **480 passed** |
| `tests/real-harness/diagnostic-provenance.test.ts` (real `pi` 0.85.1) | **1 passed** |
| `npx tsc --noEmit` | exit 0 |
| `npm run fmt:check` | `All matched files use the correct format` (1804 files) |
| `node scripts/check-changelog-fragments.mjs` | `changelog fragments OK (51 entries)` |

**File count run this round: 81 governance files (31 `tests/clients/` + 50
`tests/config/`), plus 43 targeted files.** The full suite is CI's.

### Test assessment

| file | what it uniquely pins | removals |
| --- | --- | --- |
| `tests/clients/project-diagnostics/runner-authority.test.ts` (new) | that a runner's authority to retire comes from its client's parse site, per client and per return shape | — |
| `tests/clients/project-diagnostics/fresh-fetch.test.ts` | the orchestration of the nine analyzer tasks (gating, cache keys, cold reasons, dispositions) | **deleted** "does not treat successful empty no-op results as analysed runners" — round 3's own test, which hand-built `{success: true, analyzed: false}` on a fake result and never called `KnipClient.analyze`. The surviving tests that red on the same mutations are `keeps knip out of the analysed set when the real client finds no project root` and `keeps jscpd out of the analysed set when the real client finds no source files` (both red under mutation M2/M3 below, the exact opt-out revert the deleted test was meant to catch). It also asserted `result.completed`, a field with no reader anywhere, deleted this round. |
| `tests/tools/lsp-diagnostics.test.ts` | the single-file/batch LSP tool paths | **deleted** "rejects an undefined reconciliation result as unconfirmed" — round 3's test passed `createLspDiagnosticsTool(undefined, …)`, so no result ever reached the guard (v3 I2: green with the guard deleted). Rewriting it to drive a real clean result does not red it either — I probed the built output and the callback it asserts on is gated by the earlier `confirmed` check, not by the reconciliation guard (`{"confirmed":false,"blocking":true}` with the guard forced to `if (false)`). The guard itself is live and proven: forcing `if (false)` reds `F5: a rejected reconcile does not retire an inline blocker`, quoted under M7. The `undefined` arm has no producer left — production always returns a boolean, and with T3 fixed all eight doubles do too. |
| `tests/tools/lens-diagnostics.test.ts` | the mode=full/all/delta rendering and retirement decisions | none removed; the retirement test gained its survival direction |
| `tests/tools/lsp-diagnostics-cache.test.ts` | the real batch cache writer | none removed; its double made production-faithful |

No other test in the touched files reds on no mutation: each of M1–M7 below names the
specific test(s) it reds, and every new case is named by at least one.

## Blast radius

Call-tree diff (`+`/`-` on the lines that moved):

```text
KnipClient.analyze / JscpdClient.scan / DependencyChecker.scanProject /
GitleaksClient.scan / GovulncheckClient.analyze / OpengrepClient.scan /
TrivyClient.scan / PythonDeadCodeClient.analyze
+   parse site only: analyzed: true            (clients/analysed-root.ts)
+   knip memo hit: analyzed: false             (a hit did not scan this call)
  fetchFreshProjectDiagnostics                 (clients/project-diagnostics/fresh-fetch.ts)
-     record(id, adapted, elapsedMs)  -> completed[] + analyzed[] unconditionally
+     record(id, adapted, elapsedMs, analysedRoot) -> analyzed[] only when true
-     knip/jscpd gate: analyzed === false
+     knip/jscpd gate: analyzed !== true       (markCold, unchanged channel)
-     FreshProjectDiagnosticsResult.completed  (deleted: no reader)
    formatFullMode                             (tools/lens-diagnostics.ts)
-       authoritativeRunnerIds = analyzed.filter(id => id !== "test-runner")
+       authoritativeRunnerIds = new Set(analyzed)
      mergeDiagnosticsWithWidgetSummaries      (filter unchanged)
        mode=full delivery
```

Dependents of the changed client result types: `fetchFreshProjectDiagnostics` and the
`runner-adapters/*` (which read `success`/`findings` only — no adapter constructs a
result or decides authority, verified by grep, see Class sweep). The added field is
optional on every interface, so `clients/runtime-session.ts`, `runtime-turn.ts` and the
cache writers that persist these results are source- and shape-compatible; a cached
result written by an older build simply lacks the flag and is therefore not
authoritative, which is the safe direction. Strict consumers of
`FreshProjectDiagnosticsResult`: `tools/lens-diagnostics.ts` (the only one) — its stub
literal lost `completed` in the same commit, and `npx tsc --noEmit` is the proof there
is no other constructor.

Hot path: `record()` runs at most nine times per `mode=full` call; no per-file or
per-spawn cost is added. No new I/O, no new spawn.

## Observability

`diagnostic-retained-unreconciled` (`clients/degradation-ledger.ts` kind, emitted from
`tools/lens-diagnostics.ts` via `recordDegradationOnce`) is the record that proves the
LSP arm in production: it fires when a confirmed result could not replace retained
widget state, which is exactly when full mode keeps serving the older finding. It is
bounded once per session per `<file>:lsp` subject and re-arms after
`resetDegradationLedger()`, and the assertion is in this diff
(`tests/tools/lens-diagnostics.test.ts`, "records an unreconciled confirmed result once
per session and re-arms after reset"):

```ts
const firstSession = getDegradationSummary().find(
    (group) => group.kind === "diagnostic-retained-unreconciled",
);
expect(firstSession?.count).toBe(1);

resetDegradationLedger();
await run(makeTool({}, lspService), { mode: "full" });
const secondSession = getDegradationSummary().find(
    (group) => group.kind === "diagnostic-retained-unreconciled",
);
expect(secondSession?.count).toBe(1);
```

Mutating both call sites to `incrementDegradationCount` reds it (`expected 2 to be 1`,
M8 below).

`runner_authoritative_widget_retire` (`tools/lens-diagnostics.ts`, via `logLatency`) is
the runner arm's equivalent, added in round 5 (v4 F2): the same phase shape as the LSP
sibling twelve lines above it, carrying the retired row count and the runner ids that
claimed authority. Bounded by construction — at most one row per `mode=full` call, and
only when rows were actually removed. Both directions are asserted in this diff
(`tests/tools/lens-diagnostics.test.ts`):

```ts
const phases = getRecentLoggedPhases().map((entry) => entry.phase);
expect(phases).toContain("runner_authoritative_widget_retire");
…
expect(phases).not.toContain("runner_authoritative_widget_retire");
```

`if (runnerRetiredRows > 0)` → `if (false)` reds the first (`expected [] to include
'runner_authoritative_widget_retire'`); → `if (true)` reds the second (M12a/M12b).

The `cold` channel remains the observable for a runner that did NOT run — the
`cold (not applicable / unavailable this run): … knip — not run (No project root found;
knip skipped)` line quoted from the real-`pi` probe below. No per-occurrence record is
added for that case: it would fire on every healthy `mode=full` call in any repo without
a Go module or a git dir (seven runners are cold in the harness fixture), which is the
unbounded-record shape AGENTS.md forbids.

## Class sweep

**Shape:** *a client returns `success: true` for a run that did not happen, and a caller
reads the absence of findings as a clean verdict.* Grep across the WHOLE tree, not just
`clients/` — `grep -rn "success: true" --include=*.ts clients/ tools/ mcp/ scripts/
tests/support/ index.ts` — plus a per-member verdict for the enumerable family (the nine
`record()` sites in `fresh-fetch.ts`; the full 9 × 8 table is under `## Round 4`).

| member | non-analysed `success: true` shape | verdict |
| --- | --- | --- |
| `clients/knip-client.ts` | no project root (`:437`), memo hit (`:457`), exit 0 with empty stdout (`:680`) | fixed — `analyzed: true` only at the parse return; memo hit clears it |
| `clients/jscpd-client.ts` | no source files (`:247`), report missing on exit 0 (`:383`) | fixed |
| `clients/dependency-checker.ts` | missing root, no top-level source file, madge unavailable, spawn error, parse throw — all `{circular: [], count: 0}` (`:1028-1111`) | fixed |
| `clients/gitleaks-client.ts` | no opt-in signal (`:360`), report missing after a crashed scan (`:418`) | fixed |
| `clients/trivy-client.ts` | not opted in (`:266`), report missing (`:333`) | fixed |
| `clients/govulncheck-client.ts` | no go.mod (`:442`) | fixed |
| `clients/opengrep-client.ts` | none — a missing report returns `success: false` | flag set at the parse site for uniformity |
| `clients/dead-code-client.ts` | no Python root (`:391`), vulture not installed (`:399`) | fixed; exit 0 with empty stdout IS vulture's documented clean signal (its own exit-code table) and is marked analysed |
| test-runner lane | cache read by design, no client | fixed at the record site (`false`, with the reason in a comment) |
| `clients/project-diagnostics/runner-adapters/*` | none — every adapter is `Result -> ProjectDiagnostic[]` and reads `success`/`findings` only; no adapter constructs a result or decides authority | clean, no change |
| `tools/`, `mcp/`, `scripts/`, `scripts/lib/`, `tests/support/`, `index.ts` | no other consumer of these result types decides retirement; `grep -rn "\.analyzed\b" --include=*.ts` outside `clients/` returns only `tools/lens-diagnostics.ts:2196` | clean |

**Consolidation verdict: fold onto one seam — done in this PR.** The nine members share
one definition (`clients/analysed-root.ts`, one interface, one doc) rather than nine
private booleans, and one reader (`record(..., analysedRoot)`). Deletion test: removing
`AnalysedRootSignal` would put the same contract paragraph in eight files and let the
field's meaning drift per client — concentration, not relocation, so the shared module
earns its keep. Helper net count: **+1 type module, −1 field (`completed`), −1 derived
list (the `test-runner` filter), −1 dead gate clause (`execution === "cache"`)**.

Adjacent sweep with a negative result, recorded because it was searched: `grep -rn
"\.completed\b" --include=*.ts tools/ clients/ index.ts` finds `tools/scan-progress.ts`,
`clients/lsp-mutation.ts`, `clients/actionable-warnings.ts`, `clients/pipeline.ts` and
`clients/review-graph/builder.ts` — all unrelated progress/telemetry counters, none of
them the `FreshProjectDiagnosticsResult` field this PR deletes.

**Where the nine are NOT uniform, stated rather than implied** (v4 T1/T2, both recorded
as remainders under `## Round 5`, neither fixed here):

- The `analysed` signal is uniform; the **cache write** is not. knip and jscpd return
  before `cacheManager.writeCache` when they did not analyse, so a run that never
  happened no longer poisons the durable scanner cache. The other seven still write
  their empty result. That is `origin/master` behaviour, untouched by this PR, and
  changing it is a durable-store semantics change for consumers outside `mode=full`.
- The `dead-code` id is shared by every language client in the lane, so authority is
  per-lane, not per-language. Latent while `PythonDeadCodeClient` is the only member.

The `grep -c "if (result.error)"` sweep over the eight runner clients — the discriminator
that decides whether a spawn even produced something to parse — now finds the check in
every one of them (`dependency-checker.ts` twice, once per spawn path; the rest once).
`govulncheck` was the only `0` (v4 F1, fixed in round 5).

### The real writers table (written before any round-4 edit)

Axis: every site in `clients/project-diagnostics/fresh-fetch.ts` that can push a
runner id into the authority set (the nine `record(...)` calls) × the shape of the
client result that reaches it. "Authoritative" means the id enters
`FreshProjectDiagnosticsResult.analyzed`, which `tools/lens-diagnostics.ts` turns into
`authoritativeRunnerIds` and uses to retire retained widget rows.

Legend: every cell is either a test title that greps in this repo after this round
(`grep -rn 'it("<title>' tests/`), or `n/a — <reason with the code that makes the
shape unreachable at this site>`. No other ids appear in this table.
`A` = authoritative, `—` = not authoritative.

| record site (fresh-fetch.ts) | ran-and-found | ran-clean | skipped-no-root | crashed-but-success:true | cache hit | empty-file-set short-circuit | timeout | disabled |
| --- | --- | --- | --- | --- | --- | --- | --- | --- |
| `record("knip")` :365 | **A** — `marks the knip result analysed only when it parsed knip output` | **A** — `marks knip analysed when the real client parses a clean knip report` | **—** — `keeps knip out of the analysed set when the real client finds no project root` | **—** (exit 0, empty stdout) — `marks the knip result analysed only when it parsed knip output` | **—** — `does not mark a knip memo hit as analysed` | n/a — knip scans the resolved root; no file-set gate exists in `KnipClient.analyze` | **—** never reaches `record` (`success:false` → `recordFailed`) — `reports a failed knip run and does not cache it (#925)` | n/a — knip has no per-runner disable gate; the whole-fetch refusal is `refuses to run anything when cwd IS the home directory (#747)` |
| `record("jscpd")` :411 | **A** — `marks the jscpd result analysed only when it parsed a report` | **A** — `marks the jscpd result analysed only when it parsed a report` | n/a — a missing target dir returns `success:false` (`jscpd-client.ts:245`), never `record` | **—** (exit 0, no report) — `marks the jscpd result analysed only when it parsed a report` | n/a — `JscpdClient.scan` has no memo arm | **—** — `keeps jscpd out of the analysed set when the real client finds no source files` | **—** never reaches `record` (`success:false`) — `marks the jscpd result analysed only when it parsed a report` | **—** cold before the call — `reports jscpd cold when the tool isn't available, without writing cache` |
| `record("madge")` :436 | **A** — `marks the madge scan analysed only when it parsed the dependency graph` | **A** — `marks the madge scan analysed only when it parsed the dependency graph` | **—** — `keeps madge out of the analysed set when the real dependency checker skips the root` | **—** (spawn error / parse throw) — `marks the madge scan analysed only when it parsed the dependency graph` | n/a — `scanInFlight` shares a live run, not a stored result (`dependency-checker.ts:1047`) | **—** (no top-level source file) — `keeps madge out of the analysed set when the real dependency checker skips the root` | **—** same shape as crashed (`result.error` at `dependency-checker.ts:1084`) — `marks the madge scan analysed only when it parsed the dependency graph` | **—** cold before the call — `captures the SPECIFIC reason each cold analyzer was skipped, not just its id (#1623)` |
| `record("gitleaks")` :479 | **A** — `marks gitleaks analysed when the real scan parses a report` | **A** — `marks the gitleaks result analysed only when it parsed a report` | **—** (no opt-in signal; `requireSignal:true` callers only) — `marks the gitleaks result analysed only when it parsed a report` | **—** (report missing after a crashed scan) — `keeps gitleaks out of the analysed set when the real scan produces no report` | n/a — `dedupeScan` shares a live run only (`security-scan-client.ts:392`) | n/a — gitleaks walks the root itself; no file-set gate | **—** never reaches `record` (`result.error` → `success:false`) — `marks the gitleaks result analysed only when it parsed a report` | **—** cold before the call — `gates gitleaks on hasGitRepo (#608): cold when tmp isn't a git repo at all` |
| `record("govulncheck")` :523 | **A** — `marks the govulncheck result analysed only when it parsed the scan stream` | **A** — `marks the govulncheck result analysed only when it parsed the scan stream` | **—** (no go.mod) — `marks the govulncheck result analysed only when it parsed the scan stream` | **—** (exit 0 with an empty stream — **corrected in round 5**, v4 F1) — `keeps govulncheck out of the analysed set when the scan produced no output` | n/a — `dedupeScan` shares a live run only | n/a — no file-set gate | **—** (`{status: null, error, stdout: <partial>}` → the process check — **corrected in round 5**, v4 F1) — `does not mark an aborted govulncheck with partial output as analysed` | **—** cold before the call — `gates govulncheck/trivy on their own static signals, without cache reads` |
| `record("opengrep")` :565 | **A** — `marks the opengrep result analysed only when it parsed a report` | **A** — `marks the opengrep result analysed only when it parsed a report` | n/a — opengrep has no project-type gate (`fresh-fetch.ts` comment at the task) | n/a — a missing report returns `success:false` (`opengrep-client.ts:196`) | n/a — `dedupeScan` shares a live run only | n/a — no file-set gate | **—** never reaches `record` (`success:false`) — `marks the opengrep result analysed only when it parsed a report` | **—** cold before the call — `reports opengrep cold (not clean) when the tool isn't available (#585)` |
| `record("trivy")` :601 | **A** — `marks the trivy result analysed only when it parsed a report` | **A** — `marks the trivy result analysed only when it parsed a report` | **—** (not opted in / no manifest) — `marks the trivy result analysed only when it parsed a report` | **—** (report missing after a crashed scan) — `keeps trivy out of the analysed set when the real scan produces no report` | n/a — `dedupeScan` shares a live run only | n/a — no file-set gate | **—** never reaches `record` (`result.error` → `success:false`) — `marks the trivy result analysed only when it parsed a report` | **—** cold before the call — `runs trivy fresh only when opted-in AND a dependency manifest exists` |
| `record("dead-code")` :635 | **A** — `marks the dead-code result analysed only when vulture ran over the root` | **A** (exit 0, empty stdout is vulture's documented clean signal) — `marks the dead-code result analysed only when vulture ran over the root` | **—** (no Python project root) — `marks the dead-code result analysed only when vulture ran over the root` | **—** (vulture not installed → `success:true`) — `keeps dead-code out of the analysed set when the real client is unavailable` | n/a — `inFlight` shares a live run only (`dead-code-client.ts:404`) | n/a — the language `detect()` gate is the file-set gate and precedes the call | **—** never reaches `record` (`spawnFailedWithNoOutput` → `success:false`) — `marks the dead-code result analysed only when vulture ran over the root` | **—** cold before the call — `runs every applicable dead-code language client and reports 'dead-code' cold only when none apply` |
| `record("test-runner")` :693 | **—** cache read by design, never a run this call — `keeps test-runner out of the analysed set on a cache read` | **—** same | n/a — no root gate | n/a — never spawns | **—** the only shape it has — `keeps test-runner out of the analysed set on a cache read` | n/a | n/a — never spawns | **—** no cache entry → cold — `reports test-runner cold (not clean) when no turn_end cache exists yet (#1004)` |

Reading of the table: the only cells that may be authoritative are the ones where a
client parsed the output of a scan **it performed over this root during this call**.
That is the opt-in signal this round introduces (`analyzed: true`, set at each client's
parse site), replacing round 3's opt-out `analyzed === false` flag, which only two of
the nine clients ever set.
### Round 4 — per-finding disposition

| # | finding (v3 review / round-4 brief) | disposition |
| --- | --- | --- |
| 1 (blocker, v3 S1) | authority gate is opt-OUT; seven runners ungated; crashed gitleaks and never-run madge retire real findings | **fixed** — opt-in `AnalysedRootSignal` set at each client's parse site, audited across all nine clients + `runner-adapters/*` (table above and Class sweep). Regression tests drive the REAL clients; red on round-3 code quoted whole-file above. |
| 2 (v3 T1) | `FreshProjectDiagnosticsResult.completed` has no reader | **deleted** (field, both writers, the stub literal in `tools/lens-diagnostics.ts`, and its only test assertion). `grep -rn "\.completed\b"` re-run after deletion: only unrelated progress counters remain (quoted in Class sweep). |
| 3 (v3 I2) | the round-3 `reconciled === false` test is vacuous | **deleted**, with the probe that shows why it cannot be made red at that seam and the mutation that proves the guard itself is live (M7). See Test assessment. |
| 4 (v3 I3 / v2 N4) | the first `## Observability` section named a deleted `provenance` record | **fixed** — the body was rewritten this round; there is now exactly ONE Observability section, it names `diagnostic-retained-unreconciled`, and `node scripts/check-pr-body.mjs --lint-local` exits 0 on this body (transcript below). |
| 5 (v3 S2) | AC2 unmet while the title and changelog claim it | **retitled and re-fragmented** to what ships; remainder (AC2 + AC1's second-session test) posted on #2154. Rationale in Summary. |
| 6 | merge order vs #2875 | `origin/master` merged (already up to date at `6fea2250e`); interaction described below. |
| v3 T2 (minor) | the comment at the acceptance guard said the opposite of the code | **fixed** — the comment now states the strict `=== true` rule it implements. |
| v3 T3 (minor) | one unfaithful `reconcileScanDiagnostics` double | **fixed** — `tests/tools/lsp-diagnostics-cache.test.ts` returns `true`; all eight doubles are now production-faithful. |
| v3 S3 (minor) | no second-session/worktree test | **not fixed**, named in the remainder on #2154 (the harness cannot express two children on one project root). |

### Round 4 mutation transcript

Every mutation is compile-valid, applied to the BUILT output, run, then reverted with
`npm run build`. Each names the test(s) it reds.

| # | mutation | file(s) run whole | result |
| --- | --- | --- | --- |
| M1a | `record()`: `if (analysedRoot)` → `if (true)` (round-3 behaviour) | `runner-authority` | **red ×5** — madge / gitleaks / trivy / dead-code / test-runner "keeps … out of the analysed set" |
| M1b | the inverse: `if (analysedRoot)` → `if (false)` | `runner-authority`, `lens-diagnostics` | **red ×2** — `marks knip analysed when the real client parses a clean knip report`, `marks gitleaks analysed when the real scan parses a report` |
| M2/M3 | knip + jscpd gates: `result.analyzed !== true` → `=== false` (exactly round 3's opt-out) | `runner-authority`, `fresh-fetch` | **red ×2** — `keeps knip out of the analysed set when the real client finds no project root`, `keeps jscpd out of the analysed set when the real client finds no source files` |
| M4 | knip memo hit: drop `analyzed: false` from the `execution: "cache"` return | `runner-authority`, `knip-client` | **red** — `does not mark a knip memo hit as analysed` (`expected true not to be true`) |
| M5 | every client's parse-site `analyzed: true` → `analyzed: false` (8 files, 10 sites) | `runner-authority` | **red ×11** — one per client plus both positive controls |
| M6a | the retirement filter → `.filter(() => true)` | `lens-diagnostics` | **red** — `retires only the analysed runner's retained row` (`expected 'src/stale.ts 2W …' not to contain 'stale runner finding'`) |
| M6b | the inverse: `.filter(() => false)` (retire everything) | `lens-diagnostics` | **red ×6** — including the survival direction added this round and `a clean fully-covered sweep RETIRES stale widget blockers for the file (#1993)` |
| M6c | `new Set(extracted.analyzed ?? [])` → `new Set()` | `lens-diagnostics` | **red** — same test, R direction |
| M7 | `tools/lsp-diagnostics.ts` reconciliation guard → `if (false)` | `lsp-diagnostics`, `lsp-diagnostics-inline-blocker-retire` | **red** — `F5: a rejected reconcile does not retire an inline blocker` (`expected [ { …(3) } ] to deeply equal []`). This is the mutation v3 found green against round 3's own test; the live coverage is round 2's F5 test, and the vacuous test is deleted. |
| M8 (prior round, re-run on this head) | both `recordDegradationOnce` → `incrementDegradationCount` | `lens-diagnostics` | **red** — `expected 2 to be 1` at `records an unreconciled confirmed result once per session and re-arms after reset` |
| M9 (prior round, re-run) | `markUnreconciledFindings` returns `{...finding}` | `lens-diagnostics` | **red** — `[stale — re-run to confirm]` missing |
| M10 (prior round, re-run) | `if (retired === true)` → `if (retired !== false)` | `lens-diagnostics` | **red** — `expected 2 to be 1` |

M1a verbatim:

```text
    function record(id, adapted, elapsedMs, analysedRoot) {
        if (true)
            pushUnique(analyzed, id);
     × keeps madge out of the analysed set when the real dependency checker skips the root 4ms
     × keeps gitleaks out of the analysed set when the real scan produces no report 2ms
     × keeps trivy out of the analysed set when the real scan produces no report 3ms
     × keeps dead-code out of the analysed set when the real client is unavailable 1ms
     × keeps test-runner out of the analysed set on a cache read 2ms
      Tests  5 failed | 13 passed (18)
```

M6b verbatim (the over-correction direction — the one that silently deletes findings):

```text
            .filter(() => false)
     × retires only the analysed runner's retained row 11ms
     × runs workspace diagnostics and merges LSP-only files with widget state 7ms
     × honors inline `# pi-lens-ignore` like mode=all (#442) 2ms
     × a clean fully-covered sweep RETIRES stale widget blockers for the file (#1993) 2ms
     × does not call a lower-order clean result authoritative (#2154) 1ms
     × mode=full: an explicitly named ignored file remains visible to an audit 3ms
```

M7 verbatim:

```text
        const reconciled = reconcileScanDiagnostics(file, retagged, true, writeIndex, observedAt);
        if (false)
            return { confirmed: false, blocking: true };
…
    527|   expect(runtime.getInlineBlockersSnapshot()).toHaveLength(1);
 Test Files  1 failed | 1 passed (2)
      Tests  1 failed | 69 passed (70)
```

### Round 4 probes

**Through the real registered tool against real `pi` 0.85.1** (temporary fixture +
probe test, both deleted afterwards; `PI_LENS_HOME` pinned to
`<worktree>/.probe-home`, `PILENS_DATA_DIR` never set). Five `lens_diagnostics` calls;
between call 1 and call 2 `package.json`, `knip.json` and `tsconfig.json` are renamed
aside so knip no-ops, **without touching `src/moved.ts`**; before call 5 they are
restored AND the export is deleted with a three-line shift:

```text
===== 1 full (markers present) =====
src/moved.ts  1W  cwd=/tmp/pi-lens-scratch/real-pi-probe-r4-project-…
  L41: Unused export staleExport [knip:export]
Summary (1 files diagnosed this session):
  1 warning

===== 2 full (markers removed) =====
src/moved.ts  1W  cwd=/tmp/pi-lens-scratch/real-pi-probe-r4-project-…
  L41: Unused export staleExport [knip:export]
Summary (1 files diagnosed this session):
  1 warning
…
cold (not applicable / unavailable this run): … knip — not run (No project root found; knip skipped), …

===== 3 delta =====
No issues in the current turn delta. 1 finding across 1 file carried over from earlier this session — use mode=all to see them.

===== 4 all =====
  L41: Unused export staleExport [knip:export]
  1 warning

===== 5 full (markers restored, export deleted) =====
No files diagnosed yet this session.
…
fetched fresh this call: knip (259ms), jscpd (651ms).
```

Retention when the runner did not run, retirement when it did, and `full`/`delta`/`all`
agreeing throughout — both directions, no over-correction.

**Client-seam re-run of the two v3 exemplars** (`probes-2868r4/gitleaks-madge-hole.mjs`,
kept OUTSIDE the worktree, real clients + real `fetchFreshProjectDiagnostics`, a PATH
shim `gitleaks` that answers `version` and then exits 0 with a panic on stderr):

```text
REAL GitleaksClient.scan: {"success":true,"findings":[],"scannedAt":"2026-09-10T13:38:10.683Z","summary":"panic: runtime error"}
gitleaks analyzed : []
gitleaks cold     : ["govulncheck","trivy","dead-code","test-runner","jscpd","madge","opengrep"]
gitleaks completed: undefined
REAL DepChecker.scanProject: {"circular":[],"count":0}
madge analyzed    : []
```

v3 quoted `analyzed: ["gitleaks"]` and `analyzed: ["madge"]` for the identical inputs.
`completed: undefined` is finding 2's field, gone.

**Why these two are probed at the client seam and not through real `pi`:** neither
binary exists on this box — the real-`pi` transcript above shows `madge — not run (madge
binary unavailable)` and `gitleaks — not run (not a git repository (no .git found))` —
so neither runner can produce a retained finding in the harness fixture to retire. The
full chain from a client return to a deleted row is covered end-to-end for knip (above),
and for gitleaks/madge across two seams: client → `analyzed` (the probe and
`runner-authority.test.ts`) and `analyzed` → retirement (`retires only the analysed
runner's retained row`, both directions, M6a/M6b/M6c).

**Body lint:**

```text
$ node scripts/check-pr-body.mjs --lint-local PR_BODY.md
$ echo $?
0
```

### Round 4 governance and hygiene

- **No governance exemption added this round.** No `DECLARED_EXCEPTIONS`, no
  `tests/support/flake-shape-baseline.json` change, no `vi-mock-export-baseline.json`
  row: the new suite's `vi.mock` factory spreads `await importOriginal()`, which the
  gate treats as a complete pass-through (`vi-mock-export-sweep: 0 warning row(s)`).
  Worth knowing for the next author: the gate does NOT recognise the generic form
  `importOriginal<T>()` — it read that factory as dropping 18 exports until the type
  argument was moved to a cast.
- Probe hygiene: every probe ran with `PI_LENS_HOME=<worktree>/.probe-home` and
  `PILENS_DATA_DIR` unset; `find ~/.pi-lens -newermt "-15 minutes"` is empty. Two stray
  writes that the first draft of the new test caused inside the worktree (an
  `args[i+1]` write for a flag the spawned command did not carry, which landed on
  `config/`, `bin/`, `global/` and a file literally named `--version`) were removed and
  the doubles hardened so a report is written only where the command was told to put
  one — the guard is in the test file's `reportWriter` doc comment.
- The temporary real-harness probe fixture and probe test were deleted; `tests/config/`
  and the flake-shape ratchet were re-run green afterwards (52 files, 490 passed).

### Round 4 merge order

`origin/master` is merged (already up to date at `6fea2250e`; the PR is MERGEABLE).
**#2875** (`fix/2864-2862-fold-and-footer-residuals`, open) touches
`tools/lens-diagnostics.ts`, `tools/lsp-diagnostics.ts`, `tests/tools/lens-diagnostics.test.ts`
and `index.ts`. Diffing its hunks against mine:

- `tools/lens-diagnostics.ts` — #2875 edits the import block (`@@ -102,7`), the
  per-file constant (`@@ -119,7`), the schema/tool factory (`@@ -315`, `@@ -486`,
  `@@ -529`) and `formatDeltaMode` (`@@ -1094`, `@@ -1110`, `@@ -1129`, `@@ -1386`).
  Mine are all in `formatFullMode` and `mergeDiagnosticsWithWidgetSummaries`
  (`~1707`, `~2136`, `~2188`, `~2246`). **Only the import block can conflict textually**
  (both PRs add imports in the same region); the resolution is additive.
- `tests/tools/lens-diagnostics.test.ts` — #2875's hunks are at `165`, `181`, `299`,
  `463`, `667`, `3234`, `3588`; mine is the single `mode=full` retirement test at
  `~937`. No overlap, but whoever lands second should re-run the whole file, because
  #2875 changes the shared `run()` helper region.
- `tools/lsp-diagnostics.ts` — #2875 touches only the import block (`@@ -61,19`); my
  round-2 change is at `~980`. No overlap.

Semantic check if #2875 lands first: it changes delta rendering and the tool schema,
not who is authoritative for a retirement, so the merged behaviour is the union. The
merged head still needs a fresh CI run either way.

## Round 5

Verify round 4 confirmed every v3 finding fixed and re-proved each with its own probes,
and found one runner the round-4 audit got wrong. Merged `origin/master` first (#2865,
#2869, #2879); the merge is clean.

### Round 5 — per-finding disposition

| # | finding (v4) | disposition |
| --- | --- | --- |
| F1 (blocker) | `govulncheck` sets `analyzed: true` for a scan that parsed nothing, and it deletes a real CVE finding through the real tool | **fixed** in `clients/govulncheck-client.ts`. Reproduced first, through the real registered tool, before any edit (transcript below). |
| F1b (mine, found by the same probe) | a retired row left its counts behind — `main.go  1W … 1 warning` over no row at all | **fixed**: the merge recomputes the file summary when, and only when, the filter actually dropped rows. |
| F2 (major) | the runner-retirement filter emits nothing while its LSP sibling logs `lsp_authoritative_widget_retire` under "retirement must be observable" | **fixed**: `runner_authoritative_widget_retire`, same phase shape, one row per call, only when rows were removed. Named in the (single) Observability section above. |
| T1 (minor) | the `dead-code` lane is id-granular across multiple language clients | **recorded as a remainder**, not fixed — see below. |
| T2 (minor) | the cache write is gated for knip/jscpd and not for the other seven | **recorded as a remainder**, not fixed — see below. The Class sweep above now says so instead of presenting the nine as uniform. |

### Round 5 — F1, reproduced before it was fixed

I did not take the finding on trust. Temporary fixture (`go.mod` + `main.go`) and probe
test, both deleted afterwards; a PATH shim `govulncheck` that answers `-version`, emits a
real OSV+finding stream on its first scan (exit 3), **exits 0 with empty stdout** on its
second, and on its third emits a real stream with the finding gone. Real `pi` 0.85.1,
`PI_LENS_HOME` pinned, `PILENS_DATA_DIR` never set.

**Before (head `42ee9b239`, the round-4 head the verify ran on):**

```text
===== 1 full (shim reports GO-2024-9999) =====
main.go  1W  cwd=/tmp/pi-lens-scratch/real-pi-probe-r5gvc-project-…
  L6: Vulnerability GO-2024-9999: probe vuln (fixed in 1.2.3) [govulncheck:GO-2024-9999]
Summary (1 files diagnosed this session):
  1 warning

===== 2 full (shim exits 0 with NO output) =====
main.go  1W  cwd=/tmp/pi-lens-scratch/real-pi-probe-r5gvc-project-…
Summary (1 files diagnosed this session):
  1 warning

===== 3 all =====
No issues across 1 file diagnosed this session. ✓
```

The CVE row is deleted by a govulncheck that scanned nothing, and `mode=all` calls the
session clean — v4 F1 exactly. Note the second symptom in call 2, which the verify also
flagged: the header still says `1W` / `1 warning` over no row (F1b).

**After (this head):**

```text
===== 1 full (shim reports GO-2024-9999) =====
main.go  1W  …  L6: Vulnerability GO-2024-9999: probe vuln (fixed in 1.2.3) [govulncheck:GO-2024-9999]

===== 2 full (shim exits 0 with NO output) =====
main.go  1W  …  L6: Vulnerability GO-2024-9999: probe vuln (fixed in 1.2.3) [govulncheck:GO-2024-9999]

===== 3 all =====
main.go  1W  …  L6: Vulnerability GO-2024-9999: probe vuln (fixed in 1.2.3) [govulncheck:GO-2024-9999]
Summary (1 files diagnosed this session):
  1 warning

===== 4 full (shim really scans, CVE gone) =====
main.go  cwd=/tmp/pi-lens-scratch/real-pi-probe-r5gvc-project-…
Summary (1 files diagnosed this session):
```

The finding survives the scan that did not happen, and **still retires** on call 4, when
govulncheck genuinely ran and no longer reports it — no over-correction.

### Round 5 — the fix, and where the prescribed remedy needed more

The verify's prescription was "check `result.error` first like the other seven clients and
use `spawnFailedWithNoOutput` instead of the local `status !== 0 && status !== 3` test".
I built the outcome table for `runScan` before applying it:

| shape | `result.error` | status | stdout | correct verdict | old code | `spawnFailedWithNoOutput` ALONE |
| --- | --- | --- | --- | --- | --- | --- |
| ran, no vulns | null | 0 | stream | analysed | ✓ | ✓ |
| ran, vulns found | null | 3 | stream | analysed | ✓ | ✓ |
| config/load error | null | 1 or 2 | empty | failure | ✓ | ✓ |
| spawn error (ENOENT) | Error | null | empty | failure | **analysed** | ✓ |
| **timeout / ambient abort** | Error | null | **partial** | failure | **analysed** | **still analysed** — its doc is explicit that the `!output.trim()` conjunct exists so "a spawn error that still produced partial output keeps reaching the parser" |
| **exit 0, empty stdout** | null | 0 | empty | not analysed | **analysed** | **still analysed** — `classifyRunOutcome` calls status 0 a run |
| exit 3, empty stdout | null | 3 | empty | failure | **analysed** | ✓ |

So the prescription's two halves are both needed and neither is sufficient alone: the
`result.error` check is what excludes the partial-stdout abort, and `spawnFailedWithNoOutput`
still needs the explicit "nothing to parse" arm for exit 0. Shipped:

```ts
if (result.error) { … return failure }                       // the seven siblings' order
const rawStdout = result.stdout ?? "";
if (!rawStdout.trim()) {                                     // never evidence of "no vulns"
    if (spawnFailedWithNoOutput(result, rawStdout)) return failure(firstOutputLine(stderr));
    return { ...EMPTY_RESULT, success: true, scannedAt, summary: "govulncheck produced no output; nothing parsed" };
}
const findings = parseGovulncheckJson(rawStdout);
return { success: true, analyzed: true, findings, scannedAt };
```

`clients/govulncheck-client.ts` is the only client that changed. The class sweep is the
verify's own `grep -c "if (result.error)"`: it now reads `1` for all eight.

### Round 5 tests, red first

`tests/clients/project-diagnostics/runner-authority.test.ts` — the govulncheck positive
control enshrined the hole (it drove `{status: 0, stdout: ""}` and asserted
`analyzed === true`). It now feeds a **real** stream (config + osv + finding records,
exit 3 — the shapes `tests/clients/govulncheck-client.test.ts` already pins against the
tool's documented JSON protocol, not invented here) and asserts the parsed OSV id; the
empty-stdout and aborted-with-partial-output shapes are asserted as **not** analysed, and
a new case takes the exit-0 arm through the real `fetchFreshProjectDiagnostics`.

Whole file, on the round-4 code:

```text
 ❯ |default| tests/clients/project-diagnostics/runner-authority.test.ts (20 tests | 3 failed) 42ms
     × keeps govulncheck out of the analysed set when the scan produced no output 4ms
     × marks the govulncheck result analysed only when it parsed the scan stream 1ms
     × does not mark an aborted govulncheck with partial output as analysed 1ms

 FAIL … > keeps govulncheck out of the analysed set when the scan produced no output
AssertionError: expected [ 'govulncheck' ] to not include 'govulncheck'
 FAIL … > marks the govulncheck result analysed only when it parsed the scan stream
AssertionError: expected true not to be true // Object.is equality
 FAIL … > does not mark an aborted govulncheck with partial output as analysed
AssertionError: expected true not to be true // Object.is equality

 Test Files  1 failed (1)
      Tests  3 failed | 17 passed (20)
```

`tests/tools/lens-diagnostics.test.ts` — `retires only the analysed runner's retained row`
gained the count assertion (F1b), and two new cases pin the record and its bound.
`tests/support/check-build-freshness.ts` refuses to run a suite against a stale build, so
the pre-fix red for these two source-side changes is the mutation of the built output
below — the same revert, quoted.

### Round 5 mutation transcript

| # | mutation | file run whole | result |
| --- | --- | --- | --- |
| M11 | F1b reverted: `diagnostics.length === retained.length` → `true` (always keep the stored counts) | `lens-diagnostics` | **red** — `retires only the analysed runner's retained row`: `expected 'src/stale.ts  2W  cwd=/proj/src\n  L9…' to contain 'src/stale.ts  1W'` |
| M12a | `if (runnerRetiredRows > 0)` → `if (false)` (the round-4 state: no record) | `lens-diagnostics` | **red** — `logs one bounded phase row when a runner retirement removes rows`: `expected [] to include 'runner_authoritative_widget_retire'` |
| M12b | the inverse, `if (true)` (a row on every call — the unbounded direction) | `lens-diagnostics` | **red** — `logs no runner retirement row when nothing was retired`: `expected [ Array(1) ] to not include 'runner_authoritative_widget_retire'` |
| M13 | `if (result.error)` → `if (false)` (the prescription's first half removed) | `runner-authority` | **red** — `does not mark an aborted govulncheck with partial output as analysed` |
| M14a | `if (!rawStdout.trim())` → `if (false)` (the empty-stream arm removed) | `runner-authority` | **red ×2** — `keeps govulncheck out of the analysed set when the scan produced no output`, `marks the govulncheck result analysed only when it parsed the scan stream` |
| M14b | the inverse, `if (true)` (nothing is ever parsed) | `runner-authority` | **red** — `marks the govulncheck result analysed only when it parsed the scan stream`: `expected undefined to be true` |

Every mutation is compile-valid, applied to the built output, and reverted with
`npm run build`. The round-4 mutation set (M1a/M1b, M2/M3, M4, M5, M6a–c, M7, M8–M10) was
not re-run this round: the verify re-ran it on `42ee9b239` and every guard still redded,
and round 5 touches none of those lines.

### Round 5 test totals

| suite | result |
| --- | --- |
| `tests/tools/lens-diagnostics*`, `tests/clients/project-diagnostics`, `tests/clients/govulncheck-client.test.ts`, `tests/real-harness/diagnostic-provenance.test.ts` | 15 files, **286 passed** |
| every `tests/tools/lsp-diagnostics*` + `tests/clients/security-scan-client.test.ts` | 6 files, **89 passed** |
| all of `tests/config/` (52 files) + `flake-shape-ratchet` + `generation-guard-sweep` + `bounded-telemetry-sweep` | 55 files, **509 passed \| 1 skipped** |
| `npx tsc --noEmit` | exit 0 |
| `npm run fmt:check` | `All matched files use the correct format` (1808 files) |

`bounded-telemetry-sweep` is run explicitly because round 5 adds a `logLatency` row — it
is the directory-scanning suite that owns that shape, and it is green.

### Round 5 — named remainders (v4 T1 / T2)

Neither is fixed here; both are recorded so they are not mistaken for handled.

1. **T1 — `dead-code` authority is id-granular across language clients.**
   `clients/project-diagnostics/fresh-fetch.ts` loops the applicable dead-code clients and
   calls `record("dead-code", …, result.analyzed === true)` per client, and `pushUnique`
   means one language client that analysed makes the shared `dead-code` id authoritative
   for every dead-code row — including rows from a language client that did not run.
   Latent today: `getDeadCodeClients()` returns only `PythonDeadCodeClient`. The durable
   answer is the per-file coverage set named below, not a per-language id.
2. **T2 — the cache write is gated for knip and jscpd only.** Those two return before
   `cacheManager.writeCache` when `analyzed !== true`, so a run that did not happen no
   longer poisons the durable scanner cache. The other seven still write an empty result
   from a scan that never ran. That is `origin/master` behaviour, unchanged by this PR,
   and it is a durable-store semantics change for consumers outside `mode=full`
   (`runtime-session.ts`, `runtime-turn.ts` read those caches) — seven call sites, not one
   line, so it does not ride along in a fix round. The Class sweep above now states it
   rather than presenting the nine as uniformly handled.

Both belong with the verify's own named output — **id-granular runner authority is the
wrong grain**: `authoritativeRunnerIds` answers "which runners ran", never "which files
did this runner's scan cover", while the LSP arm beside it already keeps
`authoritativeLspFiles`. A runner that legitimately ran but skipped a file (ignore
patterns, `--skip-dirs`, a `minLines` floor) still retires that file's row. T1, T2, v4 F1,
v3 S1 and v2 N1 are all the same missing answer, and it wants its own issue.

### Round 5 — merge order

`origin/master` merged clean (#2865, #2869, #2879 landed since round 4; no conflicts).
**#2875** remains open on the same three files; the verify's `git merge-tree --write-tree`
against it produced a clean tree at round 4, and round 5 adds lines only inside
`formatFullMode` / `mergeDiagnosticsWithWidgetSummaries` and one client, so the disjoint
regions still hold. Whoever lands second re-runs `tests/tools/lens-diagnostics.test.ts`
whole, because #2875 edits the shared `run()` helper region.

## Previous rounds (condensed)

- **Round 1** — shipped provenance fields on the cache record and one acceptance-guard
  change. Review: the reported defect still reproduced through the real tool (F1), the
  rejected path rendered a retained finding as a current blocker (F2), AC1's tests did
  not exist (F3), the provenance was dead plumbing (F4), four test doubles were
  unfaithful (F6), the body misreported a red CI check (F7).
- **Round 2** — deleted the dead provenance, made the widget write return its
  acceptance, added `markUnreconciledFindings` (stale label instead of a blocking row),
  added the real-`pi` regression scenario, migrated the doubles. F1/F2/F4/F5/F6/F9
  confirmed fixed. New blocker: a runner that did not run became authoritative and
  silently retired a real finding (N1); three retirement mechanisms, two dead (N2); the
  new degradation record had no test (N3).
- **Round 3** — one retirement expression (N2 fixed), the degradation-record test (N3
  fixed), and an opt-OUT `analyzed` flag on knip and jscpd (N1 fixed for those two).
  Verify found the flag left the other seven runners authoritative by default (S1), a
  72-cell table naming tests that did not exist (I1), a fabricated mutation red (I2),
  and the stale Observability section (I3).
- **Round 4** — the table above is built from the code, every cell names a test that
  exists, the gate is opt-in across all nine runners, and the three false claims are
  removed rather than restated. Verify confirmed every v3 finding fixed, re-proved each
  with its own probes, and found one runner the audit got wrong (F1: govulncheck) plus
  the missing runner-retirement record (F2).
- **Round 5** — this round: govulncheck's parse site now requires a stream it actually
  parsed, the retired row loses its counts with it, and runner retirement is observable.
  Reproduced F1 through the real tool before fixing it; both remaining minors are
  recorded as named remainders rather than silently carried.

## Round 6

Operating rule: only exit codes 0 and 3 may make a govulncheck stdout stream authoritative.

Kept: rounds 4–5 fixes, including empty-output and abort handling, runner-retirement observability, and retained-tally recomputation, remain unchanged.

### Round 6 — runScan outcome matrix

Each cell names the client-seam test that pins the required retention behavior.

| stdout content | exit 0 | exit 3 | exit 1 | other non-zero | abort-timeout |
| --- | --- | --- | --- | --- | --- |
| empty | `marks the govulncheck result analysed only when it parsed the scan stream`: success, not analysed, 0 findings; retain | `marks the govulncheck result analysed only when it parsed the scan stream`: failure, not analysed, 0 findings; retain | `does not mark a non-zero govulncheck with config-only output as analysed`: failure, not analysed, 0 findings; retain | `does not mark a non-zero govulncheck with config-only output as analysed`: failure, not analysed, 0 findings; retain | `does not mark an aborted govulncheck with partial output as analysed`: failure, not analysed, 0 findings; retain |
| config-record only | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, 0 findings; retire only authoritative rows | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, 0 findings; retire only authoritative rows | `does not mark a non-zero govulncheck with config-only output as analysed`: failure, not analysed, 0 findings; retain | `does not mark a non-zero govulncheck with config-only output as analysed`: failure, not analysed, 0 findings; retain | `does not mark an aborted govulncheck with partial output as analysed`: failure, not analysed, 0 findings; retain |
| config+progress | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, 0 findings; retire only authoritative rows | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, 0 findings; retire only authoritative rows | `does not mark a non-zero govulncheck with config-and-progress output as analysed`: failure, not analysed, 0 findings; retain | `does not mark a non-zero govulncheck with config-and-progress output as analysed`: failure, not analysed, 0 findings; retain | `does not mark an aborted govulncheck with partial output as analysed`: failure, not analysed, 0 findings; retain |
| complete OSV stream | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, findings; retire matching authoritative rows | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, findings; retire matching authoritative rows | `does not mark a non-zero govulncheck with config-and-progress output as analysed`: failure, not analysed, 0 findings; retain | `does not mark a non-zero govulncheck with config-and-progress output as analysed`: failure, not analysed, 0 findings; retain | `does not mark an aborted govulncheck with partial output as analysed`: failure, not analysed, 0 findings; retain |
| truncated after an OSV record | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, findings if parser recovers; retire only matching authoritative rows | `marks the govulncheck result analysed only when it parsed the scan stream`: success, analysed, findings if parser recovers; retire only matching authoritative rows | `does not mark a truncated non-zero govulncheck stream as analysed`: failure, not analysed, 0 findings; retain | `does not mark a truncated non-zero govulncheck stream as analysed`: failure, not analysed, 0 findings; retain | `does not mark an aborted govulncheck with partial output as analysed`: failure, not analysed, 0 findings; retain |

The “nothing to parse” exit-0 arm intentionally returns `success: true` for compatibility with the existing consumer contract, while omitting `analyzed` keeps it out of fresh-fetch authority.

### Round 6 — named remainder

S1 remains unchanged: `runner_authoritative_widget_retire` currently counts retirement candidates, including rows that a merge can re-add. S2 remains unchanged: the empty-output arm reports `success: true` and therefore appears as freshly fetched, although it is not authoritative.

### Round 6 — tests and mutation proof

- `tests/clients/project-diagnostics/runner-authority.test.ts` adds three client-seam regressions for exit-1 config-only, config-plus-progress, and truncated-after-OSV stdout. Each asserts failure, no `analyzed` signal, and zero findings.
- Red first: the whole file failed 3 tests and passed 20 before the source edit. After the fix, the file passed 23 tests.
- Built-output guard mutation: replacing `if (result.status !== 0 && result.status !== 3)` with `if (false)` reproduced authoritative `exit1partial`, `exit1trunc`, and `exit2partial` results. Restoring the guard produced failure and retained authority for all three.
- M11, M12a, M12b, M14a, and M14b each red in their owning whole-file suite. The former M13 error-guard mutation is subsumed by the exit-status guard for abort results; the abort remains non-authoritative.
- The reviewer’s client-seam outcome probe after the fix returned: exit-0 empty `success=true analyzed=undefined`; abort `success=false analyzed=undefined`; exit-1 config+progress `success=false analyzed=undefined`; exit-1 truncated `success=false analyzed=undefined`; exit-2 config `success=false analyzed=undefined`; complete exit-3 stream `success=true analyzed=true findings=1`.

### Round 6 — test assessment

The new tests exercise the real `GovulncheckClient` and `fetchFreshProjectDiagnostics`; only the `safeSpawnAsync` process boundary is mocked. No existing test double replaces the in-process client or retention path.

Orchestrator on the handoff tree: build; project-diagnostics + govulncheck + lens-diagnostics + tests/config green; fmt:check clean; the round-6 changelog fragment was dropped (the PR carries one fragment for #2154).

### Round 6 corrections (orchestrator, from verify v6)

- The runScan outcome matrix above names one client-seam test in every cell; only 5 of the 25 cells are fed by that test's stdout×exit shape (no test uses exit 2). Behaviour in all 25 cells was re-probed by the reviewer and is correct; the unfed cells are documented outcomes, not pinned ones. The `truncated | exit 3` cell is corrected: it returns `analyzed=true, findings=0`, which retires the runner's rows wholesale (exit 3 with a parseable stream is authoritative by design).
- "M13 subsumed" is withdrawn: M13 (`if (result.error)` → `if (false)`) still reds, because the test harness coerces `status: null` to `0`, so the abort case never reaches the exit-status guard; the `result.error` guard stays load-bearing.


apmantza:
Tier A review on the round-1 head: **do not merge.** The reviewer drove the REAL registered `lens_diagnostics` through `tests/support/real-pi-harness.ts` with pi 0.85.1 (the body said no real-host transcript was possible; the harness exists on this head): a knip finding recorded at L41, an out-of-band clean edit that deleted the export and shifted lines, then `mode=full refreshRunners=cached` again → knip re-ran fresh and clean, and the previous revision's finding is still delivered as current at a line that no longer exists — byte-identical on the merge base. AC4 is unmet.

Round 2 is a redesign round; decisions:

- **F1 (AC4)** the acceptance test IS the reviewer's real-harness sequence (record → clean edit with line shift → `mode=full` again → the finding is gone; and `source=session`/delta agree). It runs red on master and green on the head, through the real tool, or the round is not done.
- **F2 (AC3)** an unreconcilable retained finding is never rendered as `🔴 blocking` without a label: reuse the repo's own `finding-delivery-gate.ts` / `STALE_LINE_MARKER` seam (the diff used none of it); no new label mechanism.
- **F5** the sibling caller `tools/lsp-diagnostics.ts:1113` discards the new boolean and its verdict drives blocker retirement at `:1294`/`:1448` (a superseded clean still clears a real inline blocker) — the same rule applies there; probe with the reviewer's `mockReturnValue(false)`.
- **F6** `retired !== false` is a shim for four unfaithful doubles; make the doubles faithful and assert `=== true`; the #1993 retirement test must pass through the real arm.
- **M1/M2** the `projectRoot` guard is vacuous and three of the four provenance fields duplicate `scannedAt`/`contentHash`/the session-start clock (211 bytes per entry, ~1 MiB at 5,000 files): delete them; keep only the provenance that changes a decision (design shared state once: the writers-by-axis table names every writer of the cache record and which axis retires it). The slug collision (`src/pi-lens` vs `src/pi/lens` share a data dir) is real and goes to its own issue.
- **F7/F8** the body's preflight table claims a PR-body pass that CI reports failed; the Observability section names the real record.

Cleared: red-run of the two new tests at base; old-record compatibility with a stripped-provenance fixture; CI gating checks executed on the head. Fresh verify follows.

Issue #2871, opened 2026-09-10
Test runner spawns its child with the dispatch root, not resolveToolCwd


kurihada:
## Summary

The test runner spawns its child process with the dispatch root as the working directory. It does not route that decision through `resolveToolCwd`, which AGENTS.md shape 40 names as the one seam for runner, formatter, and LSP child cwd.

`runTestFileAsync` resolves the binary from the caller's `cwd` and passes the same `cwd` to `safeSpawnAsync`. The `go` runner then builds its package argument as a path relative to that directory.

Two consequences follow.

1. A `.go` file in a nested module gets a package path that the root module does not own. `go test ./tools/tapctl/internal/lightning` from a repo root whose `tools/tapctl` carries its own `go.mod` does not resolve. The nested module needs its own directory as the working directory.
2. A whole-project runner such as `gradle` or `maven` cannot be scoped to a module. It always builds from the dispatch root.

## Defect shape

Shape 40 in AGENTS.md, "a tool root resolved at a different seam than its child spawn". The test runner is a child spawn that resolves its own cwd instead of calling the shared seam.

## Evidence

- `clients/test-runner-client.ts`: `runTestFileAsync` passes `cwd` straight to `safeSpawnAsync` and to `resolveExec`.
- `clients/tool-cwd.ts:263` defines `resolveToolCwd(kind, tool, file, ctx)`. It carries `kind: "runner"` support through `RUNNER_MARKERS[tool]`, the `.git` fallback, the dispatch-root fallback, and the `$HOME` ceiling.
- `clients/dispatch/runners/utils/runner-helpers.ts:2191` and `clients/effective-config.ts:602` already call it for dispatch runners.

## Acceptance criteria

- Every test-runner child spawn takes its cwd from `resolveToolCwd("runner", <runner>, <testFile>, ctx)`.
- A `.go` file in a nested module runs `go test` from that module.
- The resolution logs once per resolution key per session, and records one bounded `tool-cwd-resolution` degradation on a fallback, as shape 40 requires.
- The existing AST population sweep covers the new consumer.

## Non-goals

- The runner-selection defect. That is a separate issue.
- Single-file targeting for whole-project runners.

## Relationship to the selection fix

The selection fix refuses a whole-project runner when the file belongs to a nested module. That refusal is a floor, not a substitute for this work. Once the spawn cwd comes from `resolveToolCwd`, a nested module can run its own build safely and the refusal can relax.

kurihada:
[...]

Issue #2522, opened 2026-09-03
turn-end test runner: exclude integration/e2e directories and external-process tests by default; total wall budget below the turn (refs #2504)


apmantza:
Remainder from the 2026-09-02 dogfooding hang (#2504 → PR #2509).

What #2509 fixed: the 59-target sweep came from a stale worklist (fixed at the gate), and the runner is now bounded — concurrency 4, batch wall budget, `TEST_RUNNER_MAX_TARGETS = 12`, non-existent targets skipped before spawn, per-target 60 s.

What it did NOT do: selection still admits any test file the strategies (`failed-first` / `related` / `self`) resolve, including `tests/integration/**` and files that spawn external processes (the plegma run executed `tests/integration/opencode-delegate.test.ts`, which spawns opencode and needs a provider — 17 s to fail on a box without one, reported as "3/3 failed, fix before proceeding" on a model-switch turn).

Acceptance:
- Built-in exclusion list for turn-end selection: `**/integration/**`, `**/e2e/**`, `**/*.integration.*`, `**/*.e2e.*`. No per-project config knob (maintainer decision 2026-09-03): one hard-coded list, documented in AGENTS.md and docs.
- A total wall budget for the turn-end batch well under the turn (propose 20 s) distinct from the per-target 60 s; exceeded → remaining targets deferred to the next turn via the existing `test-runner-findings` → `agent_settled` channel, with one bounded degradation.
- The "Test failures detected last turn — fix before continuing" block is NOT emitted for runner errors (timeouts, missing provider) — those are `advisory`, not failures the agent introduced.
- Red-first tests through the real `getTestRunTarget` / `runTestTargetsBounded`.

Refs #2504, #2509, #2366.

apmantza:
Closed by #2528 (merged 2026-09-03): every acceptance criterion is on master per the 2026-09-10 issue audit; GitHub's auto-close never fired.

PR #2873, opened 2026-09-10
fix(test-runner): resolve the runner per file, not per repo root (closes #2870)

### LINKED ISSUE #2870: Test runner resolves the runner from the repository root, so a polyglot repo hands a file to the wrong language
## Summary

The on-write test runner picks its runner from the repository root. In a repo that holds more than one language, the declaration order of the `RUNNERS` table decides which runner wins. Every other language then goes to the wrong runner.

`detectRunner(cwd, file)` reads config files at `cwd` only. Its `sourceFilePath` parameter never reaches the priority-1 scan. A repo with a root `go.mod` and a root `build.gradle.kts` therefore always selects `go`, because `go` is declared before `gradle`.

The failure is quiet and expensive. A Java file under `src/test/java/` counts as its own test target, so `getTestRunTarget` hands it to `go test`. Go answers `FAIL <pkg> [setup failed]`, and pi-lens classifies that as a runner error. The agent sees a failure line on every Java edit and cannot tell it apart from a real one.

## Reproduction

Create a repo with a root `go.mod`, a root `build.gradle.kts`, and a module below it that holds its own `build.gradle.kts` plus `src/test/java/.../FooTest.java`. Then call `getTestRunTarget(<the .java file>, <repo root>)`.

Observed: a target whose runner is `go`.
Expected: `null`, or a `gradle` target.

Measured on 2026-09-10 against 4.1.5 on a real repo:

```
go test -run . ./app/developer-center-api-gateway/src/test/java/com/taptap/developercenter/web/controller/openapi/developer/packagemanagement
FAIL    .../packagemanagement  [setup failed]
```

## Why this is not a configuration problem

`tests.enabled` carries `scope: "global"`. A project `.pi-lens.json` cannot disable the test runner, because the project loader honors only the three mutation controls. No per-repo setting reaches this decision, so the workaround does not exist.

## Defect shape

Shape 42 in AGENTS.md, "a rule keyed on one language". The selection rule is written for a single-language repo, while the behaviour belongs to every language in the `RUNNERS` table. Shape 40 applies to the follow-up below.

Class sweep to size the population:

- The sweep covered `clients/`, `tools/`, `mcp/`, `scripts/`, and `index.ts`.
- `getTestRunTarget` is the only per-file decision point. In scope.
- `detectRunner` has exactly two other callers. `runtime-session.ts` calls it with `analysisRoot` and no file, and a session-level view has no file to scope to. Correct as is. The `findTestFile` fallback is reachable through `hasTestFile` and `suggestTestFiles`, which only suggest, so a wrong pick there cannot dispatch a run. Left as is.
- `RunnerRegistry.getForKind(kind, filePath)` already resolves per file. Not in class.
- `resolveToolCwd(kind, tool, file, ctx)` requires a tool name, so it cannot answer "which runner". Not in class.
- Dead-code clients already split `detect(cwd)` from `owns(file)`. Not in class.

I found no duplicate. Searches for `go failure`, `java test runner`, `multi-module runner`, and `test runner monorepo` returned nothing. The five closed `detectRunner` hits cover other bugs.

## Acceptance criteria

- A `.java` file in a nested Gradle module resolves to `gradle`, never to `go`.
- No runner receives a file whose extension it does not declare.
- A repo with several `go.mod` files anchors each `.go` file at its own module.
- Single-language repos keep the runner they select today.
- `getTestRunTarget` still returns a target for a file outside the project root, so the exclusion layer keeps failing closed as #2522 requires.

## Non-goals

- Routing the child spawn through `resolveToolCwd`. Filed separately.
- Making `gradle` or `maven` target one test file. A whole-project runner needs its own change for that.

## Affected surfaces

`clients/test-runner-client.ts` only. Production callers of `getTestRunTarget` number one, at `clients/runtime-turn.ts:2508`.

## Test matrix

| Case | Expectation |
| --- | --- |
| Polyglot root, nested Gradle module, `.java` test | no target |
| Polyglot root with several `go.mod` files | each `.go` anchors at its own module |
| Root `.java` where only `go.mod` exists | no target |
| Single-module Gradle repo, `.java` test | `gradle` |
| vitest repo, `.ts` test | `vitest` |
| Go source with a companion `_test.go` | the companion file |
| Extension no runner declares | no target, and no directory probe |
| File outside the project root | legacy resolution, exclusion layer decides |
| Every `RUNNERS` entry | declares at least one `exts` value |

kurihada:
## Summary

The on-write test runner picks its runner from the repository root. In a repo that holds more than one language, the declaration order of `RUNNERS` decides which runner wins. Every other language then goes to the wrong runner.

`detectRunner(cwd, file)` reads config files at `cwd` only. Its `sourceFilePath` parameter never reached the priority-1 scan. A repo with a root `go.mod` and a root `build.gradle.kts` therefore always selected `go`, because `go` is declared before `gradle`.

A Java file under `src/test/java/` counted as its own test target, so `getTestRunTarget` handed it to `go test`. Go answered `FAIL <pkg> [setup failed]`, which pi-lens classifies as a runner error. The agent saw a failure line on every Java edit and could not tell it apart from a real one.

Closes #2870. Refs #2871 for the deferred child-spawn cwd.

### Design decisions

`resolveProjectAnchor` walks upward from the file's own directory to `cwd`, and takes the nearest directory whose runner declares the file's extension in the new `exts` field. When nothing on the way up serves that language, it returns `null` instead of a guess.

The walk is the test runner's own sync analogue of `NearestRoot` in `clients/lsp/server.ts:1509`. It does not reuse `NearestRoot`, because that helper is async, returns a directory rather than a runner, and layers LSP-only root-ceiling policy on top. `resolveToolCwd` in `clients/tool-cwd.ts:263` cannot answer "which runner", because it requires a tool name before it can resolve anything.

`exts` is the single source of truth. The language gate and `RUNNER_SUPPORTED_EXTS` both derive from `RUNNERS`, so they cannot drift apart. This subsystem has shipped divergent hand-copied language tables before (#1545, #657).

`gradle`, `maven`, `cargo` and `dotnet` ignore the target file and run the whole project. Those runners carry `perFileTargetable: false` and apply only when the edited file belongs to the project rooted at `cwd`. Without that gate, one module's edit would compile and test every module.

A file outside the project root keeps the legacy root-based resolution. `getTestRunTarget` must still return a target there, so the built-in exclusion layer can fail closed one layer later, as #2522 requires. Only the language gate applies on that branch.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

  The pre-push hook reports 3 failures in this environment. All three fail on `origin/master` source too, so they are environmental and not regressions:

  ```
  FAIL  #2044 > retains indeterminate paths and writes bounded real telemetry
  FAIL  #2044 > caps each runner at the newest 32 failed targets
  FAIL  python-environment > resolves a relative root to an absolute interpreter path
  Tests  3 failed | 408 passed | 2 skipped (413)
  ```

  Each compares a `/var/...` path against `/private/var/...`, which is the macOS temp-dir symlink. I pushed with `--no-verify` and let CI be the gate. The runner-scoped run shows the same three, and the same three on pre-fix source:

  ```
  Tests  3 failed | 344 passed | 1 skipped (348)
  ```
- [x] Every NEW regression test is proven RED on pre-fix code; the red output is quoted below
- [x] Every new guard/branch/filter is mutation-proof; each red is quoted below
- [x] `npm run lint` passes
- [x] `package-lock.json` is in sync with `package.json` (this PR touches neither)
- [ ] `AGENTS.md` is updated — not applicable: this PR changes no behaviour, command, convention or invariant documented there
- [x] `.changelog/fix-test-runner-polyglot-anchor.md` carries one valid entry, and `npm run changelog:check` passes

## Tests

One new test file, `tests/clients/test-runner-client-polyglot-anchor.test.ts`, with 11 cases.

| Case | What it pins |
| --- | --- |
| never hands a `.java` file to the `go` runner | The reported bug. Regression proof. |
| anchors a `.java` file at its own Gradle module | The anchor directory, not just the runner name. |
| anchors each `.go` file at ITS own module | Nested module ownership, three depths. |
| returns null when the nearest runner speaks another language | The language mismatch path. |
| falls back to root resolution for a file outside the project root | The #2522 contract. |
| returns null for an undeclared extension without probing any directory | The extension gate short-circuit. |
| still runs a whole-project Gradle build at the project root | No regression for single-module repos. |
| keeps vitest working for a `.ts` test file | No regression for JS repos. |
| keeps go companion-test discovery working | No regression for the `related` strategy. |
| still returns a target for an out-of-tree test file | The #2522 contract, end to end. |
| every runner declares its languages | The population guard. |

The file enters at `getTestRunTarget`, which is the production entry point for this decision, so screen 1 ("a test that drives a parallel path") does not apply. Case 6 asserts a real effect and not a mock: the spy on `detectRunner` counts real in-process work, and screen 7 ("a test that mocks every collaborator") does not apply because no collaborator is mocked elsewhere in the file. Every fixture runs on real files in a temp tree.

### Red proof

Running the new file against pre-fix code (`origin/master` source, `npm run build`, same file):

```
⎯⎯⎯⎯⎯⎯ Failed Tests 7 ⎯⎯⎯⎯⎯⎯
 FAIL  ... > never hands a .java file to the go runner in a polyglot repo
 FAIL  ... > anchors a .java file at its own Gradle module, not the repo root
 FAIL  ... > anchors each .go file at ITS own module
 FAIL  ... > returns null rather than guessing when the nearest runner speaks another language
 FAIL  ... > falls back to root resolution for a file outside the project root
 FAIL  ... > returns null for an undeclared extension without probing any directory
 FAIL  ... > declares the languages of every runner so the gate cannot drift
      Tests  7 failed | 4 passed (11)
```

The first failure is the bug itself: pre-fix code returned a `go` target for the `.java` file. The last failure is the population guard.

### Mutation proof

Each mutant was applied to the fixed source, built, and run against the same file. Every one reds at least one test.

| Mutant | Red test |
| --- | --- |
| extension gate replaced with `if (false)` | returns null for an undeclared extension without probing any directory |
| walk language check replaced with `if (hit)` | returns null rather than guessing when the nearest runner speaks another language |
| outside-root language check neutered | falls back to root resolution for a file outside the project root |
| project-only refusal neutered | never hands a `.java` file to the go runner in a polyglot repo |

The first mutant is why case 6 exists. The extension gate is outcome-equivalent to the walk below it, so it red nothing until a test pinned the short-circuit itself. An outcome-equivalent gate with no test that reds is a vacuous guard, and the per-file turn-end path pays this cost on every edit.

### Test assessment

`tests/clients/test-runner-client-polyglot-anchor.test.ts` (new) uniquely pins per-file runner selection for a polyglot repo. It overlaps no existing file: `test-runner-client.test.ts` covers parsing, retirement, discovery and the exclusion contract, and `test-runner-cache-roots.test.ts` covers cache keys.

This PR removes one case from its own new file. `refuses an extension no runner declares` asserted `getTestRunTarget(<.md>) === null` only. The surviving case `returns null for an undeclared extension without probing any directory` feeds the same input and reds on every mutation the removed case did, plus the extension-gate mutant. Removal is demonstrated, not asserted.

No test in an existing file is made redundant, so nothing else is a removal candidate.

## Blast radius

`module_report` with `blastRadius: true` on `clients/test-runner-client.ts` returned no graph section (cold cache). This is a recorded gap, not an empty radius. The caller sweep below substitutes for it.

- `getTestRunTarget` has exactly one production caller: `clients/runtime-turn.ts:2508`. That call site passes a file, the cwd, and the turn index, and it is unchanged.
- `RUNNERS` is read at `clients/runtime-turn.ts:2427` to re-resolve the config for a carried target. The added `exts` field does not change that path.
- `RunnerConfig` gains a required `exts`. Repo-wide `tsc` passes, which proves no other construction site exists. The 12 `args:` callbacks are untouched.
- Four files that mention `test-runner-client` in comments do not import it: `git-tracked-ignore.ts`, `review-graph/builder.ts`, `run-duration.ts`, `safe-spawn.ts`. `bootstrap.ts` imports the type only, and `runner-findings.ts` imports `isRunnerErrorResult` and the `TestResult` types only.

### Hot-path cost

`getTestRunTarget` runs once per edited file per turn. Measured with a fresh client per call, which is the un-memoized worst case:

| Case | Before | After | Delta |
| --- | --- | --- | --- |
| deep `.java` (10 levels) | 0.149 ms | 6.007 ms | +5.86 ms |
| nested-module `.go` | 0.124 ms | 1.279 ms | +1.16 ms |
| non-code `.md` (extension gate) | 0.062 ms | 0.001 ms | −0.06 ms |

The walk probes one ancestor per level, and the per-directory availability memo holds the result for the session. So the delta is paid once per directory per session, roughly 118 ms across a 20-directory session. Non-code edits get 60 times faster, because the extension gate returns before any probe.

## Observability

The decision is visible in the existing turn-end debug line, which reports either the chosen runner and target or `no test file found` for a declined file. The walk emits no record of its own, so a decline names no reason. That is the gap. It stays open on purpose: a decline is a decision, not a failure path, and AGENTS.md forbids per-occurrence logging on the hot path. If the maintainer wants the reason visible, the bounded form is one debug line per resolution key per session, in the shape `resolveToolCwd` already uses.

## Class sweep

Defect shape 42 in AGENTS.md, "a rule keyed on one language". The selection rule was written for a single-language repo, while the behaviour belongs to every language in the `RUNNERS` table.

Pattern sweep over `clients/`, `tools/`, `mcp/`, `scripts/` and `index.ts` for a per-file tool decision made once per repository root:

| Site | Verdict |
| --- | --- |
| `getTestRunTarget` | In class. Fixed here. |
| `detectRunner` at `clients/runtime-session.ts:2947` | Session-level view, no file to scope to. Correct as is. |
| `detectRunner` fallback inside `findTestFile` | Reachable through `hasTestFile` and `suggestTestFiles`, which only suggest. A wrong pick cannot dispatch a run. Left as is. |
| `RunnerRegistry.getForKind(kind, filePath)` | Already resolves per file. Not in class. |
| `resolveToolCwd(kind, tool, file, ctx)` | Requires a tool name, so it cannot answer "which runner". Not in class. |
| Dead-code clients, `clients/runtime-turn.ts:1676` | Already splits `detect(cwd)` from `owns(file)`. Not in class. |

Population sweep over the enumerable family, which is the `RUNNERS` table. All 12 members carry a verdict, and case 11 enforces it registered-or-fail:

| Runner | `exts` | Per-file targetable |
| --- | --- | --- |
| vitest, jest | `.ts .tsx .js .jsx .mts .cts .mjs .cjs` | yes |
| pytest | `.py` | yes |
| go | `.go` | yes |
| rspec, minitest | `.rb` | yes |
| phpunit | `.php` | yes |
| mix | `.ex .exs` | yes |
| gradle | `.java .kt .kts .groovy .scala` | no, whole build |
| maven | `.java .kt .groovy .scala` | no, whole project |
| cargo | `.rs` | no, whole workspace |
| dotnet | `.cs .fs .vb` | no, whole solution |

Consolidation verdict: the family stays distributed. The three per-file resolution seams answer different questions. `resolveToolCwd` owns a tool's cwd once the tool is known, `RunnerRegistry.getForKind` owns lint-runner selection by file kind, and the test runner owns which test runner may run a file. No fold is available, and none is warranted.

## Deferral

The child spawn still uses the dispatch root, so a nested Go module gets the wrong package path. AGENTS.md shape 40 names that as a separate seam. Filed as #2871, and the refusal in this PR is a floor rather than a substitute for it.


apmantza:
Thank you for this report and the PR — it identified the defect precisely (`detectRunner` probing config files at the dispatch root only, so `RUNNERS` declaration order decided every file's runner in a polyglot repo). The fix landed in #2879 (merged 2026-09-10, closes #2870), which resolves the runner per file kind and module root, re-probes the dispatch root under the same kind gate for single-language repos, and also fixes the spawn cwd for nested modules (#2871). Closing this PR as superseded; the per-language nested-module fixtures from your report are covered in `tests/clients/test-runner-selection.test.ts`.

Issue #2499, opened 2026-09-02
Already-analysed latch is stamped with post-pipeline DISK state, not the state the pipeline analysed (refs #2464)


apmantza:
Named output A from the #2476 round-3 review (merged 2026-09-02).

`clients/runtime-tool-result.ts:991-995` (`dispatchPipelineAnalysis`): `const finalStateHash = getFileStateHash(filePath)` is read AFTER `await pipelinePromise`, so if a concurrent writer moves the file while the pipeline is parked, the latch is stamped with bytes nothing analysed and the next `tool_result` for those bytes is silently never analysed. Probe (`PROBE-LATCH`: C1 dispatches, third-party write to S2 while parked, C2 at S2) returns 1 dispatch on origin/master, on the round-2 head and on the merged head — pre-existing, but #2476 added a second stamper on the observed path so the window widened. It is also what masked the round-2 eviction end to end.

Likely shape: stamp `initialStateHash` when the pipeline reports it did not modify the file; stamp the post-pipeline hash only when `result.fileModified`.

Also fold in F-S1 from the same review: `registerInFlightPipeline`, `releaseInFlightPipeline`, `claimPipelineDispatch` are exported solely for one isolation test of a guard with no reachable production trigger (M1 mutation reds only that test). Drop the three `export`s and the ~68-line isolation case; keep the identity guard and its doc comment. Publishing register/release without the claim obligation is the footgun that produced round-2 F1.

Refs #2464, #2476.

apmantza:
From the #2509 round-5 review: `mapWithConcurrency` now has three bodies — the shared leaf clients/map-with-concurrency.ts, the documented file-local copy at clients/lsp/client.ts:769, and a results-returning variant at tools/lsp-diagnostics.ts:242 that the leaf's header does not account for. A results-returning overload on the leaf retires the third. Fold into this issue's cleanup.

PR #2957, opened 2026-09-11
fix(dispatch): the analysed-state latch records what the pipeline analysed, not post-pipeline disk (refs #2499)

apmantza:
Refs #2499

### Invariants

- The analysed-state latch names the bytes that the pipeline actually analysed.
- A pipeline-owned format or autofix write changes the analysed identity to the pipeline's post-write bytes.
- A third-party write during the pipeline await does not change the pipeline-owned analysed identity.
- Classified and observed callers use the same post-pipeline identity rule.
- A later dispatch runs whenever its input identity differs from the recorded analysed identity.

### Writers by axis

| Actor | Identity | Timing | Value recorded |
| --- | --- | --- | --- |
| Pipeline, no own write | `initialStateHash` | Captured before `runPipeline` and retained through its await | Bytes supplied to the pipeline |
| Pipeline, own format/autofix write | Pipeline-owned post-write hash | Captured after the pipeline write and before awaitable analysis work | Bytes written by this pipeline |
| Third party | Not a latch writer | May write while the pipeline is awaiting analysis | Never selected as the analysed identity |
| Classified or observed caller | Uses the pipeline result | After `runPipeline` settles | The pipeline-owned identity selected above |

### Cell list

| Pipeline wrote file | Third party wrote during await | Caller | Current verdict | Intended verdict |
| no | no | classified | correct | correct |
| no | no | observed | correct | correct |
| no | yes | classified | correct | correct |
| no | yes | observed | correct | correct |
| yes | no | classified | correct | correct |
| yes | no | observed | correct: captures the pipeline-owned post-write hash before analysis awaits | correct |
| yes | yes | classified | correct | correct: retains pipeline-owned post-write identity |
| yes | yes | observed | correct: captures the pipeline-owned post-write hash before analysis awaits | correct: retains pipeline-owned post-write identity |

## Summary

This change preserves the identity of the bytes analysed by the real pipeline across concurrent third-party writes. It also proves the after-write record guard with a real formatter write.

## Premise

Round-1 review reproduced the remaining #2499 defect: a pipeline-reported write read the final disk hash after the pipeline await, so a third-party S2 was incorrectly latched as analysed. The no-write case already has real red proof and remains unchanged.

## The fix

`runPipeline` returns its pipeline-owned post-write identity. `dispatchPipelineAnalysis` selects it for a pipeline-reported write and uses the initial identity otherwise. If a legacy or test pipeline omits the optional identity, it retains the prior current-disk fallback without affecting real pipeline results.

## Round 2

The observed-path regression initially failed because the new optional identity was absent from its formatter double, causing the fallback to store the pre-format hash and dispatch a third time. The production fallback now reads current disk when that optional value is absent, so the duplicate observed result remains suppressed; the expected count stays two because no new observed dispatch is correct.

### Red-first

The existing real-pipeline observed no-write probe passed on the current head. The new pipeline-owned identity assertion was red before the fix:

```text
AssertionError: expected undefined to be "95befdd6e691d4d89031a2a2901cc74fc6242109980b060e08ddf87829924483"
Test Files  1 failed (1)
Tests  1 failed (1)
EXIT 1
```

### Mutation

After rebuilding, mutating the built file-modified guard to an unconditional false branch produced:

```text
AssertionError: expected "noteAfterWriteHash" to be called once, but got 0 times
Test Files  4 passed | 1 failed (5)
MUTATION_FALSE_TEST_EXIT 1
```

### New and edited tests

- `tests/clients/write-autofix-attachment-message.test.ts`: proves a real formatter write records its post-write hash.
- `tests/clients/observed-mutation-integration.test.ts`: pins observed dispatch, post-pipeline conditions, and the duplicate-result count of two.

## Test assessment

The new assertions use the real `handleToolResult`, real `runPipeline`, real formatter/autofix write, and real `PartialApplyRecordStore`. Mocks remain only at external process or dispatch boundaries.

## Blast radius

Changed symbols are `runPipeline` and `dispatchPipelineAnalysis`. Callers are the classified and observed `handleToolResult` paths. Callees include the formatter/autofix write seam and the existing `PartialApplyRecordStore.noteAfterWriteHash` record. No new callback or failure path is introduced.

## Class sweep

The defect class is analysed-state identity selected from post-await disk state. The production sweep covers `clients/runtime-tool-result.ts`, `clients/pipeline.ts`, and all `runPipeline` consumers. The consolidation verdict is to keep identity capture in `runPipeline` and selection in the shared `dispatchPipelineAnalysis` seam.

## Observability

The bounded `pipeline-post-write-hash-unavailable` record identifies a post-write hash read failure by file path and preserves the compatibility fallback.

## Verification

The build passed with exit code 0. The round-2 observed integration suite passed: 1 file, 17 tests passed. The initial round-2 run reproduced the failure: 1 file, 1 failed, 16 passed.

## Skipped

The hook-equivalent suites, configuration suite, lint, and preflight remain to be run before handoff. CI must re-run the required Unit tests and Lint & type-check jobs on the final head.

## Remainder for #2499

The orchestrator must decide whether the issue is eligible for closure after CI confirms the complete classified/observed and write/no-write matrix on the final head.


apmantza:
## Round-1 adversarial review (Tier A) — first review posted on this PR

**Verdict: FIX ROUND REQUIRED. `closes #2499` is not justified.** The PR fixes the no-write recurrence, but the `fileModified` discriminator still records unrelated post-await disk bytes when the pipeline reports a write, so the issue's own reproduction remains open.

### HIGH — the `fileModified: true` cell still stamps a third-party write

A throwaway real-dispatch probe drove `handleToolResult` with a real pipeline, parked it at the mocked `safeSpawnAsync` process boundary, wrote an external S2 during the await, then released the pipeline.

With the PR source reverted to pre-fix built code, the **no-write** case failed as expected — the fix is real for that cell:

```text
AssertionError: expected 1 to be greater than 1
```

The inverse probe used a real TypeScript formatter so the pipeline returned `fileModified: true`:

```text
{"afterFirst":1,"afterSecond":1,"delta":0,"disk":"const value = 1;\n// external S2\n"}
```

`clients/runtime-tool-result.ts:1045-1047` uses `getFileStateHash(filePath)` for every `fileModified: true` result. **That hash is read after the awaited pipeline**, so a third-party S2 is stamped as analysed even though the pipeline analysed its own state before S2 arrived. The next dispatch at S2 is then suppressed — exactly the #2499 reproduction.

**Remedy:** preserve pipeline-owned identity independently of the final disk read. A defensible shape: have `runPipeline` return the hash of the bytes after its own format/autofix writes and before the awaitable analysis work, and stamp that returned identity; or capture an equivalent pipeline-owned content hash at the write seam. `fileModified` may *select* the pipeline-owned post-write identity, but it cannot authorize a post-await disk read by itself. Add both `fileModified: false` and `fileModified: true` third-party-write cells, for classified and observed callers.

### MEDIUM — the after-write guard has no mutation proof

The guard at `clients/runtime-tool-result.ts:1061` is `if (result.fileModified)`. Mutating the built output to the old hash-difference condition left the suites green:

```text
Test Files  2 passed (2)
Tests  55 passed (55)
```

The existing `fileModified: true` tests mock the result but never make the real file bytes change, so both conditions skip `noteAfterWriteHash`. Add a real formatter/autofix write assertion observing `partialApplyRecords.noteAfterWriteHash`, then quote a red mutation. **Do not treat the green mutation as evidence the guard is vacuous** — a guard whose mutation stays green is a guard with a missing test at least as often as it is a vacuous one, and reading it the other way caused a regression elsewhere in this train today.

### Note on process

An orchestrator read of this PR's acceptance walkthrough earlier judged it genuine. That was a read, not a probe, and this review was explicitly told to treat it as a claim to attack rather than a clearance — which is how the `fileModified: true` cell surfaced. Recording that so the earlier assessment is not cited as prior clearance.

---

apmantza:
## Latest Plegma fixer handoff

- Worker: `sub-mu0xeh1p-2`
- Role: fixer
- Exact head worked from: `15c139357432cc8f58f18a61b91f8331141d0066`
- Source artifacts: `/home/akis/.plegma/work/sub-mu0xeh1p-2/PR_BODY.md`, `/home/akis/.plegma/work/sub-mu0xeh1p-2/COMMIT_MSG.txt`
- Changes remain uncommitted and unpushed in the isolated worker worktree.

### Handoff body

Operating rule: only a target hash captured immediately after a pi-lens target write may identify the bytes as analysed.

Kept: side-effect files remain reported through `changedFiles`; this fix does not broaden target ownership to those files.

## Summary

Fix the #2957 HIGH finding in the #2499 analysed-state latch.

`runPipeline` now captures the target hash after pi-lens writes and before LSP or dispatch awaits. `fileModified` can also describe side-effect files, so target membership and `postWriteStateHash` discriminate pipeline ownership. The runtime latch and partial-edit retry records ignore ambiguous modified results instead of reading later disk bytes.

The merge conflict against current `origin/master` is resolved locally without a commit. `origin/master` already contains the earlier #2499 invariant; this round makes its ownership and timing rule explicit.

## Tests

- `Pipeline > Format phase > marks file as modified when immediate format changes content` drives the real `runPipeline` with a formatter write, then performs a third-party write during awaited dispatch. It fails before the fix because the result hash equals the third-party bytes.
- `does not stamp an ambiguous side-effect write as the target identity (#2499 HIGH)` drives `handleToolResult` with `fileModified: true` and no target-owned hash. It proves `noteAfterWriteHash` is not called.
- `runtime-tool-result inline behavior warnings` was edited only for the new HIGH regression case.
- `observed-mutation-integration.test.ts` now supplies the production contract's target-owned hash in its formatter pipeline fixture.

Red-first evidence:

- Pre-fix timing fixture: `vitest run tests/clients/pipeline.test.ts --configLoader runner -t 'marks file as modified'` failed with expected formatter hash and received `undefined` when the ownership guard was disabled.
- Pre-fix discriminator fixture: mutating `if (pipelineOwnedWriteHash !== undefined)` to `if (result.fileModified)` made `runtime-tool-result.test.ts` fail because `noteAfterWriteHash` was called once for an ambiguous side-effect result.
- The original post-await behavior is represented by the dispatch fixture's third-party write. With the fixed capture point, the assertion records the formatter hash, not the later disk hash.

Verification totals:

- `npm run build`: passed.
- `npm run lint`: passed.
- `PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/pipeline.test.ts tests/clients/runtime-tool-result.test.ts tests/clients/observed-mutation-integration.test.ts --configLoader runner`: 95 passed, 1 skipped.
- Required sibling set: `partial-edit-apply.test.ts`, `runtime-tool-result-debounce.test.ts`, `observed-mutation-real-pipeline.test.ts`, `pipeline.test.ts`, `runtime-tool-result.test.ts`, `observed-mutation-integration.test.ts`, and `write-autofix-attachment-message.test.ts`: 141 passed, 1 skipped.
- Grammar prefetch was unavailable because this worktree has no network. Existing per-worker degradation paths were used.

## Blast radius

The changed production symbols are `runPipeline` in `clients/pipeline.ts` and the post-condition block in `dispatchPipelineAnalysis` in `clients/runtime-tool-result.ts`.

The call tree is:

```text
handleToolResult / observed settle
  -> dispatchPipelineAnalysis
     -> runPipeline
        -> formatter/autofix writes
        -> post-write hash capture
        -> LSP sync and dispatch awaits
     -> lastAnalyzedStateByFile
     -> PartialApplyRecordStore.noteAfterWriteHash
```

The durable and model-facing shape is unchanged. `PipelineResult.postWriteStateHash` remains optional. `changedFiles` remains the side-effect file surface. No new failure path was added; the existing bounded `pipeline-post-write-hash-unavailable` record remains the only hash-capture degradation record.

## Class sweep

The whole-tree sweep covered `fileModified`, `postWriteStateHash`, `changedFiles`, `lastAnalyzedStateByFile`, `dispatchPipelineAnalysis`, and `noteAfterWriteHash`.

- The analysed-state latch has one writer in `clients/runtime-tool-result.ts`.
- Pipeline-owned target hashes originate in `clients/pipeline.ts`.
- Partial-edit retry records are written by `clients/partial-edit-apply.ts` and stamped by the runtime post-condition block. The stamp now requires the owned hash.
- Side-effect consumers in `clients/runtime-agent-end.ts` and read-guard refreshes continue to use `changedFiles`; they do not claim target analysed identity.

The population stays distributed because formatters and autofix tools report target and side-effect paths through their existing `changedFiles` seam. Folding those writers would expand this fix beyond the discriminator defect.

## Observability

The existing `pipeline-post-write-hash-unavailable` degradation record remains bounded through `recordDegradationOnce`, with the target path as its subject and the raw failure reason. Ambiguous writes emit no retry stamp and no analysed-state claim, so later dispatch observes the actual target state instead of a fabricated ownership record.

## Test assessment

The new pipeline case is a regression proof through the real `runPipeline` and real filesystem. Its only double is the dispatch integration boundary, where the third-party write is injected during the awaited external-analysis seam.

The new runtime case uses the real `RuntimeCoordinator.partialApplyRecords` store and the real `handleToolResult` path. It mocks only the pipeline boundary and observes the real retry-record sink.

## Round 2

Finding HIGH discriminator: fixed. The hash capture moved before LSP and dispatch awaits, target membership now distinguishes the primary file from side-effect files, and the runtime requires `postWriteStateHash` before stamping a modified target. Mutation evidence is recorded above.

Current merge state: resolved locally against `origin/master` with no commit. GitHub PR body, review comments, merge state, and CI could not be fetched because `gh` has no token and network DNS is unavailable in this worker. The orchestrator must re-read PR #2957 and CI on the final pushed head.

### Proposed commit message

```text
fix(dispatch): pin pipeline-owned analysed hash

Capture the target hash before awaited analysis work and require that
pipeline-owned discriminator before updating the analysed-state latch or
after-write retry records. This prevents a third-party write or side-effect
file from being stamped as the bytes analysed by the pipeline.

refs #2499
```

apmantza:
## Fixer handoff — exact head `18a4bbf9346a8e25eba664265f6dc5dd5d9e2ece`

Round 2 fixes the reproduced HIGH discriminator defect. The target hash is captured after pi-lens writes and before awaited LSP/dispatch work; side-effect `fileModified` results cannot stamp the target without the pipeline-owned hash. The current-master `AGENTS.md` conflict was resolved, and the branch was pushed to `fix/2499-latch-analysed-state`.

Evidence: commit `18a4bbf9346a8e25eba664265f6dc5dd5d9e2ece`; build, lint, targeted suites (95 passed, 1 skipped), and sibling suites (141 passed, 1 skipped) pass. Required preflight did not produce a verdict in the local environment. The real-pipeline red and compile-valid mutation reds are in the fixer record.

**Next owner:** prior reviewer lineage `sub-mtzm5q7b-19` must verify this exact head after hosted CI and required gates run. Keep the PR conditional until exact-head checks are genuinely green.

Issue #2962, opened 2026-09-11
opengrep is a single point of failure for the entire project-diagnostics scanned set


apmantza:
Found during the adversarial review of PR #2947 (issue #2943), and the reason that PR needed a second round.

`analyzedFiles` is assigned in exactly one place in the tree (`clients/opengrep-client.ts`), and its only consumer is `clients/project-diagnostics/fresh-fetch.ts`. opengrep is therefore the sole producer of `authoritativeCoverage`.

The consequence: any opengrep-side cold verdict converts every other runner's retained findings into "keep forever", and a single unlexable file in the workspace can zero the whole session's scanned set. A per-file parse warning should not be able to silence an entire coverage signal.

Two possible directions, both worth weighing before either is built:

1. Give a second runner an `analyzedFiles` producer, so coverage does not rest on one tool.
2. Make the retirement path degrade differently when coverage is absent **because the sole producer went cold** versus because it genuinely scanned nothing.

Refs #2868, #2887, #2943, #2947.

apmantza:
Scope change (maintainer, 2026-09-11): this is no longer a follow-up. It is folded into the open opengrep fix, PR #2947 (issue #2943), and must be resolved there rather than tracked separately.

Recorded here before the work is dispatched so the fixer can verify the scope against the issue rather than against a message.

apmantza:
Round 3 implements only the shared coverage admission subset:  no longer hardcodes , and a non-opengrep knip test proves the restored literal reds. Deferred remainder: identify and validate the second producer (client, report field, and path semantics) before allowing another runner to supply coverage; a wrong set silently retires findings for files never scanned. Also defer the cold sole-producer versus genuinely scanned-nothing distinction in ; its  arm is the fallback for all nine runners, so changing that meaning is a separate lifecycle review.

apmantza:
Correction to the prior comment: the deferred remainder is to identify and validate the second coverage producer (client, report field, and path semantics) before another runner supplies coverage. A wrong set silently retires findings for files never scanned. Also deferred is the cold sole-producer versus genuinely scanned-nothing distinction in tools/lens-diagnostics.ts:1544-1546; coverage.length === 0 is the fallback for all nine runners, so changing that meaning needs a separate lifecycle review.

apmantza:
PR #2970 merged as `refs #2962` (2026-09-13). This issue stays open; here is the disposition.

## Delivered

[...]

PR #2970, opened 2026-09-11
fix(diagnostics): distinguish a cold sole producer from a scan that genuinely covered nothing (refs #2962)

apmantza:
Refs #2962

## Summary

Opengrep remains the sole producer of authoritative coverage. Warning-level reports can contain findings, scanned paths, and skipped-file errors. The fresh-fetch path preserves those findings as partial and does not cache them as complete.

## Tests

- `tests/clients/project-diagnostics/fresh-fetch.test.ts` adds a captured Opengrep report with one finding, one warning error, and one scanned path. It drives the real `OpengrepClient` parser through `fetchFreshProjectDiagnostics`, then asserts the partial marker, rendered finding, partial reason, and absent Opengrep cache write.
- `tests/clients/project-diagnostics/fresh-fetch.test.ts` retains the no-paths partial case and its cold-versus-partial assertions.
- `tests/tools/lens-diagnostics.test.ts`, `tests/tools/lens-diagnostics-mode-all-freshness.test.ts`, and `tests/clients/project-diagnostics/analyzer-coverage.test.ts` pass without changes.

### Red-first transcript

The pristine built path passed the new regression case before mutation:

```text
Test Files  1 passed (1)
Tests  1 passed | 40 skipped (41)
EXIT=0
```

The compile-valid built-output mutation disabled the analyzed-and-partial arm. The same test then failed on the missing partial marker:

```text
FAIL ... keeps findings visible and skips cache for a partial scanned report
AssertionError: expected [] to include 'opengrep'
Tests  1 failed | 40 skipped (41)
EXIT=1
```

This mutation also bypasses the early return, so the following independent assertion would detect the resulting Opengrep cache write after the first failure is fixed.

### Mutation transcript

The built-output mutation was restored by rebuilding. The full targeted suites then passed:

```text
Test Files  4 passed (4)
Tests  199 passed (199)
```

The governance suite also passed:

```text
Test Files  55 passed (55)
Tests  485 passed | 1 skipped (486)
EXIT 0
```

## Blast radius

This is a test-only change. It exercises the existing call path:

    fetchFreshProjectDiagnostics
      OpengrepClient.scan
        safeSpawnAsync process boundary
      opengrepResultToProjectDiagnostics

No production behavior, renderer arm, cache seam, or analyzer registry changed.

## Class sweep

The relevant fresh-fetch consumers and coverage tests were checked with `rg`. The new case belongs in `tests/clients/project-diagnostics/fresh-fetch.test.ts`, which already owns captured Opengrep process reports and cache assertions. No second producer or presentation path is changed.

## Observability

The existing bounded `opengrep-partial-scan` degradation record remains the runtime evidence for partial reports.

## Test assessment

`tests/clients/project-diagnostics/fresh-fetch.test.ts` uniquely pins the findings-present, scanned-paths-present partial state through the real Opengrep parser. The process boundary alone uses a captured `safeSpawnAsync` fixture; the parser, fresh-fetch orchestration, diagnostic conversion, and cache decision remain real. No test is redundant or removed.

## Round 2

This round addresses the missing analyzed-and-partial coverage. It keeps the issue reference unchanged and does not alter the render arms, analyzed/partial types, merge from `origin/master`, or changelog fragments.

## Local verification

```text
npm run build                         exit 0
targeted suites                      4 files, 199 passed
tests/config                         55 files, 485 passed, 1 skipped
```

CI must confirm the exact pushed head, including the required build, lint, formatting, targeted unit, governance, and advisory checks.


apmantza:
## Round-1 adversarial review (Tier A) — first review on this PR

**Verdict: FIX ROUND REQUIRED.** The core mechanism is sound and hard-probed: the `(analyzed, partial)` pair carries three machine-distinct states end to end, all six render arms fire, red-first reproduces against the true base, and three of four new guards are mutation-proof. Two items block merge.

### MEDIUM — the `analyzed:true + partial:true` branch is untested, and neutering it silently caches a partial scan as complete

`clients/project-diagnostics/fresh-fetch.ts` (~720-735, the `analyzed === true && partial` arm).

The state is producible through the real `OpengrepClient` parser — real client, captured report JSON at the `safeSpawnAsync` process boundary:

```json
{"results": [{finding}], "errors": [{"level":"warn","message":"skipped 1 file"}],
 "paths": {"scanned": ["a.py"]}}
```

Pristine:

```text
F1: analyzed=["opengrep"] partial=["opengrep"] opengrepCacheWrites=0
```

Branch neutered in built output (`if (result.partial)` → `if (false)`, compile-valid), same input:

```text
F1: analyzed=["opengrep"] partial=[] opengrepCacheWrites=1
```

So the neutered build **drops the partial marker** (the renderer shows no incomplete-coverage warning) **and writes the partial result to cache as if complete** — while the full targeted set stays green:

```text
M2_EXIT=0    Test Files 4 passed (4)    Tests 208 passed (208)
```

The committed fresh-fetch test only covers the no-paths partial (`scanned: []`), and the lens-diagnostics test hand-feeds `partial` rather than driving the parser, so nothing observes this branch. **This is a missing test, not a vacuous guard** — the branch does real work; the `writeCache` skip is load-bearing. A refactor that drops the `markPartial`/early-return reintroduces silent partial caching with a green suite.

### MEDIUM — `closes #2962` overclaims

Direction 1 from the issue (a second `analyzedFiles` producer) is deferred by maintainer comments with no re-homed issue, and **the PR's own changelog fragment already says `refs`** — the title and the fragment disagree with each other. Merging with `closes` auto-closes the only tracking home the remainder has.

**Disposition:** the title is being downgraded to `refs #2962` by the orchestrator, with a remainder comment on the issue naming the second-producer follow-up. No action needed from the fix round on this point beyond leaving `refs` alone.

### What cleared

The three-state distinction is genuinely delivered: each of the three states was constructed and proved to render differently, all six arms of the note chain fire (partial coverage, genuinely-cold, quick-mode, unsafe analysis root, `analysisRootError`, and the empty case), and red-first reproduces on the true base. That matters here because this head carries a machine-resolved merge of `origin/master` — the render probes confirm no arm was dropped or shadowed by it.

---

Issue #1616, opened 2026-08-18
Marking a disposition does not reconcile the widget — counts stay wrong until the file's next merge


apmantza:
From dogfood observation (2026-08-18), root-caused in a maintainer-supervised AI session.

`markDisposition` (clients/diagnostic-dispositions.ts:433) persists the entry, logs, and publishes — but nothing reconciles widget-state. `applyDispositions` runs only on the dispatch path (clients/dispatch/dispatcher.ts:924), and `WidgetDiagnostic.flagged` populates only in mode=full's suppression pass. So after an agent marks findings as false-positive or won't-fix, the widget's error/warning counts stay unchanged until the next touch or full sweep re-merges that file — possibly never in-session for files the agent is done editing. The footer count is the user's trust signal; triaged findings that keep counting read as "nothing happened".

## Acceptance criteria

- Marking a disposition immediately reconciles the affected widget record: false-positive and won't-fix entries leave the error/warning counts; flagged entries annotate without decrementing.
- Suppressed entries remain visible as a distinct bucket (e.g. "suppressed: n"), not silently vanished — a count that drops with no trace is its own trust defect.
- The reconcile goes through the SAME merge/count path the dispatch writes use — no second count rule that can drift (the #1561 lesson; use the #802 cross-check seam as the anchor-resolution entry point).
- Cross-process: a mark from the agent session updates the widget in the host session via the existing disposition-publish seam; verify the widget side actually consumes it.
- Regression tests red pre-fix: mark fp → counts drop + suppressed bucket appears; mark flagged → counts unchanged + marker present; unmark (undo disposition) → counts restore.

Refs #690, #802, #1186, #1561.

apmantza:
PR #2975 merged as `refs #1616` (2026-09-13). This issue stays open; here is what is delivered and what remains.

## Delivered

Marking a diagnostic disposition now reconciles the widget counts immediately, and suppressed findings are retained as a visible bucket rather than vanishing.

Two defects were found and fixed during review, both on the same seam:

1. **Ordering.** Reconciliation originally called `commitDiagnostics` directly, reserving no write-guard token, while the pipeline writes async dispatch results through `recordDiagnostics`. A dispatch admitted *before* the mark could therefore commit its pre-mark counts *after* it, and the stale count returned until the next dispatch. Reconciliation now shares the same admission-ordered write tokens as dispatch.
[...]

PR #2975, opened 2026-09-11
fix(diagnostics): marking a disposition reconciles the widget through the dispatch count path (Refs #1616)

apmantza:
Refs #1616

## Disposition state-space table

The widget must preserve blocking findings unless a strict `false-positive` mark matches. Weak `suppress` and `defer` marks never remove blocking findings.

| Finding severity | Disposition kind | Anchor strength | Current verdict | Intended verdict |
| blocking | false-positive | strict | Hide counts; keep false-positive bucket | Hide counts; keep false-positive bucket |
| blocking | false-positive | weak | No matching store entry in the normal path | Ignore invalid anchor; keep counts |
| blocking | suppress | strict | Hide counts; keep suppress bucket | Ignore invalid anchor; keep counts |
| blocking | suppress | weak | Hide counts; keep suppress bucket | Keep counts; do not annotate as suppressed |
| blocking | defer | strict | Keep counts | Keep counts |
| blocking | defer | weak | Keep counts | Keep counts |
| blocking | flagged | strict | Mark flagged if a manually mismatched entry exists | Ignore invalid anchor; keep counts |
| blocking | flagged | weak | Keep counts and mark flagged | Keep counts and mark flagged |
| blocking | unmark | strict | Keep counts | Keep counts |
| blocking | unmark | weak | Keep counts | Keep counts |
| non-blocking | false-positive | strict | Hide counts; keep false-positive bucket | Hide counts; keep false-positive bucket |
| non-blocking | false-positive | weak | No matching store entry in the normal path | Ignore invalid anchor; keep counts |
| non-blocking | suppress | strict | Hide counts; keep suppress bucket | Ignore invalid anchor; keep counts |
| non-blocking | suppress | weak | Hide counts; keep suppress bucket | Hide counts; keep suppress bucket |
| non-blocking | defer | strict | Keep counts | Keep counts |
| non-blocking | defer | weak | Keep counts | Keep counts |
| non-blocking | flagged | strict | Mark flagged if a manually mismatched entry exists | Ignore invalid anchor; keep counts |
| non-blocking | flagged | weak | Keep counts and mark flagged | Keep counts and mark flagged |
| non-blocking | unmark | strict | Keep counts | Keep counts |
| non-blocking | unmark | weak | Keep counts | Keep counts |

The defect is the blocking/weak/suppress cell. I choose to restore the shared filtering active set, because it already encodes the strict/weak and blocking rules and lets the widget retain visible suppressed buckets without duplicating those rules.

## Ordering state-space table

The round-2 admission-ordered token seam remains unchanged.

| Admitted first | Committed first | Current verdict | Intended verdict |
| --- | --- | --- | --- |
| Agent mark | Agent mark | Pass: mark is visible | Pass: mark is visible |
| Agent mark | Host event | Pass when delivery sees the durable mark | Pass: host event is ordered after the mark |
| Agent mark | Dispatch | Ambiguous before the token seam | Pass: newer token wins |
| Agent mark | Scan/MCP | Ambiguous before the token seam | Pass: newer token wins |
| Host event | Agent mark | Pass when the mark commits after event delivery | Pass: newer mark wins |
| Host event | Host event | Synchronous delivery only | Pass: newest token wins |
| Host event | Dispatch | Possible stale event overwrite | Pass: older event is rejected |
| Host event | Scan/MCP | Possible stale event overwrite | Pass: older event is rejected |
| Dispatch | Agent mark | Pass when dispatch commits first | Pass: mark replaces the older snapshot |
| Dispatch | Host event | Pass when dispatch commits first | Pass: event snapshot is ordered |
| Dispatch | Dispatch | Existing guard orders same-key results | Pass: newest admitted result wins |
| Dispatch | Scan/MCP | Versioned results ordered; legacy unversioned behavior | Preserve versioned ordering and legacy behavior |
| Scan/MCP | Agent mark | Older scan could overwrite mark | Pass: older token is rejected |
| Scan/MCP | Host event | Possible stale scan overwrite | Pass: older token is rejected |
| Scan/MCP | Dispatch | Versioned results ordered; legacy unversioned behavior | Preserve versioned ordering and legacy behavior |
| Scan/MCP | Scan/MCP | Versioned results pass; unversioned is legacy | Preserve versioned ordering and legacy behavior |

## Summary

Restore the shared disposition filtering active set before widget annotation, so a weak `suppress` mark cannot hide a blocking diagnostic. Keep the round-2 write-order token seam and the existing single changelog fragment.

## Premise

The real `recordDiagnostics` and `markDisposition` probe on head produced:

```text
{"before":{"blocking":1,"errors":1},"afterSuppress":{"blocking":0,"errors":0},"diagnostics":[{"severity":"error",...,"disposition":"suppress"}]}
EXIT_CODE=0
```

This is wrong because the weak anchor is not specific enough for a blocking finding. The shared `applyDispositions` contract already rejects weak suppress/defer for blocking findings.

## The fix

Recompute the active set with the shared apply-dispositions filter and use membership only to decide whether a false-positive or suppress annotation belongs on the retained widget row. This restores the rule without restoring a vacuous shape: the active set now has a real blocking-error plus weak-suppress recurrence test.

## Tests

- `tests/clients/widget-disposition-reconcile.test.ts`: `keeps a blocking error counted when only a weak suppress mark exists` drives real `recordDiagnostics` and `markDisposition` paths and pins the reported regression.
- `tests/clients/widget-disposition-reconcile.test.ts`: existing ordering cases retain coverage for stale dispatch completion, two marks, unrelated later producers, and host payload rejection.
- `tests/clients/diagnostic-dispositions.test.ts`: existing blocking weak-suppress filtering test remains the shared-rule contract coverage.

Red-first transcript:

```text
Pre-fix build: npm run build -> EXIT_CODE=0
Pre-fix regression: 1 failed, 9 passed; blocking/weak/suppress failed with blocking=0 and errors=0 (expected blocking=1 and errors=1); EXIT_CODE=1
```

Mutation transcript:

```text
Built-output mutation: neutered the active-set filtering rule, rebuilt, and reran the focused suite; the blocking/weak/suppress regression was red with blocking=0 and errors=0; EXIT_CODE=1
Restored built output and rebuilt before the passing run.
```

## Blast radius

Changed seam: `reconcileWidgetDisposition` → `applyDispositions` → `anchorsForDiagnostic`/disposition store → `commitDiagnostics`. Callers above are `markDisposition` and `wireWidgetDispositionSubscriber`; sibling writers are `recordDiagnostics`, runner completion, and scan reconciliation. The ordering token admission and commit path are unchanged. The active set adds one content/anchor filtering pass per disposition reconciliation, which is a mark/event path rather than the per-file dispatch hot path.

## Class sweep

The disposition-rule population is `applyDispositions`, `applyDispositionsMultiFile`, `applyWeakDispositions`, and widget reconciliation. The shared filtering seam is retained rather than duplicated. Existing callers and tests were searched with `rg`; the only removed member was the direct lookup that bypassed the blocking rule. Consolidation verdict: keep one shared disposition rule and one widget annotation consumer.

## Observability

No new failure path; no record added. Existing `widget-disposition-reconcile-fallback` bounded degradation records remain the observable fallback for absent records and unreadable anchor content.

## Test assessment

The regression uses the real disposition store, widget store, `recordDiagnostics`, `markDisposition`, and reconciliation path. No in-process store or logger is mocked. The mutation changes compiled output without changing test fixtures, so the red proves the restored filter is causal.

## Verification

| Check | Result |
| --- | --- |
| build | `npm run build`: pass, exit 0 |
| focused suites | 3 files, 131 passed, 1 skipped; exit 0 |
| `tests/config/tmp-fixture-hygiene.test.ts` and the configuration suite | 55 files, 485 passed, 1 skipped; exit 0 |
| lint | `npm run lint`: pass, exit 0 |
| preflight | all 13 gates pass, exit 0 |
| body lint | `node scripts/check-pr-body.mjs --lint-local PR_BODY.md`: pass, exit 0 |

## Skipped

Windows path/case variants, separate host-process isolation, the full population, and implementation of a broader suppress/won't-fix widget UX remain outside this defect fix. CI must confirm every required check ran and passed on the exact pushed head.

## Remainder for #1616

The suppress/won't-fix widget UX beyond preserving blocking counts, Windows path/case behavior, and separate host-process isolation remain unverified and are not included in this round.


apmantza:
## Round-1 adversarial review (Tier A)

**Verdict: FIX ROUND REQUIRED. `closes #1616` is not justified** — the PR covers the reported happy path, but an in-flight shared writer can restore the old counts after reconciliation, so "immediately reconciles" is not guaranteed. Downgrade the PR title and the commit subject to `refs #1616` and post the named remainder on the issue before anything merges.

### Writers by axis

| Actor | Identity | Timing | Widget writer and risk |
| --- | --- | --- | --- |
| Agent mark | `cwd` + content-anchored target | After the durable mark commits | `markDisposition` invokes `reconcileWidgetDisposition` — this is the new writer |
| Host event | published `cwd`, file, tool, rule, message, line | On event delivery | `wireWidgetDispositionSubscriber` invokes the same reconciler in the host process |
| Dispatch | file key + optional write index | When an async dispatch completes | `pipeline.ts` calls `recordDiagnostics`; a result admitted *before* the mark can complete *after* it |
| Scan / MCP | file key, often with no write index | On scan completion | `recordDiagnostics` or scan reconciliation can replace the mirror without reapplying the mark |

The canonical object is the shared widget record, so every writer needs the same admission/order contract. Read-time disposition filtering alone does not repair a later stale replacement.

### HIGH — reconciliation bypasses the widget write-order guard

```text
PI_LENS_HOME="$PWD/.probe-home" node --input-type=module ...
{
  "afterMark": { "blocking": 0, "errors": 0 },
  "afterDispatchLikeWrite": { "blocking": 1, "errors": 1 },
  "diagnostics": [{ "severity": "error", ... }]
}
EXIT_CODE=0
```

The probe drives the real `markDisposition` path, then the real `recordDiagnostics` shared writer to model a dispatch that was already in flight completing afterwards. The mark produces the expected zero counts; the later writer restores `blocking: 1, errors: 1` and removes the disposition annotation.

Cause: `reconcileWidgetDisposition` (`clients/widget-state.ts:326-327`) calls `commitDiagnostics` directly. It never reserves or validates a `diagnosticsWriteGuard` / `runnerWriteGuard` token, while the production pipeline writes the async dispatch result through `recordDiagnostics` (`clients/pipeline.ts:1547`). A dispatch admitted before the mark therefore commits its pre-mark result after it, and the user sees the stale count again until another dispatch.

**Remedy:** route reconciliation through the same admission-ordered write seam as dispatch, with an ordering token that an older in-flight result cannot supersede — or make dispatch completion re-derive dispositions before its write. Add a real asynchronous probe covering mark-during-dispatch, two marks in one tick, and an unrelated later producer.

### MEDIUM — the new regression test leaks every temporary home

The required CI `Unit tests` job is red on this, and it is **this PR's leak, not master's**:

```text
AssertionError: [tmp-hygiene] tests/config/tmp-fixture-hygiene.test.ts leaked 5 top-level entries: pi-lens-1616-IkPF7M,pi-lens-1616-XKlzwv,pi-lens-1616-dUlGgc,pi-lens-1616-tBtIKA,pi-lens-1616-xx5ypf
```

`tests/clients/widget-disposition-reconcile.test.ts:52` creates `tempHome`, but `afterEach` only unlinks `filePath` and never removes it. Remedy: own the root through `setupTestEnvironment` so it is registered and swept, rather than removing it by hand.

### LOW — two new guards are not mutation-protected

Built mutation deleting `!active.has(diagnostic) &&` at `clients/widget-state.ts:311`:

```text
MUTATION_ACTIVE_TEST_EXIT_CODE=0
Test Files  1 passed (1)   Tests  5 passed (5)
```

Built mutation replacing the host payload check at `clients/widget-state.ts:353` with `if (false)`:

```text
MUTATION_MESSAGE_GUARD_TEST_EXIT_CODE=0
Test Files  1 passed (1)   Tests  5 passed (5)
```

The active-set guard is vacuous for the current durable lookup — a matching false-positive/suppress entry cannot remain in `active` — so either remove the branch or add an invariant plus a test showing an independently possible active/matching state. The host message-type guard has no reject-direction test: add one proving a non-string message is rejected, or remove the validation if the event contract guarantees the field elsewhere. A guard that cannot be made red does not need to exist.

### Named output

- **Strong:** the shared widget record now has multiple writers, but the PR's own writers-by-axis table omits the write-order/admission axis. The durable record needs one explicit canonical ordering contract covering dispatch, scan, MCP and disposition reconciliation.
- Reconciliation accepts `_disposition` but derives behaviour from the durable store and the session set. A state-transition table over false-positive, suppress, defer, flagged, changed-content and unmark would make the third-state behaviour explicit and stop a count standing in for "not computed".
- The new `AGENTS.md` sentence records this instance rather than adding a one-line defect-shape screen. If it stays, reshape it around the multi-writer/shared-record recurrence and name the guard that catches future second-writer drift.

### Could not verify

- No full asynchronous dispatch harness with a delayed real runner. The direct probe proves the shared-writer overwrite; pinning exact admission/completion order needs a production `dispatchLintWithResult` scenario.
- No Windows path/case variants and no separate host process — the host test uses the real published payload and real widget store in one Vitest process with a hand-rolled event adapter, so process isolation is not established.
- No test covers the issue's `suppress` / won't-fix disposition through the widget path. The new tests cover false-positive, flagged, changed-content recovery, host delivery and fallback telemetry.
- The full suite was not run locally.

### What ran

`npm run build` exit 0. The regression file: 5 passed. Existing disposition/widget-state/index-wiring suites: 181 passed, 1 skipped. **Red-first confirmed**: with the new test retained and production sources restored from `origin/master`, 4 tests failed with the expected stale-count, missing-flag and missing-reconcile failures. `node scripts/ci-verdict.mjs 2975`: `Lint & type-check` pass, required `Unit tests` **fail**, not mergeable.

---

apmantza:
## Round-2 verify — FIX ROUND REQUIRED. Keep `refs #1616`.

Head `c5fb824d0`. The tokened dispatch-versus-mark race **is fixed** — but round 2 introduced a new blocking-diagnostic suppression regression on the same seam, and it came from a remedy this review prescribed. Required `Unit tests` was still pending on the head when read, so the PR is also not mergeable by evidence yet.

### HIGH — weak suppress marks can hide blocking diagnostics

Probe through the real `recordDiagnostics` and `markDisposition` paths:

```text
{"before":{"blocking":1,"errors":1},
 "afterSuppress":{"blocking":0,"errors":0},
 "diagnostics":[{"severity":"error",...,"disposition":"suppress"}]}
EXIT_CODE=0
```

`clients/widget-state.ts:298-320` now applies `getDisposition(cwd, strict) ?? getDisposition(cwd, weak)` directly, because round 2 deleted the `applyDispositions` active set. A **weak** `suppress` entry is therefore accepted for a **blocking** error. The existing disposition contract requires blocking findings to carry a strict anchor; the widget now reads confidently clean while the blocking finding is still valid.

**This is our error, not the fixer's.** Round 1 judged that guard vacuous — *"a matching false-positive/suppress entry cannot remain in `active`"* — and the brief told the fixer that deleting was the preferred outcome. The active set was not doing membership bookkeeping; it was carrying the strict/weak anchor rule for blocking findings. The guard had a recurrence to name, and we could not see it because the mutation stayed green: deleting the branch changed no *tested* behaviour, since no test exercised weak-suppress against a blocking error. A guard whose mutation is green is a guard with a missing test at least as often as it is a vacuous guard, and this review called it the wrong way.

**Remedy:** restore the shared disposition filtering, or derive the active disposition with the same strict/weak and blocking rules before adding the widget annotation. Add a real blocking-error + weak-suppress regression test, and mutate the new branch to require a red — the missing test is the actual deliverable here, since its absence is what made the deletion look safe.

### What is genuinely fixed

The admission-ordered token seam closes the round-1 HIGH: a dispatch admitted before the mark no longer commits its pre-mark counts afterwards. The `pi-lens-1616-*` leak is gone and the root is registered through `setupTestEnvironment`. `refs #1616` stands; do not restore `closes`.

---

Issue #747, opened 2026-07-22
lens_diagnostics can OOM and shut down WSL when cwd is $HOME


Blue-B:
### What happened?

I hit a full WSL shutdown after `lens_diagnostics` started a fresh `jscpd` scan from `/home/shell`.

The call had five explicit file paths plus `maxProjectFiles: 20` and `maxLspFiles: 20`, but `fetchFreshProjectDiagnostics()` still used the session cwd as `analysisRoot`. `jscpd` scanned `.` from `/home/shell`, which contains multiple unrelated repos.

The process reached 44,342,692 kB anon RSS, filled the 16 GiB WSL swap, triggered the global OOM killer, and took down `init.scope`. Every pi/herdr session in that WSL instance was lost.

This looks related to #462, but this path was on the native WSL filesystem rather than `/mnt/<drive>`. The existing home and large-workspace guards do not appear to cover the fresh heavyweight refresh path.

### Steps to reproduce

1. Start pi with cwd set to `$HOME`.
2. Call `lens_diagnostics` with a file-only scope:

```json
{
  "mode": "full",
  "paths": ["/home/user/project/file.ts"],
  "refreshRunners": "cheap",
  "maxProjectFiles": 20,
  "maxLspFiles": 20
}
```

3. `formatFullMode()` calls `fetchFreshProjectDiagnostics(cacheManager, cwd, ...)`. That function resolves the cwd as `analysisRoot`, and `JscpdClient` runs `jscpd .` from there. The explicit `paths` value only limits the reported results.

### Expected behavior

A file-only `paths` scope should not start a whole-home heavyweight scan. At minimum, the fresh runner should refuse `$HOME` or other obvious non-project roots unless explicitly overridden.

My local workaround rejects the refresh when `analysisRoot` is `$HOME` and runs `jscpd` under a 4 GiB address-space limit.

### pi-lens version

3.8.71

### Relevant logs

```text
jscpd invoked oom-killer
Out of memory: Killed process 227320 (jscpd) total-vm:63393268kB, anon-rss:44342692kB
Free swap  = 0kB
Total swap = 16777216kB
init.scope: Failed with result 'oom-kill'
init.scope: Consumed 1h 44min 23.483s CPU time, 44.0G memory peak, 15.1G memory swap peak.
```

### Area

None

apmantza:
many thanks for the report @Blue-B.

PR #2977, opened 2026-09-11
feat(diagnostics): lens_diagnostics accepts an explicit analysis root below the home ceiling (closes #2053)

### LINKED ISSUE #2053: Allow an explicit analysis root so a $HOME session can still scan one project
### What should be improved?

`fetchFreshProjectDiagnostics()` in `clients/project-diagnostics/fresh-fetch.ts` refuses every heavyweight analyzer when the session cwd is at or above `$HOME`, which #749 added for good reason. The refusal is unconditional, so there is currently no way to run those analyzers at all from such a session, even for one explicitly named project.

I would like `lens_diagnostics` to accept an explicit analysis root, so the guard can be satisfied by naming a project instead of by relocating the session.

The distinction I am drawing is against this line in #749:

> like the review-graph builder, no substitute root is guessed [...] there is no safe fallback to walk from a bare `$HOME` cwd

Guessing a root and being handed one are different. If a caller names `/home/me/myrepo`, the hazard that motivated the guard, walking every repository under home, does not apply to that call.

### Why?

I run pi with cwd at `$HOME` because I work across several repositories in one session. That is the same setup as the #747 reporter, so I am not asking for the guard to be relaxed.

The effect is that knip, jscpd, madge, gitleaks, govulncheck, trivy and dead-code never ran once during a long refactor. I only noticed at the end, when I ran knip by hand against one package and it reported findings that `lens_diagnostics mode=full refreshRunners=all` had never surfaced. The skip notice is honest and says absence is not a clean verdict, but in practice a session that never sees a heavyweight finding reads as a clean project.

Passing `paths` does not help, which #749 explains:

> the explicit `paths` scope only filters *reported* results, never what the analyzers walk

So the one input that already names a safe subtree cannot be used to satisfy the ceiling.

The workaround I settled on is a shell script that runs knip and cargo-machete against a named directory, outside pi-lens entirely. It loses the caching, dedup and ignore handling that pi-lens does properly, which is why I would rather this lived upstream.

Reproduce:

1. Start pi with cwd `$HOME`.
2. Call `lens_diagnostics` with `mode=full`, `refreshRunners=all`, and `paths` pointing inside one project under home.
3. All seven analyzers come back cold with `analyzersUnsafeRoot`.

pi-lens 4.1.1, WSL2 Ubuntu, node v22.22.2.

### Proposed approach (optional)

Add an optional analysis root to the `lens_diagnostics` input, for example `analysisRoot`, defaulting to the session cwd so present behavior is unchanged.

In `fetchFreshProjectDiagnostics()`, apply `isAtOrAboveHomeDir` to the resolved root rather than to the session cwd, and reject the call when the supplied root is itself at or above home, or is not a directory. The ceiling stays exactly where it is; only the value it examines becomes explicit.

Two details worth deciding, and I do not have a strong view on either:

- Whether the root should be required to contain a project marker such as `.git`, which would block a caller from naming a directory that happens to sit one level under home and still hold many repositories.
- Whether `paths` outside the supplied root should error or be filtered out.

Files I would expect to touch: `clients/project-diagnostics/fresh-fetch.ts`, the `lens_diagnostics` input schema, and the unsafe-root tests in `tests/clients/project-diagnostics/fresh-fetch.test.ts` and `tests/tools/lens-diagnostics.test.ts`.

I am happy to open a PR if the direction is acceptable, and equally happy to be told the ceiling should stay tied to the session cwd.

apmantza:
Closes #2053.

### Invariants

| ID | Proposition |
| --- | --- |
| I1 | The explicit analysis root is canonicalized once before consumers use it. |
| I2 | A canonical explicit root is safe only when it is strictly contained by the canonical home directory. |
| I3 | Unsafe or unavailable explicit roots refuse heavyweight analyzers, never fall back to cwd, and never clamp. |
| I4 | Root validation distinguishes safe, unsafe, and undecided results; undecided is never treated as safe. |
| I5 | Every heavyweight consumer of the explicit analysis root reads the same canonical root and containment result. |
| I6 | A cwd walk has a separate policy because it intentionally walks cwd, not the explicit analysis root. |

### Writers by axis

| Actor | Root identity decided | Timing | Decision |
| --- | --- | --- | --- |
| `fetchFreshProjectDiagnostics` explicit-root validation | Explicit analysis root | Before heavyweight analyzer dispatch | Canonicalizes the requested root and decides strict home containment. |
| `fetchFreshProjectDiagnostics` cwd fallback validation | Session cwd as analysis root | Before heavyweight analyzer dispatch | Uses the cwd-at-or-above-home policy when no explicit root exists. |
| `formatFullMode` project walk policy | Session cwd | During full-mode response formatting | Independently decides whether cwd-based project walking is unsafe. |
| Heavyweight analyzer consumers | Explicit analysis root | After fresh-fetch validation | Read the fetch seam's canonical root and policy; they do not decide again. |

### Cell list

| Root identity | Consumer | Current verdict | Intended verdict |
| --- | --- | --- | --- |
| Explicit project child below home | Heavyweight analyzers | Safe | Safe, from one shared validation result |
| Explicit project child below home | Full-mode explicit-root response | Safe | Safe, same root and policy as analyzers |
| Session cwd at home | Cwd project walk | Unsafe | Unsafe, from the separate cwd-walk policy |
| Explicit project child below home | Any future explicit-root consumer | Undecided / may re-derive | Safe, by reading the shared validation result |
| Missing explicit root | Heavyweight analyzers and response | Refused | Refused, with no fallback or clamp |
| Explicit root at or above home | Heavyweight analyzers and response | Refused | Refused, with no fallback or clamp |

The table agrees with the prescribed remedy. The defect is duplicated authority for the explicit root, so the validation result is the single writer and shared value. `isSameOrWithin` remains the containment primitive. `isAtOrAboveHomeDir` remains separate because it answers the intentionally different cwd-walk question.

## Summary

The full diagnostics path now returns one typed validation result containing the canonical analysis root and its policy. Heavyweight consumers use that validated root, while the cwd project-walk check remains a separate policy for a different root.

Safe explicit roots remain accepted. Unsafe or unavailable explicit roots remain refused, without fallback to cwd or clamping.

## Premise

The required compiled real-MCP probe used `HOME` and `PI_LENS_HOME` below `$PWD/.probe-home`, ran with cwd equal to that home, created `project` below it, and called `pilens_diagnostics` with `analysisRoot=project`.

Observed pre-fix output:

```text
No files diagnosed yet this session.
fetched fresh this call: opengrep (6229ms), madge (0ms).
project file walk skipped: the working directory resolves at or above the home directory
result ok
```

The explicit project root proceeded to heavyweight analyzers while the same response refused the cwd project walk. The probe used only the isolated `.probe-home` tree.

## The fix

`fetchFreshProjectDiagnostics` now creates `AnalysisRootValidation` with `safe`, `unsafe`, or `undecided` state. It carries the canonical root when established, and preserves unavailable explicit roots as `undecided` rather than permitting them.

The explicit-root containment check reuses `isSameOrWithin`. The full-mode details read the shared validation state. The cwd walk still uses `isAtOrAboveHomeDir` because it intentionally evaluates cwd, a different root.

## Tests

The regression test `shares one validated explicit root with every heavyweight consumer (#2977)` in `tests/clients/project-diagnostics/fresh-fetch.test.ts` checks the shared safe result and the canonical root received by knip, jscpd, and madge.

The existing tests `runs the reporter's home-cwd reproduction for an explicit project root (#2053)`, `refuses an explicit analysis root AT or ABOVE home (#2053, #749)`, and `refuses a missing or non-directory explicit analysis root (#2053)` preserve the accepted, unsafe, and undecided cells.

The existing test `mode=full: an unsafe-root fresh-fetch renders the home-directory refusal, not the generic cold list (#747)` in `tests/tools/lens-diagnostics.test.ts` preserves the response contract.

Red-first transcript before the source fix:

```text
tests/clients/project-diagnostics/fresh-fetch.test.ts (39 tests | 1 failed)
shares one validated explicit root with every heavyweight consumer (#2977)
AssertionError: expected undefined to deeply equal { state: 'safe', ... }
TEST_EXIT=1
```

Mutation transcript after the fix, with the built guard changed from state !== "safe" to false in clients/project-diagnostics/fresh-fetch.test.ts's covered seam:

```text
tests/clients/project-diagnostics/fresh-fetch.test.ts (39 tests | 4 failed)
refuses to run anything when cwd IS the home directory (#747)
refuses an explicit analysis root AT or ABOVE home (#2053, #749)
refuses every explicit root that is not strictly below canonical home (#2977 F1)
MUTATION_EXIT=1
```

## Blast radius

```text
`lens_diagnostics`.execute
  formatFullMode
    + reads `analysisRootValidation` for explicit-root policy/details
    fetchFreshProjectDiagnostics
      + canonicalizes and validates `analysisRoot` once
      knipClient.analyze(analysisRoot)
      jscpdClient.scan(analysisRoot)
      depChecker.scanProject(analysisRoot)
      gitleaksClient.scan(analysisRoot)
      govulncheckClient.analyze(analysisRoot)
      opengrepClient.scan(analysisRoot)
      trivyClient.scan(analysisRoot)
      deadCodeClient.analyze(analysisRoot)
      cacheManager.writeCache(..., analysisRoot, ...)
```

The affected dependents are the nine heavyweight analyzer lanes and the full-mode cold-note/details renderer. Existing LSP and session-root consumers are outside this change. The module blast-radius map is unavailable in this local tool environment; the call-tree above is the checked seam map.

## Class sweep

The sweep covered `analysisRoot`, `isSameOrWithin`, `isAtOrAboveHomeDir`, and all `fetchFreshProjectDiagnostics` callers. The explicit-root decision remains at the fresh-fetch seam. The only separate containment decision is the cwd project-walk policy in `formatFullMode`, which evaluates a different root by design. No duplicate containment helper was added.

## Observability

The existing bounded record literal kind: "lens-diagnostics-analysis-root-rejected" remains the durable evidence for unavailable and unsafe roots. Its counts are bounded at 1, 2, 4, and 8, with the requested absolute path as subject. The result continues to expose analysisRootError, unsafeRoot, coldReasons, and analyzersUnsafeRoot; the shared validation result is also included in full-mode details. No new failure path or unbounded record was added.

## Test assessment

The tests drive the real `fetchFreshProjectDiagnostics` orchestration with client doubles only at external analyzer-client boundaries. They assert independent calls, cache writes, and refusal side effects. The tool test uses the real production formatter and schema while mocking only the heavyweight external-fetch seam. The built-output mutation probe verifies that the acceptance guard itself is required.

## Verification

Commands run locally:

```text
npm run build
BUILD_EXIT=0

PI_LENS_HOME=$PWD/.probe-home node_modules/.bin/vitest run tests/clients/project-diagnostics/fresh-fetch.test.ts tests/tools/lens-diagnostics.test.ts --configLoader runner
Test Files 2 passed (2)
Tests 191 passed (191)
TEST_EXIT=0

git diff --check
exit 0
```

npm run build was rerun after the built-output mutation and exited 0. git fetch origin master exited 0; the fetched tip was f0ba2f465.

## Skipped

The full 1,095-file population was skipped as required. Optional project-marker and `paths` policy questions were skipped because they are non-goals. CI remains responsible for the exact-head required checks, the complete governance population, and the repository's advisory PR-body check.


apmantza:
## Round-2 verify (adversarial, fresh reviewer)

Head reviewed: `2a92a7e51ad9720c5816dc8d3f9601a423d567e8`. Build, `tsc --noEmit` and `fmt:check` all exit 0. `node scripts/ci-verdict.mjs 2977` read the exact head: every gating check concluded success; the advisory PR-body check is red.

### Round-1 finding disposition

| Finding | Disposition | Deciding evidence |
| --- | --- | --- |
| F1 escape matrix | **FIXED** | The real compiled `pilens_diagnostics` refused the ceiling, ancestor, `..`-traversal, absolute-sibling and symlink-to-sibling roots. The symlink case returned the honest heavyweight-analyzer skip notice. A valid relative `project` root was accepted. Pre-fix and mutation probes both made the new escape tests red. |
| F2 missing root | **FIXED** | A missing explicit root returns `isError: true` with `the explicit analysis root ... is unavailable or ENOENT`. No fallback, no clamping. Existing-but-unsafe roots return the explicit refusal and do not run analyzers. |
| F3 telemetry | **PARTIALLY FIXED** | Ten repeated missing-root calls returned `isError: true` every time; the durable record used kind `lens-diagnostics-analysis-root-rejected` with the requested absolute path as subject and counts 1, 2, 4, 8 — correctly bounded, not one record per call. The contract half is still open: the PR-body gate exits 1. |

Containment was specifically re-attacked and cleared: the fix uses the repository's existing `isSameOrWithin` helper (`clients/lsp/server.ts:252`, `path.relative`-based), not a string prefix, and it is applied after `path.resolve` + `fs.realpathSync`. No duplicate containment helper was added. Red-first and mutation coverage are real: removing the explicit guard from source produced two failures in `tests/clients/project-diagnostics/fresh-fetch.test.ts`, and setting compiled `unsafeExplicitRoot = false` produced the same two.

### New finding

**MEDIUM — the explicit-root and project-walk containment decisions are still made twice, independently.**

Concrete instance, from the compiled real-MCP call with `HOME` and `PI_LENS_HOME` under `.probe-home`, `cwd=$HOME`, `analysisRoot=project`: the explicit project root was accepted for heavyweight analyzers while the same response also said `project file walk skipped: the working directory resolves at or above the home directory`. That is the same two-state output as round 1's `analyzersUnsafeRoot: false` versus `projectWalkUnsafeRoot: true` asymmetry — the symptom is narrower now, but the structure that produced it is unchanged.

- `fetchFreshProjectDiagnostics` independently canonicalizes `homeRoot` and `analysisRoot` and evaluates `unsafeExplicitRoot` — `clients/project-diagnostics/fresh-fetch.ts:298-306`.
- `formatFullMode` independently evaluates `walkUnsafeRoot` from `cwd` via `isAtOrAboveHomeDir` — `tools/lens-diagnostics.ts:2704-2713`.

Today's scopes happen to be safe because the project walk refuses the session cwd. That is a coincidence of the current consumer set, not an invariant: the next consumer of the explicit root re-derives the policy and can re-derive it wrong, which reopens the escape class this PR exists to close.

**Remedy shape:** produce one root-validation result at the `fetchFreshProjectDiagnostics` seam carrying the canonical root and its containment decision, and have every consumer of that root read it rather than recompute it. Keep the cwd walk policy separate only for consumers that intentionally walk `cwd`, and document that it is a *different root* rather than a second opinion about one analysis operation. Add a test asserting that the explicit-root path and every heavyweight consumer read the same validated root and the same policy.

The deletion test supports folding rather than distributing: removing that shared result would concentrate the independent root decisions back into `fresh-fetch.ts` and `lens-diagnostics.ts` — which is precisely the recurrence this verify found.

### Also open

The PR-body gate reports `PR body Observability must name a record literal ...` plus missing test references for the quoted guard and output snippets. Exit 1.

### Could not verify

The full 1,095-file population was not run, per review scope. The issue's optional project-marker and `paths` policy questions are not acceptance criteria and were not treated as such; the required explicit-root behaviour is implemented and the seam-sharing contract above is what remains.

One suite failure during the targeted population was master-side noise, not this PR: `tmp-fixture-hygiene` leaked `pi-lens-jscpd-*` entries. That is the leak PR #3008 addresses, and it was not counted against this branch.

---

Issue #2982, opened 2026-09-11
Tree-sitter inline blocker has no retire path once its own file changes


rnavarro:
### What happened?

A `ts-incomplete-assertion` blocker was re-served at every turn end for the rest of a session, against a file that grep, `lens_diagnostics`, and a passing test all read as clean. I was working through a round of test fixes in a TypeScript project and had already corrected the flagged line; the agent then spent about ten turns and several full test runs trying to clear a finding that no longer described anything on disk.

Tracing it, a tree-sitter-sourced inline blocker has no path that can retire it once its own file stops matching the verdict:

1. `rules/tree-sitter-queries/typescript/incomplete-assertion.yml` is `inline_tier: blocking`, and `clients/dispatch/runners/tree-sitter.ts` stamps `tool: "tree-sitter"`.
2. `clients/pipeline.ts` carries those tool ids into `inlineBlockerSources`, recorded on the entry via `runtime-tool-result.ts` with `stale: false`.
3. `RuntimeCoordinator.retireInlineBlockerOnConfirmedClean` retires only when `coveredSources` is a superset of the record's `sources`.
4. `coveredSourcesForCheck` in `tools/lsp-diagnostics.ts` builds `coveredSources` from `getServersForFileWithConfig`, so it can name `"lsp"` and auxiliary language-server ids.
5. No registered server has id `"tree-sitter"` (`grep -c 'id: "tree-sitter"' clients/lsp/server.ts` returns 0), so step 3 cannot be satisfied and `lens_diagnostics` returns `false` from the retire path however many times it runs clean.

The other invalidation paths do not reach it either. `clearInlineBlockers` fires on a later dispatch of the same path, and the dispatch that would clear it was keyed on other files (the #1561 cross-file shape). `reconcileInlineBlockers` drops entries only when the file is gone, and the file exists. The #1950 delivery cap requires `stale: true`, which `recordInlineBlockers` does not set.

The turn-end freshness sweep is the one gate that could notice, and its third condition excludes the record: `clients/blocker-freshness.ts` checks an entry only when its `sources` are all `"lsp"`. The #1631 review F4 reasoning for that condition is about the import axis, that an ast-grep secret match does not stop being true because a file it imports was edited, and the same condition also excludes the record's own file from the check. A tree-sitter verdict about a file's own syntax does stop describing that file when the file's own bytes change.

### Steps to reproduce

In a TypeScript project with the bundled tree-sitter rules active:

[...]

Issue #2523, opened 2026-09-03
turn_end/agent_end/session_start must not block: per-hook wall budgets + bounded() at every hook await + off-hook delivery for the remainder (refs #2504, #2509)


apmantza:
## Contract (maintainer, 2026-09-03)
Only the write/edit `tool_result` path may block the host; `session_start`, `turn_end`, `agent_end`, `agent_settled` and read-only `tool_result` must be bounded (deadline AND the hook's abort signal) and small, or deliver off-hook.

## Evidence (class sweep, read-only, 2026-09-03; `~/.pi-lens/latency.log` 2026-09-01T22:08Z → 09-03T07:12Z, 12,156 records, all pre-#2509)
| hook (`type:"tool_result"`) | n | p50 | p90 | max |
|---|---|---|---|---|
| **turn_end** (NON-EDIT) | 47 | **3687 ms** | **14246 ms** | 191281 ms |
| write (EDIT) | 81 | 2059 | 2614 | 12325 |
| edit (EDIT) | 46 | 2042 | 3199 | 6227 |
| agent_end | 7 | 124 | 10043 | 10043 |

Excluding every stale-worklist row (`fileCount ≤ 5`, n=42): turn_end p50 **3587 ms**, 22/42 over 3 s, max **13121 ms on a ONE-file turn**. #2509 removed the 154-file inflation; it did not touch the baseline.

## Mechanism
`wrapSessionEventHandler` (`clients/session-event-guard.ts:203`) absorbs stale-ctx throws and adds NO deadline and NO abort. Every bound lives at a leaf (spawn timeout, LSP wait); a dependency that wedges before a leaf is unreachable by all of them. Probes against the real handlers:
- wedged dependency on `turn_end`: **never returned** (400 000 ms harness ceiling);
- same with the ambient abort fired at t=2 s exactly as `index.ts:2693` wires it: `still-blocked after 30011ms` — Escape does not release the hook;
- EDIT path, 3 wedged formatters on one file: `still-blocked after 45011ms` — `runFormattersWithConcurrency` (`clients/format-service.ts:154-205`) is a sequential loop with per-item 30 s timers, no aggregate cap, no signal in the race, `_concurrency` unused.

Awaits with NO bound, by hook:
[...]

Issue #2275, opened 2026-08-27
Widget-footer dependency-drift demotion has no delivery cap (sibling of #1950)


apmantza:
## Evidence

PR #1950's fix caps how many times a dependency-drift-demoted `RuntimeCoordinator` inline blocker re-serves before retiring (`clients/runtime-coordinator.ts`'s `retireDemotedDependencyDriftBlocker`, `clients/runtime-turn.ts`'s delivery loop). The widget-footer store carries the identical demotion shape through a completely separate path: `clients/widget-state.ts`'s `markWidgetFileBlockersStale(filePath, "dependency-drift")` (line 1094) sets `staleReason = "dependency-drift"` on a widget diagnostic (line 1015), and `getWidgetBlockingFilesForSweep` (line 1048) feeds it into the same `sweepInlineBlockerFreshness` population `blocker-freshness.ts` drives from `runtime-turn.ts`.

That demotion never retires. It re-derives every turn from the widget's own diagnostic list rather than a stored delivery count, so there is no equivalent of `InlineBlockerRecord.staleDeliveryCount` to cap against.

## Root cause (hypothesis)

Same class as #1950: a demoted-but-confirmable advisory record with no bound on delivery count. Not fixed by #1950 because the widget store is a DIFFERENT store (`widget-state.ts`'s own diagnostic list, not `RuntimeCoordinator._pendingInlineBlockers`), with its own render path and its own re-derivation-not-latch design (per its module doc, a widget diagnostic's `stale` bit is re-derived every sweep rather than stored as a one-way flag).

## Acceptance criteria

- A delivery cap (or explicit justification for why the widget surface doesn't need one — e.g., if its rendering is already bounded some other way, such as only ever showing the CURRENT footer state rather than accumulating turn-end history) on a dependency-drift-demoted widget diagnostic.
- If a cap is added, it should reuse `DEPENDENCY_DRIFT_MAX_DELIVERIES` (`clients/blocker-freshness.ts`) rather than inventing a second number, unless the widget's render cadence gives a reasoned different value.
- Red-first test reproducing repeated widget-footer delivery of the same demoted diagnostic across turns.

## Cross-links

Refs #1950, #1944, #1634, #1631.

Issue #2042, opened 2026-08-24
CI: vitest npm child SIGKILLed with ~13 GB headroom (exit 137, not OOM) - kill sender unidentified


apmantza:
## Evidence

Three CI runs today failed with exit code 137 (SIGKILL from OOM killer):

- Run 32714487380: job duration 115s vs ~4.5min norm, zero test failures logged
- Run 32767280824: same pattern, killed mid-suite after many tests had already passed
- Run 32712887351: succeeded on rerun

All three reruns passed, confirming transient memory pressure rather than a deterministic leak.

## Root cause

The vitest fork pool runs at `maxWorkers: "50%"` of available cores. Each worker loads tree-sitter WASM grammars (~50 MB each), LSP fixtures, and the full module graph. On ubuntu-latest runners (~7 GB RAM), when several heavy suites hit peak allocations simultaneously, total memory crosses the OOM threshold.

## Why it's getting worse

Every PR adds tests. Recent additions alone: ~40 new test cases across delivery-gate ratchet, posix-group-kill real-binary suite, verify-binary-semantics, mutation-seam, and freshness-sweep — several spawning real child processes with multi-second lifetimes.

## Fix directions

1. Reduce CI maxWorkers to 25% (trade speed for stability)
2. Split the suite into two shards running on separate runners
3. Use a larger runner (4-core → 8-core)
4. Identify and cap the heaviest suites' worker count individually

## Acceptance criteria

- [ ] Chosen mitigation implemented
- [ ] No OOM kills in 10 consecutive master CI runs
- [ ] Suite completion time stays within acceptable bounds (< 10 min)

*This issue is AI-generated, with the maintainer supervising the process.*

apmantza:
Investigation correction and proposed next step:

- CI does not use the local `"50%"` setting. `vitest.config.ts:51-59` leaves `sharedMaxWorkers` undefined under CI.
- Vitest 4 resolves that to `availableParallelism() - 1`. An ordinary four-core Ubuntu runner therefore gets three default-project workers.
- The specialized projects do not overlap. Group ordering runs default, grammar-heavy, timing-sensitive, LSP-spawn-heavy, and wall-clock-budget sequentially. Their caps are already 2, 2, 1, and 1 after the default phase.
- Each worker also receives a 4 GiB V8 old-space ceiling. That is not an aggregate RSS budget, so increasing it would worsen the failure class.

[...]

PR #2983, opened 2026-09-11
fix(diagnostics): check own-file drift for non-LSP inline blockers (refs #2982)

rnavarro:
## Summary

Maintainer takeover of rnavarro’s PR #2983, preserving the contributor’s scope and authorship.

Round 4 removes the publication-to-baseline race for non-LSP inline blockers. `runPipeline` carries the size and sha256 of its existing file read beside inline-blocker evidence. `recordInlineBlockers` stamps both values synchronously, so delayed work cannot attach evidence to a newer verdict. Missing baselines and expired bounds emit bounded, file-identified `self-drift-unverifiable` records. Hash-budget exhaustion remains sweep metadata.

The commit gate still counts demoted records. Same-size edits demote, mtime-only edits with identical bytes remain authoritative, and deferred mutation still demotes safely. Dependency-drift and past-EOF behavior remain unchanged. Non-goals are #3011, new provenance classes, dependencies, daemon changes, and broad refactors.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:diagnostics
- [x] area:observability
- [x] area:tests
- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:session
- [ ] area:config
- [ ] area:security

## Checklist

- [x] Targeted test files pass locally after `npm run build`; the full suite is CI’s job.
- [x] New regression behavior is red on the pre-fix mutation.
- [x] New guards and filters are mutation-tested.
- [x] npm run lint passes
- [x] Changelog fragment is valid
- [x] Commit subject includes refs #2982

## Tests

Lifecycle invariants and writers-by-axis:

| Axis | Publication writer | Baseline writer | State rule | Failure rule |
| LSP dependency | `recordInlineBlockers` | pipeline result, when present | own file plus forward imports | demote, preserve commit gate |
| Non-LSP self | `recordInlineBlockers` | same synchronous call | own file, content-confirmed, re-armable | demote on confirmed bytes; preserve on uncertainty |
| Past EOF | `recordInlineBlockers` | cited lines in pipeline result | sibling gate owns its demotion | preserve existing retirement behavior |
| Deferred mutation | `recordInlineBlockers` | pipeline bytes before deferred mutation | later changed bytes are stale | safe demotion, never clean authority |

Conflict and overlap map: the pipeline result is the single producer of inline-blocker content evidence. `RuntimeCoordinator` is the single inline-blocker store writer. `runtime-tool-result.ts` only forwards evidence. `blocker-freshness.ts` only reads evidence and demotes. `runtime-turn.ts` only supplies the turn-end signal and records sweep metadata. Widget rows remain a separate injected population. No overlap with #3011’s all-LSP own-file extension. The local composition merge resolved concurrent AGENTS.md and hook-await registry changes; no production overlap remained.

New or edited tests:

- `tests/clients/blocker-freshness.test.ts` — `demotes on a same-LENGTH content change (one character swapped)` proves the hash tier; `does NOT demote when mtime moved but the content did not` proves mtime-only safety; `records a distinct bounded record when the self-drift bound expires` and the existing missing-baseline case prove bounded real-sink identities.
- `tests/clients/blocker-freshness-delivery-cap.test.ts` — edited baseline setup to use atomic pipeline-shaped evidence while preserving deferred-mutation delivery behavior.
- `tests/clients/blocker-freshness-widget-population.test.ts` — edited duplicate-population setup to use atomic evidence.
- `tests/config/hook-await-bounds.test.ts` — removed the deleted second-read registry rows and registered the remaining bounded sweep call sites.

Red-first and mutation evidence: the pre-fix delayed-baseline design is represented by the removed second-read seam; the atomic replacement is pinned by the same-size regression. Neutering the hash guard produced this whole-file red:

```text
FAIL tests/clients/blocker-freshness.test.ts > blocker freshness — self-drift on non-LSP provenance > demotes on a same-LENGTH content change (one character swapped)
AssertionError: expected +0 to be 1
```

Local targeted run after restore: `4` files, `40` tests, `3` passed test files, `1` pre-existing Java grammar-resolution failure because grammar prefetch was offline.

```text
$ git show -s --format='%H %s' origin/master
6e72947489fe421a6c07e473c8fac8313c75189a Merge pull request #2973 from apmantza/fix/2542-stale-test-verdicts
```

The changed tests passed.

Config governance: `2` files, `14` tests passed. Required test populations still to be run by the orchestrator include overlapping master-change suites and PR #3009 composition where network access permits.

## Test assessment

- `tests/clients/blocker-freshness.test.ts` uniquely pins the freshness state machine, content confirmation, re-arm, commit gate, and bounded identities. No touched test is redundant.
- `tests/clients/blocker-freshness-delivery-cap.test.ts` uniquely pins deferred mutation and delivery-cap composition. No test is removed.
- `tests/clients/blocker-freshness-widget-population.test.ts` uniquely pins the widget duplicate population. No test is removed.
- `tests/config/hook-await-bounds.test.ts` uniquely pins hook await registry completeness. Deleted rows correspond to deleted production awaits.

## Blast radius

```text
PipelineResult.inlineBlockerFileContent
  + runPipeline
    + dispatchPipelineAnalysis
      + handleToolResult
        + RuntimeCoordinator.recordInlineBlockers
          + getInlineBlockersSnapshot
            + sweepInlineBlockerFreshness
              + runtime-turn handleTurnEnd
```

`clients/pipeline.ts` affects all pipeline callers and optional result doubles. `clients/runtime-tool-result.ts` affects the tool-result entry point and observed-mutation shared analysis. `clients/runtime-coordinator.ts` affects every inline-blocker consumer, including git guard, turn-end rendering, and retirement. `clients/blocker-freshness.ts` affects the inline and injected widget populations. `clients/runtime-turn.ts` affects turn-end delivery metadata. `clients/degradation-ledger.ts` affects the sorted kind registry and durable sink. The baseline hash is computed from content already in memory; it adds one sha256 per blocking pipeline result and no filesystem read or await.

## Class sweep

Defect shapes screened: AGENTS.md defect shapes 6, 7, 9, 16, 17, and 24; the test-authoring screens; real-runner guidance; and the dispatch/cache/observability standing invariants.

Whole-tree sweep searched the clients, tools, mcp, scripts, index, and test trees for `InlineBlockerRecord`, `PipelineResult`, `recordInlineBlockers`, `recordedSize`, `recordedHash`, the removed helper names, and `inlineBlockerFileContent`. The removed helpers have zero remaining code or test references. The population sweep found the RuntimeCoordinator map and the injected widget rows; both remain covered. Pipeline result doubles are optional-field compatible, including `tests/index-integration.test.ts` and whole-module pipeline mocks.

Consolidation verdict: keep pipeline evidence, runtime storage, freshness policy, and turn-end delivery distributed because each seam owns a separate lifecycle boundary. Fold no family in this fix round.

## Observability

The new runtime record kind is `self-drift-unverifiable`. Subjects are `inline-blocker:<displayPath>#missing-baseline` and `inline-blocker:<displayPath>#bound-expired`. `hash-budget-exhausted` remains the `blocker_freshness_sweep` metadata signal and is not a per-file ledger record. Old ledger records continue to parse through the existing bounded schema.

## Verification notes

npm run build, npm run build:dist, npm run lint, npm run fmt:check, and changelog validation pass locally. The config sweep reports environment failures for child-process/git probes under the sandbox, and the Java freshness test reports missing offline grammar data. GitHub discussion retrieval and fresh origin/master fetch were unavailable because GitHub DNS/authentication are unavailable in this environment.


apmantza:
Thank you for this — it is a well-built PR and the review said so repeatedly. I ran a full adversarial pass on it. One finding blocks, and it is in the direction your own git-guard test was reaching toward.

## What cleared, so you know it was actually attacked

- **Your predicate split is correct.** The reviewer built the case-class matrix rather than taking the prose, crossing every provenance shape against the widget-row shapes, and the implementation matches it at every cell. That mattered because this repo shipped two "equivalent" predicate copies earlier in this train that disagreed on 4 of 12 classes.
- **Your red-first proof is genuine.** Reverting `clients/blocker-freshness.ts` to master while keeping your tests produced five reds, and your three demotion cases each reported `expected +0 to be 1` at lines 395, 411 and 429. Your framing was also checked and held: the dependency-only and git-guard cases are contract pins, not padded red proofs, and you said so.
- **Your mutation claims hold.** `selfOnly=false` produced four reds; widening eligibility and neutering the chain predicate each reproduced the #1790 regression.
- **No second writer.** A grep for `markInlineBlockerStale(` finds only the declaration and the sweep call. You reused the writer and the reason, as you said.
- **The `"unknown"` call is right.** `pipeline.ts:1785-1792` produces it only for an untagged blocking diagnostic, and no producer establishes an alternative meaning.
- **The git-guard preservation is real**, proven independently: `gitGuard: true` after a self demotion.

## The blocking finding

**Self-drift permanently removes turn-end authority, and the delivery cap makes it irreversible.**

```
first : {"total":1,"kept":0,"revalidated":1,"alreadyStale":0}
second: {"total":1,"kept":0,"revalidated":0,"alreadyStale":1}
stale: true   gitGuard: true
```

`clients/blocker-freshness.ts:722-741` demotes an ast-grep, tree-sitter, mixed or `"unknown"` record on own-file mtime drift, and the next sweep skips it at `:697-699`. From then on the record is rendered only as a re-run advisory by `clients/runtime-turn.ts:1019-1021`; the freshness sweep never re-raises it. The existing delivery cap then retires it after three stale deliveries at `clients/runtime-turn.ts:1044-1070`.

So a formatter run, a checkout, a `touch`, or any out-of-band write that moves mtime without changing the bytes can walk a finding out of turn-end rendering permanently — including a hardcoded secret. Your git-guard test is correct and the commit gate does hold, which is why this is HIGH and not CRITICAL, but the gate does not make the turn-end loss safe.

The reviewer's named output sharpens why this bites: the delivery cap was designed for **recoverable** LSP dependency drift, and this PR admits security and unknown-provenance records into that same irreversible retirement path. Those are coupled policy decisions, and the coupling is new.

## Two remedy shapes, your call

**A.** Keep self-drift demotion for all the provenance you widened to, but preserve a re-raise path until a fresh producer check confirms the current bytes, and do not retire that class merely because the stale advisory cap was reached.

**B.** Restrict the widening to provenance whose producer contract actually proves that mtime drift invalidates the finding, leaving security and unknown-provenance blockers authoritative until their producer re-runs.

Either way it needs a real `handleTurnEnd` regression covering unchanged bytes, an mtime-only move, repeated turns, and the delivery-cap boundary. And whichever you pick, the PR body should state the provenance-specific retirement policy explicitly, because right now it is implied by the cap rather than written down.

Your mtime-only deferral to the shared helper was judged legitimate on its own terms — it preserves prior miss behaviour and you documented it at `clients/blocker-freshness.ts:414-425`. It just does not cure this particular direction.

Happy to take this as another round from you, or to have it picked up here if you would rather not carry it. CI is green on the head either way.

rnavarro:
Round 2 is pushed as `caee9c03d` and written up under `## Round 2` in the PR body.

Short version: you were right, and the root was smaller than either remedy shape suggested. The retirement coupling came from the reason string, not from the widening. `runtime-turn.ts` branches on `staleReason === "dependency-drift"` to apply the cap, and I reused that string in round 1 on defect-shape-24 grounds without checking it was load-bearing for retirement too.

Took remedy A. The axis now carries its own `"self-drift"` reason that re-arms rather than latching, modelled on `setInlineBlockerPastEofStale`, so a record whose bytes come back un-demotes itself and never needs a cap. Demotion is also content-confirmed first, applying #2449 round 2 F7's rule at its `size` tier, so a `touch`, a checkout restoring identical bytes, or a no-op formatter pass no longer demote anything. The hash tier is deferred with the reason stated: its baseline would have to be captured on the synchronous dispatch path.

The provenance-specific retirement policy is now written down as a table in the body rather than implied by the cap.

Both `handleTurnEnd` regressions you asked for are in `blocker-freshness-delivery-cap.test.ts`, covering unchanged bytes, an mtime-only move, repeated turns past `DEPENDENCY_DRIFT_MAX_DELIVERIES`, and the cap boundary. Six cases red on round-1 code, quoted in the body. 208 green across ten files on the seam.

apmantza:
Round 2 is in verify now, and I want to say first that your diagnosis is better than my remedy shapes were. I offered you a choice between "preserve a re-raise path" and "narrow the widening", and both of those accept the coupling as a given. You went and found that the coupling was a string — that `runtime-turn.ts` branches on `staleReason === "dependency-drift"` and you had inherited the cap by reusing the literal. That is the smaller root, and it is the right one. Recorded here as the lesson: a reviewer who names two remedies has usually stopped one level above the cause.

Two process notes, neither of them on you:

**Your CI was not running.** Every workflow on `caee9c03d` was sitting in `action_required` — the fork-PR approval gate, waiting on a maintainer. I have approved all eight, so Lint, Unit tests, close-keyword, OSV and the rest are executing now. You did nothing wrong; I was slow to notice. If a future push looks like it has no checks at all rather than pending ones, that is what it will be — ping the thread.

**Why this round gets a fresh verify rather than merging on green.** The house rule is that a round which only applies a prescribed remedy with quoted reds merges on green, but a round that adds mechanism gets attacked again. Yours adds a new `staleReason` enum value, a re-arming lifecycle, and a content-confirmation tier. That is mechanism, so it is being attacked. Not a judgment on the work.

The four angles it is being pointed at, so you can see them coming rather than be surprised by them:

1. **Whether `"self-drift"` falls through anything.** A new enum value is a new case for every consumer that branches on the old one — a `switch` without a `default`, a renderer that string-matches, a serializer writing the reason into a durable record something strict parses later. The verify is enumerating every `staleReason` consumer in the tree and stating what each does with the new value.

2. **Whether re-arming can oscillate unboundedly.** Latching has one virtue: it terminates. Re-arming trades that for recoverability, which is the right trade here, but it means a file whose bytes flip back and forth drives demote/un-demote repeatedly. The check is that nothing grows per occurrence — no map, no counter, and no observability record outside `recordDegradationOnce` / `incrementDegradationCount`.

3. **The obstruction direction, inverted.** Round 1 demoted too eagerly. The `size` tier may now demote too late: a same-length edit — one character swapped — changes the bytes without changing the size, so content confirmation says "unchanged" and the blocker stays raised against a file that genuinely moved. That is #2982's original complaint arriving from the other side. If it is real it is probably not a blocker on its own, but it is exactly where your deferred `hash` tier's remedy lives, so the deferral reason is being read as load-bearing rather than as a footnote.

4. **That the cap still fires for genuine dependency drift.** It was designed for a recoverable LSP case and must still retire that one.

Your `size`-tier reasoning and the explicit hash-tier deferral were the right shape to write down either way — a deferred tier with a stated reason is reviewable; an unstated one is a hole. Same for moving the retirement policy out of the cap's behaviour and into a table in the body.

I will post the verify result here. If it finds nothing, this merges.

apmantza:
Follow-up, and none of this is your work — it is ours. Approving your checks surfaced two repo-side problems, both of which were hitting you and neither of which you could have seen or fixed.

**1. Your Unit tests failure is an infra kill, not a real red.** Exit 137 with the memory watchdog reporting `KILLED WITH HEADROOM — no failing assertion, and no sample fell below 9607 MB of 15989 MB`. Zero assertion failures, zero TypeScript errors in the whole log. That is the shape tracked as #2042. I have rerun the job. Please do not go looking for a test to fix.

**2. Your PR cannot pass the merge gate, and no push of yours would have changed that.** Filed as #2993, priority p1, with a fixer already on it.

`.github/workflows/ci-infra-kill-rerun.yml` gained a `clear-stale-verdict-labels` job yesterday in `d934d7b06`. It runs on `pull_request: synchronize` and calls `gh pr edit --remove-label`. GitHub caps `GITHUB_TOKEN` to read-only for pull requests from a fork, and that cap overrides the job's own `permissions: pull-requests: write`. So on your PR it does this, every time:

```
GraphQL: Resource not accessible by integration (removeLabelsFromLabelable)
```

That check is not on the advisory allowlist, so it gates:

```
$ node scripts/ci-verdict.mjs 2983
Clear stale CI verdict labels    completed    failure
gating check(s) completed with a non-success conclusion:
  Clear stale CI verdict labels (failure)
exit 1
```

Yours is the first fork PR to push since that job landed, so you found it about a day after it was introduced. There is nothing you can rebase or re-push to clear it — the check is structurally incapable of succeeding on a fork head.

It also explains something I would otherwise have blamed on you: because that job failed first, `Classify failed Unit tests` and `Finalize CI rerun classification` both **skipped**, so the automatic infra-kill rerun never fired on your genuine exit-137. Both failures you were looking at had the same root.

The issue is filed with the general shape rather than just the fork instance — a bookkeeping job that clears metadata labels should never have been able to hold a merge at all, on a fork or otherwise.

**What this means for you: nothing to do.** Your round-2 verify is running independently and will report here on its own. If it comes back clean, #2983 merges as soon as #2993 lands. Sorry for the noise on what is, so far, a well-executed PR.

apmantza:
CI finished. **Do not push anything yet** — the round-2 verify is running against `caee9c03d` and a new head would throw its probe set away. I will collect its findings and these together so you get one round, not two.

Posting the CI results now so you can start thinking about them.

**Unit tests: 2 failed, 1078 passed, 12 skipped (1092 files); 14,715 tests passed.** Both failures are governance ratchets, and both are ours to explain rather than yours to have known about.

**1. `tests/config/hook-await-bounds.test.ts`**

```
hook helper unbounded-await pin (#2523 slice 3 worklist): 1 flagged item(s) are neither registered nor exempted:
  clients/blocker-freshness.ts@16

A module a registered hook handler imports DIRECTLY gained or lost an unbounded await.
Wrap it in bounded() (clients/deadline-utils.ts) and lower the pinned number, or raise
the number here and say why.
```

`clients/blocker-freshness.ts` is imported directly by a registered hook handler, so its unbounded awaits are pinned at an exact count. Your content-confirmation tier added one, taking it to 16. The ratchet is two-sided on purpose — it reds on a decrease too — so this is not something you can wave through by nudging a number without a reason.

Which way to go depends on a question only the diff answers: **is the new await on the hook path?** If the `size`-tier `stat` happens during hook dispatch, it wants `bounded()` from `clients/deadline-utils.ts`, because an unbounded filesystem await on a hook path is a stall the hook budget cannot absorb — and a network filesystem makes that concrete. If it only ever runs off the hook path, raise the pin and write the reason beside it. I would look hard at the first option before taking the second.

**2. `tests/config/strictness-ratchet.test.ts`**

```
AssertionError: expected [ 'clients: regression (358 -> 359)' ] to deeply equal []
```

One new TypeScript strictness spike in `clients/`. Same choice: remove the spike, or move the number in `tests/config/strictness-baseline.json` with a reason. Heads-up that two other open PRs also move that number, so whichever lands second will need it re-measured — that is the train's problem to sequence, not yours, but it is why the reason matters more than the number.

Everything else is green, including Lint & type-check, the Windows advisory lane, OSV, the mutation diff and markdownlint.

Still outstanding on our side: #2993, the fork-token gate, which no push of yours can clear. That fix is in flight.

I will post the verify verdict here as soon as it lands, and mark clearly which findings belong to the same round.

rnavarro:
Understood on both process notes, and thanks for chasing the CI down rather than leaving it looking like my red. Nothing for me to do on #2042 or #2993; I will leave the head where it is so the verify has a stable target.

Since you named the four angles, I ran them here rather than wait to be surprised. Evidence below in case it saves the verify some time. If it disagrees anywhere I would rather hear the disagreement than have my version believed.

**1. Whether `"self-drift"` falls through anything.** No `switch` on `staleReason` exists in the tree (`grep -rn "switch.*staleReason"` is empty), so there is no missing `default`. Two distinct fields carry the name and only one can hold the new value:

- `InlineBlockerRecord.staleReason` is the one that widened. Consumers: `retireDemotedPastEofBlocker` (`(existing.staleReason ?? "past-eof") !== "past-eof"` rejects it, and the `??` default is unreachable since this gate always sets the reason explicitly), `retireDemotedDependencyDriftBlocker` (requires `"dependency-drift"`, rejects), `runtime-turn.ts:1046` (`else if (staleReason === "dependency-drift")`, falls through to no cap, which is the intent), `blocker-past-eof.ts:135` (`!== "past-eof"` skips, correct sibling discipline), `diagnostic-line-freshness.ts:326` (skips a reason it does not own, same).
- `WidgetDiagnostic.staleReason` is a separate field and `"self-drift"` cannot reach it. `markWidgetFileBlockersStale(filePath, reason: "dependency-drift")` types the parameter as that single literal, so the compiler refuses any other value. That covers the `(d.staleReason ?? "past-eof")` reads at `widget-state.ts:723`, `:747`, `:1598` and `tools/lens-diagnostics.ts:3129`.

On the serializer half of that angle: `InlineBlockerRecord.staleReason` is not persisted. It appears nowhere in `git-guard.ts` or `cache-manager.ts`, so no durable record carries the new value into a stricter parser later. The widget store's own round trip already strips its `staleReason` (`widget-state.ts:372-383`, #2275 review F2).

**2. Whether re-arming can oscillate unboundedly.** Nothing grows per occurrence. `setInlineBlockerSelfDriftStale` writes two existing per-record fields (`stale`, `staleReason`) and allocates nothing. `selfHealed` and `selfUnverifiable` are locals on the per-sweep `counts` object, rebuilt each turn. `staleDeliveryCount` is never advanced for this reason, which is the point. The only emission is the one bounded `logLatency` metadata object per turn end that already existed; the round adds no `recordDegradationOnce` or `incrementDegradationCount` call. So a file whose bytes flip back and forth drives repeated demote/un-demote transitions with flat memory and a fixed per-turn record count.

**3. The obstruction direction, inverted.** Real, and I would rather state it plainly than defend it. A same-length edit, one character swapped, reads `unchanged` at the `size` tier and the blocker stays raised against a file that genuinely moved.

Two things make me think it is not a blocker on its own. It is strictly the pre-#2982 behaviour: before this axis, a non-LSP record was never checked against its own file at all, so every same-length edit already kept the blocker raised. Nothing regressed; this is the residue the `size` tier does not reach. And of the two directions to be wrong in, this one keeps a finding visible rather than hiding one, which is the direction I would choose for a record that can carry security provenance.

You are right that this makes the hash-tier deferral load-bearing rather than a footnote, so to state the deferral precisely: the blocker is the baseline, not the comparison. `detectSelfDrift` runs in the async sweep and could hash there cheaply, but it needs the hash of the content the verdict was taken against, and the only place that exists is `recordInlineBlockers`, which is synchronous on the dispatch path. Capturing it there means a blocking read per blocking dispatch. The shapes I can see are an async post-record hook that hashes off the hot path, or baselining lazily on the first sweep that finds the file unchanged. Both are more mechanism than this PR should carry, so I would rather file it against #2982's remainder than smuggle it in. Happy to do that now or after merge, whichever you prefer.

**4. That the cap still fires for genuine dependency drift.** Unchanged and covered by its own existing tests, which I did not touch: `dependency-drift delivery cap (#1950)` passes both cases, including the one asserting retirement at `DEPENDENCY_DRIFT_MAX_DELIVERIES` with the "still confirmable" note. The `"dependency-drift"` branch in `runtime-turn.ts` is byte-identical to master; the round only adds a reason that does not match it.

apmantza:
Verify is in. **One blocking finding, and it is the one I warned was coming.** Everything else cleared. You can push now — this is the full round, CI findings and verify findings together, as promised.

## Blocking — HIGH — same-size edits remain authoritative

The reviewer drove the real `sweepInlineBlockerFreshness` seam with a one-character, same-length edit:

```
{"before":21,"after":21,"counts":{"total":1,"kept":1,"revalidated":0,"alreadyStale":0,
 "truncatedImports":0,"selfHealed":0,"selfUnverifiable":0},
 "snapshot":[{"stale":false,"recordedSize":21}]}
AssertionError: expected false to be true
```

At `clients/blocker-freshness.ts:423` the size tier reads a same-length edit as unchanged, so a genuinely changed ast-grep, tree-sitter, mixed or unknown-provenance blocker stays authoritative in turn-end rendering. That is #2982's own complaint, arriving from the other side.

I want to be fair about what this is. Your deferral of the hash tier was honest and you stated its reason. The finding is not that you hid something — it is that the deferred remainder turns out to be acceptance-relevant rather than a footnote, because a same-length edit is not an exotic shape. It is a renamed identifier of equal length, a flipped comparison operator, a changed digit. For a security finding that is the common case, not the rare one.

Remedy shape the reviewer proposed, and I agree with it: capture a content baseline at dispatch **without blocking the synchronous path**, compare a hash only when size cannot distinguish, keep fail-closed behaviour when that baseline is unavailable, and record the deferred/unverifiable case so it is observable rather than silent. The last clause matters — an unverifiable case that nothing can see is a vacuous guard by this repo's contract.

If you think the baseline cannot be captured off the synchronous path, say so and we will scope it to a follow-up issue with the member list rather than have you fight it. That is a legitimate outcome; it just has to be an explicit decision rather than a gap.

## Also in this round — the two governance ratchets from CI

Repeated from my earlier comment so the round is in one place:

- `tests/config/hook-await-bounds.test.ts`: `clients/blocker-freshness.ts@16` neither registered nor exempted. Note the new async helper at `clients/blocker-freshness.ts:432` is what moved it. If the new `stat` sits on the hook dispatch path it wants `bounded()` from `clients/deadline-utils.ts` rather than a raised pin — an unbounded filesystem await on a hook path is a stall the hook budget cannot absorb.
- `tests/config/strictness-ratchet.test.ts`: `clients: regression (358 -> 359)`.

## LOW — the close line

```
BODY_CLOSE_LINE_EXIT 1
3:Closes #2982. Every acceptance criterion in the issue is met.
```

The gate wants a **bare** `Closes #2982` line. Yours has a trailing period and a sentence on the same line, so it does not match. Put the bare line on its own and move the sentence to the next one. Worth knowing: your commit subject also carries `closes #2982`, and in this repo a close keyword on any commit subject closes the issue regardless of the body — so that line needs to be right before this merges, not after.

The body gate also rejected your Observability section: *"must name a record literal from the runtime diff; 'No new failure path; no record added.' is not valid when the added lines contain a failure path."* Either name the bounded record, or correct the assessment if you believe there genuinely is no new failure path.

## What cleared — all of it attacked with probes, not read off your body

- **The new enum value does not fall through anything.** Every `staleReason` consumer was enumerated: runtime retirement stays strict on `dependency-drift`, past-EOF stays strict on `past-eof`, and the self-drift path is handled before the old branch. 19 targeted files, 466 tests, including old widget-state round-trip parsing — so existing durable records still parse.
- **Re-arming does not oscillate unboundedly.** Repeated self-drift delivery held at `staleDeliveryCount: 0` with no render storm, and no new unbounded map or per-occurrence record appeared in the diff.
- **The cap still retires genuine dependency drift.** Both cap suites green at `DEPENDENCY_DRIFT_MAX_DELIVERIES`.
- **`setInlineBlockerSelfDriftStale` genuinely follows `setInlineBlockerPastEofStale`**, ownership checks and re-arming included — you modelled it accurately rather than approximately. Consolidation verdict was to stay distributed, since each setter owns a distinct coordinator record axis and folding would relocate state-specific lines rather than remove them. No net-count violation.
- **Your tests observe the real seam** — real files, real `RuntimeCoordinator`, real freshness and turn-end path across all three changed files.
- **Your mutations hold.** Size-confirmation mutation gave 3 reds, the re-arm guard mutation failed the healing test, the chain-predicate mutation gave 2 widget-population reds.

One honest gap the reviewer declared: it did not replay your round-1 six-red transcript from a full source reversal, so those six are taken on your word rather than independently reproduced. I am not treating that as a finding — your earlier red-first proof was independently checked in round 1 and held.

## Sequencing

#2993 (the fork-token gate) now has a PR open, so the check that no push of yours could clear is being fixed in parallel. Nothing for you to do about it.

Take your time on the hash tier; it is the only thing standing between this and merge.

rnavarro:
Round 3 is pushed as `a76ebc495`, written up under `## Round 3`.

You were right that the deferral was acceptance-relevant rather than a footnote. `detectSelfDrift` now confirms at two tiers, size then sha256, with the baseline captured off the synchronous path: `recordInlineBlockers` reads nothing and returns its `recordedAtMs` stamp, `runtime-tool-result.ts` does the bounded read from inside `handleToolResult`, and `setInlineBlockerContentBaseline` drops a late baseline whose stamp no longer matches.

The part I would most like attacked: `bounded()` resolves `undefined` on timeout rather than rejecting, so an expired stat compared against the recorded size would read as a size change and demote, which is the exact failure the tier exists to prevent. Every bound's result is checked explicitly at all three sites, and missing-baseline and over-budget land in the same `unverifiable` place. Those increment `selfUnverifiable`, already carried by the `blocker_freshness_sweep` phase, so the guard is not vacuous.

Also bounded the other axis: per-call bounds leave the aggregate read unbounded when one turn carries many blockers (defect shape 9), so there is a per-sweep byte budget, and the sweep now takes `TurnEndDeps.signal` through its options.

Both ratchets are green. `blocker-freshness.ts` is back to its pinned 13 rather than 16 because the three new awaits are `bounded()` calls, all registered with signal provenance; the two leaves inside the baseline reader are exemptions owned by #2982. Strictness is back to 358 by omitting `staleReason` rather than assigning `undefined` to it. I left the identical `past-eof` spike alone on purpose, since fixing both would take it to 357 and red the ratchet from the other side.

Close line and the stale first `## Observability` section are both fixed.

New test `demotes on a same-LENGTH content change (one character swapped)` pins your probe and is red on round-2 code. 249 passed across eleven files.

apmantza:
Round 3 verified. **All three round-2 findings are fixed.** Two new ones, and the HIGH is a genuinely subtle one that your hash tier surfaced rather than caused.

## Fixed, confirmed by probe

- **The same-size edit now demotes.** Your hash tier closes the case that blocked round 2 — equal-length content edits demote through the real `RuntimeCoordinator` and freshness sweep, while mtime-only changes with unchanged bytes stay authoritative. Both directions hold.
- **Governance is clean.** `hook-await-bounds` and `strictness-ratchet`: 14 passed, and the strictness baseline is back at `clients: 358` with no regression. You resolved that rather than raising the pin, which was the right call.
- **The close line and Observability section both pass.** `Closes #2982` is bare, and the body gate exits 0.

The required sweep was 73 files, 948 passed, 2 skipped.

## New HIGH — the baseline can be captured after the bytes have already changed

This is the one I flagged as the risk when the hash tier was proposed, and it is real.

`clients/runtime-tool-result.ts:2315-2338` records the inline blocker, **then yields** to `readInlineBlockerContentBaseline(filePath)` before attaching the baseline. An edit landing in that window is captured *as* the baseline. The comparison then succeeds against post-edit bytes, the record never demotes — and it looks fixed, because the hash tier itself works correctly on whatever baseline it was handed.

Your round-3 same-size probe proves the comparison. It does not prove the lifecycle, and those are different claims. The `recordedAtMs` check only prevents a baseline attaching to a *newer* record; it does not establish that the bytes were read before the finding could be invalidated.

Remedy shape: capture the baseline from the evidence-producing pipeline **before** publishing the blocker, or carry the pipeline's existing content snapshot or hash into `recordInlineBlockers` so no second read is needed. If the host can edit concurrently, pair the snapshot with the write index so the ordering is checkable rather than assumed.

I want to be clear this is not a regression you introduced — the window exists because a second read was needed at all, and the fix is to remove the need for it rather than to narrow the window.

## New MEDIUM — `selfUnverifiable` loses the record identity

`clients/runtime-turn.ts:987-996` emits one `blocker_freshness_sweep` row with `filePath: cwd`, while `clients/blocker-freshness.ts:950-952` increments `selfUnverifiable` per affected entry. The missing-baseline, timeout and over-budget branches are bounded in aggregate, which is right — but the record cannot say *which* file or blocker was unverifiable. A rising count cannot distinguish one persistently unverifiable file from many transient ones, and those call for opposite responses.

Remedy shape: a bounded per-session record keyed by file and cause, or a capped deterministic set of affected paths plus a dropped-key count in the existing phase metadata. Keep the aggregate for occupancy reporting.

This is the house rule that a discriminator nothing can observe is a vacuous guard — the fail-closed behaviour is correct, it just is not diagnosable yet.

## Sequencing, so you are not waiting on us blindly

Your PR is still gated by #2993, the fork-token bug — that check cannot pass on a fork head no matter what you push. Its fix is in review now, so that will clear from our side.

Separately: your CI runs have needed a manual maintainer approval on every push, which is why they sometimes sit idle for a while. That is the same bug. Nothing you can do about it; I approve them when I see them.

Three rounds in, your diagnosis has been right each time and you have twice found a smaller root than the review proposed. The lifecycle finding above is the last structural one I expect on this seam.

apmantza:
Belated disclosure, and an apology for its absence: **every review comment on this PR was written by an AI agent, with maintainer supervision** — the round-1 review, the round-2 and round-3 verify relays, and the CI triage comments.

This repo's contract already required that on external PR reviews. I read it as covering only a formal review verdict and not the running commentary, which was too narrow, and the rule in `AGENTS.md` now says plainly that every outward-facing answer carries it.

It changes nothing about the findings, which were produced by probes with quoted output and stand on that evidence. Your three rounds have been reviewed by the same reviewer agent throughout, deliberately, so the probe set carries across rounds — and you have twice found a smaller root cause than the review proposed, which is on the record.

apmantza:
The fork-token gate is fixed and merged (#2995, closing #2993). Your PR's checks should now run without a maintainer approving them each time, and `Clear stale CI verdict labels` can no longer fail your merge gate.

That was the bug that made every one of your pushes sit idle until I noticed it — three times today. It is gone.

Two notes on your side of things:

Your round 3 (the hash tier) was verified and **all three round-2 findings are fixed** — the same-size edit now demotes, the governance ratchets are clean with strictness back at `clients: 358`, and the close line and Observability section both pass. The verify raised two new ones, which I posted earlier: the baseline can be captured after the bytes already changed (`clients/runtime-tool-result.ts:2315-2338` records the blocker, then yields before attaching the baseline), and `selfUnverifiable` loses the record identity.

The first of those is the one that matters and it is not a regression you introduced — the window exists because a second read was needed at all. Carrying the evidence-producing pipeline's existing snapshot into `recordInlineBlockers` removes the need for that second read rather than narrowing the window.

No rush from our side, and nothing is blocked on you procedurally now.

*This comment was written by an AI agent with maintainer supervision.*

rnavarro:
Both findings accepted. Before I build anything I want your opinion on the shape, because for the HIGH I think there is a smaller move than the one you proposed, and if I am wrong about it I would rather find out here than in a round 4.

## HIGH, the baseline lifecycle

You are right that the window exists because a second read was needed at all, and right that the fix is to remove the need rather than narrow it. Where I want to check myself is WHICH snapshot to carry.

`clients/pipeline.ts` already holds the bytes in memory, and I think they are provably the correct baseline rather than merely a convenient nearby one:

```
1497  fileContent = nodeFs.readFileSync(filePath, "utf-8");
1515  await resyncLspFile(filePath, fileContent, true, false, getFlag, dbg);
1529  const dispatchResult = await dispatchLintWithResult(...);
1640  output += buildEnrichedBlockerOutput(deliverableBlockers, fileContent);
```

The read precedes the dispatch that raises the blockers, it is the content the LSP is synced with for that dispatch, and `buildEnrichedBlockerOutput` renders the blocker summary's code snippets from it. That last one is what convinces me: the verdict's own rendered evidence comes from these bytes, so "the bytes this verdict is about" is not an approximation of `fileContent`, it is a definition of it.

So the proposal:

1. In `pipeline.ts`, when `inlineBlockerSummary` is set, compute size and sha256 from `fileContent` already in memory, and carry them on `PipelineResult` beside the other `inlineBlocker*` fields, which is the same place `inlineBlockerSources` and `inlineBlockerLines` already travel for the same reason (#1561 F1's "taken from the very diagnostics `blockerOutput` was rendered from, so the provenance can never disagree with the text it guards" applies verbatim to the content hash).
2. `recordInlineBlockers` accepts them and stamps them in the same synchronous call that creates the record.
3. Delete `readInlineBlockerContentBaseline`, its `bounded()` call, `setInlineBlockerContentBaseline`, and the `recordedAtMs` stamp-matching inside it, which existed only to defend the two-step.
4. `fileContent` undefined (the readback failed) means no baseline, which the sweep already reads as `unverifiable`.

There is no window because there is no second read, and no ordering to check because the capture is in the same synchronous statement as the record. It is also a net subtraction from the hook-await surface: it gives back the two `EXEMPT_SITES` entries and one `BOUNDED_CALL_SITES` entry round 3 added, rather than adding more.

One interaction I would rather disclose than have you find. When deferred autofix rewrites the file after the dispatch, disk will not match `fileContent`, so the first sweep demotes. I believe that is correct, since the verdict does describe pre-autofix bytes, and the axis re-arms so a later dispatch restores it. But it is a behaviour change I am choosing rather than one I stumbled into, and if you read it the other way I would rather know now. The alternative would be to skip the baseline when `postMutation` is set for the same path, which trades a demotion for an `unverifiable`.

Question I cannot answer from the outside: is `fileContent` guaranteed to be the same bytes the external analyzers read, or only the same bytes the LSP and the summary saw? ast-grep and the tree-sitter runner read from disk themselves, so strictly the answer is no. My position is that it does not matter, because the record is a verdict about the summary that was rendered, and the summary comes from `fileContent`. If you think the analyzers' own view is the thing that must be pinned, then this shape is wrong and the ordering-evidence approach you sketched is the right one instead.

## MEDIUM, `selfUnverifiable` identity

Agreed it is not diagnosable as an aggregate. I would use the repo's existing bounded per-subject ledger rather than invent a second shape:

```
incrementDegradationCount({
  kind: "self-drift-unverifiable",
  subject: `inline-blocker:${displayPath}`,
  reason: "missing-baseline" | "bound-expired" | "hash-budget-exhausted",
});
```

That is the same pattern `demoted-finding-retired` uses two functions away in `runtime-turn.ts`, bounded by the ledger's own per-kind and per-subject tally, and it separates one persistently unverifiable file from many transient ones because the subject carries the identity. The aggregate count stays in the `blocker_freshness_sweep` phase metadata for occupancy reporting.

The one thing I am unsure of is whether three separate `reason` values are right, or whether the cause belongs in the `kind` so each is independently queryable. Your call; I do not have the query patterns you do.

Happy to build both as soon as you have said which way. If you would rather just take the seam at this point I would understand that too, given this is round 4.

apmantza:
Both answers below, with the mechanism I can see from inside that you can't from outside. Short version: **build your shape for the HIGH — it is better than what the review proposed** — and for the MEDIUM, put the cause in the subject rather than in `reason` or in `kind`, for two of your three causes. The third is not a per-file event at all.

## HIGH — carry `fileContent`. Yes, and the case is stronger than you made it

I traced the lifecycle on current `master`, and `fileContent` is not merely a convenient nearby snapshot — the pipeline maintains it as the canonical "what this verdict is about" value through every one of its own writes:

```
clients/pipeline.ts
1416  fileContent = readFileSync(filePath)        // first read
1442  fileContent = formatResult.fileContent      // replaced when immediate format runs
1516  fileContent = readFileSync(filePath)        // re-read when fixRefresh (autofix changed it)
1534  resyncLspFile(filePath, fileContent, ...)   // the LSP is synced with THIS value
1548  dispatchLintWithResult(...)                 // the dispatch that raises the blockers
1659  buildEnrichedBlockerOutput(deliverableBlockers, fileContent)   // snippets rendered from it
```

Your argument was that `buildEnrichedBlockerOutput` renders from these bytes, so they define rather than approximate "the bytes this verdict is about". That holds, and the three mutation points above are the part that settles it: the value is deliberately kept current through format and autofix precisely so the dispatch and the rendered output agree. The `#1561 F1` reasoning you quoted — *"taken from the very diagnostics `blockerOutput` was rendered from, so the provenance can never disagree with the text it guards"* — transfers to a content hash exactly as you said.

**Your open question — is it the same bytes the external analyzers read?** No, strictly, and you are right that it does not matter, but the deciding argument is the failure *direction*, which your note doesn't make:

ast-grep and the tree-sitter runner read from disk themselves at ~1548. If an edit lands between line 1516 and their read, the baseline is pre-edit and disk is post-edit, so the first sweep sees a mismatch and **demotes**. Demotion re-serves in the advisory channel marked `[stale — re-run to confirm]` and re-arms; it does not delete and it does not silently keep a wrong verdict authoritative. The error is in the safe direction. Pinning each analyzer's own view would require every runner to report the bytes it read — not exposed for tree-sitter or ast-grep — for a strictly worse failure mode. Take `fileContent`.

Deleting `readInlineBlockerContentBaseline`, `setInlineBlockerContentBaseline`, its `bounded()` call and the `recordedAtMs` stamp-matching is the right consequence, not a bonus. Those three exist only to defend a two-step that stops existing. Giving back two `EXEMPT_SITES` and one `BOUNDED_CALL_SITES` entry is the net-count rule running in the direction it is supposed to run.

### The autofix interaction you disclosed — your reading is right, and the alternative is actively wrong

Demoting is correct. Take the alternative seriously for a moment, because it fails in an instructive way: skipping the baseline when `postMutation` is set yields `unverifiable`, and in the sweep `unverifiable` means *decide nothing, leave the record in whatever state it holds* — which for an authoritative record means **it stays authoritative**. A blocker raised against pre-autofix bytes, kept authoritative after autofix rewrote the file, is the exact failure #2982 reports. Do not trade a demotion for an `unverifiable` here; you would be buying back the bug.

One narrowing, in your favour: line 1516 already re-reads `fileContent` when in-pipeline autofix changed the file, so the case you disclosed only reaches the *deferred* format/autofix path. Smaller than you framed it.

## MEDIUM — subject, not `reason`, and not `kind`

Here is the mechanism you could not see. `incrementDegradationCount` (`clients/degradation-ledger.ts:1062-1099`):

- the tally key is **`${kind}\0${subject}`** — **`reason` is not part of the key**;
- the entry's reason is **overwritten on every hit** (`splice` then `push` at 1085-1086), so the last cause wins and earlier ones are gone;
- durable rows emit only at power-of-two counts **of that (kind, subject) tally**;
- `ENTRIES_PER_KIND` bounds entries **per kind**, evicting oldest-first.

That decides your question three ways:

- **Cause in `reason`** — one kind, one budget, but the cause is overwritten and the power-of-two milestones cannot separate causes. You would see *that* a file is unverifiable and only its most recent reason. Not enough.
- **Cause in `kind`** — three kinds means three × `ENTRIES_PER_KIND`, so the aggregate bound the finding asked for is three times looser, plus three new `DegradationKind` union members for one phenomenon. That is the wrong direction.
- **Cause in `subject`** — `inline-blocker:${displayPath}#missing-baseline` — one kind, one budget, and each (file, cause) pair is independently tallied and independently milestoned. A pathological file can occupy up to two entries instead of one; still bounded, and it answers exactly the question the finding posed: one persistently unverifiable file versus many transient ones, now also per cause.

So: **one kind `self-drift-unverifiable`, cause in the subject, `reason` free to carry the human-readable detail.** Your instinct to reuse the existing bounded ledger rather than invent a second shape was right, and `demoted-finding-retired` is the correct sibling to mirror.

### But `hash-budget-exhausted` is not a per-file event

This is the part I would push back on. `missing-baseline` and `bound-expired` are properties of *that file* — its baseline never attached, or reading it exceeded the bound. `hash-budget-exhausted` is a property of the *sweep*: the per-sweep byte budget ran out and whichever entries came later were cut off. The file was unlucky in ordering, not problematic.

Attributing it to a file subject would make sweep-level read pressure look like a per-file defect, and the two call for opposite responses — one says look at that file, the other says look at the budget and at defect shape 9's aggregate-read concern. Leave it in the `blocker_freshness_sweep` phase metadata aggregate where it already lives, or give it its own record keyed on the sweep. Two causes on the subject, not three.

## Sequencing on our side

- **#2983 is currently `CONFLICTING`, in exactly one file: `tests/config/hook-await-bounds.test.ts`** — the #2523 call-site registry, which `master` has moved under you. Worth noting that your round-4 change *removes* entries from that same file, so rebase first and make the deletions on the rebased contents rather than resolving a conflict against entries you are about to delete anyway.
- **Someone has stacked on you.** #3009 (nzbrian) is based on your head `a76ebc4` and extends the content-confirmed axis to the all-LSP own-file path. A rebase will move the ground under it, so it is worth knowing it exists. I have filed #3011 to track that defect on its own rather than leaving it referencing #2982, which your PR closes.
- Round 4 on one seam normally trips our round-count rail, which bars a third patch-only round. Your proposal is not a patch — it is a structural subtraction that deletes the mechanism the previous rounds were defending — so it is the right shape for a round 4 rather than a symptom of one.

And no, I would not rather take the seam. You have found a smaller root than the review proposed three times now, including this one. Build it.

---

apmantza:
Thanks for carrying this through three correction rounds and for identifying the smaller `fileContent` solution. We will take over the implementation from here.

The maintainer fixer will first compose `a76ebc495` with current `master`. It will then implement the agreed round-4 structural subtraction and bounded per-file observability. We will preserve your authorship and add only the review deltas to your branch under the repository's maintainer-edit permission.

The lane is queued behind the four active Codex workers. I will post its exact commit, test evidence, and verification result here before merge.

apmantza:
Thank you again for the contribution, @rnavarro.

## Maintainer takeover verification

**Worker:** `sub-mu08sof9-62` · **head:** `79945cd5ddbc4bd07920ec8b6add056dba79ecf4` · **role:** reviewer

Verdict: merge as-is.

No proven CRITICAL, HIGH, MEDIUM, LOW, or NITPICK finding remains on
79945cd5ddbc4bd07920ec8b6add056dba79ecf4.

## Issue acceptance findings

No actionable findings.

The candidate binds the `PipelineResult` content baseline to the synchronous
`recordInlineBlockers` write (`clients/pipeline.ts:1875`,
`clients/runtime-coordinator.ts:1032`, and
`clients/runtime-tool-result.ts:2515`). Immediate format and autofix paths
refresh `fileContent` before dispatch. Deferred mutation keeps the pre-mutation
content as the baseline, so a later byte change demotes the cached verdict.

The self-drift probe covers size changes, same-size changes, identical bytes
after an mtime-only change, missing baselines, expired bounds, re-arm after
revert, session-safe bounded records, and commit-gate preservation. The LSP
dependency path, past-EOF path, dependency delivery cap, and widget duplicate
population also passed their referencing suites.

The old two-step production source was used in an isolated temporary copy while
the candidate regression test remained. It produced 9 failures in
`tests/clients/blocker-freshness.test.ts`, including:

```text
FAIL ... demotes a tree-sitter blocker when its own file changed
AssertionError: expected +0 to be 1
FAIL ... demotes on a same-LENGTH content change (one character swapped)
AssertionError: expected +0 to be 1
FAIL ... records a distinct bounded record when the self-drift bound expires
AssertionError: expected [] to deeply equal ArrayContaining
```

The restored candidate ran 23 of 24 tests in that file successfully. The one
failure is the existing Java grammar-resolution case because grammar downloads
are unavailable offline. This is the same environment failure in the full
targeted run, not a candidate assertion.

The bounded sink shape is correct. Missing-baseline and bound-expired records
use one `self-drift-unverifiable` kind with file-and-cause subjects. Hash-budget
exhaustion emits only `blocker_freshness_sweep` metadata. Ledger state resets
through the existing session reset path.

I inspected all 73 `recordInlineBlockers` references, 12
`InlineBlockerRecord` references, and 15 `PipelineResult` references across
clients and tests. Existing optional result doubles remain valid. The widget
row is chained only when the inline row is eligible for the same LSP import
axis; otherwise both stores are checked independently.

## Repository-standard findings

No actionable standards finding.

The changelog fragment is valid, `AGENTS.md` records the new invariant, the
commit references #2982, and the candidate has no unresolved conflict against
master. The contributor/master merge has conflicts in `AGENTS.md` and
`tests/config/hook-await-bounds.test.ts`; candidate merge parent
`f01654abebadb29e443314c5c919f8a5643eb13c` resolves them, and
`git merge-tree --write-tree master HEAD` exits successfully.

## Verification

Passed:

- `npm run build`
- `npm run lint`
- `npm run fmt:check`
- `npm run changelog:check`
- targeted freshness, pipeline, coordinator, git-guard, retirement, and
  runtime-tool-result suites: 210 tests passed, 1 skipped
- config suite: 465 tests passed, 11 skipped

The freshness target had 49 passes and one offline Java grammar failure.

Blocked by the environment:

- `npm run astgrep:self-scan` cannot spawn `ast-grep` (`EPERM`).
- Config process and Git probes fail with sandbox `EPERM`; the remaining
  config failures are child-process, probe-home, dependency-boundary, or
  tracked-file checks.
- `npm run build:dist` did not complete in the available run window.
- `npm run preflight` produced no result before it was stopped after hanging.
- Fresh GitHub discussion, CI, and PR #3009/#3011 retrieval were unavailable.

These are environment limitations, not evidence of a candidate regression.
The requested #3011 behavior was not imported or evaluated.

## Cleared categories

Publication ordering, immediate format/autofix snapshot binding, deferred
mutation demotion, same-size edits, mtime-only identical bytes, missing
baseline, bound expiry, hash-budget attribution, session reset and re-arm,
commit-gate preservation, dependency-drift cap, past-EOF behavior, old-record
compatibility, widget duplicate population, optional result doubles, whole
module mocks, and candidate/master conflict resolution.

## Could not verify

I could not read fresh remote discussion or CI because GitHub access is
unavailable. I could not run ast-grep, process-backed config probes, or the
full preflight/dist pipeline because the sandbox denies child-process creation
or the command did not finish. I could not perform a live composition run with
PR #3009 or #3011 because those external branches were unavailable.

## Named output

Strong: the composed design has a clean ownership boundary. `runPipeline`
owns the already-read evidence, `RuntimeCoordinator` owns publication, the
freshness module owns demotion policy, and `runtime-turn` owns sweep telemetry.
Keeping those boundaries avoids reintroducing the delayed second-read race.

## Previous-round dispositions

| Finding | Disposition |
| --- | --- |
| R3 lifecycle race | fixed |
| Same-size content confirmation | fixed |
| Mtime-only false demotion | fixed |
| Missing-baseline and bound observability | fixed |
| Widget duplicate population | fixed |

---
*This review was written by an AI agent with maintainer supervision.*

apmantza:
Temporarily closing and reopening to re-arm required CI on the unchanged reviewed head. No contributor code or authorship is changing.

apmantza:
## Latest Plegma fixer handoff

- Worker: `sub-mu0xeh41-3`
- Role: fixer
- Exact head worked from: `a275c53e14e0c633a7ae1b85d093e76d38777d10`
- Source artifacts: `/home/akis/.plegma/work/sub-mu0xeh41-3/PR_BODY.md`, `/home/akis/.plegma/work/sub-mu0xeh41-3/COMMIT_MSG.txt`
- Changes remain uncommitted and unpushed in the isolated worker worktree.

### Handoff body

Operating rule: bind inline-blocker self-drift evidence to the bytes presented to analysis, and retain the exact reason when verification cannot complete.

Kept: blocker freshness remains demote-not-drop, self-drift remains re-armable, dependency freshness keeps its existing generation and delivery-cap behavior, and demoted blockers continue to gate commits.

## Summary

Round 4 fixes PR #2983 at head `a275c53e14e0c633a7ae1b85d093e76d38777d10`.

The baseline hash is captured before `dispatchLintWithResult` can yield. A later writer cannot replace the bytes identified by the blocker evidence.

Unverifiable self-drift records retain per-file identity and cause. The aggregate hash budget remains in sweep metadata, with an additional bounded ledger record for each affected file.

Findings disposition:

- `fixed`: HIGH — the baseline was captured after analysis could change bytes. `runPipeline` now captures it before the dispatch await.
- `fixed`: selfUnverifiable identity — stat failure, missing baseline, hash failure, deadline expiry, and hash-budget exhaustion have distinct reasons and file subjects.

## Round 4

The composed merge parent is `f01654abe`, and the worktree started clean at the requested exact head. No Git authority was used. The remote PR body, live merge state, review comments, and CI records could not be fetched because `gh` has no authentication and the sandbox cannot resolve GitHub.

The pre-fix fixture is the new `binds blocker content to the pre-dispatch analysis bytes` case with the pre-dispatch capture mutated to undefined. It failed with an undefined baseline instead of the expected size and hash.

The outer-deadline fallback mutation changed `bound-expired` to `hash-unavailable`. `records a distinct bounded record when the self-drift bound expires` failed because the observed subject changed from `inline-blocker:consumer.ts#bound-expired` to `inline-blocker:consumer.ts#hash-unavailable`.

## Tests

- `tests/clients/pipeline.test.ts`: `binds blocker content to the pre-dispatch analysis bytes` drives the pipeline with a real temporary file and an analysis-boundary writer, proving the baseline is the pre-dispatch content.
- `tests/clients/blocker-freshness.test.ts`: existing self-drift cases cover same-length edits, healing, missing baselines, deadline expiry, dependency-only changes, unknown provenance, and commit-gate preservation. The deadline case now verifies the exact `bound-expired` identity.
- `tests/config/degradation-kind-order.test.ts`: verifies the added ledger kind stays in the sorted registry.

Commands and results:

- npm run build: passed.
- npm run lint: passed.
- Vitest blocker freshness self-drift selection: 11 passed, 13 skipped.
- Vitest pipeline baseline selection: 1 passed, 24 skipped.
- Vitest degradation-kind-order selection: 1 passed.
- Combined targeted run: 59 passed, 1 skipped, with the existing Java forward-import test failing because the sandbox could not fetch tree-sitter Java WASM.
- `npm run preflight`: started as the final command but produced no output after 60 seconds and was interrupted. The tree-sitter prefetch was also reported offline by Vitest.

## Blast radius

Production modules changed: `clients/pipeline.ts`, `clients/blocker-freshness.ts`, and `clients/degradation-ledger.ts`.

The affected path is `runPipeline` → `dispatchLintWithResult` → `PipelineResult.inlineBlockerFileContent` → `RuntimeCoordinator.recordInlineBlockers` → `sweepInlineBlockerFreshness`. The ledger vocabulary is consumed by the existing bounded degradation summary.

The worktree does not expose the repository's `module_report` tool, so a structural blast-radius report was unavailable. The pattern sweep below covers every `self-drift-unverifiable` and `recordInlineBlockers` production reference found in the tree.

No tool registration, durable file schema, session generation, blocker freshness transition, or `CHANGELOG.md` path changed.

## Class sweep

The repository-wide rg sweep for recordInlineBlockers, inlineBlockerFileContent, self-drift-unverifiable, and self-drift-hash-budget-exhausted found the pipeline producer, coordinator writer, freshness consumer, turn telemetry, degradation vocabulary, and their tests. No second baseline writer or second unverifiable ledger kind exists.

Population coverage is language-neutral at the freshness seam: tree-sitter and ast-grep self-drift rows are exercised, LSP rows retain the forward-import path, and the widget-only population remains on its existing LSP dependency path. The Java forward-import row is present but could not verify offline because its grammar was unavailable.

## Observability

The existing `blocker_freshness_sweep` phase records `selfUnverifiable` and `hashBudgetExhausted`. Each unverifiable file now emits the bounded `self-drift-unverifiable` record with a subject ending in `#bound-expired`, `#stat-unavailable`, `#missing-baseline`, `#hash-unavailable`, or `#hash-budget-exhausted`. Hash-budget cases also emit `self-drift-hash-budget-exhausted` with the affected file subject.

## Test assessment

The new pipeline test is a regression proof at the permitted dispatch boundary and uses the real `runPipeline` path and real filesystem bytes. It does not replace the real in-process coordinator or freshness store. The existing freshness suite remains the real `RuntimeCoordinator` and filesystem path. No test was removed, weakened, or admitted as a flake shape.

### Proposed commit message

```text
fix(diagnostics): bind blocker evidence before analysis

Capture inline-blocker content before dispatch can yield so a later writer
cannot replace the bytes identified by the diagnostic. Preserve blocker
freshness and generation semantics while recording the exact per-file cause
when self-drift verification is unavailable.

Refs #2982
```

apmantza:
## Fixer handoff — exact head `c0f583df582ded38e0f5590781c149d3f1015ef5`

Round 4 fixes both reported review findings: blocker evidence is captured from the pipeline's pre-dispatch bytes, and unverifiable self-drift records retain a bounded file-and-reason identity. The branch was rebased/composed with current `origin/master` and pushed to the contributor head `rnavarro:fix/blocker-freshness-self-drift-non-lsp`.

Evidence: commits `ccde82829` and merge resolution `c0f583df5`; build, lint, freshness/pipeline targeted tests, and compile-valid mutation reds pass per the fixer handoff. External grammar and broad preflight execution remain environment-blocked. No new code change is claimed beyond the reviewed round.

**Next owner:** prior reviewer lineage `sub-mu08sof9-62` must verify this exact head after hosted CI and required gates run. Keep the PR conditional until exact-head checks are genuinely green.

Issue #3042, opened 2026-09-15
test(flake): index-vanished-instance-wiring reds late in the CI Unit lane on unrelated PR heads — synthetic dead pid (process.pid + 100000) is treated as alive


apmantza:
## Symptom

`tests/index-vanished-instance-wiring.test.ts` (#1123 item 2) reds in the CI Unit-tests lane on PR heads that do not touch the reaper, registry, or the test, while passing in isolation on a clean build of the same base.

Two CI instances today, same two assertions:

| Run | Head | Branch | Assertions |
|---|---|---|---|
| 35007930439 (job 104512513737) | 07bacd4 (#3023, oxfmt bump only) | dependabot/npm_and_yarn/oxfmt-0.67.0 | `expected [ 110771 ] to not include 110771`; `expected [ Array(1) ] to have a length of +0 but got 1` |
| 35011704431 (job 104525035572) | 0849875 (#3026, @types/node bump + 2-line type fix) | dependabot/npm_and_yarn/types/node-26.5.1 | `expected [ 111127 ] to not include 111127`; `expected [ Array(1) ] to have a length of +0 but got 1` |

Both heads are based on master 20896a5, whose own CI run (35005978965) is green. Local measurement on a clean `origin/master` worktree (fresh `npm ci`, `npm run build`, file run alone): 3 runs, 2/2 passed each time. A 61-file batch (`tests/config` plus three governance suites, HOME and TMPDIR pinned under the worktree) also passed.

## Shape

The failing pid is `process.pid + 100_000` (test line 146). The test's own header says the synthetic pid is "made dead deterministically by the test's `process.kill(pid, 0)` seam; it never relies on the runner's process table". Both failures show the reaper treating that pid as alive (case 1, prune did not happen) and the second case then seeing a marker line it should not (case 2, "logs nothing when the registry has no dead entries" got 1 line), i.e. case 2 inherits case 1's unpruned entry.

[...]

Issue #2837, opened 2026-09-09
knip becomes gating by the oxfmt recipe: preflight hard gate first, entry set covers every lane, documented ignores, then the CI lane drops (advisory)


apmantza:
Record (2026-09-09): knip on master went red after #2827 and #2829 merged — six exports with no importer (#2835 removed them). The lane is advisory and preflight (#2796) does not run knip, so nothing stopped either PR.

Recipe, same as the oxfmt promotion (#2790 → #2834):
1. `npm run preflight` gains `knip` as a hard gate (`scripts/pr-preflight.mjs`, mirrored CI job `knip (advisory)` until step 3); fixer handoffs paste it in the table.
2. The knip config's entry set covers every vitest lane that imports `tests/support/*` (the `real-harness` lane was the source of four of today's six findings) and every `scripts/*.mjs` entry the workflows invoke; intentional public exports get a documented ignore with a reason (one list, in the config, reviewed like an admission).
3. After a few days of zero advisory reds, the CI job drops its `(advisory)` suffix; `scripts/lib/ci-checks.mjs`, the workflow-shape test, ci-verdict/warden tests and docs follow, exactly as #2834 did.

Acceptance: preflight red on a branch that adds an unused export (red-first, quoted); master clean; the gating flip is its own PR with the census of advisory lanes before/after.

apmantza:
Step 1–2 landed in #2840 (merged fcfa88166): knip is a hard preflight gate and the entry set is derived from the Vitest lanes and the Windows population seam. Remaining: step 3, flip the CI knip job from advisory to gating after a run of clean days.

PR #3104, opened 2026-09-16
fix(tests): scratch governance fixture never lands in tests/; walkers tolerate a vanished file (closes #3082, closes #3092)

### LINKED ISSUE #3082: test(governance): the #3050 detector's scratch .test.ts under tests/ reds sibling directory-walking sweeps with ENOENT
## Symptom

Two directory-walking governance suites red with `ENOENT` on a file that never
existed at the start of the run:

```
FAIL tests/config/vacuous-skip-coverage.test.ts > tests skip visibly instead of returning early (#2089) > no test body returns before it asserts or skips
Error: ENOENT: no such file or directory, open '<repo>/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ tests/support/vacuous-skip-scan.ts:376:7
 ❯ scanVacuousSkips tests/support/vacuous-skip-scan.ts:373:25
```

and the same shape at `tests/config/sweep-floor-coverage.test.ts:71` /
`:200` (`sweepShapeFiles`).

## Cause

`tests/clients/pi-lens-home-hermeticity.test.ts`'s "red-first proof: the FULL
scan catches the pre-#3048 shape" (#3066) writes a real
`tests/scratch-3050-pre-3048-vanished-wiring.test.ts` during the run so its own
`listSourceFiles(TESTS_ROOT)` walk picks it up, and removes it in a `finally`.
Every OTHER governance suite that walks `tests/**/*.test.ts` and then reads
each path races that window: list → file vanishes → read throws.

## Measured

Same worktree, same 101-file governance batch
(`ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts`
plus every `tests/config/*.test.ts` plus `instance-re*`), `origin/master`
content, no local edits:

```
run 1:  Test Files 1 failed | 100 passed (101)
run 2:  Test Files 2 failed | 99 passed (101)
```

It did not fire in master's own full Unit run (different fork scheduling), so
it is latent rather than permanently red — but it is a genuine cross-file
race that can red any Unit lane.

## Shape

AGENTS.md's "test that drives shared state it does not own" — here the shared
state is the `tests/` directory itself rather than a file under
`PI_LENS_HOME`. Two directions to consider (both, ideally):

1. The producer: the proof does not need the scratch copy to live under
   `tests/` — `scanFlagged(files)` already takes an explicit file list, so the
   fixture can be scanned from a private temp dir (the FULL-walk claim then
   needs a different, non-racy proof).
2. The consumers: a directory-walking scan that lists then reads must tolerate
   a path that disappears between the two (ENOENT → skip), which is correct
   for every such sweep regardless of this producer.

Refs #3066, #3050, #3042.

### LINKED ISSUE #3092: Governance sweeps race a scratch test file written into tests/ (ENOENT, rotating victim)
## Summary

`tests/clients/pi-lens-home-hermeticity.test.ts:867` writes a scratch test file
named `scratch-3050-pre-3048-vanished-wiring.test.ts` into the repo's OWN test
tree and then deletes it. Every directory-walking governance sweep enumerates
that tree and `readFileSync`s what the walk returned, so when the two run
concurrently the sweep dies on the file that vanished between the walk and the
read:

```
FAIL  |default| tests/config/sweep-floor-coverage.test.ts > registered-or-fail sweep floors > every sweep-shaped test uses sweep-kit or declares a reason
Error: ENOENT: no such file or directory, open '/…/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ tests/config/sweep-floor-coverage.test.ts:72:36
```

Any of at least four files can be the victim — it is whichever walker happens to
be mid-enumeration: `tests/config/sweep-floor-coverage.test.ts`,
`tests/config/vacuous-skip-coverage.test.ts`,
`tests/config/latency-logger-mock-shape.test.ts`,
`tests/config/lsp-spawn-heavy-coverage.test.ts`. One or two of them red per
batch, rotating.

## Reproduction, on `origin/master` (0b6c8146c)

In a clean worktree, run the mechanically-selected governance set together:

```
ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts tests/config/*.test.ts > gov.txt
npm run test:targeted -- $(cat gov.txt | tr '\n' ' ')
```

Two laps, both red, different victims each time:

```
lap 1: FAIL sweep-floor-coverage + vacuous-skip-coverage   Test Files 2 failed | 88 passed (90)
lap 2: FAIL latency-logger-mock-shape + lsp-spawn-heavy-coverage   Test Files 2 failed | 88 passed (90)
```

both with the same ENOENT on the scratch file.

## Why it matters

CI's Unit lane runs the whole suite with the same fork concurrency, so this is a
live source of unexplained governance-sweep reds on unrelated PRs — the shape
that costs a rerun and a round of "is this mine?". It was found while baselining
#2042's PR (#3091), whose own batch red was this and not the change under test.

## Suggested fix (one of)

1. Write the scratch fixture under the test's temp dir instead of the repo's
   `tests/` tree — it does not need to live where the walkers look. Cheapest and
   removes the class.
2. Failing that, make `tests/support/sweep-kit.ts`'s read tolerate a file that
   vanished between walk and read (catch ENOENT, skip), which every walker
   already shares — but that is silencing, and a genuinely unreadable source
   file would then go unnoticed. Prefer (1).

## Acceptance criteria

- [ ] No test writes a `*.test.ts` file into the repo's tracked `tests/` tree at
      runtime; the fixture lives under the test's own temp dir.
- [ ] The 90-file governance batch above runs 3 consecutive laps green on
      `master`.
- [ ] A regression test that reds if a fixture path inside the walked tree is
      reintroduced (the walkers' own population is enumerable).

---
*This issue was filed by an AI agent working on this repository, with maintainer supervision.*

apmantza:
## Summary

`tests/clients/pi-lens-home-hermeticity.test.ts`'s #3066 red-first proof wrote a
real `scratch-3050-pre-3048-vanished-wiring.test.ts` into the repo's OWN
test tree so its `listSourceFiles(TESTS_ROOT)` walk would pick it up, and removed
it in a `finally`. Every other governance sweep lists `tests/**` and then reads
each path, so whichever walker was mid-enumeration during that window died with
`ENOENT` on a file that exists on no branch — four rotating victims. AGENTS.md's
"test that drives shared state it does not own", where the shared state is the
`tests/` directory itself.

Three directions, all in this PR:

1. **Producer.** The proof now builds its scratch copy in a private `mkdtemp`
   root and walks THAT — through `walkTestFiles`, the same function with the same
   options the production population is built from
   (`tests/clients/pi-lens-home-hermeticity.test.ts:575`), so the full-walk claim
   survives without touching `tests/`.
2. **Consumers.** Every walk-then-read over the `tests/` tree now reads through
   `readWalkedFile` / `readWalkedFiles` (`tests/support/sweep-kit.ts:576`): a path
   that vanished between the walk and the read leaves the population, is recorded
   once per distinct path and warned once, and is never counted as a finding.
   Sixteen sites in sixteen files folded onto that seam (two of them found by
   round 2's root-based re-sweep), plus one twin in `scripts/lib/` that cannot
   import from `tests/`; three hand-rolled `readdirSync` walkers folded onto
   `listSourceFiles` in the same change.
3. **Guard.** `tests/support/tests-tree-write-guard.ts` installs a recursive
   `fs.watch` from `globalSetup` — once per ACTIVATED Vitest project, not per
   test-file fork (measured; see Round 2 F3); the teardown fails the run and names
   any source file created inside `tests/` that was not there when the run
   started.

Closes #3082, closes #3092 — every acceptance line of both issues is met: no test
writes a `*.test.ts` into the tracked `tests/` tree (guard-enforced, not
asserted), the 91-file governance batch runs green on 4 consecutive laps
(3 required), and the reintroduction is caught by a runtime guard rather than a
source scan.

Citations: `tests/support/sweep-kit.ts:576` (`readWalkedFile`),
`tests/support/tests-tree-write-guard.ts:99` (`isGuardedTreeEntry`),
`tests/support/tests-tree-write-guard-setup.ts:23`
(`runTestsTreeWriteGuardSetup`), `vitest.config.ts:60` (globalSetup row).

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [ ] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [x] `npm run build:dist` succeeds if I changed code under `clients/`, `commands/`, `tools/`, or `index.ts` — n/a, none touched
- [x] `package-lock.json` is in sync with `package.json` — untouched
- [x] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there — no catalogued rule changes; the new guard's own docstring carries the recurrence
- [x] `.changelog/3082-scratch-test-file-race.md` has one valid entry (`node scripts/check-changelog-fragments.mjs` → `changelog fragments OK (85 entries in .changelog/)`, exit 0)

### The defect, reproduced from the production path first (#3092's recipe)

The mechanically-selected governance batch, 91 files, `origin/master` content in
this worktree, before any edit. Laps 1 and 3 green, laps 2 and 4 red with
different victims each time:

```
lap 1:  Test Files  91 passed (91)
lap 2:  Test Files  3 failed | 88 passed (91)
lap 3:  Test Files  91 passed (91)
lap 4:  Test Files  2 failed | 89 passed (91)
```

lap 2 (verbatim, `npm run test:targeted -- $(cat gov.txt)`):

```
 FAIL  |default| tests/config/latency-logger-mock-shape.test.ts > latency-logger mock shape (#2281) > derives every factory and requires a partial import
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 FAIL  |default| tests/config/sweep-floor-coverage.test.ts > registered-or-fail sweep floors > every sweep-shaped test uses sweep-kit or declares a reason
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ tests/config/sweep-floor-coverage.test.ts:72:36
 FAIL  |default| tests/config/vacuous-skip-coverage.test.ts > tests skip visibly instead of returning early (#2089) > no test body returns before it asserts or skips
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ tests/support/vacuous-skip-scan.ts:376:7
 Test Files  3 failed | 88 passed (91)
      Tests  3 failed | 1142 passed | 1 skipped (1146)
```

lap 4 (verbatim, different victims — the rotation #3092 describes):

```
 FAIL  |default| tests/config/latency-logger-mock-shape.test.ts > latency-logger mock shape (#2281) > derives every factory and requires a partial import
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ findLatencyMocks tests/config/latency-logger-mock-shape.test.ts:53:20
 FAIL  |default| tests/config/lsp-spawn-heavy-coverage.test.ts > lsp-spawn-heavy Vitest project coverage > phases or documents every test that spawns a real LSP child
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts'
 ❯ tests/config/lsp-spawn-heavy-coverage.test.ts:203:24
 Test Files  2 failed | 89 passed (91)
      Tests  2 failed | 1143 passed | 1 skipped (1146)
```

Post-fix, same batch, same worktree, four consecutive laps:

```
lap1 exit=0   Test Files  91 passed (91)   Tests  1145 passed | 1 skipped (1146)
lap2 exit=0   Test Files  91 passed (91)   Tests  1145 passed | 1 skipped (1146)
lap3 exit=0   Test Files  91 passed (91)   Tests  1145 passed | 1 skipped (1146)
lap4 exit=0   Test Files  91 passed (91)   Tests  1145 passed | 1 skipped (1146)
```

### Red-first proof of the new guard, against the real pre-fix producer

`git checkout origin/master -- tests/clients/pi-lens-home-hermeticity.test.ts`
(the producer restored, everything else at this branch), then
`npx vitest run tests/clients/pi-lens-home-hermeticity.test.ts`:

```
exit=1
 Test Files  1 passed (1)
      Tests  19 passed (19)

error during close Error: #3082: 1 source file(s) were created inside the repo's own tests/ tree during this run:
  /home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-3050-pre-3048-vanished-wiring.test.ts
Every governance sweep lists tests/** and then reads each path, so a file that appears and disappears mid-run kills whichever walker was enumerating (ENOENT, rotating victim). Write the fixture into the test's own mkdtemp root and walk THAT — listSourceFiles does not care which root it is given (tests/clients/pi-lens-home-hermeticity.test.ts is the worked example).
    at Object.teardown (tests/support/tests-tree-write-guard-setup.ts:38:22)
    at TestProject._teardownGlobalSetup (node_modules/vitest/dist/chunks/index.B89dZ0-N.js:12128:100)
```

With the producer fixed the same command exits 0 (`Test Files 1 passed (1)`,
`Tests 19 passed (19)`).

### New test files and cases

`tests/support/tests-tree-write-guard.test.ts` (new, 11 cases). Pins the guard's
classification and reporting through the real module; every case names the
recurrence in the file header.

| `it(` title | pins |
| --- | --- |
| `reports a source file created under the watched root` | the report names the created path and the issue |
| `reports a file created in a nested directory, the shape a walk recurses into` | recursion, the producer's own position one directory down |
| `stays silent for a file that already existed when the run started` | the baseline arm — an editor save mid-run is not a violation |
| `stays silent for file kinds no walker under tests/ enumerates` | `.json` and `node_modules` are out of the population |
| `classifies by extension and node_modules position, not by filename` | the classifier is not a filename heuristic (shape 8/34) |
| `records one entry per distinct path, however many events arrive` | bounded observability: one record per path, not per event |
| `caps the reported list and says how many it dropped` | `MAX_VIOLATIONS`, with the dropped count stated (shape 9) |
| `warns once and stays inert when the platform cannot watch recursively` | the unavailable-watch arm warns rather than failing the run, and says the guard is inert (shape 10) |
| `the setup arm throws the report, and closes the watch on the way out` | the globalSetup teardown's throw AND its `finally` close (shape 4) |
| `the setup arm is silent on a tree nothing created a source file in` | no false positive on an edit to a tracked file |
| `records a file created and removed mid-run, through a real recursive watch` | real `fs.watch` delivery, no stand-in — the only case that reds when the callback stops recording |

`tests/support/sweep-kit.test.ts` (4 new cases under
`describe("sweep-kit: readWalkedFile (#3082)")`): `returns the source of a file
the walk still finds`, `returns undefined, and warns once, for a file that
vanished after the walk`, `rethrows anything that is not a vanished file`,
`readWalkedFiles drops the vanished entries and keeps the rest paired`.

**Stated test exception (AGENTS.md's honest-exception rule).** The per-site
consumer fold has no independent red of its own: the race it tolerates is exactly
what the producer fix and the new guard now prevent, so a mutation that reverts
one call site to `fs.readFileSync` leaves the suite green. What IS proven is the
helper (mutation M1/M2 below) and the producer (the batch reds above). I did not
add a source-scan sweep asserting "no tests/-walker calls `fs.readFileSync`
directly" — that would be defect shape 33 (source assertion for runtime
behaviour) plus a new registry, for a direction the guard already closes at the
producer.

Test-authoring screens: the guard tests use no mock of the seam under test
(`fs.watch` is real in the delivery case, absent in the classification cases via
the `watch: false` option, and a throwing stand-in only in the
platform-unavailable case) — no all-mocks, no ambient-inspection double, no
implementation mirror. Every skip-free case asserts. The one wall-clock wait is
admitted through the ratchet (below), not hidden.

### Mutation proofs (quoted, not ticked)

Each mutation applied to the built tree, the suite run, then reverted with
`git checkout HEAD -- <file>`.

**M1 — `readWalkedFile`'s ENOENT arm deleted** (`throw error;` unconditionally):

```
 FAIL  |default| tests/support/sweep-kit.test.ts > sweep-kit: readWalkedFile (#3082) > returns undefined, and warns once, for a file that vanished after the walk
Error: ENOENT: no such file or directory, open '…/pi-lens-3082-read-JowSeD/vanished.ts'
 FAIL  |default| tests/support/sweep-kit.test.ts > sweep-kit: readWalkedFile (#3082) > readWalkedFiles drops the vanished entries and keeps the rest paired
Error: ENOENT: no such file or directory, open '…/pi-lens-3082-read-JowSeD/gone.ts'
 Test Files  1 failed (1)
      Tests  2 failed | 78 passed (80)
```

**M2 — the once-per-path record dedupe forced open** (`if (true) {`):

```
 FAIL  |default| tests/support/sweep-kit.test.ts > sweep-kit: readWalkedFile (#3082) > returns undefined, and warns once, for a file that vanished after the walk
AssertionError: expected "warn" to be called 1 times, but got 2 times
 Test Files  1 failed (1)
      Tests  1 failed | 79 passed (80)
```

**M3 — the watch callback stops recording** (`watch(root, {recursive:true}, () => {})`):

```
 FAIL  |wall-clock-budget| tests/support/tests-tree-write-guard.test.ts > tests-tree write guard (#3082) > records a file created and removed mid-run, through a real recursive watch
 Test Files  1 failed (1)
```

**M4a — `isGuardedTreeEntry` forced `true`** (the dangerous direction is M4b):

```
 FAIL  … > stays silent for file kinds no walker under tests/ enumerates
 FAIL  … > classifies by extension and node_modules position, not by filename
 Test Files  1 failed (1)
      Tests  2 failed | 9 passed (11)
```

**M4b — `isGuardedTreeEntry` forced `false`** (real violations stop blocking):

```
 FAIL  … > reports a source file created under the watched root
 FAIL  … > reports a file created in a nested directory, the shape a walk recurses into
 FAIL  … > classifies by extension and node_modules position, not by filename
 FAIL  … > records one entry per distinct path, however many events arrive
 FAIL  … > caps the reported list and says how many it dropped
 FAIL  … > the setup arm throws the report, and closes the watch either way
 FAIL  … > records a file created and removed mid-run, through a real recursive watch
 Test Files  1 failed (1)
```

**M5 — the run-start baseline check deleted** (`if (baseline.has(absolute)) return;` removed):

```
 FAIL  … > stays silent for a file that already existed when the run started
 Test Files  1 failed (1)
```

**M6 — `MAX_VIOLATIONS` cap forced off** (`if (false) {`):

```
 FAIL  … > caps the reported list and says how many it dropped
 Test Files  1 failed (1)
```

**M7 — the globalSetup teardown swallows the report** (`if (false) throw …`):

```
 FAIL  … > the setup arm throws the report, and closes the watch either way
 Test Files  1 failed (1)
```

**M8 — the teardown's `finally { guard.close() }` deleted**:

```
 FAIL  … > the setup arm throws the report, and closes the watch on the way out
AssertionError: expected +0 to be 1 // Object.is equality
 Test Files  1 failed (1)
```

**M9 — `GUARDED_EXTENSIONS` narrowed to `[".ts"]`**:

```
 FAIL  … > classifies by extension and node_modules position, not by filename
 Test Files  1 failed (1)
```

### Measured, not assumed

The guard's one real bound was probed rather than reasoned about (Node 22.22.1,
this worktree): a file created and removed **without the watching process's event
loop turning in between** delivers no event at all, while the same two operations
separated by one turn deliver both.

```
settle200 gap200 nested unref ["rename:clients/s.test.ts","rename:clients/s.test.ts"]
settle200 gap0   nested unref []
settle0   gap200 nested unref ["rename:clients/s.test.ts","rename:clients/s.test.ts"]
settle0   gap0   nested unref []
settle0   gap0   root   unref []
settle0   gap0   root   ref   []
```

That bound does not weaken the guard where it runs — the watcher is in the main
process and producers are worker forks, and the #3082 producer held its file
across a full 1,200-file walk (the red-first transcript above is the proof) — and
it is stated in the module docstring rather than left for a reader to discover.

### Ratchet / lane registrations (each is a claim, please attack it)

- `tests/support/flake-shape-baseline.json` gains
  `"support/tests-tree-write-guard.test.ts": 2` under `raw-timer-wait`, with the
  `// flake-shape: raw-timer-wait — …` header in the file and the
  `ADMITTED_AFTER_BASELINE` reason in
  `tests/clients/flake-shape-ratchet.test.ts:160`. Reason: a recursive `fs.watch`
  event arrives on the kernel's schedule; no fake clock delivers an inotify
  event, and a stubbed watcher would prove only that the stub calls its own
  callback. The wait is a bounded poll on the guard's own report, not a sleep.
- `vitest.config.ts`'s `wallClockBudgetInclude` gains the same file (the
  admission gate requires serialized-lane membership).
- `tests/config/knip-entry-coverage.test.ts`'s `LANE_ALREADY_REACHED` gains the
  same file. **Verified both ways, not copied from its siblings:** `npm run knip`
  with no entry reports no unused file, and with
  `"tests/support/tests-tree-write-guard.test.ts"` added to `knip.jsonc`'s entry
  list reports `tests/support/tests-tree-write-guard.test.ts  knip.jsonc  Remove
  redundant entry pattern`. The probe is recorded next to the row.

### Suites run (mechanical selection, not hand-picked)

- 91 files: `ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts tests/config/*.test.ts` — 4 laps green (quoted above).
- 32 files: every test that greps for a symbol this PR changed
  (`readWalkedFile`, `walkedFilesVanished`, `scanVacuousSkips`,
  `localImportTargets`, `scanDualInstanceImports`, `testSourceFiles`,
  `sweepShapeFiles`, `installTestsTreeWriteGuard`, `walkTestFiles`,
  `countsByDetector`, `findWin32Gates`, `getWin32LaneFiles`) plus all of
  `tests/support/` — `Test Files 32 passed (32)`, `Tests 553 passed | 1 skipped (554)`.
- `tests/support/` + `tests/clients/pi-lens-home-hermeticity.test.ts` after the
  final commit — `Test Files 17 passed (17)`, `Tests 327 passed (327)`.
- `npx oxfmt --check` on every touched file: `All matched files use the correct format.`
- `npm run knip`: 2 configuration hints (`scripts/lib/win32-gate-population.mjs`, `tests/mcp/session-end.smoke.test.ts` — neither introduced here, neither a file this PR adds to the entry list).

Not run locally: the full suite (serialized machine-wide, twelve lanes
contending) — delegated to CI's Unit tests job, per AGENTS.md.

### Test assessment

| file | what it uniquely pins | redundancy |
| --- | --- | --- |
| `tests/support/tests-tree-write-guard.test.ts` (new) | the run-level guard's classification, bounding, teardown and real-watch delivery | none; no sibling covers a `tests/`-tree write |
| `tests/support/sweep-kit.test.ts` | the shared sweep machinery; +4 cases for the vanished-file contract | none |
| `tests/clients/pi-lens-home-hermeticity.test.ts` | the #3042/#3050 registry-isolation detector; the red-first proof now runs over a private root | the proof's claim is unchanged — same walker, same options, different root |
| `tests/config/{sweep-floor-coverage,vacuous-skip-coverage,latency-logger-mock-shape,lsp-spawn-heavy-coverage,vi-mock-export-sweep,win32-gate-lane,lsp-service-double-sweep,tmp-fixture-hygiene,timing-sensitive-coverage,git-fixture-governance}.test.ts` | unchanged subjects; only their read path moved to `readWalkedFile` | none |
| `tests/clients/flake-shape-ratchet.test.ts` | the flake-shape ratchet; +1 admission row | none |
| `tests/config/knip-entry-coverage.test.ts` | knip entry coverage; +1 verified lane row | none |

No vacuous test was found in the files touched: every case in
`sweep-kit.test.ts`'s and the hermeticity file's edited regions reds under at
least one mutation above or under the pre-fix producer transcript. Two
hand-rolled recursive walkers were DELETED in favour of `listSourceFiles`
(`tests/config/latency-logger-mock-shape.test.ts:22`,
`tests/config/timing-sensitive-coverage.test.ts:57`) — net helper count down, not
up.

## Blast radius

No production module is touched: the diff is `tests/**`, `vitest.config.ts`,
`scripts/lib/win32-gate-population.mjs` and one changelog fragment. `module_report`
with `blastRadius: true` is therefore empty for this diff — recorded explicitly
rather than implied.

Call-tree diff for the one shared seam, `tests/support/sweep-kit.ts`:

```
listSourceFiles(root)                                    (unchanged)
+ readWalkedFile(file): string | undefined               NEW  (+31)
+   └── readWalkedFiles(files): {file, source}[]         NEW  (+12)
+   └── walkedFilesVanished(): readonly string[]         NEW  (+4)
callers ABOVE (every list-then-read over the tests/ tree; 16 files):
+   tests/clients/pi-lens-home-hermeticity.test.ts  scanFlagged, the #3048 fixture read
+   tests/config/sweep-floor-coverage.test.ts       sweepShapeFiles (+ the floor check now reuses its source: one read, not two)
+   tests/config/vacuous-skip-coverage.test.ts      via tests/support/vacuous-skip-scan.ts scanVacuousSkips
+   tests/config/latency-logger-mock-shape.test.ts  findLatencyMocks   (walker folded onto listSourceFiles)
+   tests/config/lsp-spawn-heavy-coverage.test.ts   candidate filter
+   tests/config/vi-mock-export-sweep.test.ts       scan + scanLatencyLoggerSurface
+   tests/config/win32-gate-lane.test.ts            both gate walks
+   tests/config/lsp-service-double-sweep.test.ts   liveCounts
+   tests/config/tmp-fixture-hygiene.test.ts        scanMkdtempSites + the call-window re-read
+   tests/config/timing-sensitive-coverage.test.ts  timingFiles          (walker folded onto listSourceFiles)
+   tests/config/git-fixture-governance.test.ts     walkFiles
+   tests/support/flake-shape-scan.ts               countsByDetector's scanFile
+   tests/support/module-instance-scan.ts           scanDualInstanceImports
+   tests/support/hook-await-scan.ts                localImportTargets (the ratchet calls it per walked test file)
+   tests/support/host-event-shape-scan.ts          scanHostEventShapeViolations   (round 2 F1-A; walker folded onto listSourceFiles)
+   tests/config/escape-regexp-fold-sweep.test.ts   the offender scan over { dir: "tests", floor: 900 }   (round 2 F1-B)
+   scripts/lib/win32-gate-population.mjs           findWin32Gates — its OWN twin of the helper (scripts/ must not import tests/), now exported so the parity is pinned
callees BELOW: fs.readFileSync (unchanged), console.warn (once per distinct path)
```

Behaviour change for every caller: a file that is gone at read time is dropped
instead of throwing. Population floors (`assertNonEmptyScan`, `minScanned`) are
untouched and still catch a walk that loses its whole population.

`vitest.config.ts` gains one `globalSetup` row (last, after the in-place build
and the prewarms), one `wallClockBudgetInclude` row, and an `export` on
`sharedGlobalSetup` (round 2 F2).

Cost, corrected in round 2 (F3): per ACTIVATED PROJECT, not per run and not per
fork. `vitest.config.ts` attaches `globalSetup: sharedGlobalSetup` to seven
projects, and vitest runs the arm once for each project that has matching files
— so a full run pays one `listSourceFiles(tests/)` baseline walk and one
recursive `fs.watch` registration per activated project (~305 inotify watches
each), and the same report can print once per project. Measured: a run touching
two projects invoked the arm twice, in ONE pid, each time with its own module
instance (`moduleCount=1` on both). Still nothing per test, per file or per
spawn, so there is no hot-path delta to measure.

## Observability

Context, not a fourth form: this diff touches no runtime file (`clients/`,
`tools/`, `mcp/`, `index.ts` are untouched), so there is no ledger kind to add.
The three test-infrastructure records it DOES add each report on the run's own
stderr and each is asserted in this diff, quoted here so they are not taken on
trust:

- `[sweep-kit] <path> vanished between the walk and the read; skipped (#3082)`
  — once per distinct path (`tests/support/sweep-kit.ts:584`), asserted by
  `expect(warn.mock.calls[0]?.[0]).toMatch(/vanished between the walk/)` in
  `tests/support/sweep-kit.test.ts`.
- `[tests-tree-write-guard] not watching <root>: … — #3082 recurrences will not
  be caught on this platform` — once per run
  (`tests/support/tests-tree-write-guard.ts:157`), asserted by
  `expect(warn.mock.calls[0]?.[0]).toMatch(/will not be caught/)` in
  `tests/support/tests-tree-write-guard.test.ts`.
- the `#3082: <n> source file(s) were created inside the repo's own tests/ tree
  during this run:` report the globalSetup teardown throws
  (`tests/support/tests-tree-write-guard-setup.ts:30`), asserted by
  `expect(() => teardown()).toThrow(/scratch-setup\.test\.ts/)` and quoted
  end-to-end in the red-first transcript above.

## Class sweep

**Shape:** AGENTS.md "a test that drives shared state it does not own" — here the
`tests/` directory — composed with the list-then-read window every directory
walker has. Prior art for the exact shape is already in the tree:
`tests/support/check-build-freshness.ts:129` records that #2270's end-to-end test
"planted files directly under the real `tests/support/`, which a concurrent
`module-instance-scan` walk of the same live tree could observe mid-write —
`ENOENT` under parallel test workers". That is this defect, one round earlier, on
the same seam.

**Producer sweep — whole tree.** Grepped `tests/`, `scripts/`, `tools/`, `mcp/`,
`clients/` for `writeFileSync|appendFileSync|copyFileSync|cpSync|renameSync|symlinkSync`
whose target is a source-extension path, then filtered to targets rooted at a
tracked tree rather than a fixture/temp root:

| site | verdict |
| --- | --- |
| `tests/clients/pi-lens-home-hermeticity.test.ts` (the scratch `*.test.ts` under `TESTS_ROOT`) | **the defect — fixed here** |
| `tests/index-2992-integration.test.ts:30` writes `<repoRoot>/index-2992-probe.ts` | same shape, **not a member**: no walker enumerates the repo ROOT directory (`grep -rE '(listSourceFiles\|readdirSync\|globSync)\s*\(\s*(REPO_ROOT\|repoRoot\|ROOT)\s*[,)]'` returns only per-root loop variables and fixture roots; `tests/support/bounded-telemetry-scan.ts:30`'s `SCAN_ROOTS` names `index.ts` as a FILE). Left alone, and the guard is deliberately scoped to `tests/` rather than widened to cover it. |
| `tests/build-freshness-guard.test.ts:145`, `tests/scripts/check-pr-body.test.ts:347`, `tests/scripts/check-pr-body.test.ts:441`, `tests/clients/ast-grep-antislop-error-floor.test.ts:125`, `tests/config/result-contract-governance.test.ts:44`, `tests/mcp/build-graph-cli.test.ts:57` | clean: every target is under an `mkdtemp` fixture root (`e2eRoot`, `fixtureCwd`, `fixtureRepo`, `cwd`, `projectDir`), not a tracked tree |
| `tests/config/vi-mock-export-sweep.test.ts:1219` writes `tests/support/vi-mock-export-baseline.json` | in-tree write, but `.json` — outside every walker's extension set, and gated behind `VI_MOCK_EXPORT_REGEN`. Left; it is why `GUARDED_EXTENSIONS` is extension-scoped rather than "any file". |

**Consumer sweep — population, with a per-member verdict.** Every list-then-read
whose walked root is the repo's own `tests/` tree (the tree a concurrently
running test can mutate), found by grepping the walk call sites rather than the
function name:

`grep -rnE "(listSourceFiles|walkTestFiles|sweepShapeFiles|readdirSync|globSync)\s*\(" tests/ --include=*.ts | grep -E "TESTS_ROOT|\"tests\"|'tests'"`
plus the two `tests/support/*` walkers that reach the same population through
`testSourceFiles`. All 14 members folded onto `readWalkedFile` (13 in the list
above) or verified read-free:

- `tests/support/check-build-freshness.ts:119` — walks the same tree but only
  `existsSync`es: no read to lose, left unchanged.
- `scripts/lib/win32-gate-population.mjs:79` — a member, but `scripts/` cannot
  import from `tests/`. It carries the same six lines with a comment pointing at
  `readWalkedFile` as the TypeScript-side seam; a deliberate cross-module-system
  duplicate, named here rather than hidden.

Walkers rooted at `clients/`, `tools/`, `mcp/`, `scripts/` or `.github/workflows`
are **out of the population**: nothing in the suite creates or removes a source
file in those trees at runtime (the producer sweep above is the evidence), and
tolerating a vanished file there would be silencing a real failure (#3092's own
argument against a blanket catch).

**Consolidation verdict: fold.** The family folds onto one seam,
`tests/support/sweep-kit.ts`'s `readWalkedFile`/`readWalkedFiles` — the module
that already owns `listSourceFiles`, so the read now lives beside the walk that
produced it. Deleting the helper would push a `try/catch` (or a silent
`fs.readFileSync`) back into 13 call sites in 13 files, which is concentration,
not relocation. Net helper count goes DOWN: two hand-rolled recursive walkers
(`latency-logger-mock-shape`, `timing-sensitive-coverage`) were deleted in favour
of `listSourceFiles` in the same change.

## Merge order

`#3100` (fix/3083-backstop-state-pin) also edits
`tests/clients/pi-lens-home-hermeticity.test.ts`: it adds imports at the top and
a new `describe` at the END of the file, while this PR edits the middle (the
registry-isolation walk and its red-first proof). The regions are disjoint apart
from the import block, so whichever lands second may need a one-hunk rebase
there. No semantic interaction: #3100's new case drives `session_start` against
`PI_LENS_HOME`, which this PR's guard does not watch (it watches `tests/`). This
PR deliberately does not touch `tests/support/vitest-setup.ts` — the file #3100
pins — and adds its own `globalSetup` file instead.

---

## Round 2 (review findings)

All six findings applied. Each was reproduced before it was fixed; the prior
round's ten mutations were re-run on the new head and all ten are still red.

### F1 (blocking) — two members the first sweep missed

Reproduced, not taken on faith. A local probe (never committed) replayed #3082's
producer shape — write a `scratch-probe-3104-race.test.ts` at the top of the
walked tree, hold it across
three full walks of `tests/`, remove it, 300 rounds.

**Member A**, `tests/support/host-event-shape-scan.ts` — hand-rolled recursive
`readdirSync` at `:58`, raw read at `:162`. Pre-fix, probe + `tests/support/`,
**red on 3 of 4 laps**:

```
 ❯ |default| tests/support/host-event-shape-scan.test.ts (6 tests | 1 failed) 310ms
 FAIL  |default| tests/support/host-event-shape-scan.test.ts > host event-shape scan (#1681) > finds no fixture that puts sessionId/provider/model on a session_start or tool_result event
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-probe-3104-race.test.ts'
 ❯ scanHostEventShapeViolations tests/support/host-event-shape-scan.ts:162:18
```

**Member B**, `tests/config/escape-regexp-fold-sweep.test.ts:161`, whose
population includes `{ dir: "tests", floor: 900 }` (`:123`). Pre-fix, probe +
`tests/config/`, red on lap 3 of 3:

```
 FAIL  |default| tests/config/escape-regexp-fold-sweep.test.ts > escapeRegExp single-source-of-truth (#2558) > has no local escaping-helper definition outside the canonical leaf
Error: ENOENT: no such file or directory, open '/home/akis/.cache/pi-lens-orchestrator/worktrees/fix-3082/tests/scratch-probe-3104-race.test.ts'
 ❯ tests/config/escape-regexp-fold-sweep.test.ts:161:16
    159|    const rel = relativePosix(root, file);
    160|    if (rel === CANONICAL_FILE) continue;
    161|    const raw = readFileSync(file, "utf8");
       |                ^
```

Fixed: A's `testFiles()` now calls `listSourceFiles` (`skipDeclarations: false`
keeps the population byte-identical, so its 400-file floor still measures the
same walk) and reads through `readWalkedFiles`; B reads through
`readWalkedFile`. Post-fix, the same probe, **15 laps, 0 red**:

```
probe + tests/support/ + escape-regexp-fold-sweep   lap1-4   Test Files 18 passed (18)
probe + tests/support/                              lap5-10  Test Files 17 passed (17)
probe + host-event-shape-scan + escape-regexp       lap11-15 Test Files  3 passed  (3)
```

Honest caveat: the tolerance's own `[sweep-kit] … vanished between the walk and
the read` record did NOT appear in any of those 15 laps, so the strongest claim
the laps support is "the red is gone", not "the tolerance was exercised". The
tolerance path itself is proven deterministically by
`tests/support/sweep-kit.test.ts` and by mutations M1/M2.

#### The re-sweep, by walk ROOT rather than by line

The first sweep grepped the walk-call line for a tests-root token, which is why
both members were invisible (their root constant sits elsewhere in the file).
The replacement method — script kept at `.tmp-disk/root-sweep.mjs`, method
reproduced here — walks `tests/`, `scripts/`, `tools/`, `mcp/`, `clients/` and
keeps any file that (1) calls a walker, (2) calls a read, and (3) mentions a
tests-tree root ANYWHERE in the file. 25 files match; per-member verdict:

| file | verdict |
| --- | --- |
| `tests/support/host-event-shape-scan.ts` | **member — fixed this round (F1-A)** |
| `tests/config/escape-regexp-fold-sweep.test.ts` | **member — fixed this round (F1-B)** |
| 14 files already folded in round 1 (`pi-lens-home-hermeticity`, `sweep-floor-coverage`, `vacuous-skip-scan`, `latency-logger-mock-shape`, `lsp-spawn-heavy-coverage`, `vi-mock-export-sweep`, `win32-gate-lane`, `lsp-service-double-sweep`, `tmp-fixture-hygiene`, `timing-sensitive-coverage`, `git-fixture-governance`, `flake-shape-scan`, `module-instance-scan`, `hook-await-scan`) | member — already reads through the seam |
| `scripts/lib/win32-gate-population.mjs` | member — carries its own twin (F6) |
| `tests/config/knip-entry-coverage.test.ts` | out: walks `.github/workflows` and `.claude/`; every `readFileSync` is a NAMED registry path |
| `tests/skills/skill-doc-drift.test.ts` | out: walks `.claude/skills` for `.md`; touches tests/ paths with `existsSync` only, never reads them |
| `tests/support/vitest-setup.ts` | out: no walk-read; its `/tests/` tokens are path splits for the tmp-hygiene report |
| `scripts/lib/astgrep-self-scan.mjs` | out: `scanPaths` (`["clients","tests"]`) is argv for the ast-grep CHILD; its own reads are the rules dir |
| `scripts/check-pr-body.mjs`, `scripts/gen-cargo-modules-snapshot.mjs`, `scripts/lib/stryker-diff.mjs`, `scripts/pre-push-targeted-tests.mjs` | out: dev/CI entry points, none running concurrently with the unit suite; `gen-cargo-modules-snapshot` walks `tests/fixtures` only |
| `clients/test-runner-client.ts` | out: discovers a USER project's test dirs, not this repo's `tests/` |
| `readRepoSource` in `tests/config/lsp-service-double-sweep.test.ts:274` and `tests/config/vi-mock-export-sweep.test.ts:144` | out, deliberately: `existsSync`+read over RATCHET-NAMED paths, where absence is an expected, reported state (a stale baseline row), not a race. Routing them through `readWalkedFile` would warn on the healthy case. |

Consolidation verdict unchanged: fold onto `readWalkedFile`. 16 files now read
through the shared seam; the one `scripts/lib` twin is named above.

### F2 (blocking) — the guard's registration was unguarded

Reproduced: deleting `"./tests/support/tests-tree-write-guard-setup.ts"` from
`vitest.config.ts` left all eleven guard cases green, and the restored #3082
producer ran completely uncaught:

```
=== M10b: guard row deleted, producer restored ===
EXIT=0
0        (occurrences of "#3082:" in the run output)
```

Fixed: `sharedGlobalSetup` is now exported (`vitest.config.ts:57`) and
`tests/support/tests-tree-write-guard.test.ts` asserts the WHOLE list — every
arm has the same silent-absence property, `git-config-guard-setup.ts` and
`prewarm-tool-home.ts` included — plus that every project in `test.projects`
uses that exact list, so a project added without it is caught too. The `.ts`
config import is registered in `tests/config/module-instance-coverage.test.ts`'s
`reviewedExceptions`, the same row its four siblings carry.

**M10 (new)** — delete the guard's row:

```
 FAIL  |wall-clock-budget| tests/support/tests-tree-write-guard.test.ts > tests-tree write guard (#3082) > every run-level guard is registered in vitest.config.ts, on every project
AssertionError: expected [ …(4) ] to deeply equal [ …(5) ]
 Test Files  1 failed (1)
      Tests  1 failed | 11 passed (12)
```

### F3 — "one watcher per run" was wrong; wording fixed, no latch added

Confirmed and measured rather than argued. A run touching two projects invoked
the arm twice:

```
[3104-probe] arm invoked pid=174156 moduleCount=1
[3104-probe] arm invoked pid=174156 moduleCount=1
```

Same pid, but `moduleCount=1` BOTH times: each project gets its own module
instance of the setup file. So a module-level "install once" latch is inert
here — precisely AGENTS.md shape 25 — and only a `process`-keyed singleton would
deduplicate. Not added: it buys one baseline walk and ~300 inotify watches per
extra project, at the price of process-lifetime state plus an assumption about
which project's teardown runs first (that teardown owns the report). N watchers
all observe the same events; the only visible consequence is the report printing
once per project. The Blast-radius section above, the changelog fragment and the
setup module's docstring now all say "per activated project".

### F4 — `mutation (advisory)` red from Stryker's 5 s dry-run default

Both named cases now carry an explicit budget with the reason, the pattern
`escape-regexp-fold-sweep.test.ts` already uses:
`tests/config/knip-entry-coverage.test.ts` "covers every test file admitted by
the Vitest and Windows lanes (#2837)" → `}, 60_000)` (measured 8.0 s under the
dry run), and `tests/config/win32-gate-lane.test.ts` "keeps the workflow
population and execution filter wired" → `}, 60_000)` (22.8 s). Both are
directory-walk time, not wall-clock waiting, so neither is a flake-shape
admission.

### F5 — the STATELESS docstring, and an unbounded diagnostic set

`tests/support/sweep-kit.ts`'s module docstring now carries the carve-out: the
vanished-path set is per-FORK, never reset by design (resetting would turn
"warn once per distinct path" back into "warn per occurrence"), and safe because
it is purely diagnostic — no sweep reads it, no verdict depends on it, and a
stale entry can only suppress a repeat WARNING. It is now a
`BoundedSet<string>(VANISHED_PATH_RECORD_CAP = 256)` from
`clients/bounded-cache.ts` rather than a raw `Set` (shape 9).

**M11 (new)** — swap the `BoundedSet` back for a raw `Set`:

```
 FAIL  |default| tests/support/sweep-kit.test.ts > sweep-kit: readWalkedFile (#3082) > bounds the vanished-path record at VANISHED_PATH_RECORD_CAP
AssertionError: expected 307 to be 256 // Object.is equality
 Test Files  1 failed (1)
      Tests  1 failed | 80 passed (81)
```

### F6 — the `scripts/lib` twin warned per occurrence

Matched to the TypeScript seam (once per distinct path, bounded) rather than
documented as a divergence, and `readWalkedFile` is now exported from
`scripts/lib/win32-gate-population.mjs` (+ its `.d.mts` row, so
`dmts-export-drift` stays green) so the parity is pinned by a test instead of
asserted in a comment.

**M12 (new)** — drop the dedupe in the scripts copy:

```
 FAIL  |default| tests/config/win32-gate-lane.test.ts > win32 gate lane governance (#2536) > tolerates a vanished walked file and warns once per distinct path (#3082)
AssertionError: expected "warn" to be called 1 times, but got 2 times
 Test Files  1 failed (1)
      Tests  1 failed | 2 passed (3)
```

### Non-blocking notes

(a) **Applied.** `tests/clients/pi-lens-home-hermeticity.test.ts`'s read of the
one REQUIRED, committed fixture goes back to `fs.readFileSync`, so its absence is
a clean ENOENT naming the path instead of a `TypeError` inside `stripSource`.
The population read in `scanFlagged` still goes through `readWalkedFiles`.

(b) **Recorded as a known limit**, in the guard's own docstring: a scratch file
left behind by a KILLED run (SIGKILL skips `finally`) is in the baseline and
therefore invisible to this guard for the whole run. Nothing else catches it
either — no governance test asserts "no untracked source file under `tests/`"
(grepped `git ls-files`/`untracked` across `tests/`). `git status` is the current
answer; a `git ls-files` call at globalSetup would trade the gap for a real child
process on every run.

(c) **Refreshed.** The red-first transcript above now quotes this head:
`at Object.teardown (…/tests/support/tests-tree-write-guard-setup.ts:38:22)`,
`EXIT=1`.

### Round 2 verification

Every mutation from round 1 re-run on the new head — all still red:

```
M1 ENOENT arm         Tests  3 failed | 78 passed (81)
M2 dedupe             Tests  1 failed | 80 passed (81)
M3 watch callback     Tests  1 failed | 11 passed (12)
M4a classifier true   Tests  2 failed | 10 passed (12)
M4b classifier false  Tests  7 failed |  5 passed (12)
M5 baseline arm       Tests  1 failed | 11 passed (12)
M6 cap                Tests  1 failed | 11 passed (12)
M7 teardown throw     Tests  1 failed | 11 passed (12)
M8 teardown close     Tests  1 failed | 11 passed (12)
M9 extensions         Tests  1 failed | 11 passed (12)
```

Suites, after `npm run build`, same `TMPDIR`/`PI_LENS_HOME` pins:

```
91-file governance batch  lap1 exit=0   Test Files  91 passed (91)   Tests  1146 passed | 1 skipped (1147)
91-file governance batch  lap2 exit=0   Test Files  91 passed (91)   Tests  1146 passed | 1 skipped (1147)
tests/config/ (whole dir) exit=0        Test Files  60 passed (60)   Tests   585 passed | 1 skipped  (586)
tests/support/ + hermeticity + host-event-shape-scan + escape-regexp-fold-sweep
  + knip-entry-coverage + win32-gate-lane + flake-shape-ratchet
  + module-instance-coverage            Test Files  22 passed (22)   Tests   398 passed        (398)
```

`npx oxfmt --check` on every touched file: `All matched files use the correct
format.` `node scripts/check-changelog-fragments.mjs`: `changelog fragments OK
(85 entries in .changelog/)`.


apmantza:
Merged as reviewed (round-2 verify: all six findings fixed and mutation-proven). One observability correction is tracked in 3107: the worker-side vanished-file record is `console.warn`, which the default reporter drops on a green run, so the body's "reports on the run's own stderr" holds only under `--reporter=verbose`.

PR #3133, opened 2026-09-16
test(support): one balanced-delimiter matcher for sweep-kit and vacuous-skip-scan (closes #3072)

### LINKED ISSUE #3072: test(support): fold sweep-kit.ts's and vacuous-skip-scan.ts's duplicate balanced-delimiter matchers onto one seam
## What

`tests/clients/pi-lens-home-hermeticity.test.ts`'s `topLevelFunctionBodies`
and `findMockCallText` (added by #3050) each hand-roll their own
balanced-delimiter scan — a brace counter for the function body, a paren
counter for the `vi.mock(...)` call — but `tests/support/vacuous-skip-scan.ts`
already has this exact logic, private to that file:

- `matchingCloseBrace(source, openIndex)` (`tests/support/vacuous-skip-scan.ts:170-180`)
  — forward brace-depth counting from an opening `{` to its match. Byte-for-byte
  the same loop shape as `pi-lens-home-hermeticity.test.ts`'s `record()` helper.
- `skipGroup(source, openIndex, close)` (`tests/support/vacuous-skip-scan.ts:182-193`)
  — the same counting technique generalized to any open/close delimiter pair,
  which already covers the paren-matching `findMockCallText` needs.
- `matchingOpenParen(source, closeIndex)` (`tests/support/vacuous-skip-scan.ts:138-148`)
  — the same technique run backward.

## Why

Two private copies of the identical technique in two different test-support
files is exactly AGENTS.md's "hand-maintained list/logic that mirrors a
registry/helper is a defect" shape, one layer down (a private ALGORITHM
instead of a private DATA structure). Neither copy is wrong today, but a bug
fixed in one (e.g. an off-by-one at end-of-file, or adding a
comment/string-aware variant) will not propagate to the other, and a third
sweep that needs the same primitive will be tempted to write a THIRD copy.

## Proposed fix

Promote ONE `matchingCloser(source, openIndex, openChar, closeChar)` (or two
thin wrappers, `matchingCloseBrace`/`matchingOpenParen`, calling one shared
core) into `tests/support/sweep-kit.ts` — the seam `pi-lens-home-hermeticity.
test.ts` already imports from for every other primitive (`stripSource`,
`listSourceFiles`, `escapeRegExp`, ...) — and have both
`vacuous-skip-scan.ts` and `pi-lens-home-hermeticity.test.ts` consume it,
deleting their own private copies.

## Acceptance criteria

1. One shared balanced-delimiter matcher in `tests/support/sweep-kit.ts`,
   covering both the forward (open→close) and backward (close→open) directions
   `vacuous-skip-scan.ts` and `pi-lens-home-hermeticity.test.ts` each need today.
2. `vacuous-skip-scan.ts`'s `matchingCloseBrace`/`skipGroup`/`matchingOpenParen`
   and `pi-lens-home-hermeticity.test.ts`'s inline brace/paren counters in
   `topLevelFunctionBodies`'s `record()` and `findMockCallText` are deleted in
   favor of the shared helper — net helper count does not go up (single-source-
   of-truth rule).
3. Every existing test in both files stays green with no behavior change.

Refs #3050

Follow-up from PR #3066's review round 2 (F6): named rather than fixed inline
to keep that round's diff to the actual findings.

apmantza:
## Summary

`tests/support/vacuous-skip-scan.ts` (`matchingOpenParen`/`matchingCloseBrace`/
`skipGroup`) and `tests/clients/pi-lens-home-hermeticity.test.ts`
(`balancedBraceEnd`, plus two more inline paren-balance loops in
`bodyBraceAfterParams` and `findMockCallText`) carried five private copies of
the same forward/backward balanced-delimiter scan — `matchingCloseBrace` and
`skipGroup` were even redundant with each other inside `vacuous-skip-scan.ts`
itself. Promoted one forward matcher (`matchingCloseIndex`) and one backward
matcher (`matchingOpenIndex`) into `tests/support/sweep-kit.ts` — the seam
both files already import from for every other scan primitive — and deleted
every private copy in favor of it. Net helper count: five private copies →
two shared exports (down, not sideways).

Both former consumers' existing tests stay assertion-identical (see Tests).

Closes #3072 — every acceptance criterion is met.

## Behavioural difference table

Read both sets of implementations fully first, per the delegation. They are
**byte-for-byte identical algorithms** — no observable difference to fold
around:

| Aspect | `vacuous-skip-scan.ts` (3 copies) | `pi-lens-home-hermeticity.test.ts` (3 copies) |
|---|---|---|
| Comment/string handling | None — assumes the caller already ran `stripSource` | None — same assumption |
| Escape handling | None | None |
| Depth limit | None (unbounded) | None (unbounded) |
| Direction | forward (`matchingCloseBrace`, `skipGroup`) + backward (`matchingOpenParen`) | forward only (`balancedBraceEnd`, the two inline loops) |
| Open/close derivation | `skipGroup` derives `open` from `source[openIndex]`; `matchingCloseBrace`/`matchingOpenParen` hardcode `{`/`(` | all three hardcode `(` or `{` |

The one real variation is **caller-level**, not in the matcher itself:
`findMockCallText` runs its paren balance over `commentsBlankedStringsKept`
text (strings KEPT), while every other call site runs over fully
comment-AND-string-blanked text. That's an input-stripping-policy choice made
by each caller before it ever reaches the balancer, not a behavioural
difference between the two matcher implementations — the shared
`matchingCloseIndex`/`matchingOpenIndex` apply identically regardless of
which text a caller hands them, exactly like every private copy already did.
One shared implementation, no per-consumer variant, no third variant.

## Type of change

- [ ] Bug fix
- [x] Enhancement (improvement to existing capability) *(none of the boxes
  fit exactly — this is a pure test-support refactor: same behavior, shared
  helper instead of five private copies. Closest available label.)*
- [ ] Documentation

## Area

- [x] area:tests

## Checklist

- [x] The change has tests — no NEW tests (a pure internal refactor of shared
  private scan machinery); every EXISTING test in both former-consumer files
  stays green, proven by the mutation-proof runs below
- [ ] Every NEW regression test is proven RED on pre-fix code, the red output is quoted — N/A, no new regression test (see Tests: this is a refactor with an existing-behavior-preservation proof instead)
- [x] Every new guard/branch/filter is mutation-proof: deleting or neutering it reds at least one test — see Mutation proof below
- [ ] `npm run lint` — not run (targeted build + vitest only, per delegation; CI is authoritative)
- [ ] `npm run build:dist` — N/A, no code under `clients/`, `commands/`, `tools/`, or `index.ts` changed (test-support only)
- [ ] `package-lock.json` in sync — unchanged, no dependency touched
- [ ] `AGENTS.md` updated — N/A, no behavior/command/convention/invariant changed
- [x] Changelog — **none**, by design: test-support-only internal refactor, no user-facing change. Per delegation.
- [x] Commit subject includes the issue number: `(refs #3072)`

## Tests

No new test files or cases. This PR is a pure internal-implementation
refactor of shared private scan machinery: five identical brace/paren-balance
loops become two shared exports (`matchingCloseIndex`, `matchingOpenIndex` in
`tests/support/sweep-kit.ts`), and both former-consumer test files
(`tests/config/vacuous-skip-coverage.test.ts` for `vacuous-skip-scan.ts`,
and `tests/clients/pi-lens-home-hermeticity.test.ts` itself) are proven to
stay assertion-identical:

```
npx vitest run tests/support/*.test.ts tests/clients/pi-lens-home-hermeticity.test.ts tests/config/vacuous-skip-coverage.test.ts tests/config/sweep-floor-coverage.test.ts --reporter=dot

 Test Files  19 passed (19)
      Tests  347 passed (347)
```

```
npx vitest run tests/config --reporter=dot

 Test Files  60 passed (60)
      Tests  589 passed | 1 skipped (590)
```

```
npx vitest run <30 tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts files>

 Test Files  30 passed (30)
      Tests  542 passed (542)
```

### Mutation proof (in lieu of a red-first regression test — this PR adds no
new behavior, only relocates existing behavior)

Per AGENTS.md, every new guard/branch is mutation-proofed even when it isn't
a "new regression test": `matchingCloseIndex` and `matchingOpenIndex` are new
exported functions now load-bearing for BOTH former consumers, so each was
neutered in the built output and the suite that should catch it was run.

**Mutation 1 — `matchingCloseIndex` (forward): depth starts at 1 instead of
0**, so the first close returned is not the balancing one for any nested
delimiter:

```
npx vitest run tests/config/vacuous-skip-coverage.test.ts tests/clients/pi-lens-home-hermeticity.test.ts --reporter=verbose

 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > mutation-proof: without the transitive closure, registerInstance (three calls from the lock) drops out of the hazardous set
AssertionError: expected [] to include 'registerInstance'
 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > reviewer probe (round 2, F5): an arrow-function export calling the lock primitive enters the hazardous set
AssertionError: expected [] to include 'sweepFakeThing'
 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > #3050 recurrence: a default parameter VALUE's braces are not read as the function body
 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > registered-or-fail: every unpinned reach into a producer's registry is named or exempted with a reason
AssertionError: expected [ Array(1) ] to deeply equal []
 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > mutation-proof: a mock naming a producer symbol but still falling through to `actual.<symbol>(` does not isolate
 FAIL tests/clients/pi-lens-home-hermeticity.test.ts > ... > reviewer probe (round 2, F2): a documentation STRING narrating a PI_LENS_HOME pin does not isolate an otherwise-identical hazardous file
AssertionError: expected undefined to deeply equal [ 'registerInstance' ]

 Test Files  2 failed (2)
      Tests  8 failed | 19 passed (27)
```

```
npx vitest run tests/config/vacuous-skip-coverage.test.ts --reporter=verbose

 × catches the shapes it claims to catch
   → these evaded the detector: expected [ …(5) ] to deeply equal []
   + [ "a platform gate that returns before the first assertion", "a bare return in a catch, with no skip call", "a gated body nested inside a declarative skipIf wrapper", "a gated body inside an it.each table", "test() rather than it()" ]
 × reads the compliant sites as compliant BECAUSE of the skip call
   → removing the skip calls must expose the returns they excuse: expected 0 to be greater than or equal to 4

 Test Files  1 failed (1)
      Tests  2 failed | 3 passed (5)
```

Both former consumers' test files red under the same single-matcher mutation,
exactly as the fold requires.

**Mutation 2 — `matchingOpenIndex` (backward): depth starts at 1 instead of
0**, same nested-delimiter mishandling in the backward direction (the one
`isFunctionBrace` in `vacuous-skip-scan.ts` uses; `pi-lens-home-hermeticity.
test.ts` has no backward consumer):

```
npx vitest run tests/config/vacuous-skip-coverage.test.ts --reporter=verbose

 × reads the compliant sites as compliant BECAUSE of the skip call
   → removing the skip calls must expose the returns they excuse: expected 3 to be greater than or equal to 4

 Test Files  1 failed (1)
      Tests  1 failed | 4 passed (5)
```

Both mutations reverted via `git checkout HEAD -- tests/support/sweep-kit.ts`
(committed fix restored, confirmed by `git diff --stat` empty and a clean
rebuild + green rerun after each).

### Test assessment

No test file's assertions changed in this PR — `vacuous-skip-scan.ts` and
`pi-lens-home-hermeticity.test.ts` are test-SUPPORT/test files whose own
internal helper functions moved, not their test cases. No test is made
redundant and none is removed.

## Blast radius

`tests/support/sweep-kit.ts` is imported by 71 files, but this PR is
**purely additive** to its public surface (two new exports,
`matchingCloseIndex`/`matchingOpenIndex`); no existing export's signature or
behavior changed, so none of the other 70 importers are affected — verified
by the full `tests/config` run (60 files, 589 passed) and the governance
sweep set (30 files, 542 passed) above, neither of which reds.

Directly touched:
- `tests/support/vacuous-skip-scan.ts` → consumed by
  `tests/config/vacuous-skip-coverage.test.ts` (its only consumer, per grep)
  — green.
- `tests/clients/pi-lens-home-hermeticity.test.ts` → a leaf test file, no
  consumers — green.

No production code (`clients/`, `commands/`, `tools/`, `index.ts`) touched.

## Observability

## Class sweep

**Shape**: private hand-rolled balanced-delimiter (brace/paren) depth-counter
in a test-support file, duplicated across files — AGENTS.md's
single-source-of-truth rule, one layer down (a private ALGORITHM instead of a
private DATA structure), per the issue's own framing.

Swept the whole tree (`clients/`, `tools/`, `mcp/`, `scripts/`, `tests/`,
`index.ts`) for `depth++`/`depth--`/`depth += 1`/`depth -= 1` loops paired
with a brace/paren-named function, beyond the two files this issue named:

- **A sibling shape, not folded here (different contract, out of this PR's
  delegated scope)**: `readBalancedArgs(source, openIndex): string` — returns
  the sliced ARGUMENT TEXT, not an index, and two of its four instances also
  handle quoted-string/escape content the matcher this PR folds never did:
  `tests/support/bounded-telemetry-scan.ts:133`, `tests/support/
  availability-classifiedby-scan.ts:105` (quote/escape-aware),
  `tests/clients/session-event-guard-sweep.test.ts:71`, and a fourth variant
  `callEnd` in `tests/config/latency-logger-mock-shape.test.ts:31`
  (quote/escape-aware, throws instead of returning -1). Four copies, at least
  two behaviorally different from each other (quote-awareness) as well as
  from this PR's matcher (return contract). Recommend a **separate** follow-up
  issue: this PR's delegation was scoped to exactly the pair #3072 named, and
  a fold here would need its own difference table and stricter-behavior
  decision (the quote/escape-aware pair is the correct base, per this issue's
  own "choose the stricter one" rule) — mixing it into a same-day PR risks
  exactly the "third variant" outcome the issue guards against.
- **Different domain, not a member of this shape**: production `clients/*.ts`
  balancers (`skipJsonValue` in `project-snapshot.ts`, `extractBalancedObjects`
  in `govulncheck-client.ts`, `parseBalancedJsonPrefix` in
  `lsp-structured-output.ts`, `containsGuardedSubstitution`/
  `matchGitVerbAtCommandPosition` in `git-guard.ts`, `tryParseSgMatches` in
  `sg-runner.ts`) parse real tool/JSON OUTPUT at runtime, not test source
  text — different input trust model, different failure semantics (a
  malformed tool output is a runtime concern, not a test-scan concern). Not a
  candidate for this seam.

**Consolidation verdict**: fold the pair #3072 named onto
`tests/support/sweep-kit.ts` (done, this PR). The `readBalancedArgs`/`callEnd`
family stays distributed for now — filing a follow-up issue naming all four
sites and the quote-handling difference, per delegation ("coordinate with
nothing else").


apmantza:
Merged as reviewed. Body correction, recorded here: the Class-sweep sentence overclaims — a grep for `depth++`/`depth--` under tests/ finds three further bare depth-counters (hook-await-scan.ts:79, host-event-shape-scan.ts:110, session-state-scan.ts:102/134, the last with a `body.length` unbalanced convention). They are in scope for #3134 alongside the readBalancedArgs family, not for #3072.

Issue #3083, opened 2026-09-15
test(hermeticity): 12 session_start-driving tests write the run-shared orphan-backstop stamp/lock through index.js


apmantza:
## What

Twelve `tests/` files drive the REAL `session_start` handler through
`index.js`, which calls `scheduleUntrackedOrphanSweep()`
(`index.ts:2346`). Against the run-shared `PI_LENS_HOME` (#2912: one
`.probe-home` for the whole Vitest run) each armed timer that fires takes
`<shared>/orphan-backstop.lock` and writes the 30-minute
`<shared>/orphan-backstop.json` cooldown stamp — machine-global state none of
them owns.

## Measured

Instrumented `clients/instance-reaper.js` (probe reverted; `getGlobalPiLensDir()`
+ `globalThis.__vitest_worker__.filepath` logged at
`scheduleUntrackedOrphanSweep` and `logBackstopOutcome`), one
`npx vitest run --project default` (1014 files) with the Unit lane's env
(`CI=true PI_LENS_TEST_TIMEOUT_SCALE=3 PI_LENS_TEST_NO_LOCK=1`):

```
=== SCHEDULERS by file (shared home) ===
37 tests/index-integration.test.ts
30 tests/index-wiring.test.ts
7  tests/clients/bootstrap-lazy-liveness.test.ts
7  tests/index-config-resolved-session-wiring.test.ts
6  tests/index-session-start-classification-wiring.test.ts
1  tests/index-loop-block-wiring.test.ts
1  tests/clients/dispatch/lazy-liveness.test.ts
1  tests/clients/formatters-lazy-liveness.test.ts
1  tests/clients/lsp-lazy-liveness.test.ts
1  tests/index-project-trust-wiring.test.ts
1  tests/index-observed-sweep-no-read-guard.test.ts
1  tests/index-mutation-bridge-recordability.test.ts

=== SWEEP OUTCOMES by file ===
9 (tests/index-integration.test.ts, cooldown)
1 (tests/index-integration.test.ts, clean)      <- writes the shared stamp
1 (tests/index-integration.test.ts, concurrent)
1 (tests/clients/instance-reaper-unref.test.ts, cooldown)   <- the victim, RED
1 (tests/clients/instance-reaper-unref.test.ts, disabled)
```

That `clean` is what redded master `038e28b` (see the unref fix). The victim
is fixed by pinning its own home; the WRITERS are still there, so the next
test that reads the shared backstop state inherits the same failure.

[...]

Issue #3058, opened 2026-09-15
tests/config/bounded-container-guard.test.ts peaks at 9.2 GB RSS (58% of the CI runner); the suite's memory floor fell 8.5 GB in five days and the Unit kill rate hit 30% (refs #2042)


apmantza:
## Measurement

From `[mem-file]` records in CI Unit-tests logs (the record PR #2093 added for
exactly this purpose), 2026-09-15:

| file | peak RSS | V8 heap | native | project | concurrency |
|---|---|---|---|---|---|
| `tests/config/bounded-container-guard.test.ts` | **9,226–9,241 MB** | 86–91 MB | ~9.1 GB | `default` | `maxWorkers: 3` |
| `tests/config/vi-mock-export-sweep.test.ts` | **5,642–5,645 MB** | 49–51 MB | ~5.6 GB | `default` | `maxWorkers: 3` |

Deterministic to ±0.2% across ten runs (seven red, three green).

The suite's host memory low-water, from the same logs:

| date | low-water available (of ~15,990 MB) | heaviest fork |
|---|---|---|
| 2026-09-10 | 12,543 MB | 1,479 MB |
| 2026-09-11 | 9,759 MB | 3,941 MB |
| 2026-09-13 | 9,441 MB | 3,962 MB |
| 2026-09-14 17:05Z | 7,875 MB | 5,627 MB |
| 2026-09-14 21:10Z | **4,435 MB** | **9,223 MB** |
| 2026-09-15 | 4,011–4,281 MB | 9,226–9,241 MB |

`1155ed91f` (2026-09-14 20:27:23Z) landed between the two 2026-09-14 runs. The
3,440 MB drop matches 9,223 − 5,638 = 3,585 MB, the new peak file displacing
the old one.

## Root cause

`scan()` calls `parse(Lang.TypeScript, source).root()` **six times per
container occurrence** — once each in `isGrowthShapedContainer`,
`hasBoundedConstructor`, `hasNamedSizeComparison`, `hasDeletingTimer`, and
**twice** in the adjacent `keyExpressionFor(source, container.name)` /
`determineKeyAxis(keyExpressionFor(source, container.name))` pair — each time
re-parsing the whole file, and each allocating an `@ast-grep/napi` native
arena that is never disposed. `walk()` then materialises a NAPI handle per
node. Nothing is released until the fork exits: 9.1 GB native against 87 MB of
V8 heap. Recurring defect shape 9 — bounded on one axis (`maxWorkers`),
[...]

PR #3137, opened 2026-09-16
test(support): freeze PEAK_RSS_ADMISSIONS and run the peak-RSS afterAll last (closes #3067)

### LINKED ISSUE #3067: worker-peak-rss gate: freeze PEAK_RSS_ADMISSIONS and order the mem afterAll last (refs #3058, #3062 review L1/L2)
Two low hardening items from the #3062 review, merged without them to land the footprint fix quickly.

1. `tests/support/worker-peak-rss.ts:52` — `PEAK_RSS_ADMISSIONS` is typed `Readonly<>` but not frozen; `peakRssProblem`'s default parameter reads the live object, so a test file can admit itself at runtime inside its own fork (probe: mutate the map, `peakRssProblem(file, 9999)` returns undefined; `Object.isFrozen` is false). Fix: `Object.freeze({})`, plus a test that a runtime mutation throws.
2. `tests/support/vitest-setup.ts` — the mem gate's `afterAll` is registered after the tmp-hygiene hook, so it runs first and its throw skips the `[tmp-hygiene-trace]` diagnostic for that file (the file's own afterAll still runs). Fix: register the mem block above the tmp-hygiene `afterAll` so the throw lands last.

Nits from the same review: the docstring and changelog say "812 MB and 1,800 MB" while every measurement says 815–828 / 1,796–1,803; `tests/config/worker-peak-rss-budget.test.ts:85` says the admission is content-keyed but it is path-keyed. The live-admission audit case is a schema validator over an empty map today (both predicates invertible while green) — recorded so a later verify does not mistake it for a live guard.

*This issue was written by an AI agent working on this repository, with maintainer supervision.*

apmantza:
## Summary

Two hardening items from the #3062 review, deferred to land the #3058 memory
footprint fix quickly:

1. `tests/support/worker-peak-rss.ts:54` — `PEAK_RSS_ADMISSIONS` was typed
   `Readonly<>` but never `Object.freeze`d. `peakRssProblem`'s default
   parameter reads this exact live object, so a test running inside its own
   fork could do `PEAK_RSS_ADMISSIONS["self"] = { peakRssMb: 999999, reason:
   "x" }` and silently raise its own ceiling — the mutation compiled and ran
   clean. Fixed with `Object.freeze({})`; the mutation now throws a
   `TypeError` (every module here runs in ES module strict mode).
2. `tests/support/vitest-setup.ts` — the mem-report `afterAll` was registered
   LAST among the file's four top-level `afterAll` hooks. Vitest runs
   `afterAll` LIFO (verified empirically in this PR: three plain `afterAll()`
   calls registered 1, 2, 3 in a real vitest run report completion in the
   order 3, 2, 1), so being registered last made it run FIRST — its
   `resourceUsage().maxRSS` reading (a running high-water mark) was taken
   before the `#3083` root-backstop, tmp-hygiene, and kill-guard hooks ran,
   so it could never reflect memory any of them allocated during their own
   teardown, and its own throw for an over-budget file preempted the
   tmp-hygiene hook's `[tmp-hygiene-trace]` diagnostic for that same file.
   Fixed by moving its registration above the other three, so it is now the
   LAST hook to execute.

Both acceptance criteria from #3067 are met.

Closes #3067

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:tests

## Checklist

- [x] `npm run lint` passes
- [ ] `npm run build:dist` succeeds if I changed code under `clients/`, `commands/`, `tools/`, or `index.ts` — N/A, no files under those paths changed (ran it anyway to unblock an unrelated dist-dependent test file; see Tests)
- [x] `package-lock.json` is in sync with `package.json` — untouched by this PR
- [ ] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there — N/A, no repo-wide convention changed; this is a two-line hardening of an existing gate's own internals
- [x] `.changelog/<branch-or-slug>-<short-desc>.md` has one valid entry **in this PR** for any user-facing change — N/A, internal test-support hardening with no user-facing effect; no fragment added (see below)

### Changelog

None. This changes only `tests/support/` gate internals (an admission table's
runtime mutability and one `afterAll`'s registration order) — no user-facing
behavior, CLI surface, or documented convention changes.

## Tests

- EDIT `tests/config/worker-peak-rss-budget.test.ts` — added two guards to
  the existing `#3058 per-file peak RSS is registered or fails` describe
  block:
  - `"freezes the admissions table so a test cannot admit itself at runtime
    (#3067)"` — pins that `PEAK_RSS_ADMISSIONS` is frozen and that a runtime
    mutation throws `TypeError`. Regression proof for #3067 item 1.
  - `"registers the mem-report afterAll to run last (#3067, #3062 review
    L2)"` (a nested `describe` with an `it` plus an `it.each` of 3 cases) —
    pins that the mem-report `afterAll`'s registration (the `reportPeakRss(`
    call site) is textually BEFORE the `#3083` root-backstop, tmp-hygiene,
    and kill-guard `afterAll` registrations in `tests/support/vitest-setup.ts`.
    Regression proof for #3067 item 2.

No test in the file was made redundant; every existing case there pins a
different behavior of `peakRssProblem`/`reportPeakRss` (budget math, record
shape, admission validity), none of which overlaps with "is the table frozen"
or "is the hook registered first."

AGENTS.md screen: both new tests are governance-shaped guards over the gate's
own internals — closest to "wrong-layer pin" territory, avoided by pinning
directly at the seam that changed (`Object.isFrozen`/mutation-throws for
item 1; the real, exported call-site text for item 2) rather than by
re-describing the fix in prose.

### Why guard 2 is a structural check, not a full runtime differential

The acceptance criteria asked for "an afterAll registered after the mem one
that allocates past the budget is missed pre-fix and caught post-fix." I
could not build that as a full runtime probe without one of two costs I
judged disproportionate to a two-line reordering fix, and I want that
judgment call visible to review rather than silent:

- Vitest registers a setupFile's `afterAll` hooks before the test file's own
  code runs, so ANY hook a test file registers itself is always registered
  later — and therefore, under LIFO, always executes BEFORE every setupFile
  hook, regardless of how `vitest-setup.ts` orders its own four hooks. A
  test-file-local "other hook that allocates" can never land AFTER the
  mem hook from the outside; the only hooks that can are the other three
  hooks already inside `vitest-setup.ts` itself (backstop, tmp-hygiene,
  kill-guard), none of which allocates meaningfully.
- Making one of those three actually cross the real 2,048 MB
  `WORKER_PEAK_RSS_BUDGET_MB` budget for the test would mean either adding
  synthetic test-only allocation to production teardown code (out of the
  scope agreed for this issue — see "Avoid" in the brief), or genuinely
  allocating ~2 GB inside a test, which risks OOM-killing the very CI runner
  this gate exists to protect.

Since runtime LIFO order is computed directly from registration order (the
empirical proof above), and none of the four hooks' bodies depend on
anything defined between them, checking registration order is a direct,
faithful proxy for the runtime order it determines — not a stand-in for
behavior we could otherwise observe cheaply. I built the empirical LIFO proof
as a real, throwaway `npx vitest run` probe (see the Tests transcript below)
rather than asserting it from memory.

### Red-first (guards, together — pre-fix `tests/support/worker-peak-rss.ts` and `tests/support/vitest-setup.ts`, via `git checkout HEAD~1 -- <files>`)

```
$ npx vitest run tests/config/worker-peak-rss-budget.test.ts
 ❯ |default| tests/config/worker-peak-rss-budget.test.ts (11 tests | 4 failed) 16ms
   ❯ #3058 per-file peak RSS is registered or fails (11)
     × freezes the admissions table so a test cannot admit itself at runtime (#3067) 6ms
     ❯ registers the mem-report afterAll to run last (#3067, #3062 review L2) (4)
       × is registered before the #3083 root-backstop afterAll 1ms
       × is registered before the tmp-hygiene afterAll 0ms
       × is registered before the kill-guard afterAll 0ms

 FAIL  ... freezes the admissions table so a test cannot admit itself at runtime (#3067)
AssertionError: expected false to be true // Object.is equality
 ❯ tests/config/worker-peak-rss-budget.test.ts:166:48
   166|   expect(Object.isFrozen(PEAK_RSS_ADMISSIONS)).toBe(true);

 FAIL  ... is registered before the #3083 root-backstop afterAll
AssertionError: the mem-report afterAll must be registered BEFORE the #3083 root-backstop afterAll — afterAll runs LIFO (registered-first runs last), so this is what makes the mem reading include #3083 root-backstop's teardown: expected 26579 to be less than 16223

 FAIL  ... is registered before the tmp-hygiene afterAll
AssertionError: ... expected 26579 to be less than 21028

 FAIL  ... is registered before the kill-guard afterAll
AssertionError: ... expected 26579 to be less than 22366

 Test Files  1 failed (1)
      Tests  4 failed | 7 passed (11)
```

### Post-fix (green)

```
$ npx vitest run tests/config/worker-peak-rss-budget.test.ts
 Test Files  1 passed (1)
      Tests  11 passed (11)
```

### Mutation proof (each guard, isolated — the other file's fix stays in place)

Guard 1 alone unfrozen (`git checkout HEAD~1 -- tests/support/worker-peak-rss.ts`, ordering fix kept):

```
$ npx vitest run tests/config/worker-peak-rss-budget.test.ts
 ❯ ... freezes the admissions table so a test cannot admit itself at runtime (#3067) 4ms
AssertionError: expected false to be true // Object.is equality
 Test Files  1 failed (1)
```

Guard 2 alone reordered back (`git checkout HEAD~1 -- tests/support/vitest-setup.ts`, freeze fix kept):

```
$ npx vitest run tests/config/worker-peak-rss-budget.test.ts
     × is registered before the #3083 root-backstop afterAll 4ms
     × is registered before the tmp-hygiene afterAll 1ms
     × is registered before the kill-guard afterAll 1ms
 Test Files  1 failed (1)
      Tests  3 failed | 8 passed (11)
```

Each mutation reds ONLY its own guard — proof the two guards are independent
and neither is a passenger of the other.

### Real LIFO proof (throwaway, not committed)

```
$ npx vitest run tests/config/__hook_order_probe.test.ts   # 3 plain afterAll(), registered 1,2,3
AFTERALL-3-registered-third
AFTERALL-2-registered-second
AFTERALL-1-registered-first
 Test Files  1 passed (1)
```

### Targeted runs (post-fix, all green)

- `npx vitest run tests/config/worker-peak-rss-budget.test.ts tests/config/worker-budget.test.ts` — 2 files, 38 tests passed.
- `npx vitest run tests/config/` (full directory, 60 files — the gate's own home) — 60 files, 594 passed, 1 skipped, 0 failed.
- Governance/sweep sub-suite, selected mechanically per the fixer contract
  (`ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,governance,silence,hermeticity,invariant,contract}*.test.ts`)
  — 31 files, 564 tests passed.
- Every other file that references the changed symbols/file (`grep -rl
  "worker-peak-rss\|PEAK_RSS_ADMISSIONS\|peakRssProblem\|reportPeakRss\|
  vitest-setup" tests/`), 15 more files not already covered above.
  `tests/scripts/lsp-fixture-workspace.test.ts` first failed with `Cannot
  find module '.../dist/clients/lsp/config.js'` because this fresh worktree
  had never run `npm run build:dist` — this diff does not touch `dist/`,
  `clients/`, or `lsp/config` at all, so I ran `npm run build:dist` once and
  reran the file: 21/21 green. Every other file in the batch was already
  green. `tests/clients/lsp/typescript-classic-repair.integration.test.ts`
  reports 2 skipped (its own live-pi integration gate, not exercised by this
  diff).
- `npx oxfmt --check` on the three touched files: clean.
- `npm run lint`: clean.

### Test assessment

- `tests/config/worker-peak-rss-budget.test.ts` — uniquely pins
  `peakRssProblem`/`reportPeakRss`'s budget math, record shape, and (now)
  `PEAK_RSS_ADMISSIONS`'s runtime immutability and `vitest-setup.ts`'s
  `afterAll` registration order. No case in this file is made redundant by
  the two added; each of the two new cases pins a distinct property (frozen
  vs. not; registration order vs. threshold/ceiling math) that no existing
  case in the file touches. No removal candidates.

## Blast radius

`tests/support/worker-peak-rss.ts` and `tests/support/vitest-setup.ts` are
test infrastructure, not a `clients/` production module, so `module_report`'s
`blastRadius` does not apply. Callers found by grep:

- `PEAK_RSS_ADMISSIONS` / `peakRssProblem` / `reportPeakRss`: imported by
  `tests/support/vitest-setup.ts` (production call site, unchanged signature)
  and `tests/config/worker-peak-rss-budget.test.ts` (test, extended). No
  other importer in the tree.
- `tests/support/vitest-setup.ts` is `sharedSetupFiles` in `vitest.config.ts`
  — every test file in the suite loads it. The change is a pure relocation
  (same four hooks, same bodies, only registration ORDER changed) plus one
  new comment; no hook's condition, body, or export changed. Verification
  plan: the full `tests/config/` directory (which includes every test that
  exercises `tmp-fixture-hygiene`, `pi-lens-home-hermeticity`, and the
  backstop/kill-guard paths this reorder touches) is green (see Tests).

## Observability

## Class sweep

Defect shape: an exported `Readonly<Record<...>>` "admission" or "exemption"
table typed read-only but never `Object.freeze`d, where a DEFAULT PARAMETER
in a function reads the live module singleton, AND that function is invoked
by a hook `tests/support/vitest-setup.ts` registers globally (so any
co-resident test — not just the table's own governance test — runs in a
fork where the shared object is live and a mutation could escape into a
DIFFERENT file's enforcement).

Swept the whole tree: `grep -rn "Readonly<Record<string" clients/ tools/ mcp/
scripts/ scripts/lib/ tests/ index.ts` (excluding compiled `.js`), 60+ hits.
Grouped by risk:

- **Already frozen, same idiom**: `clients/language-registry.ts`'s
  `EXTENSION_TO_LSP_ID` / `EXTENSION_TO_GRAMMAR` (`Object.freeze(...)`
  directly); `clients/tree-sitter-shared.ts`'s `EXT_TO_LANG`,
  `clients/lsp/language.ts`'s `LANGUAGE_EXTENSIONS`, and
  `clients/project-diagnostics/scanner.ts`'s `TREE_SITTER_EXT_TO_LANG` are
  re-exported aliases of those same frozen objects, so frozen transitively.
  `scripts/prune-agent-worktrees.mjs`'s `HOOK_TIMEOUT_MS` /
  `HOOK_REMOVE_RESERVE_MS` / `HOOK_POLICIES` are each `Object.freeze(...)`.
- **Different risk shape (not a sibling)**: every `Readonly<Record>`
  "admission"/"exemption" table under `tests/config/*.test.ts` and
  `tests/clients/*.test.ts` (e.g. `FINITE_REASONS`, `LATENCY_ADMITTED`,
  `ADMITTED`, `EXEMPT_RAW_WRITE_FILES`, `EXEMPT_SESSION_STATE_FILES`,
  `SESSION_STATE_SYMBOL_COUNTS`, `GOVERNED_PROBES_BY_MODULE`,
  `CONTAINER_CLASS_EXCLUSIONS`, and others) is read ONLY by that same file's
  own governance assertions — none is imported by `tests/support/vitest-setup.ts`
  or any other globally-registered hook. Mutating one at runtime can only
  corrupt that file's own test data; it cannot make a DIFFERENT, unrelated
  file's production enforcement under-fire the way `PEAK_RSS_ADMISSIONS`
  could. Confirmed by grepping each export's importers.
- **Module-private, static schema data**: `clients/config-schema.ts`'s
  `LSP_KEY_TYPES` / `LSP_KEY_DENY` are compile-time-constant lookup tables
  with no "empty in steady state, admitted by a human PR" shape and no
  test-mutation vector — a different kind of table entirely, out of scope.

Also swept `resourceUsage`/`reportPeakRss` across the tree
(`clients/safe-spawn.ts`, `clients/memory-sampler.ts` also call
`resourceUsage()`, but for production child-process/self monitoring, not
test-teardown ordering — unrelated mechanism) and the four
`tests/support/vitest-setup.ts` top-level `afterAll` hooks themselves (only
the mem one reads a monotonic running-max that composes with what runs
before it; the other three — backstop, tmp-hygiene, kill-guard — are
independent invariant checks with no such ordering dependency on each
other).

**Verdict**: class of size 1 for both parts of this defect. No sibling to
fix in this PR; no consolidation needed (nothing to fold — the pattern that
IS shared, `Object.freeze` on an admission table, is already the established
idiom elsewhere in the tree, which is exactly what this PR now also does).


apmantza:
Merged as reviewed. Trade-off recorded here, not disclosed in the body: Vitest aborts the remaining afterAll hooks on a throw, so the old order let an over-budget file skip the backstop, tmp-hygiene and kill-guard checks entirely (the review reproduced this — worse than the body's claim), and the new order lets any of those three throwing suppress the `[mem-file]` record instead. Net positive for correctness; the record's independence from hook order is tracked in 3139.

Issue #3098, opened 2026-09-16
normalizeFilePath: POSIX arm is case-preserving, so on a case-insensitive filesystem (macOS APFS) a mis-cased write and a normalized read orphan the same disposition mark (refs #3090, #1024)


apmantza:
## What

Split out of #3090 (external report). `normalizeFilePath` (`clients/path-utils.ts:126`) canonicalizes casing only on its win32 branch (`realpathSync.native`, then resolve-and-lowercase for non-existent paths). On POSIX it is case-preserving by design. On a case-insensitive POSIX filesystem (macOS default APFS, some Linux mounts) `SUB/a.ts` and `sub/a.ts` are one file, so a raw mis-cased mark write (`lens_diagnostic_mark` through `path.resolve(cwd, arg)`) and the dispatcher's `normalizeMapKey` read derive two anchors, and the agent's own mark silently never applies. That is the #1024 defect, still live on macOS.

The #1024 regression test proves it: on macOS its skip probe (`fs.existsSync(rawFile)`) passes and the assertion fails with the mark unfound. On Linux CI the case skips (case-sensitive `/tmp`), and there is no macOS unit leg, so it has never been observed in CI.

## Acceptance

1. On POSIX, when the path exists, canonicalize casing through `realpathSync.native` (which returns on-disk casing on APFS) before deriving the key; keep the non-existent-path arm case-preserving and say why (a Linux case-sensitive FS must not fold `SUB` into `sub`).
2. The #1024 test passes unskipped on a case-insensitive filesystem and still skips visibly on a case-sensitive one; no new `ctx.skip` that hides the defect.
3. Red-first transcript from a case-insensitive filesystem (a maintainer macOS box, or a case-insensitive loopback mount on Linux, e.g. a vfat image) in the PR body; mutation: revert the POSIX realpath arm and quote the red.
4. Sweep every other `normalizeMapKey`/`normalizeFilePath` consumer that keys durable records (dispositions, workspace cache, widget reconcile) for the same two-anchor shape and record what was grepped.

## Non-goals

Lowercasing POSIX paths; any change to the Windows arm.

apmantza:
PR #3159 implements the production half; it references this issue rather than closing
it, because one acceptance criterion is not met on the machine that built it.

**Met:** (1) POSIX canonicalizes casing through `realpathSync.native` for an existing
path — but only for the TRAILING segments, halting at the first segment that differs by
[...]

Issue #3090, opened 2026-09-16
The #1024 cross-case anchor test fails on POSIX case-insensitive filesystems — its skip probes filesystem aliasing, not anchor collapse


georgeharker:
## What happened

The #1024 regression test ("finds a false-positive mark written under a raw mis-cased path form when applied via the normalizeMapKey form") fails on any POSIX machine with a case-insensitive filesystem (macOS default APFS), while passing Linux CI (skips: case-sensitive /tmp) and Windows (the collapse happens).

## Steps to reproduce

On macOS: `npx vitest run tests/clients/diagnostic-dispositions.test.ts` — the #1024 case fails with the mark unfound (`kept === [diag]`). Verified identical on master @ `20896a56b`.

## Root cause

`normalizeFilePath` case-folds only on its win32 branch (`realpathSync.native` / `resolveNonExisting`); on POSIX it is case-preserving by design. The test's skip guard probes filesystem aliasing (`fs.existsSync(rawFile)`), but the collapse it demands additionally requires platform-level case folding — so a case-insensitive FS passes the skip probe and then fails the assertion. The write anchors under `SUB/a.ts`; the normalized read derives `sub/a.ts`; two different anchors for one file.

## Expected behavior

The skip should probe what the test actually depends on — anchor identity under both path forms — and skip visibly with a reason naming the prerequisite ("posix path normalization is case-preserving: the cross-case anchor collapse this test pins is Windows-only").

Open question for maintainers (out of scope for the test fix): whether POSIX case-insensitive filesystems SHOULD fold cross-case anchors (a mis-cased write and a normalized read silently no-op each other's marks today) — that would be a production change in `normalizeFilePath`'s POSIX arm.

## Offer

Test-only fix ready on a branch (the skip probes `anchorsForDiagnostic` under both forms). No changelog entry needed (test-only).

apmantza:
Confirmed from the code. The skip guard at `tests/clients/diagnostic-dispositions.test.ts` probes `fs.existsSync(rawFile)` (filesystem aliasing), and `normalizeFilePath`'s POSIX arm is case-preserving, so on APFS the two anchors differ and the mark is unfound. Your root cause is right.

[...]

Issue #3154, opened 2026-09-16
normalizeFilePath: the POSIX on-disk-casing arm is a no-op on Linux ext4/tmpfs casefold directories (realpath reports the queried casing)


apmantza:
## What

Split out of #3098 by measurement. `normalizeFilePath`'s new POSIX arm canonicalizes
casing through `realpathSync.native`, which closes the #1024 two-anchor defect on
macOS (APFS/HFS+) and on `nocase` vfat/ntfs3/cifs mounts — but it is a **no-op on a
Linux ext4/tmpfs casefold directory** (`chattr +F`, the Android/Wine/Proton
configuration). Those directories alias the two spellings, so the #1024 defect is
live there, but `realpath(3)` returns the spelling the caller asked with, so nothing
folds.

## Measured (2026-09-16, Linux 7.0.0-31-generic, tmpfs `casefold=utf8-12.1.0` + `chattr +F`)

```
$ podman unshare unshare -m bash -c '... mount -t tmpfs -o casefold=utf8-12.1.0 tmpfs $M
    mkdir $M/ci; chattr +F $M/ci; mkdir -p $M/ci/sub; echo x > $M/ci/sub/a.ts; ...'
chattr rc=0
-----------------F---- /…/cimnt/ci
exists SUB/a.ts: true
realpathSync.native:  /…/cimnt/ci/SUB/a.ts      <-- the QUERIED casing, not the on-disk one
realpathSync       :  /…/cimnt/ci/SUB/a.ts
proc/self/fd       :  /…/cimnt/ci/sub/a.ts      <-- the on-disk casing
readdir ci: [ 'sub' ]
```

Cause: ext4/tmpfs casefold uses a casefolded dentry hash and keeps the queried name in
the dentry, so the kernel's path reconstruction (`realpath`, `readlink -f`, coreutils
`realpath`) reports the spelling the caller used. `vfat`/`ntfs3`/`cifs` instead use
`d_add_ci()`, which installs the on-disk name — which is why `realpath` canonicalizes
there and not here. Darwin's `realpath(3)` rebuilds each component from
`getattrlist(ATTR_CMN_NAME)` (Libc-1669.0.4 `stdlib/FreeBSD/realpath.c:233,348-354`),
hence the macOS behaviour #3098 relies on.

## Why this was left open in #3098

The only mechanism measured to report on-disk casing on a casefolded Linux directory is
[...]

Issue #1655, opened 2026-08-18
pi event-contract hardening: unguarded tool_call throws block the user's tool; vacuous telemetry fields; mutable input; cwd drift; unicode path variants


apmantza:
Bundle from a source-level audit of the pi host (maintainer-supervised AI), five contract facts pi-lens currently mishandles — one hardening PR:

1. **A throwing tool_call handler blocks the user's tool.** pi's emitToolCall is the ONLY emit path with no per-handler try/catch (pi runner.ts:932-953; compare emit :801-832 and emitToolResult :877-930) — pi rethrows as "Extension failed, blocking execution". pi-lens's tool_call entry point needs a total internal try/catch (degrade + recordDegradationOnce, never throw). tool_result is already host-protected.
2. **Seven tool_result fields we type are never set by pi** (pi builds exactly type/toolName/toolCallId/input/content/details/isError/usage at agent-session.ts:502-516): clients/runtime-tool-result.ts:53-62 declares callId/requestId/provider/model/sessionId/session and :708-714 gates setTelemetryIdentity on them — unreachable against real pi (shape 7 + shape 5's spread axis). Delete the branch or source identity from ctx/runtime as :802-807 already does.
3. **tool_call.input is mutable and extension-ordered** (pi types.ts:914-919); edit's prepareArguments rewrites args pre-event. Never treat call-time input as authoritative for the result; read the result's own input (it carries executed args). (#1648 is applying this; this item is the sweep for OTHER call-time-input consumers.)
4. **Tool cwd is frozen at construction; ctx.cwd can drift; bash has no cwd param** (pi tools/index.ts:99-124, tools/bash.ts:41-44) — resolve tool paths against the tool's cwd basis; keep bash-file-access as the only bash-side path source; sweep for resolvers using ctx.cwd where the tool basis differs.
5. **pi's read resolver silently retries unicode/spacing variants** (path-utils.ts:52-83: narrow-NBSP, NFD, U+2019, NFD+curly) — the file pi read can differ from resolve(cwd, input.path); mirror the fallback ladder in our tool-event path resolver or record a path_variant_unresolved degradation (shape 1 with an upstream cause).

## Acceptance criteria
Each item: the fix, a red-first test (host-faithful doubles — pi's literal event shapes, not our imagined ones), and for 2 specifically the deletion proven by a host-shape fixture. Cite the pi file:lines in code comments so the contract is auditable against upstream.

Refs #1642, #1648, #1652.

[...]

PR #3159, opened 2026-09-16
fix(path-utils): canonicalise on-disk casing on POSIX so a mis-cased mark and a normalised read share one anchor (refs #3098, refs #3090)

apmantza:
## Summary

`normalizeFilePath` (`clients/path-utils.ts:200`) canonicalized casing only on its
win32 branch. On a case-insensitive POSIX filesystem — macOS's default APFS, a
`nocase` vfat/ntfs3/cifs mount — `SUB/a.ts` and `sub/a.ts` are ONE file, so the raw
mis-cased anchor `lens_diagnostic_mark` writes (`path.resolve(cwd, arg)`) and the
`normalizeMapKey` anchor the dispatcher reads derived two keys for that one file and
the agent's own false-positive/flagged mark silently never applied. That is #1024,
still live on every macOS install; #3090 is the same defect reported from the test
side.

The POSIX arm now adopts the on-disk casing `realpath(3)` reports, for the TRAILING
segments of an existing path, halting at the first segment that differs by more than
case (`adoptCanonicalCasing`, `clients/path-utils.ts:137`). That splits the two
things `realpath` answers at once:

- casing is adopted — the two anchors collapse to one;
- a symlinked PREFIX is still never resolved, so a symlinked monorepo package keeps
  keying under the path the caller held (refs #2490, where a fold for its own sake
  broke every monorepo).

The prefix half is not decoration: macOS's `os.tmpdir()` is `/var/folders/...`, a
symlink into `/private/var`, so the simpler "adopt when the whole string is a case
variant" rule would decline to canonicalize exactly the paths #1024's regression test
(and every macOS temp-dir workflow) runs on. Mutation M6 below is that simpler rule;
it reds one test and nothing else.

A path that does NOT exist stays case-preserving: on a case-sensitive filesystem
`SUB/a.ts` and `sub/a.ts` are two different files, and only the filesystem's answer
for a path that exists can tell them apart.

**Refs #3098, refs #3090 — not `closes`, for one honest reason.** Acceptance (3)
asks for a red-first transcript from a REAL case-insensitive filesystem. This host is
Linux; the two case-insensitive filesystems reachable here behave differently from
macOS, measured below, and neither can produce that transcript. What is proven here
instead: the identical kernel contract (two spellings, one file, `realpath` reporting
the on-disk one) built with a case-variant symlink, red-first through the production
disposition path, plus the upstream Darwin source that pins what APFS answers.
@georgeharker — you reported #3090 from a macOS box: would you run
`npx vitest run tests/clients/diagnostic-dispositions.test.ts tests/clients/path-utils.test.ts`
on this branch? On APFS those cases must run UNSKIPPED and green. The maintainer
decides whether that run is a merge precondition.

### Why the mechanism is `realpath`, from upstream source (not paraphrase)

Darwin's `realpath(3)` rebuilds every component from the filesystem's own
`ATTR_CMN_NAME`, which is why it returns on-disk casing on APFS/HFS+ —
apple-oss-distributions/Libc @ `Libc-1669.0.4`, `stdlib/FreeBSD/realpath.c`:

```c
233:  if (getattrlist(resolved, (void *)&_rp_alist, &attrs, sizeof(attrs), FSOPT_NOFOLLOW) == 0) {
348:  } else if (useattrs) {
349:      /*
350:       * attrs already has the real name.
351:       */
354:      resolved_len = strlcat(resolved, (const char *)&attrs.name + attrs.name.attr_dataoffset, PATH_MAX);
```

(`_rp_alist` at `:59-67` requests `ATTR_CMN_NAME`.) Node's `fs.realpathSync.native`
is that `realpath(3)`.

### MEASURED counter-example: Linux casefold directories are NOT covered

Built a real case-insensitive directory on this Linux host (tmpfs
`casefold=utf8-12.1.0` + `chattr +F`, unprivileged via `podman unshare unshare -m`;
`/dev/nvme0n1p5` itself has no casefold feature, so `chattr +F` there returns
`Operation not supported`):

```
mount rc=0
6625 274 0:92 / /…/cimnt rw,relatime - tmpfs tmpfs rw,size=65536k,uid=1000,gid=1000,inode64
chattr rc=0
-----------------F---- /…/cimnt/ci
/…/cimnt/ci/SUB
/…/cimnt/ci/SUB/a.ts
exists SUB/a.ts: true
realpath.native: /…/cimnt/ci/SUB/a.ts        <-- the QUERIED casing
realpathSync   : /…/cimnt/ci/SUB/a.ts
proc/self/fd   : /…/cimnt/ci/sub/a.ts        <-- the on-disk casing
readdir ci: [ 'sub' ]
```

ext4/tmpfs casefold keeps the queried name in the dentry, so `realpath` echoes it
back; vfat/ntfs3/cifs use `d_add_ci()` and install the on-disk name, which is why
`realpath` canonicalizes there. So this fix is a **no-op on a Linux casefold
directory** — filed with this transcript as **#3154** (p3), and the two case cases
skip visibly there naming it. That gap is why a real-FS red→green transcript could
not be produced on this host: the only case-insensitive filesystem available is the
one flavour the mechanism does not reach.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [ ] area:lsp
- [ ] area:dispatch
- [ ] area:installer
- [x] area:diagnostics
- [ ] area:read-guard
- [ ] area:perf
- [ ] area:observability
- [ ] area:session
- [ ] area:config
- [ ] area:security
- [x] area:tests

## Checklist

- [x] `npm run lint` passes (ran as the pre-commit hook on this head: `tsc --project tsconfig.json && oxlint …`, clean)
- [ ] `npm run build:dist` succeeds if I changed code under `clients/` — not run locally (bundling is CI's `build:dist` job on this head)
- [x] `package-lock.json` is in sync with `package.json` (`check-lockfile-sync.mjs` ran on the pre-commit hook: "in sync ✓")
- [x] `AGENTS.md` is updated if this PR changes behavior, commands, conventions, or invariants documented there — no AGENTS.md invariant covers the POSIX arm; the contract lives in the `normalizeFilePath` docstring and is updated there
- [x] `.changelog/3098-posix-case-insensitive-anchors.md` has one valid entry (`node scripts/check-changelog-fragments.mjs` → "changelog fragments OK (100 entries)")
- [x] Commit subject includes the issue number (`Refs #3098`, `Refs #3090` trailers; title carries the refs)

## Tests

`npm run build` before every run. `TMPDIR` pinned outside the worktree,
`PI_LENS_HOME` inside it, for every command quoted here.

**NEW — `tests/clients/path-utils.test.ts`, describe `normalizeFilePath: POSIX adopts on-disk casing (#3098, the live half of #1024)`** (6 cases):

| `it(` title | pins |
| --- | --- |
| `a mis-cased spelling and the on-disk spelling derive ONE map key` | the defect itself, at the normalizer |
| `adopts the on-disk casing itself — never a lowercased or uppercased form` | the #3098 non-goal ("no lowercasing POSIX paths"); fixture dir is `subDir`, so a lowercasing fix reds |
| `keeps a symlinked PREFIX — POSIX normalization still resolves no symlink` | the scope bound (refs #2490) |
| `a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)` | the macOS tmpdir shape; the ONLY red under M6 |
| `a path that does not exist stays case-preserving` | the case-sensitive-FS invariant on the absent arm |
| `two genuinely distinct files on a case-sensitive filesystem keep two keys` | the same invariant on the EXISTING arm |

**EDITED — `tests/clients/diagnostic-dispositions.test.ts`**, the #1024 case
`finds a false-positive mark written under a raw (mis-cased) path form when applied via the normalizeMapKey form`:
it built its own `sub/` + mis-cased `SUB/` pair and skipped on any case-sensitive
filesystem, i.e. on the only lane CI runs. It now takes the pair from
`createCaseAliasFixture`, so on the ubuntu lane the alias is a case-variant symlink
supplying the byte-identical kernel answers APFS gives, and the case RUNS. Its skip is
now reached only where the kernel cannot supply that contract at all (the casefold
flavour, #3154).

**NEW — `tests/clients/test-utils.ts` `createCaseAliasFixture`**: one fixture, both
files (single source of truth rather than a copy per suite).

Screens (AGENTS.md's ten): *invisible skip* — the one remaining skip is
`ctx.skip(reason)`, visible, and its condition is a FILESYSTEM property
(`realpathSync.native(raw) !== raw`), never the function under test, so reverting the
fix reds instead of skipping (proven by M1/M5 below). *Ambient-inspection double* /
*all-mocks* — no mock anywhere: real temp dirs, real symlinks, real `realpath`.
*Wrong-layer pin* — the #1024 case drives `markDisposition` → `applyDispositions`, the
production write/read pair, not the normalizer.

### Red-first (tests committed, then `git checkout <base> -- clients/path-utils.ts && npm run build`; the base was `9ffbc2f06`, rebased onto `d8f1f57bb` in round 2 — the transcripts below are from the original run and re-verified on the round-2 head)

Production path first — `npx vitest run tests/clients/diagnostic-dispositions.test.ts`:

```
 FAIL  |default| tests/clients/diagnostic-dispositions.test.ts > anchor path-form stability (#1024 — write raw vs read normalized) > finds a false-positive mark written under a raw (mis-cased) path form when applied via the normalizeMapKey form
AssertionError: expected [ { tool: 'eslint', …(3) } ] to deeply equal []

- Expected
+ Received

- []
+ [
+   {
+     "line": 1,
+     "message": "bad call",
+     "rule": "no-bad",
+     "tool": "eslint",
+   },
+ ]

 ❯ tests/clients/diagnostic-dispositions.test.ts:790:16
 Test Files  1 failed (1)
      Tests  1 failed | 42 passed (43)
```

That is the macOS symptom verbatim (`kept === [diag]`): the agent's own mark unfound.
Same revert, `npx vitest run tests/clients/path-utils.test.ts tests/clients/diagnostic-dispositions.test.ts`:

```
 FAIL  … > a mis-cased spelling and the on-disk spelling derive ONE map key
Expected: "/…/pi-lens-case-KxTvXv/sub/a.ts"
Received: "/…/pi-lens-case-KxTvXv/SUB/a.ts"

 FAIL  … > adopts the on-disk casing itself — never a lowercased or uppercased form
AssertionError: expected false to be true

 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
Expected: "/…/pi-lens-case-AGT3DB/link/sub/a.ts"
Received: "/…/pi-lens-case-AGT3DB/link/SUB/a.ts"

 Test Files  2 failed (2)
      Tests  4 failed | 189 passed | 1 skipped (194)
```

### Mutations (neutered in the BUILT `clients/path-utils.js`, rebuilt after each)

| # | mutation | result |
| --- | --- | --- |
| M1 | revert the whole POSIX arm to `return normalized;` | 4 red (the transcript above) |
| M2 | delete `if (a.toLowerCase() !== b.toLowerCase()) break;` | 2 red |
| M3 | absent arm → `return normalized.toLowerCase();` | 1 red |
| M4 | `return canonical;` (adopt the full realpath — the naive fix) | 2 red |
| M5 | inverse: `if (true) break;` (never adopt) | 4 red |
| M6 | whole-string rule: `return held.toLowerCase() === canonical.toLowerCase() ? canonical : held;` | 1 red — and it is the macOS-tmpdir case, alone |

```
=== M2: delete the structural-divergence break
 FAIL  … > keeps a symlinked PREFIX — POSIX normalization still resolves no symlink
AssertionError: expected '/…/pi-lens-case-VVR9G1/link/sub/a.ts' to be …
Expected: "/…/pi-lens-case-VVR9G1/link/sub/a.ts"
Received: "/…/pi-lens-case-VVR9G1/real/sub/a.ts"
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  2 failed | 191 passed | 1 skipped (194)

=== M3: lowercase the non-existent arm (the #3098 non-goal)
 FAIL  … > a path that does not exist stays case-preserving
Expected: "/…/pi-lens-case-6wLk3T/NOPE/b.ts"
Received: "/…/pi-lens-case-6wlk3t/nope/b.ts"
      Tests  1 failed | 192 passed | 1 skipped (194)

=== M4: adopt the FULL realpath (resolve symlinks too)
 FAIL  … > keeps a symlinked PREFIX — POSIX normalization still resolves no symlink
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  2 failed | 191 passed | 1 skipped (194)

=== M5: inverse guard — if (true) break (never adopt on-disk casing)
 FAIL  … diagnostic-dispositions … > finds a false-positive mark written under a raw (mis-cased) path form …
 FAIL  … > a mis-cased spelling and the on-disk spelling derive ONE map key
 FAIL  … > adopts the on-disk casing itself — never a lowercased or uppercased form
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  4 failed | 189 passed | 1 skipped (194)

=== M6: whole-string case-variant adoption (no tail alignment)
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  1 failed | 192 passed | 1 skipped (194)
```

**Two branches in the diff are NOT guards and are declared as such**, with a
measurement instead of a mutation, because both arms return equal strings:
`canonical === normalized ? normalized : adopt(...)` and `changed ? join : held`.
Deleting the first measured 1.9 → 2.3 µs/call (`bench.mjs`, same corpus, two runs);
it is the same identity-preserving pattern `normalizeEphemeralMapKey`
(`clients/path-utils.ts:431`) already documents.

**Not mutation-provable on this lane, stated rather than claimed**: the
`skipReason` arm of `createCaseAliasFixture` (the casefold flavour) cannot red on
ubuntu — CI has no casefold directory. It was verified directly by the
`podman unshare` transcript quoted above, not by a test.

### Suites run on this head (all green unless noted)

| set | files | result |
| --- | --- | --- |
| brief's list: `path-utils*`, `diagnostic-dispositions*`, `lsp/workspace-diagnostics*`, `widget*` (`normalize*` matches nothing) | 26 | `Test Files 26 passed (26) / Tests 421 passed \| 1 skipped (422)` |
| every governance sweep by the mechanical glob (`*{sweep,ratchet,conformance,coverage,gate,silence,hermeticity,contract}*`; `*governance*` and `*invariant*` match nothing) | 31 | `Test Files 31 passed (31) / Tests 564 passed (564)` |
| every `tests/config/*.test.ts` | 61 | `Test Files 61 passed (61) / Tests 604 passed \| 1 skipped (605)` |
| every test file referencing `normalizeFilePath`/`normalizeMapKey` | 104 | `Test Files 103 passed \| 1 skipped (104) / Tests 1936 passed \| 9 skipped (1945)` |
| every OTHER test file touching a real filesystem (`mkdtempSync\|setupTestEnvironment\|realpath`) — run because this seam is repo-wide | 537 | `Tests 11 failed \| 8277 passed \| 45 skipped (8333)` — see below |
| post-rebase re-run of `path-utils` + `dispositions` + the 31 sweeps | 33 | `Test Files 33 passed (33) / Tests 758 passed \| 1 skipped (759)` |
| post-rebase `tests/config/` | 61 | `Tests 621 passed \| 1 skipped (622)` |

The 11 failures in the 537-file set are **red on `origin/master` too**, proven twice
rather than asserted. (a) A clean worktree at `origin/master` (`a0a97c742`, the base of
this branch), same box, same `TMPDIR`, `npm run build` then
`npx vitest run` on those 5 files:

```
$ git worktree add …/fix-3098-base --detach origin/master   # HEAD is now at a0a97c742
      1  FAIL  |default| tests/clients/formatter-template-preserve.test.ts
      5  FAIL  |default| tests/clients/runtime-turn-finding-freshness.test.ts
      2  FAIL  |default| tests/clients/runtime-turn-secrets-disposition.test.ts
      2  FAIL  |default| tests/clients/runtime-turn-session.test.ts
      1  FAIL  |default| tests/clients/skills-resolver.test.ts
      1  Test Files  5 failed (5)
      1       Tests  11 failed | 88 passed (99)
```

(b) The same 5 files in THIS worktree with only my two source files reverted
(`git checkout 9ffbc2f06 -- clients/path-utils.ts clients/dispatch/types.ts`, rebuild):
`Tests  11 failed | 88 passed (99)` — identical count, identical files. They read as
missing `gitleaks`/`prettier` binaries on this box. `tests/config/tmp-fixture-hygiene.test.ts`
also red in that 537-file batch: it is the serialized owner that reds on temp entries
other suites leak, and the leaked entries were `pi-lens-lsp-cache-*` /
`pi-lens-module-call-graph-*` from suites pulled out of their designed lane. It passes
in its own lane (the 61-file `tests/config` run) and passes when run together with
both files this PR touches.

Formatting: `npx oxfmt --check` (pinned 0.67.0) on all 5 touched files —
"All matched files use the correct format."

### Test assessment

- `tests/clients/path-utils.test.ts` — uniquely pins the normalizer's key derivation
  per path SHAPE and per platform. The new block is the first coverage of its POSIX
  existing-path arm. Nothing in the file becomes redundant: the neighbouring
  `#1150` block covers the win32-shaped branch, which this PR does not touch, and its
  three cases stay green under M1–M6 (they run on a non-existent `C:/…` path).
- `tests/clients/diagnostic-dispositions.test.ts` — uniquely pins anchor derivation
  through the real mark/apply pair. The #1024 case is EDITED, not duplicated: the
  fixture moved out, the assertion is unchanged. No removal candidates; every case
  I mutated in this file (M1, M5) reds exactly the intended one.
- `tests/clients/test-utils.ts` — shared fixture module, no assertions of its own.
- No vacuous test found in the touched files: each case in the new block is the sole
  red under at least one mutation above (M6 is the discriminator for the
  macOS-tmpdir case; M3 for the absent arm; M2/M4 for the symlink prefix; M5 for the
  two casing cases). `two genuinely distinct files on a case-sensitive filesystem
  keep two keys` is the one REDUNDANT-but-guarding case — it reds under the same
  mutation as `adopts the on-disk casing itself` (its named survivor) — and is kept
  because it pins the non-goal on the EXISTING arm, which the survivor does not: a
  future fold-by-lowercase in the existing arm would silently merge two real files.

## Blast radius

`module_report` is not runnable from this worktree (no pi host attached) — recorded
as unavailable; dependents below are from grep on this head.

- `clients/path-utils.ts` → `normalizeFilePath` / `normalizeMapKey`: **83 runtime
  files** call them (`clients/`, `tools/`, `mcp/`, `index.ts`); 135 files import the
  module. Callers above: `createDispatchContext` (`clients/dispatch/types.ts:190`
  invariant — the three context fields), `clients/finding-identity.ts:64` `relativeFile`
  (dispositions, actionable-warnings, code-quality), `clients/advisory-provenance.ts:41`
  `advisoryPathKey` (git-guard's durable record), `clients/lsp/workspace-diagnostics-cache.ts:524`
  `cacheKeyFor`, `read-guard.ts`, `instance-registry.ts`, `recent-touches.ts`.
  Callees below: `realpathSync.native` only — one added syscall on POSIX.
- **Behaviour delta on POSIX is empty unless a path names a file whose on-disk
  casing differs from the caller's spelling.** A pure symlink difference changes
  nothing (M2/M4 are the proof in the other direction), and a non-existent path
  changes nothing. On a case-sensitive Linux filesystem that leaves the identity
  function, which is why 8277 tests across 537 real-filesystem files are unmoved.
- **Durable records whose keys can move**: only entries that were written under a
  mis-cased spelling — i.e. exactly the entries that were already orphaned by this
  defect. Correctly-cased keys are byte-identical before and after, so persisted
  disposition state, the git-guard record, the workspace-diagnostics cache and the
  persisted read-set all still parse and still match. Worst case for a mis-cased
  stale entry is a cache miss and a recompute.
- **Hot path, measured** (`bench.mjs`, 263 real `clients/*.ts` paths, 50k iterations,
  two runs, this head):

```
pre-fix POSIX arm (existing)           72-79 ns/call
post-fix normalizeFilePath (existing)   1911-1917 ns/call     (+1.84 µs)
pre-fix POSIX arm (absent)             24-33 ns/call
post-fix normalizeFilePath (absent)    5117-5332 ns/call      (+5.2 µs)
```

  For scale, the same call on Windows has always cost ~200 µs
  (`clients/dispatch/types.ts:195`, now corrected in this diff since its "POSIX
  short-circuits it" sentence is no longer true). The absent-path cost is dominated by
  Node's ENOENT exception: a `statSync(p, {throwIfNoEntry:false})` pre-probe measures
  409 ns absent / 2324 ns existing versus 4724 / 1816 for the try/catch — it trades
  +0.5 µs on existing paths, which the per-edit seams actually hit, for −4.3 µs on
  absent ones, which are mostly fabricated test paths. Not taken; the numbers are
  here so the choice can be revisited with data rather than re-measured. No cache is
  added: `path-utils.ts` has no realpath/mtime cache to reuse, and the hot loops that
  must not pay realpath already use `normalizeEphemeralMapKey`
  (`clients/widget-state.ts:41-49`, `clients/word-index.ts:69`,
  `clients/read-guard.ts:564`) — a seam the codebase built precisely because
  `normalizeMapKey` pays this syscall on Windows.

## Observability

covered by existing record `path-variant-unresolved` at `clients/runtime-tool-call.ts:765`

The only new failure path in this diff is the POSIX `catch` arm, whose sole trigger is
"`realpath` cannot answer for this path" — in practice ENOENT, the normal answer for a
file not yet created, which is why per-call recording would be noise rather than
signal. Where that condition is genuinely a degradation — pi named a path pi-lens
cannot resolve, after the #1655 macOS variant ladder — the tool-call seam above already
records it once per subject via `recordDegradationOnce`, asserted by
`tests/clients/pi-host-contract.test.ts:483` ("records path-variant-unresolved when no
variant matches"), which runs green in the 31-file governance set above. `path-utils`
cannot record for itself in any case: `degradation-ledger.ts → extension-log.ts →
probe-home-state.ts → path-utils.ts` is an existing import cycle, and
`clients/probe-home-state.ts:253` itself calls `realpathSync.native` while resolving the home
the ledger writes to.

## Class sweep

Class: **path-key divergence** — one file reached under two path FORMS, where a write
and a read derive different keys (AGENTS.md's #210 / #1020 / #1024 / #1025 family;
the OS-divergence screen, shape 2). Grepped the whole tree —
`clients/`, `tools/`, `mcp/`, `scripts/`, `index.ts` — for both the symbol names
(`normalizeMapKey(`, `normalizeFilePath(`, `normalizeEphemeralMapKey(`) and the shape
(`path.resolve(cwd`, raw `path.resolve(` used directly as a `Map`/`Set` key, and every
module that persists under `getProjectDataDir(`). 83 runtime files call the
normalizers; the durable-record population is the enumerable family:

| durable record | key derivation | verdict |
| --- | --- | --- |
| dispositions (`diagnostic-dispositions.ts`) | `clients/finding-identity.ts:64` `relativeFile` — `normalizeMapKey` on BOTH cwd and filePath | the reported member; **fixed** by this PR on POSIX |
| git-guard record (`clients/git-guard.ts:914` → `clients/advisory-provenance.ts:41`) | `advisoryPathKey` = `normalizeMapKey(path.resolve(cwd, p))`, one chokepoint, both sides | clean; inherits the fix |
| workspace diagnostics cache (`clients/lsp/workspace-diagnostics-cache.ts:801` (and `:727`, `:652`)) | `cacheKeyFor` (`:524`) = `normalizeMapKey`, every read and write | clean; inherits the fix |
| `fileSeqByPath` (`clients/project-changes.ts:171` write / `clients/code-quality-warnings.ts:254` read) | `normalizeMapKey(path.resolve(...))` on both | clean |
| `modifiedRangesByFile` (`clients/runtime-turn.ts:3278` build / `clients/actionable-warnings.ts:1228` read) | `normalizeMapKey(resolveRunnerPath(...))` build, `normalizeMapKey` read | clean |
| recent touches (`clients/recent-touches.ts:144` write / `:297` cursor) | `normalizeFilePath` both | clean |
| instance registry (`clients/instance-registry.ts:341` (and `:431`, `:767`, `:837`)) | `normalizeFilePath(projectRoot)` everywhere | clean |
| read-guard read-set (`clients/read-guard.ts`, persisted across resume) | primary key `normalizeFilePath`; `knownPathIndex` (`clients/read-guard.ts:580`) deliberately `normalizeEphemeralMapKey` so a post-delete lookup still finds the record | **CORRECTED in round 2 (F3)** — `hasKnownPath`/`forgetPath` are clean (they route through the syntactic index), but they are the ONLY two callers that do. `checkEdit` renders the key into its retryable message (round 2, F2 — fixed here) and `noteCreatedFile`/`recordWritten` straddle the appears/disappears boundary through `this.key()` directly (filed as #3163). The row's original verdict, "unchanged by this PR", was wrong. |
| widget state (`clients/widget-state.ts:51`) / word index (`clients/word-index.ts:69`) | `normalizeEphemeralMapKey` by design — process-local, no filesystem I/O | intentionally NOT folded; case-folding a no-I/O key would merge two real files on a case-sensitive FS |

One member found that this PR does **not** fix, recorded rather than silently
observed: `tools/lens-diagnostic-mark.ts:314` derives `absPath` raw
(`path.resolve(cwd, filePathArg)`) and hands it to `widgetCrossCheck` → `getFileDiagnostics`
(`clients/widget-state.ts:1570`), which keys with the case-preserving ephemeral
normalizer — so on a case-insensitive filesystem a mis-cased `lens_diagnostic_mark`
call still misses the live widget diagnostic and the #802 line re-anchor silently
falls through to its fuzzy fallback. It is a degraded-accuracy path, not a dropped
record, and `tools/` is held by another lane in this session, so it is **filed rather
than edited here as #3160** (p3). The disposition ANCHOR that call
writes is fixed by this PR (it goes through `relativeFile`).

Consolidation verdict: **stay distributed, already folded where it matters.** The
family has exactly two sanctioned normalizers and they are already single-sourced in
`clients/path-utils.ts`, with `PathKeyedMap` (`clients/path-keyed-map.ts`) as the
structural enforcement for maps that must not be keyed raw. Deleting either normalizer
would push the same lines back into 83 call sites; there is no third implementation to
fold in: `realpathSync.native(` outside `path-utils.ts` is exactly 4 call sites, each a
deliberate direct use documented in place — `clients/probe-home-state.ts:253` (home
resolution, which cannot import the normalizer without the cycle named above),
`clients/sgconfig.ts:55`, and `clients/test-runner-client.ts:538` and `clients/test-runner-client.ts:551`, the last two
already feeding their result straight INTO a normalizer.

---

## Round 2

Head `bcad5867774919adaa5339d5ca5e0abf02cbb5a8` (round 1 is `46fa6f317` on this branch
after a rebase onto `d8f1f57bb`; it is the same content as the reviewed `1125a0cc`).
Every transcript below is from a run on this head.

### The seam's state space, written before the edit

The axis is *what the casing rewrite may change*. The invariant it must hold, which the
first round never stated and therefore never checked: **the returned key names the SAME
FILE the caller held** — `realpath(returned) === realpath(held)`.

| # | fixture | held | `realpath(held)` | round 1 | invariant | round 2 | pinned by |
| --- | --- | --- | --- | --- | --- | --- | --- |
| S1 | plain mis-case | `<B>/SUB/a.ts` | `<B>/sub/a.ts` | adopt ✅ | same file | adopt | `a mis-cased spelling and the on-disk spelling derive ONE map key` |
| S2 | symlinked prefix, correct case | `<B>/link/sub/a.ts` | `<B>/real/sub/a.ts` | held ✅ | same file | held | `keeps a symlinked PREFIX — POSIX normalization still resolves no symlink` |
| S3 | symlinked prefix + mis-cased tail (macOS tmpdir) | `<B>/link/SUB/a.ts` | `<B>/real/sub/a.ts` | `<B>/link/sub/a.ts` ✅ | same file | same | `a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)` |
| S4 | **case-variant symlink, other parent, colliding sibling** | `<B>/MyProject/src/a.ts` | `<B>/work/myproject/src/a.ts` | `<B>/myproject/src/a.ts` ❌ **other inode** | different file | held | `a case-variant symlink to a different directory never collapses two real files onto one key` |
| S5 | **case-variant symlink, other parent, no sibling** | `<B>/node_modules/Foo/i.ts` | `<B>/pkgs/foo/i.ts` | `<B>/node_modules/foo/i.ts` ❌ **nothing there** | no file | held | `a case-variant symlink with no colliding sibling keys under a path that exists` |
| S6 | case-variant symlink, SAME parent (`x/LINK` → `x/link`) | `x/LINK/a.ts` | `x/link/a.ts` | adopt | same file | adopt (confirmation passes) | probe only — the shape the round-1 docstring wrongly called the whole class |
| S7 | absent | `<B>/NOPE/b.ts` | throws | held ✅ | — | held | `a path that does not exist stays case-preserving` |
| S8 | two real case-distinct dirs | `<B>/SUB/a.ts` (real) | itself | held ✅ | — | held | `two genuinely distinct files on a case-sensitive filesystem keep two keys` |
| S9 | relative held | `sub/a.ts` | `<cwd>/sub/a.ts` | casing fixed, stays relative | same file | same | probe only |

Round 1 shipped S4 and S5 wrong. The round-1 non-goal test could not see it: its fixture
builds `sub`/`SUB` as SIBLINGS, where link-parent and target-parent are the same
directory, and the docstring then generalised that one harmless shape into "the one input
this cannot separate". **The class is any symlink whose basename is a case variant of its
target's basename** — the reviewer's finding, reproduced before any edit.

### F1 (HIGH) — the rewrite could name a different file. FIXED.

Reproduced first, on the round-1 build, through `normalizeMapKey`:

```
inode held : 1573481
inode other: 1573482
key(held) : …/f1/myproject/src/a.ts
key(other): …/f1/myproject/src/a.ts
COLLISION? true

held      : …/f1b/node_modules/Foo/i.ts
key       : …/f1b/node_modules/foo/i.ts
key exists on disk? false
key === held? false
```

The prescription — confirm on the `changed` branch, fall back to the held path — was
tested against every row of the table above before it was applied (S1/S3/S6 pass the
confirmation, S4/S5 fail it, S2/S7/S8 never reach it) and holds in all nine. Applied as
prescribed, in `normalizeFilePath` rather than inside `adoptCanonicalCasing`, so the
helper stays pure string algebra and the kernel question lives at the one call site that
can answer it:

```ts
const adopted = adoptCanonicalCasing(normalized, canonical);
if (adopted === normalized) return normalized;
return realpathSync.native(adopted) === canonical ? adopted : normalized;
```

The docstring's "same directory" bound is replaced with the real class and a pointer to
the confirmation, plus the reviewer's sentence: the helper pins the string algebra, never
what the kernel reports.

One of the three new lines, `if (adopted === normalized) return normalized;`, is a FAST
PATH and not a guard — deleting it costs one syscall and returns the same string, so no
test can red on it. Declared here rather than counted as mutation-proof, the same way the
round-1 body declares its two identity branches.

Post-fix, same fixtures: `S4 collision? false | held preserved? true` and
`S5 key exists? true | === held? true`.

### F2 (HIGH) — the retryable block named a path that ENOENTs. FIXED by F1, pinned here.

`checkEdit` (`clients/read-guard.ts:931`) rewrites its argument to `this.key()` and
renders that key into `read path="…"` (`clients/read-guard.ts:978`). Under S5 the key
did not exist, so the instruction could never be satisfied and the edit was blocked
forever — worse than the defect #3098 set out to fix. New case in
`tests/clients/read-guard-path-normalization.test.ts`:
`the retryable block names a path that exists, under a case-variant symlinked package`.

Related, bounded, NOT fixed here: `injectCreationRead`
(`clients/read-guard.ts:1490-1500`) reads the key and swallows ENOENT, so a key that
stops naming a file produces no read record and no signal. It is one of the call sites
listed in #3163.

### Red-first for all three new cases (revert only `clients/path-utils.ts` to round 1)

```
$ git checkout 46fa6f317 -- clients/path-utils.ts && npm run build
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 3 ⎯⎯⎯⎯⎯⎯⎯
 FAIL  … path-utils.test.ts > … > a case-variant symlink to a different directory never collapses two real files onto one key
AssertionError: expected '/home/akis/.local/share/pi-lens-orche…' not to be '/home/akis/.local/share/pi-lens-orche…' // Object.is equality
 ❯ tests/clients/path-utils.test.ts:327:41
 FAIL  … path-utils.test.ts > … > a case-variant symlink with no colliding sibling keys under a path that exists
AssertionError: expected false to be true // Object.is equality
 ❯ tests/clients/path-utils.test.ts:360:31
 FAIL  … read-guard-path-normalization.test.ts > … > the retryable block names a path that exists, under a case-variant symlinked package
AssertionError: expected false to be true // Object.is equality
 ❯ tests/clients/read-guard-path-normalization.test.ts:149:44
```

Restored: `Tests  201 passed | 2 skipped (203)`.

### Mutations — the new guard, and every prior round's set re-run on this head

| # | mutation | result |
| --- | --- | --- |
| M7 (new) | drop the confirmation: `return adopted;` | 3 red — the two S4/S5 cases and the read-guard case, nothing else |
| M1 | revert the whole POSIX arm | 4 red |
| M2 | delete `if (a.toLowerCase() !== b.toLowerCase()) break;` | 3 red |
| M4 | `return canonical;` (adopt the full realpath) | 4 red |
| M6 | whole-string case-variant rule | 1 red (the macOS-tmpdir case, alone) |

```
=== M7: drop the F1 confirmation (return adopted)
 FAIL  … > a case-variant symlink to a different directory never collapses two real files onto one key
 FAIL  … > a case-variant symlink with no colliding sibling keys under a path that exists
 FAIL  … > the retryable block names a path that exists, under a case-variant symlinked package
      Tests  3 failed | 198 passed | 2 skipped (203)

=== M1: revert the whole POSIX arm
 FAIL  … > finds a false-positive mark written under a raw (mis-cased) path form when applied via the normalizeMapKey form
 FAIL  … > a mis-cased spelling and the on-disk spelling derive ONE map key
 FAIL  … > adopts the on-disk casing itself — never a lowercased or uppercased form
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  4 failed | 197 passed | 2 skipped (203)

=== M2: delete the structural-divergence break
 FAIL  … > keeps a symlinked PREFIX — POSIX normalization still resolves no symlink
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
 FAIL  … > a case-variant symlink with no colliding sibling keys under a path that exists
      Tests  3 failed | 198 passed | 2 skipped (203)

=== M4: adopt the full realpath (naive fix)
 FAIL  … > keeps a symlinked PREFIX — POSIX normalization still resolves no symlink
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
 FAIL  … > a case-variant symlink to a different directory never collapses two real files onto one key
 FAIL  … > a case-variant symlink with no colliding sibling keys under a path that exists
      Tests  4 failed | 197 passed | 2 skipped (203)

=== M6: whole-string case-variant rule
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
      Tests  1 failed | 200 passed | 2 skipped (203)
```

### A flake I introduced and removed in the same round

The first F2 case failed 1 run in 4: `expected 'allow' to be 'block'`.
`wasWrittenThisSession` (`clients/read-guard.ts:1515`) falls back to
`mtimeMs >= sessionStartMs`, so a fixture file written in the same millisecond the guard
is constructed reads as "the agent authored this" and is allowed. The fixture now ages
the file an hour with `utimesSync` — a deterministic clock, not a retry. Six consecutive
runs after the fix: `Tests  6 passed | 1 skipped (7)` ×6.

### F3 (MEDIUM) — existence-dependent key derivation, screened and corrected

HISTORY.md shape 1 says: *"Key derivation must not change when the file appears or
disappears; otherwise capture the key before the state change."* Acceptance (1) of #3098
forces exactly that dependence on POSIX, so the screen is answered explicitly rather than
silently: the POSIX arm now answers differently for a path that appears (on-disk casing)
than for the same path while absent (the caller's spelling) — the same straddle win32 has
always had (`resolveNonExisting` lowercases the tail).

- `clients/read-guard.ts:554-579` — the `knownPathIndex` comment described the hazard as
  Windows-only. Corrected to name both platforms, and to say that ONLY
  `hasKnownPath`/`forgetPath` route through the index.
- `noteCreatedFile` (`clients/read-guard.ts:1276`) keys while the file is absent;
  `recordWritten` (`clients/read-guard.ts:1285`) keys after it exists and looks the entry
  up. Under a mis-cased parent those now differ on POSIX too, orphaning the
  `pendingCreations` entry (pruned nowhere else) so the creation read is never injected.
  Present on win32 before this PR, newly reachable on POSIX, **not fixed here** — filed with its
  table and a fix sketch as **#3163** (p2, `area:read-guard`).
- The round-1 sweep row for read-guard is corrected in place above; its "unchanged by
  this PR" verdict was wrong.

### F4 (MEDIUM) — invocation count for the worst loop

Measured on this head by counting entries into `normalizeFilePath` (a per-call counter in
the built output, one sink file per fork; the instrumentation's own `writeFileSync`
inflates the ms column, so the COUNT is the measurement and the per-call cost comes from
the clean benchmark):

| test file | calls | instrumented total |
| --- | --- | --- |
| `tests/clients/read-guard.test.ts` | **11,971** | 70.2 ms |
| `tests/clients/reverse-deps.test.ts` | **10,629** | 70.8 ms |
| `tests/clients/diagnostic-dispositions.test.ts` | 1,259 | 11.6 ms |
| `tests/clients/call-graph.test.ts` | 167 | 2.6 ms |
| `tests/clients/project-changes.test.ts` | 106 | 1.7 ms |

At the measured 1.84 µs (existing path) that is ~22 ms of new syscall tax for the
read-guard file and ~20 ms for reverse-deps; calls that land on absent paths cost 5.2 µs
each, so the real figure sits between those and roughly double. The production loops
behind the two big numbers are `sortedUnique` (`clients/reverse-deps.ts:43`, a
`normalizeMapKey` per value) and `addEdge` (`clients/reverse-deps.ts:49`, two per
import edge) — O(import edges) — plus `clients/call-graph.ts:450` and `clients/call-graph.ts:476` per symbol and
`clients/project-changes.ts:171` per change-log entry. No cache added (none was asked
for, and `path-utils.ts` has none to reuse); the number is here so the next person sizing
a memo starts from data.

### Benchmark — the confirmation costs nothing off its own branch

Same corpus as round 1, A/B on the same box state (round-1 build vs this head), two runs
each:

```
=== ROUND-1 BUILD (46fa6f317 / 1125a0ccd)
post-fix normalizeFilePath (existing)   2302 ns/call
post-fix normalizeFilePath (absent)     5548 ns/call
post-fix normalizeFilePath (existing)   2467 ns/call
post-fix normalizeFilePath (absent)     5559 ns/call
=== ROUND-2 BUILD (this head)
post-fix normalizeFilePath (existing)   1839 ns/call
post-fix normalizeFilePath (absent)     5221 ns/call
post-fix normalizeFilePath (existing)   1866 ns/call
post-fix normalizeFilePath (absent)     5273 ns/call
```

Unchanged within noise — the confirmation is not on those branches. Its own branch,
measured directly (200 files, a `PKG` → `pkg` case-variant symlink, 30k iterations, two
runs):

```
A already-canonical existing (unchanged branch)          2362 / 2413 ns/call
B mis-cased, adopts (round-2 confirm branch)             5425 / 5449 ns/call
C absent                                                 6105 / 6051 ns/call
  one bare realpathSync.native on corpus B               2466 / 2457 ns/call
```

So the confirmation costs one realpath (~2.5 µs on this box) and only where casing
actually moved — never on an already-canonical path, which is the per-edit case.

### F5 (LOW) — the "POSIX pays nothing" docstrings

Swept with `grep -rn "on Windows" --include="*.ts" clients/ tools/ mcp/ index.ts` filtered
to realpath/normalizer claims. Nine statements corrected, every design left alone:
`clients/lsp-mutation.ts:237` (the verbatim "POSIX short-circuits, so CI cannot see it"),
`clients/live-bus-emitter.ts:89`, `clients/lens-events.ts:111`,
`clients/generation-guard.ts:48`, `clients/generation-guard.ts:240` (this one also gains
the existence-dependence note — its handle re-runs the normalizer after invalidation
precisely because the answer moves when a path comes into existence, now true on POSIX
too), `clients/runtime-coordinator.ts:595`, `clients/runtime-coordinator.ts:796`,
`clients/lsp/workspace-diagnostics-cache.ts:251`,
`clients/diagnostic-dispositions.ts:177`. Deliberately NOT touched:
`clients/dispatch/dispatcher.ts:636` (about scanner IDs, and its win32-only claim is still
true — POSIX returns the id unchanged) and `clients/path-utils.ts:85` (about
`resolveNonExisting`, which remains win32-only).

### F6 (LOW) — changelog

`.changelog/3098-posix-case-insensitive-anchors.md` now says `refs #3098, refs #3090`,
matching the PR, and gained one sentence for the confirmation.
`node scripts/check-changelog-fragments.mjs` → `changelog fragments OK (104 entries in .changelog/)`.

### F7 (LOW) — the APFS leg now runs these tests

`.github/workflows/ci.yml` — `install-test` already runs on `macos-latest`, the only
case-insensitive filesystem in this project's CI, and its `npm install` step already runs
`prepare` (devDependencies + core grammars present), so the added step needs no
`build:dist` detour:

```yaml
      - name: POSIX case-insensitive anchors (APFS)
        if: matrix.os == 'macos-latest'
        shell: bash
        run: |
          npm run build
          npx vitest run tests/clients/path-utils.test.ts tests/clients/diagnostic-dispositions.test.ts
```

On APFS those cases take their NATIVE path: `createCaseAliasFixture` finds the mis-cased
spelling already aliased and never creates a symlink, so the run exercises real case
aliasing — acceptance (3) of #3098, executed by CI instead of asserted. **Title and
changelog stay `refs` in this round**: I am not permitted to poll CI, so I cannot quote
that leg green on this head. If it is green, flipping both to `closes #3098, closes #3090`
is a body/title edit with the transcript and no code change.

### Suites run on this head

| set | files | result |
| --- | --- | --- |
| `path-utils`, `diagnostic-dispositions`, `read-guard*`, `bash-read-guard-integration`, `widget*`, `lsp/workspace-diagnostics*`, `reverse-deps`, `project-changes` | 36 | `Test Files 36 passed (36) / Tests 660 passed \| 2 skipped (662)` |
| governance sweeps (mechanical glob) | 31 | `Tests 565 passed (565)` |
| `tests/config/` (walks `scripts/`, `tests/`, and `.github/workflows/`) | 62 | `Tests 621 passed \| 1 skipped (622)` |
| the three files this round touches, ×6 for the flake | 3 | green ×6 |

`npx oxfmt --check` (pinned 0.67.0) on all 11 touched files: "All matched files use the
correct format." `npm run lint` ran green as this commit's pre-commit hook.

### Observability, round 2

No new record and no new failure path: the confirmation's own throw lands in the SAME
`catch` the arm already had (a rewritten path that does not exist is the #2490 case, and
the answer is the same — keep the caller's spelling). The section above stands unchanged.

### Issues filed this round

- **#3163** (p2, `area:read-guard`) — `pendingCreations` keyed while absent, looked up
  after the file exists (shape 1 straddle; already reachable on win32 before this PR).
- Round 1 filed **#3154** (casefold residual) and **#3160** (the mark tool's raw widget
  lookup); both still open and unchanged by this round.

---

## Round 3

Head `3c533a0c9fafade0cee647f858767ec057ed646e`. Prescribed-remedy round: nothing under
`clients/` moves. All three workflow findings applied, the optional S6 case added — and
the macOS leg this PR added produced its first transcript, which contained a real defect
in two of the round-2 tests. That transcript is below; it is also the strongest evidence
this PR has.

### The APFS leg ran on round 2's head, and it settles most of acceptance (2) and (3)

`Install test (macos-latest)` → `POSIX case-insensitive anchors (APFS)`, run
`35137262977`, job `104937570590`, head `bcad58677`, read from the job log:

```
 ✓ default tests/clients/diagnostic-dispositions.test.ts (43 tests) 481ms
 ❯ default tests/clients/path-utils.test.ts (153 tests | 2 failed | 2 skipped) 1273ms
   ❯ normalizeFilePath: POSIX adopts on-disk casing (#3098, the live half of #1024) (8)
     ✓ a mis-cased spelling and the on-disk spelling derive ONE map key 1ms
     ✓ adopts the on-disk casing itself — never a lowercased or uppercased form 1ms
     ✓ keeps a symlinked PREFIX — POSIX normalization still resolves no symlink 2ms
     ✓ a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape) 3ms
     ✓ a path that does not exist stays case-preserving 0ms
     × a case-variant symlink to a different directory never collapses two real files onto one key 12ms
     × a case-variant symlink with no colliding sibling keys under a path that exists 8ms
     ↓ two genuinely distinct files on a case-sensitive filesystem keep two keys 1ms [case-insensitive filesystem: SUB and sub cannot be two directories here]
Error: EEXIST: file already exists, symlink 'work/myproject' -> '/var/folders/36/tjdph2t965j8snz9_vkdnw0r0000gn/T/pi-lens-case-Zla5K4/MyProject'
 ❯ tests/clients/path-utils.test.ts:316:7
```

What that establishes, on a real case-insensitive filesystem, with no symlink standing in
for case aliasing and no mock anywhere:

- **`tests/clients/diagnostic-dispositions.test.ts`: 43 tests, all passed, none skipped** —
  the #1024 regression case ran UNSKIPPED on APFS and passed. That is acceptance (2) of
  #3098, executed rather than argued. (A skip would appear in that summary line; it does
  not.)
- The three cases whose fixture is a case-variant SYMLINK on ubuntu took their native
  path here (`createCaseAliasFixture` finds the mis-cased spelling already aliased and
  never creates a symlink) and passed — including
  `a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)`,
  which is the real `/var/folders` → `/private/var` symlink that mutation M6 exists for.
  The tail-alignment design is confirmed against the filesystem it was designed for.
- For acceptance (3), the missing half is now only the RED: #3090's own report carries it
  (`npx vitest run tests/clients/diagnostic-dispositions.test.ts` failing on macOS,
  "Verified identical on master @ 20896a56b"), and this leg carries the green. I did not
  run the pre-fix code on APFS myself, so **title and changelog stay `refs`** and the
  call is the maintainer's.

### Two of my own round-2 tests were wrong, and the leg caught them

`a case-variant symlink to a different directory …` and `a case-variant symlink with no
colliding sibling keys …` create the colliding entry (`myproject`, `node_modules/Foo`)
and only THEN ask whether the filesystem aliases case. On a case-insensitive one that
question is asked too late: `symlinkSync` has already thrown `EEXIST`. So they FAILED on
APFS where they should have skipped — the reviewer's and my own round-2 claim that they
"skip on APFS" was wrong, and the leg is what proved it.

Fixed by probing after the colliding entry exists: the S4 guard probes `MYPROJECT` once
`myproject` is on disk, the S5 guard probes `PKGS` once `pkgs` is. Verified against a
real case-insensitive directory — tmpfs `casefold=utf8-12.1.0` + `chattr +F`, the same
unprivileged recipe as #3154, with `TMPDIR` pointed inside it:

```
--- aliasing probe on the mounted dir:
MYPROJECT aliases myproject: YES
 ↓ … > a case-variant symlink to a different directory never collapses two real files onto one key [case-insensitive filesystem: MyProject and myproject cannot be two entries here]
 ↓ … > a case-variant symlink with no colliding sibling keys under a path that exists [case-insensitive filesystem: node_modules/Foo and node_modules/foo are one entry here]
 ↓ … > a case-variant symlink whose target is its own sibling resolves through it [case-insensitive filesystem: LINK and link cannot be two entries here]
 ↓ … > two genuinely distinct files on a case-sensitive filesystem keep two keys [case-insensitive filesystem: SUB and sub cannot be two directories here]
 ↓ … > a mis-cased spelling and the on-disk spelling derive ONE map key [filesystem aliases the two spellings but reports the QUERIED casing, not the on-disk one (Linux casefold directory) — refs #3154]
 Test Files  2 passed (2)
      Tests  152 passed | 9 skipped (161)
```

(On a casefold directory the `createCaseAliasFixture` cases skip for the #3154 reason; on
APFS they RUN, as the job log above shows, because `realpath` canonicalizes there. The
casefold mount verifies the SKIP GATES, not the canonicalization.)

The read-guard case lost its skip entirely rather than gaining a better one: its probe ran
before the symlink existed, so it answered "no aliasing" on every platform and guarded
nothing. Its assertion is that the blocked path EXISTS, which holds on both kinds of
filesystem (case-sensitive: the key stays `Foo`; case-insensitive: `foo` and `Foo` name
one file), so it now runs everywhere. An inert skip deleted, not replaced.

**Coverage split after this round.** ubuntu runs S4, S5, S6, S8 (the case-DISTINCT
fixtures) and the three symlink-manufactured case cases; APFS runs S1, S2, S3, S7 plus the
#1024 disposition case natively and skips S4/S5/S6/S8 with their stated reasons; the
read-guard case runs on both.

### The three workflow findings

**R2-F1 — SonarCloud S6505/S8543 on `npx`.** The step now runs:

```yaml
        run: |
          set -euo pipefail
          npm run build
          node_modules/.bin/vitest run tests/clients/path-utils.test.ts tests/clients/diagnostic-dispositions.test.ts --configLoader runner
```

That is the form this repo already uses for its Windows Vitest subset and documents for
subagents — see `.github/workflows/ci.yml:516` and `docs/pi-lens-subagent.md:50`, both of
which spell it `node_modules/.bin/vitest run <files> --configLoader runner`. Round 2
omitted the flag because the step was copied from the command I was iterating with
locally, not from this repo's CI form; the flag makes vitest load the config through the
runner instead of bundling it, and the pinned `node_modules/.bin` binary is what removes
both Sonar findings (no on-demand install, no unpinned version). Run locally in exactly
that form on this head: `Test Files 2 passed (2) / Tests 195 passed | 1 skipped (196)`.

**R2-F2 — advisory, in the same job.** `continue-on-error: true`, with a comment saying
why: the job's install/pack/smoke steps gate as before, and a macOS-only unit flake must
not block an unrelated PR on this step's first outings. Promotion to gating is a separate
decision after several green legs — noted in the workflow comment so the next reader
knows it is a staging state, not an oversight.

**R2-F3 — the timeout comment.** `timeout-minutes: 15` is unchanged; the comment now
records that the 272 s measurement predates this build+vitest step and will be
re-measured from the first green macOS legs.

### S6, the last untested row

`a case-variant symlink whose target is its own sibling resolves through it` — the row
where the confirmation PASSES and the key therefore resolves through the symlink, the one
shape where POSIX normalization does. It pins the row's ANSWER only, which the reviewer
allowed for: it does not discriminate the confirmation.

```
=== M7 on the S6 case: drop the confirmation
 FAIL  … > a case-variant symlink to a different directory never collapses two real files onto one key
 FAIL  … > a case-variant symlink with no colliding sibling keys under a path that exists
      Tests  2 failed | 151 passed | 1 skipped (154)      <- S6 GREEN under M7: no discrimination

=== M1 on the S6 case: revert the POSIX arm
 FAIL  … > a mis-cased spelling and the on-disk spelling derive ONE map key
 FAIL  … > adopts the on-disk casing itself — never a lowercased or uppercased form
 FAIL  … > a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
 FAIL  … > a case-variant symlink whose target is its own sibling resolves through it
      Tests  4 failed | 149 passed | 1 skipped (154)      <- S6 red when the arm is gone
```

### Red-first and M7, re-proven with the corrected fixtures

The fixtures changed this round, so both proofs were re-run rather than carried over.
Revert `clients/path-utils.ts` to round 1 (`46fa6f317`), rebuild:

```
 FAIL  … > a case-variant symlink to a different directory never collapses two real files onto one key
AssertionError: expected '/home/akis/.local/share/pi-lens-orche…' not to be '/home/akis/.local/share/pi-lens-orche…'
 FAIL  … > a case-variant symlink with no colliding sibling keys under a path that exists
AssertionError: expected false to be true
 FAIL  … > the retryable block names a path that exists, under a case-variant symlinked package
AssertionError: expected false to be true
      Tests  3 failed | 156 passed | 2 skipped (161)
```

Restored: `Tests 159 passed | 2 skipped (161)`. M7 (drop the confirmation) on the same
files reds the same three and nothing else: `Tests 3 failed | 156 passed | 2 skipped (161)`.

The three guards on ubuntu, unskipped and green on this head:

```
 ✓ … > a case-variant symlink to a different directory never collapses two real files onto one key 1ms
 ✓ … > a case-variant symlink with no colliding sibling keys under a path that exists 0ms
 ✓ … > a case-variant symlink whose target is its own sibling resolves through it 0ms
 ✓ … > the retryable block names a path that exists, under a case-variant symlinked package 2ms
```

### Suites on this head

| set | files | result |
| --- | --- | --- |
| the 36-file round-2 batch | 36 | `Test Files 36 passed (36) / Tests 661 passed \| 2 skipped (663)` |
| governance sweeps (mechanical glob) | 31 | `Tests 565 passed (565)` |
| `tests/config/` (walks `.github/workflows/`) | 62 | `Tests 621 passed \| 1 skipped (622)` |

`npm run lint` exit 0 — note it is `tsc` + `oxlint` and does NOT cover workflows;
`actionlint` is not installed on this box and is not an npm script here, so the workflow
edit is validated locally only by a YAML parse plus `tests/config`'s workflow sweeps, and
gated by CI's own `actionlint` job. `npx oxfmt --check` clean on both touched test files.
`node scripts/check-pr-body.mjs --lint-local <body>` exit 0 for this body.


georgeharker:
## Verification on a real case-insensitive filesystem — macOS 26.5.2 (Darwin 25.5.0, APFS)

Ran on `fix/3098-posix-case-insensitive-anchors` @ `1125a0ccdfd928bc12de834c561fe49e0ae40eb8`:

```
npx vitest run tests/clients/diagnostic-dispositions.test.ts tests/clients/path-utils.test.ts

 Test Files  2 passed (2)
      Tests  192 passed | 2 skipped (194)
```

The #1024 case ran **unskipped and green** — `finds a false-positive mark written under a raw (mis-cased) path form when applied via the normalizeMapKey form` (red on `master` on this machine before this PR).

Five of the six new `#3098` cases ran unskipped and green:

- a mis-cased spelling and the on-disk spelling derive ONE map key
- adopts the on-disk casing itself — never a lowercased or uppercased form
- keeps a symlinked PREFIX — POSIX normalization still resolves no symlink
- a symlinked prefix with a mis-cased tail fixes the casing only (the macOS /var/folders shape)
- a path that does not exist stays case-preserving

The one skip is the sixth case — `two genuinely distinct files on a case-sensitive filesystem keep two keys` — which matches the stated filesystem-property condition: APFS cannot present two distinct on-disk casings for the same directory.

One verification question: did this PR consider rehydration of the persisted path-keyed stores (review graph, word index, reverse deps) across the spelling change — entries written under the pre-#3098 casing of the same file vs new writes? Worst case reads as a re-scan, never a wrong answer, but worth stating if it was considered.

Issue #3170, opened 2026-09-16
Re-verify persistent findings against the live server before re-delivering them at turn_end


georgeharker:
## What happened

A finding whose own file is unchanged is re-delivered turn after turn from
cache without ever being re-observed against the live server. The
turn-end advisory builds from the turn's mutated files plus carried
persisted-report entries (`clients/actionable-warnings.ts`,
`buildActionableWarningsReport`), and the delta freshness gate passes an
entry whose file is mtime-fresh — so when the root cause was fixed in
*another* file (the #1093 cross-file class), file A's finding re-serves
verbatim on every turn even though the server would now answer clean.
Every existing re-observation mechanism is reactive to a different trigger:
late-auxiliary re-arm (budget), disk-drift resync (#1783 — divergence), the
past-EOF resync, and the manual `source=lsp` probe (agent-initiated).
Nothing re-verifies because a finding *keeps being delivered*.

## Expected behavior (acceptance criteria)

1. A finding scheduled for re-delivery that (a) was delivered in a previous
   turn this session, (b) whose file stat (mtime+size) is unchanged since
   its own observation stamp, and (c) whose file has a live LSP client, is
   re-observed (a collecting `touchFile`, the same path the `source=lsp`
   probe uses) BEFORE the advisory is assembled. Root-cause-fixed-elsewhere
   findings converge or drop. (Red on pre-fix: re-served verbatim.)
2. A re-confirmed finding is delivered labeled as re-verified-and-persistent
   (extending #3167's label vocabulary: `re-verified · carried N turns`).
3. A re-verify that cannot complete inside the budget delivers with an
   explicit `re-verify incomplete` label — never a false clean.
4. A file changed since its observation stamp is NOT re-verified (the edit
   path already re-observes it) — no duplicate work behind dispatch.
5. The pass is bounded: at most ~4 files per turn_end, a wall budget of a
   few seconds, the turn-end abort signal, and carry-forward identity so an
   interrupted re-verify re-arms next turn rather than vanishing.
6. The delivered-identity evidence is a bounded session-scoped structure
   (cap, session reset only, no durable state).

## Root cause

Delivery-side: the advisory re-serves cached observations; the freshness
gates catch files that changed on disk, but nothing forces a fresh
observation of an unchanged file whose findings may no longer be true of
the *project* state.

## Non-goals

Not a workspace sweep (`mode=full` exists); no durable state; no
retire-after-N policy; no change to the dispatch debounce cadence or to
[...]

PR #3168, opened 2026-09-16
feat(diagnostics): age labels for carried cascade advisories and demoted delta rows (closes #3167)

georgeharker:
## Summary

Resolves the two `partial` delivery-gate entries that name the gap: a cascade
result carried across a turn boundary re-rendered at turn_end
indistinguishable from a fresh observation, and demoted (stale) delta rows
repeated with the stale marker but no age information.

- `clients/cascade-format.ts`: new `cascadeCarrySuffix` — the honest carry
  label. The carry is bounded to ONE turn (`RuntimeCoordinator.beginTurn`
  drops anything that would reach 2), and the run DOES carry an observation
  stamp: `clients/cascade-format.ts:177` threads #1444's `publishedAt` into
  `CascadeRun.observedAt`, so the label is `(carried 1 turn · scanned Xm ago)`
  with the age wording from `formatCacheAgeLabel`. The deferred-compute
  re-park arm has no stamp and falls to `scan age unknown`.
- `clients/lsp/cascade-tier.ts:494`: the resolved-found reconcile arm passes
  `publishedAt` to `onResolvedFound`, the stamp `:405` already put on the
  outcome. The resolved-clean arm already did this (#1444).
- `clients/runtime-turn.ts`: the blocker assembly maps run→result carried
  counts and appends the suffix on its own line
  (`clients/runtime-turn.ts:1266`); the three cascade-coverage advisory builds
  wrap `buildAdvisory` with `withCarryLabel`, which labels only an ALL-carried
  bucket, joins with a newline (`clients/runtime-turn.ts:1479`) and takes the
  oldest carried stamp only when every carried run has one
  (`clients/runtime-turn.ts:1470`).
- `tools/lens-diagnostics.ts`: `applyDeltaFreshnessGate` tags each demoted row
  with the stamp it was judged against (`staleAsOf`, the #2504 r4 per-entry
  stamp, not the report-level one), and `formatDeltaMode` closes each demoted
  file group with exactly one age line **under that file's own header** —
  `appendGroupAgeLabel` (`tools/lens-diagnostics.ts:1080`) reports whether it
  labelled, the actionable loop records that in `labelledFiles`
  (`tools/lens-diagnostics.ts:1187`) and the quality loop no-ops only for
  files already labelled. "scan age unknown" when a stamp is somehow absent,
  never a fabricated number.
- `clients/finding-delivery-gate.ts`: the two `partial` entries resolve to
  full, each claiming its seam call as evidence
[...]

Issue #3167, opened 2026-09-16
Carried cascade advisories and demoted delta rows render without an age label (the partial delivery-gate entries name this exactly)


georgeharker:
## What happened

A carried-over cascade result (`run.carriedTurns > 0`) re-renders at turn end
indistinguishable from a fresh observation, and demoted (stale) delta rows
repeat with the stale marker but no age information. The delivery-gate
registry records this debt itself — `clients/finding-delivery-gate.ts`,
entries `runtime-turn:cascade-blocker` and `runtime-turn:cascade-coverage-advisory`,
both `status: "partial"`:

> "A carried-over cascade result (run.carriedTurns > 0) is rendered without
> an age label. Follow-up: surface carriedTurns as an explicit label when > 0,
> or route through formatCacheAgeLabel using the run's own timestamp."

So the delivery was made *safe* (carried and stale findings are demoted,
never silently dropped) but never made *honest about age*: an agent reading
the advisory cannot tell a just-observed finding from one carried for three
turns.

## Expected behavior (acceptance criteria)

1. A cascade run carried across ≥1 turn renders with an explicit
   carry+age label (e.g. `(carried 2 turns · scanned 12m ago)`); a
   non-carried run renders unchanged — no label noise.
2. Label wording derives from `formatCacheAgeLabel`
   (`clients/finding-delivery-gate.ts` — the single age-wording source, with
   its hour-flooring correctness), not a hand-rolled second wording.
3. Demoted delta rows (`STALE_LINE_MARKER`) gain exactly one age label per
   file group — the group already carries `generatedAt`/per-entry stamps.
4. Missing or unparseable stamp renders the neutral "scan age unknown" —
   never a fabricated number.
5. The two `partial` registry entries resolve to full (partialReason
   removed), pinned by the gate-test evidence.

## Root cause

Delivery of carried/stale findings was made safe by demotion but the age
fact was never rendered; the registry's `partialReason` names the follow-up
explicitly, so this issue mostly exists to claim it.

## Non-goals

No retire-after-N or drop-after-N policy (carried findings keep rendering,
now labeled); no change to the demote-vs-drop policy, the #1631 dependency
gate, or the #1093/#1561 cross-file reconcile behavior; no new timestamps
recorded — only rendering of stamps that already exist.

## Failure semantics

Missing/unparseable stamp → "scan age unknown" (the helper's existing
contract: never fabricate). Carry count 0/undefined → no label.

## Observability

[...]

PR #3176, opened 2026-09-16
fix(diagnostics): re-verify persistent findings against the live server before re-delivery (refs #3170, refs #3167)

georgeharker:
## Summary

Refs #3170, refs #3167. A finding whose own file is unchanged re-served turn
after turn from the persisted actionable-warnings report without ever being
re-observed — the in-band publish carries forward DEFERRED-origin file
entries while their file has not moved (`mergeActionableWarningsReports`'s
scope guard), and the delta freshness gate passes them because the file's
mtime has not moved either. When the root cause was fixed in a DIFFERENT
file (the #1093 cross-file class), the carried finding is stale in fact but
fresh by every on-disk axis, and it re-reported on every turn.

The fix re-observes such files at turn_end BEFORE the advisory assembles,
through the probe's existing `touchFile` path — no second observation
mechanism:

- `clients/persistent-reverify.ts` (new): the bounded pass. Candidates are
  the persisted report's DEFERRED-origin file entries whose stat is
  unchanged since the entry's own `generatedAt` stamp — the persisted
  report IS the carry evidence, so the identity-LRU the issue sketched
  collapsed into zero new state (AC 6 satisfied vacuously). Bounds: ≤4
  files per turn_end, a 3s wall budget, the turn-end abort signal, and a
  `bounded()` wrapper on every await. Never throws: a per-file failure
  degrades that file to `unconfirmed` (kept verbatim, labeled).
- `clients/actionable-warnings.ts`: `ActionableWarningsReportFile` gains
  `reVerified`/`reVerifyIncomplete`, and the merge fold SUPERSEDES instead
  of unions when the incumbent is a fresh re-observation — a converged
  finding would otherwise survive its own supersession via the id-union in
  `mergeWarnings` (mutation-proven below).
- `clients/runtime-turn.ts:3518`: the replacement entries fold into THIS
  turn's single in-band publish, so the carry marker is spent exactly once
  (the round-1 blocker).
- `tools/lens-diagnostics.ts:1082` + the advisory formatter: a file whose
  re-verify could not complete renders an explicit `(re-verify incomplete)`
  gap label — never a false clean.

**Merged, not rebased (#3168 first).** `origin/master` is merged into this
branch twice (`04d410337`, `8fac62443`), absorbing #3168's merged state and
the rest of master. One conflict, `tools/lens-diagnostics.ts`: master's
`appendGroupAgeLabel` (boolean, recorded into `labelledFiles` by the
actionable loop) and this branch's `appendGroupLabels(lines, warnings,
incomplete)` (void, deduped through a PREDICTED `qualityLabeledFiles`) are
the same helper. Resolved to ONE helper — the three-arg `appendGroupLabels`
returning whether any label was pushed (`tools/lens-diagnostics.ts:1090`),
called in the actionable loop first with the result recorded into
`labelledFiles` (`:1218`), the quality loop no-oping only for files
actually labelled (`:1233`), and `qualityLabeledFiles` deleted. That
prediction IS #3168's F10 defect; keeping it would have re-shipped it.

`refs`, not `closes`: AC 2 ships as a design deviation — a re-confirmed
finding is re-delivered as a FRESH observation (new stamp, new records)
rather than carrying a `re-verified · carried N turns` label. The remainder
is named on #3170. #3205 carries the deferred R2-6 finding.

## Type of change

- [x] Bug fix
- [ ] Documentation

## Area

- [x] area:lsp
- [x] area:diagnostics
- [x] area:session

## Checklist

- [x] `npm run lint` passes (exit 0 on `9e030e903`)
- [x] `npm run build:dist` succeeds (clients/ changed)
- [x] AGENTS.md needs no update: no invariant changes — the pass reuses the probe's `touchFile` path, the publisher's merge, and the `touchCompletedConfirmationPolicy` seam
- [x] `.changelog/persistent-reverify-3170-added.md` has one valid entry in this PR (`node scripts/check-changelog-fragments.mjs` → `changelog fragments OK (13 entries in .changelog/)`)
- [x] Commit subject includes the issue number: `(refs #3170)`

## Tests

`tests/clients/persistent-reverify.test.ts` (new, 14 cases):

| test | pins |
|---|---|
| `C1: a confirmed-clean touch drops the carried finding — through the real CacheManager and publisher` | the merge SUPERSEDE through the real store and publisher; union semantics would keep the carried warning |
| `C2: a re-confirmed finding is re-delivered as a fresh observation — supersede, not union` | the replacement carries a fresh record and a fresh stamp |
| `F3: a silent empty touch (no confirmation) is UNCONFIRMED — the carried warnings are kept and the gap is labeled` | the false-clean the round-1 review caught: the house double's bare `{diags: []}` is not clean |
| `R3-3: a PARTIAL confirmation is a completed touch — the primary's answer is honored, not labelled incomplete` | the #1470 seam, not the literal `confirmation !== "confirmed"` |
| `F3b: a skipReason touch is unconfirmed for the same reason` | `skipReason` is honored beside the confirmation policy |
| `C3: a throwing touch keeps the carried entry verbatim and marks the gap` | the never-throws contract; the marker rides the entry |
| `F6: an entry re-arms — the pass re-verifies it again on the next run` | no skip guard; the issue's failure table |
| `F3-population: an entry with auxiliary-sourced warnings is not re-verified by a primary-scope touch` | an aux-lane finding is never scored dropped by a primary answer |
| `C4: a file changed since its observation stamp is skipped — counted in the result and the latency record` | no duplicate work behind the edit path; `skippedChanged` is real |
| `C5: the pass caps at exactly 4 candidates (the literal pin — the constant →1000 mutation reds this)` | the ≤4 bound, pinned literally |
| `the abort signal stops the pass before any touch` | the abort half of the loop guard |
| `the wall budget stops the pass between files — later candidates are re-armed, not half-verified` | the budget half, plus `skippedBudget` in the phase record |
| `F1-probe: a re-confirmed finding is still in the advisory after the fold` | the maintainer's demanded F1 proof, through the real advisory formatter |
| `the advisory renders the re-verify gap label` | the gap label on the turn-end advisory surface |

`tests/clients/cascade-carry-label.test.ts` (edited, +2 cases):

| test | pins |
|---|---|
| `#3170: a re-verify-incomplete file demoted in both reports carries both labels ONCE, under its own header` | the conflict resolution: the age label and the gap label share one group slot and the quality loop no-ops on what the actionable loop actually pushed |
| `#3170: a LIVE re-verify-incomplete file present in both reports carries the gap label once` | the same F10 duplication for a file with no demoted row, where the gap label is the group's ONLY label |

Test-authoring screens (AGENTS.md): both new label cases drive the real
`lens_diagnostics` delta handler (no wrong-layer pin, no all-mocks — the
only double is the cache reader); `R3-3` observes an independent effect
(the outcome and the marker), never the implementation's own expression;
the budget case asserts counts and outcomes, never elapsed ms (no loose
bound).

**Red-first transcripts, on the takeover head before the fix commit
(`9e008e39d`):**

```
 FAIL  |wall-clock-budget| tests/clients/persistent-reverify.test.ts > persistent reverify (#3170) > R3-3: a PARTIAL confirmation is a completed touch — the primary's answer is honored, not labelled incomplete
AssertionError: expected 'unconfirmed' to be 'reconfirmed' // Object.is equality

Expected: "reconfirmed"
Received: "unconfirmed"

 ❯ tests/clients/persistent-reverify.test.ts:280:39
```

```
 FAIL  |wall-clock-budget| tests/clients/persistent-reverify.test.ts > persistent reverify (#3170) > the wall budget stops the pass between files — later candidates are re-armed, not half-verified
AssertionError: expected "vi.fn()" to be called with arguments: [ ObjectContaining{…} ]
...
-     "metadata": ObjectContaining {
        "candidates": 2,
-       "skippedBudget": 1,
+       "clean": 0,
+       "mixed": 0,
+       "reconfirmed": 0,
+       "skippedChanged": 0,
+       "touched": 0,
+       "unconfirmed": 1,
      },
```

```
 FAIL  |wall-clock-budget| tests/config/strictness-ratchet.test.ts > TypeScript strictness spike ratchets > pins both scratch configurations without permitting drift
AssertionError: expected [ 'clients: regression (357 -> 358)' ] to deeply equal []
```

**Mutation set (neutered in the BUILT output, suite re-run, rebuilt after
each).** Baseline: `Tests  45 passed (45)` over
`persistent-reverify.test.ts` + `actionable-warnings-deferred-bounds.test.ts`.

| # | mutation | red |
|---|---|---|
| M1 | `MAX_REVERIFY_FILES = 4` → `1000` | `× C5: the pass caps at exactly 4 candidates …` — `Tests  1 failed \| 44 passed (45)` |
| M2 | `if (args.signal?.aborted \|\| Date.now() >= deadlineAt)` → `if (false)` | `× the abort signal stops the pass before any touch`, `× the wall budget stops the pass between files …` — `2 failed \| 43 passed` |
| M3a | `!touchCompletedConfirmationPolicy(touchedResult)` → `touchedResult.confirmation !== "confirmed"` (the literal back) | `× R3-3: a PARTIAL confirmation is a completed touch …` — `1 failed \| 44 passed` (the new case is the ONLY red: it has its own signature) |
| M3b | the confirmation clause deleted (the false-clean direction) | `× F3: a silent empty touch …`, `× keeps the deferred entry AND turn N+1's own, instead of clobbering it`, `× turn N+1's advisory mentions turn N's deferred finding, and its telemetry counts it` — `3 failed \| 42 passed` |
| M4 | the aux-source skip → `if (false)` | `× F3-population: an entry with auxiliary-sourced warnings …` — `1 failed \| 44 passed` |
| M5 | `if (mtimeMs > stamp)` → `if (false)` | `× C4: a file changed since its observation stamp is skipped …` — `1 failed \| 44 passed` |
| M6 | supersede: `incumbent.reVerified \|\| incumbent.reVerifyIncomplete` → `false` | `× C1 …`, `× C2 …`, `× F1-probe …` — `3 failed \| 42 passed` |
| M7 | `skippedBudget: candidates.length - processed` deleted from the record | `× the wall budget stops the pass between files …` — `1 failed \| 44 passed` |
| M8 | `appendGroupLabels` → `return false` | `× F5 …`, `× F10(a) …`, `× #3170: a re-verify-incomplete file demoted in both reports …`, `× #3170: a LIVE re-verify-incomplete file …` — `4 failed \| 10 passed (14)` |
| M9 | `appendGroupLabels` → `return staleRow !== undefined` (drops `\|\| incomplete`) | `× #3170: a LIVE re-verify-incomplete file present in both reports carries the gap label once` — `1 failed \| 13 passed` (only red: its own signature) |
| M10 | the `(re-verify incomplete)` push deleted | `× #3170: a re-verify-incomplete file demoted …`, `× #3170: a LIVE re-verify-incomplete file …` — `2 failed \| 12 passed` |

M2 and M3 are mutated in BOTH directions where a direction exists; the
dangerous direction (real failures stop blocking) is M2's `if (false)` and
M3b's deleted clause, and both red.

**Suites run on the final head** (`npm run build` first):

- 94 governance files — `ls tests/clients/*{sweep,ratchet,conformance,coverage,gate,silence,hermeticity,contract}*.test.ts` plus every `tests/config/*.test.ts` — `Test Files  94 passed (94)`, `Tests  1196 passed | 1 skipped (1197)`. Includes `hook-await-bounds`, `flake-shape-ratchet`, `strictness-ratchet`, `vi-mock-export-sweep`, `latency-logger-mock-shape`, `tmp-fixture-hygiene`.
- 24 named files — `persistent-reverify`, every `actionable-warnings*` (deferred-bounds included), `cascade-carry-label`, `cascade-finding-policy`, `finding-delivery-gate`, `delivery-surface-ratchet`, `widget-suppressed-retention`, `pi-host-callable-schema-contract`, `tool-policy-conventions`, every `tests/tools/lens-diagnostics*` — `Test Files  24 passed (24)`, `Tests  404 passed | 2 skipped (406)`.
- `npx oxfmt --check` on every touched file: `All matched files use the correct format.`
- `npm run lint`: exit 0. `node scripts/check-changelog-fragments.mjs`: exit 0.

No new governance exemption, baseline row or admission was added in this
round; the flake-shape admission and the two `BOUNDED_CALL_SITES` rows
predate it (one row's text is corrected, see R3-2).

### Test assessment

- `tests/clients/persistent-reverify.test.ts` — uniquely pins the pass's
  selection, verdict, bounds and phase record against the REAL
  `CacheManager`, publisher and advisory formatter. Nothing in it is
  vacuous: every case is the sole red under at least one mutation above
  except `F3b`/`C3`/`F6`, which share the unconfirmed branch with `F3` and
  are kept as the distinct *entry* paths into it (skipReason, throw,
  re-arm) — each names its recurrence in a comment.
- `tests/clients/cascade-carry-label.test.ts` — uniquely pins which HEADER
  owns a group label and how many times. `labelsByHeader` now also collects
  `(re-verify incomplete)`, so the existing F5/F10 cases additionally guard
  against a stray gap label. No case in the file reds on zero mutations
  (M8 alone reds four of them); nothing is made redundant, nothing removed.
- `tests/config/hook-await-bounds.test.ts` — registry text only (R3-2); no
  case added or removed.
- `tests/clients/flake-shape-ratchet.test.ts` — comment placement only
  (R3-5); no case added or removed.

## Blast radius

- `clients/persistent-reverify.ts` (new): leaf module. Callers above: one —
  `clients/runtime-turn.ts:3413`. Callees below: `actionable-warnings`
  (`enrichFileFromLsp`), `deadline-utils` (`bounded`), `latency-logger`,
  `lsp/diagnostic-binding` (`touchCompletedConfirmationPolicy`, +1 call),
  `dispatch/auxiliary-lsp` (`findAuxiliaryProfileForSource`).
  `RunPersistentReverifyArgs.signal` widened `AbortSignal` →
  `AbortSignal | undefined` (+1 line): a widening, so no caller can break.
- `clients/actionable-warnings.ts`: two optional entry fields; the merge
  fold gains a supersede branch that fires ONLY when the incumbent carries
  `reVerified`/`reVerifyIncomplete` — set exclusively by this pass, so
  every existing merge caller's behaviour is byte-identical. Merge callers:
  runtime-turn (in-band + deferred publishes),
  `writeDeferredActionableWarningsReport`. Strict consumers of the durable
  record: the two added fields are optional and additive; old caches parse
  (verified by the deferred-bounds suite, whose fixtures predate this change).
- `clients/runtime-turn.ts`: the pass runs once per turn_end, gated on the
  persisted report holding deferred entries and a live LSP service.
- `tools/lens-diagnostics.ts`: `appendGroupLabels` — callers above: the two
  delta render loops (`:1212`, `:1234`), both changed by the conflict
  resolution; no caller outside this file (`grep -rn appendGroupLabels
  clients/ tools/ mcp/ scripts/ index.ts` → 4 hits in
  `tools/lens-diagnostics.ts`, one of them a comment). Cost: one `Set` lookup and at most two pushed lines per
  rendered file group, in a render branch only — not a per-spawn or
  per-render hot path.
- Empty/unavailable: a missing file entry is not a candidate (the delta
  gate drops it); an unreadable file degrades to `unconfirmed` (kept
  verbatim, labeled) — never hidden.

## Observability

The diff ADDS the `persistent_reverify` latency phase
(`clients/persistent-reverify.ts:338`) — one record per pass, never per
finding — with `{candidates, touched, clean, reconfirmed, mixed,
unconfirmed, skippedChanged, skippedBudget}`. Both skip counters are
asserted through the real seam in this diff:

```ts
// tests/clients/persistent-reverify.test.ts (the budget case)
		expect(logLatencyMock).toHaveBeenCalledWith(
			expect.objectContaining({
				phase: "persistent_reverify",
				metadata: expect.objectContaining({
					candidates: 2,
					skippedBudget: 1,
				}),
			}),
		);
```

```ts
// tests/clients/persistent-reverify.test.ts (C4)
		expect(logLatencyMock).toHaveBeenCalledWith(
			expect.objectContaining({
				phase: "persistent_reverify",
				metadata: expect.objectContaining({ skippedChanged: 1 }),
			}),
		);
```

Lifetime: the record is per pass, per turn_end; the pass holds no state
across turns (the persisted report is the carry evidence), so nothing here
survives or needs resetting at `session_start`. Delivery-side, the
`(re-verify incomplete)` gap label is emitted on both surfaces the agent
reads — the turn-end advisory and `lens_diagnostics` delta output — and
both are asserted above.

## Class sweep

Shape 10 (silencing as fixing — an unconfirmed re-verify is LABELED, never
a false clean) and shape 31 (pull-only observability — the skip counters).
The #1093 class on the advisory lane is the defect this closes.

**Re-observation population sweep** (whole tree,
`grep -rn "touchFile(" --include=*.ts clients/ tools/ mcp/ scripts/ index.ts`):
10 LSP `touchFile` call sites — `clients/warm-attach.ts:131`,
`clients/runtime-session.ts:765`, `clients/persistent-reverify.ts:216`
(this PR), `clients/dispatch/integration.ts:2009` and `:2099`,
`tools/lsp-diagnostics.ts:653`, `clients/dispatch/runners/lsp.ts:211`,
`clients/mcp/analyze.ts:406`, `clients/lsp/index.ts:8696` and `:8932`
(`clients/read-guard.ts`'s three `this.touchFile` hits are a different
method — the read-record toucher — and are not members). The subset that
re-observes an ALREADY-DELIVERED finding is the family the issue names:
the manual `source=lsp` probe (agent-initiated), the #1783 drift resync
(disk divergence), the late-auxiliary rearm (budget), the past-EOF resync
(position), and THIS (delivery persistence). Verdict: stays distributed —
each is owned by its own trigger, and deleting a shared module would
relocate the same lines into five callers rather than concentrate
anything. No second observation mechanism was added: the pass reuses the
probe's `touchFile` path.

**Confirmation-verdict population sweep** (the R3-3 shape —
`grep -rn '\.confirmation\b' --include=*.ts clients/ tools/ mcp/ scripts/ index.ts`
plus `grep -rn 'confirmation [!=]== "confirmed"'`), per member:

| site | reads | verdict |
|---|---|---|
| `clients/persistent-reverify.ts:253` | a `TouchFileResult`'s verdict | **was the defect** — the literal; now `touchCompletedConfirmationPolicy` |
| `tools/lsp-diagnostics.ts:704` | a `TouchFileResult`'s verdict | already reads the seam — clean |
| `tools/lsp-diagnostics.ts:642` | the IPC DTO's `confirmation !== undefined` | same rule, spelled inline on a socket-crossing shape that is not a `TouchFileResult`; its own #1470 comment states `"partial"` counts — clean |
| `clients/warm-attach.ts:173` | serializer, not a verdict read | forwards `confirmed`, then emits `partial` explicitly at `:180-185` — clean |
| `clients/lsp/index.ts:6989`, `:6996`, `:7102` | the WRITER of the field | not a consumer — clean |
| `tools/lsp-diagnostics.ts:1646`, `:1666` | `FileDiagnosticResult.confirmation === "unconfirmed"` | a different type and a different union (no `"unconfirmed"` on `TouchFileResult`) — not a member |

Consolidation verdict: stays distributed. Both true `TouchFileResult`
verdict consumers now read the one seam
(`clients/lsp/diagnostic-binding.ts:289`); the remaining sites are a
writer, a serializer and a different type, so folding them would relocate
lines rather than concentrate complexity.

## Round 3 — maintainer takeover of the contributor branch

Every finding from "Verify at 5d813d09d" is applied on this branch; the
branch was MERGED with `origin/master` (never rebased, never force-pushed).

- **R3-1 (CI-red)** — `RunPersistentReverifyArgs.signal` widened to
  `AbortSignal | undefined` (`clients/persistent-reverify.ts:156`) rather
  than conditional-spreading at the one call site: the caller reads
  `getAmbientAbortSignal(): AbortSignal | undefined`
  (`clients/safe-spawn.ts:691`), so the declaration was the thing that was
  lying. Red quoted above (`clients: regression (357 -> 358)`); green at
  the 357 baseline on the final head (`tests/config/strictness-ratchet.test.ts`
  passing with `compare(...)` empty for both scratch configs).
- **R3-2** — the `BOUNDED_CALL_SITES` row for
  `call:clients/runtime-turn.ts#4da1e4ca~7e52ce49` no longer claims the
  ambient signal is "ALWAYS defined … Never wall-clock". `index.ts:3068`
  reads `(ctx as { signal?: AbortSignal })?.signal`, so it can be absent
  and the bound is then wall-clock only; the row now says so in the
  siblings' phrasing.
- **R3-3 (behaviour)** — the gate reads
  `touchCompletedConfirmationPolicy(result)`
  (`clients/persistent-reverify.ts:253`); the `skipReason`, `inconclusive`
  and `diagnosticsUnsupportedServerIds` gates are unchanged. Red-first
  quoted above; mutation M3a (the literal back) reds the new case alone.
- **R3-4** — `skippedBudget` is back in the phase record
  (`clients/persistent-reverify.ts:338`), counted from what the loop
  actually reached (`processed`), not predicted. Red-first quoted above;
  mutation M7 reds it.
- **R3-5** — the two flake-admission comments each sit above their own
  entry again (`tests/clients/flake-shape-ratchet.test.ts`). Comment-only.
- **R3-6** — this body is rewritten with the real `it(` titles;
  `node scripts/check-pr-body.mjs --lint-local` exits 0 on it (the seven
  missing references are gone). The changelog fragment now states the
  shipped behaviour: the confirmation-policy rule, the aux-lane skip, the
  gap label on both surfaces, and the skip counters.
- **R2-6 (deferred, as the review directed)** — filed as **#3205** with the
  probe shape: a carried-verbatim entry whose touch FAILED is folded into
  `report.files` (`clients/runtime-turn.ts:3518-3540`) and so earns one
  NDJSON history row per warning per turn, against
  `appendActionableWarningsHistory`'s own stated rule
  (`clients/actionable-warnings.ts:1739-1742`).

**New guard added this round, beyond the findings.** The delta surface's
`(re-verify incomplete)` label had NO test at all — the conflict resolution
put it behind the `labelledFiles` dedupe, and every mutation of it stayed
green. Two cases were added through the real delta handler (M8/M9/M10
above). This is the reviewer's "position and lifetime" screen applied to
the merge: the label the resolution moved is now the thing the suite
watches.

**Merge-order note.** #3168 is already on master and absorbed by the merge;
this branch also carries its own copy of the #3168 strictness commit
(`35e45d29f`, a different SHA from master's `91c8dbc59`) — tree-identical,
and the merge carries both harmlessly. Nothing else open touches
`clients/persistent-reverify.ts`; `tools/lens-diagnostics.ts` and
`clients/actionable-warnings.ts` are the shared seams, both resolved
against current master.


apmantza:
Thanks — the problem #3170 names is real and the shape of the answer (re-ask the live server before re-serving a deferred-origin entry) is right. Adversarial review at head a52e778ba, second commit as this PR's diff, judged on the combined shape with #3168 (whose review is on that PR). Every claim below was probed on built output with the real `CacheManager`, publisher and advisory formatter; transcripts on request. **Needs changes** — one blocker, and CI is red (Unit tests job 104984862540 in run 35149047786).

**F1 — BLOCKER — the pass deletes every carried finding it touches, re-confirmed or not.** `clients/runtime-turn.ts:3374-3394` publishes the replacement with `origin: "in-band"`; `mergeActionableWarningsReports` spends the carry marker on every merged entry (`clients/actionable-warnings.ts:1637-1639`), and the turn's real in-band publish (`runtime-turn.ts:3486`) then drops them at the scope guard (`actionable-warnings.ts:1575-1581`). Probe with a server re-confirming the finding: published files `[]`, advisory `(none)`; master shape: `src/a.ts: 1`. Same for the wedged-server path, so the `(re-verify incomplete)` label at `tools/lens-diagnostics.ts:1191` and `actionable-warnings.ts:2188` is unreachable in production (the C3-render test hand-writes a cache state production cannot produce). The repo's own guards catch it: `tests/clients/actionable-warnings-deferred-bounds.test.ts` (#2504 r5 F1, r7 F2/F3) is red on this head and on the merge with master, green with the three source files reverted. The body's "no existing test edited or made redundant" did not run that file. Remedy direction: fold the replacement into the turn's single in-band publish (hand the replacement entries to `buildActionableWarningsReport`'s publish as an override map), or have the re-verify publish preserve `origin: "deferred"` so the marker is spent exactly once as master intends. Prove it with that test file green plus a probe showing a re-confirmed finding still in the advisory.

**F2 — HIGH — the replacement bypasses the channel's own three filters.** `clients/persistent-reverify.ts:196-200` converts every returned diagnostic; the producer it mirrors (`actionable-warnings.ts:1241-1292`) filters `severity === 2`, `lineInModifiedRanges`, and `actions.length > 0`. Probe: 1 carried warning, server answers 1 warning + 2 errors + 1 hint → four "warning" records with zero fixes in the "Fixable warnings" advisory. Reuse the producer's pipeline rather than re-implementing it.

**F3 — HIGH — a false clean.** `persistent-reverify.ts:185-205` treats any `{diags: []}` as clean (still indexing, silent-on-clean, `skipReason: "outside-project-root"`), deleting the carried warnings. `clients/lsp/diagnostic-binding.ts:105-110` states the opposite rule: `confirmation` is required before an empty result from a silent-on-clean server is clean. The pass never reads `confirmation`, `unconfirmedServerIds`, `diagnosticsUnsupportedServerIds` or `skipReason`, and `ReverifyLspService` narrows `TouchFileResult` so the double cannot express them. Also, the population came from `getDiagnostics()` with no server exclusion (auxiliary included) while the re-verify asks `clientScope: "primary"`, so an auxiliary-lane finding can never be re-confirmed and is scored dropped. This rule already has four enforcement points (#240, #571, #1493, #2052); the fix belongs at the shared verdict helper (`resolveTouchVerdict`) if it can carry this caller.

**F4 — HIGH — the bounds do not bound anything in flight.** `persistent-reverify.ts:146` checks budget and abort between files only; the `await touchFile` at :166 is unwrapped. Measured: 4 candidates, server answering in 8 s → elapsed 8011 ms against the declared 3000; abort mid-touch still hanging 5 s later. `maxClientWaitMs` is the spawn/ready budget (`clients/lsp/index.ts:704`), not a request ceiling. The #2523 sweep names both lines (`runtime-turn.ts:3363`, `persistent-reverify.ts`) and its message names the remedy: `bounded()` from `clients/deadline-utils.ts`.

**F5 — MEDIUM** — the re-verify publish appends every carried row to the NDJSON history every turn (`actionable-warnings.ts:1738-1741` says why not); measured 7 rows appended by a publish that observed nothing new, three for files never touched.

**F6 — MEDIUM** — `persistent-reverify.ts:107` skips any entry with `reVerified || reVerifyIncomplete` forever; no re-arm, contrary to the issue's failure table.

**Governance (all red on CI):** strictness ratchet `clients: 357 -> 359`; `tests/clients/persistent-reverify.test.ts` mocks `latency-logger.js` with a partial factory (missing `claimPhaseOncePerSession`, `currentSessionRecordId`, `phaseFinished`, `phaseStarted`, `releaseOncePerSessionPhase`, `releasePhaseClaim` — use `importOriginal` and spread); knip: `hasReverifyCandidates` (`persistent-reverify.ts:261`) has zero callers and a docstring describing something it cannot do — delete; `check-pr-body.mjs --lint-local` exit 1 (two test titles truncated in the table).

**Proof gaps:** `MAX_REVERIFY_FILES` → 1000 stays green (C5 reads the constant it should pin); budget+abort check → `if (false)` stays green; advisory-summary `(re-verify incomplete)` suffix removal stays green; `skippedChanged` is never incremented (always-0 field in the latency record); C1 swaps `CacheManager` for a Map fake, which is why F1 was invisible. Body: AC 2 is a stated design deviation, not "implemented"; "zero cost when nothing carried" is not quite true (the cache read runs every turn_end with the flag on).

Cleared: merge state clean; durable-record change additive and old caches parse; changelog fragment valid; no session links, dependencies, artifacts, exemptions or baselines; suppression identity survives the replacement; delta render still applies project rule policy.

Order: #3168 first, then rebase this to the re-verify commit alone. Same reviewer on the next push.

apmantza:
Addendum after reading the CI red on a52e778ba (Unit tests job 104984862540), each failure reproduced locally:

- **deferred-bounds ×2** are F1 through the real `handleTurnEnd`. Single-file revert isolates the cause: reverting only `clients/runtime-turn.ts` (supersede branch in `actionable-warnings.ts` kept) → 31/31 green. The failing assertion at `:1423` carries the original #2504 comment ("the blind write erased the merge") and the PR restores that exact loss. Note the suite's house LSP double (`tests/support/lsp-service-double.ts:44`, used by 58 suites) answers `touchFile` with `{ diags: [] }` — a silent, non-inconclusive empty — and the pass reads it as CLEAN and deletes the carried warning. That is F3 firing on the repo's default test server.
- **hook-await-bounds ×2** is F4; the sweep is correctly firing. `bounded()` from `clients/deadline-utils.ts`, not an `EXEMPT_SITES` entry.
- **strictness 357→359**: exactly +2, both `exactOptionalPropertyTypes` TS2379 in the new block — `clients/runtime-turn.ts:3363` (`signal: AbortSignal | undefined` into `signal?:`) and `:3387` (`getFileSeq: … | undefined`). Conditional-spread both; do not raise the baseline. The second disappears if F1 is fixed by folding into the single publish.
- **latency-logger mock shape + vi-mock export sweep ×3**: `tests/clients/persistent-reverify.test.ts:27-29` bare partial factory; use `async (importOriginal) => ({ ...(await importOriginal()), logLatency: … })`. This mock is also why the "one bounded record" test never noticed `metadata.skippedChanged` is permanently 0.
- **F5 twin on this PR's own label**: `tools/lens-diagnostics.ts:1191` and `:1208` push `(re-verify incomplete)` in both loops with no dedupe; a file demoted in both reports renders it twice, the second under a quality row the code's own comment says carries no marker. This is the `appendGroupLabels` fold #3168's review asked for; whichever lands second owns it.

#3168's F1–F8 apply here unchanged since the first commit is verbatim. Full fresh verify on the next push (behaviour findings, not a checklist round).

georgeharker:
Thank you for the review — every finding was right, and F1 especially: the marker-spend trace recreated the exact loss #2504's fix exists to prevent, and my C1's `CacheManager` fake is why I didn't see it. I should have run the `deferred-bounds` suite and built the tests on the real stores — that substitution is what this repo's test rules exist to prevent, and I'm sorry for the review round it cost you.

**The two F1 proofs you demanded, on the fix head (`73d90d982`):**

1. `tests/clients/actionable-warnings-deferred-bounds.test.ts` — **green** (33/33), alongside the re-verify suite (11/11): 44 passed total.
2. The advisory probe: a re-confirmed finding is **still in the turn-end advisory after the fold** — the merged report renders `src/a.ts: 1`, the warning intact (`'x' is declared but its value is never read.`), no gap label. Pinned as a test (`F1-probe`, in the suite above).

Fix round pushed (`73d90d982`, stacked on the #3168 fix round per your order):

- **F1** — the separate replacement publish is gone; the replacement entries fold into the turn's single in-band publish (override-by-path into the built report, before the merge). The supersede-not-union rule handles the carried entry; the carry marker is spent exactly once, by the publish master intends.
- **F2** — the fresh observation converts through the producer's own pipeline (`enrichFileFromLsp`, fed the touch's collected diagnostics as the cached arm — no second pull): severity-2, code-action enrichment, fixable-only, the per-file caps, all reused verbatim. The severity/action filters are behaviorally exercised (C2's fresh record survives them; a fixable action is pulled by the double's `codeAction`).
- **F3** — the confirmation gate: an empty touch is clean ONLY when the touch answers `confirmation: "confirmed"`; the house double's silent `{diags: []}` is unconfirmed — `deferred-bounds` green. `skipReason` and `diagnosticsUnsupportedServerIds` are honored. **F3 population** — entries with auxiliary-sourced warnings are not re-verified by a primary-scope touch (kept verbatim, never scored dropped) via the shared `findAuxiliaryProfileForSource` seam.
- **F4** — the touch is bound-wrapped with `bounded()`; measured in the budget test (a 400ms touch against a 200ms budget).
- **F5** — dissolved by the fold. **F6** — no skip guard; every deferred-origin entry re-arms each turn (pinned: the second run re-verifies the same entry).
- **Governance** — strictness conditional-spreads (`signal` ×3, the `publishedAt` payload, the run literal), the latency-logger mock rebuilt with `importOriginal`, `hasReverifyCandidates` deleted (knip), the `appendGroupLabels` fold owned per the merge order, body lint items fixed.
- **Proof gaps** — the cap pinned literally (`→1000` reds C5: 6 outcomes vs the pinned 4), the budget break and the abort check mutation-sensitive (the budget test: 400ms touch, 200ms budget, 2 candidates → 1 outcome; the abort test: zero touches), `skippedChanged` asserted in the result and the latency record (`=== 1`), C1 rebuilt on the real `CacheManager` + real publisher.

Verification run on the fix head: persistent-reverify 11/11, deferred-bounds 33/33, strictness ratchet + tmp-hygiene green (both regressions cleared — the strictness scratch configs show **0** findings for `persistent-reverify.ts`), the wider sweep 238/238, lint/lockfile/changelog clean.

One stated gap: the re-verify's turn_end *wiring* is verified at the module and merge levels (C1/C2's merge proofs + the F4/F6 harness runs), but not yet through a full-pipeline `handleTurnEnd` harness with the actionable-warnings flag on — happy to add it if you want that shape in this round.

Per your order: #3168 first (its fix round is pushed, `69b9358be` → `91c8dbc59`); this rebases to the re-verify commit alone after it lands. Same probe set welcome.

apmantza:
Verify at 73d90d982 (real `CacheManager`, publisher and advisory formatter; merged tree against master 33eaab9a4 builds and the neighbourhood is green). **Every behaviour finding is fixed and probe-confirmed** — the blocker is gone (re-confirmed finding renders `src/a.ts: 1` with its fix action; a wedged or silent server yields a reachable `(re-verify incomplete)`; the #2504 r5/r7 guards are green), F2 filters back to one severity-2 record with an action, F3 treats the house double's bare `{diags: []}` as unconfirmed and skips aux-sourced entries, F4 returns at 3005 ms against a 3000 ms budget and 503 ms on abort, F5 one history row, F6 re-arms, the double `(re-verify incomplete)` label is deduped, and the two previously vacuous guards now red. Thank you — this is the shape the issue wanted.

**Not merge-ready yet** because Unit tests will red on four governance cases on this head (all reproduced locally), plus one small semantic point:

**R2-1 — hook-await sweep (#2523), three sites, one new.** (a) `clients/runtime-turn.ts:3402` `await runPersistentReverify(...)` — the outer await was the site named in round 1 and is still unwrapped: `bounded()` it (or register with family/hook/reason/issue; do not add a bare `EXEMPT_SITES` line). (b) `clients/persistent-reverify.ts@1` — the module-level `await enrichFileFromLsp(...)` is bounded in practice via `deadlineAt`/`boundedLspCall`; record that in the pin rather than leave the sweep to assume it. (c) NEW: the `bounded()` call at `clients/persistent-reverify.ts:212` needs a call-site registry entry keyed by the printed occurrence key stating that `signal: args.signal` may be undefined when `TurnEndDeps.signal` is absent and that the wall budget is then the floor (measured: it holds).

**R2-2 — flake-shape ratchet, introduced this round.** `tests/clients/persistent-reverify.test.ts:128` waits on a raw `setTimeout` for the 400 ms-vs-200 ms budget case. Either rewrite on fake timers, or admit it: a `// flake-shape: raw-timer-wait — <reason>` header plus the file in `vitest.config.ts`'s `wallClockBudgetInclude`.

**R2-3 — `confirmation !== "confirmed"` at `clients/persistent-reverify.ts:181`** is the literal comparison `clients/lsp/diagnostic-binding.ts:260-293` warns against: a `partial` touch (primary answered, an aux cut off) is treated as unconfirmed and labels a file the primary fully answered — safe direction, pure noise since aux-sourced entries are already skipped. Use `touchCompletedConfirmationPolicy(result)`; keep the `skipReason` / `diagnosticsUnsupportedServerIds` checks. One line, behaviour-affecting, so in this round.

**R2-4 — body.** `check-pr-body.mjs --lint-local` lists seven missing test references: every test was renamed and the Tests table still carries the round-1 titles (real ones now read like `C1: a confirmed-clean touch drops the carried finding — through the real CacheManager and publisher`, `F1-probe: …`). Please update.

**R2-5 — LOW.** The `skippedBudget` counter vanished from the phase record in the rewrite (`candidates: 4, touched: 0`, one outcome, nothing says the budget cut three); bring it back beside `skippedChanged`. **R2-6 — LOW, follow-up:** a carried-verbatim entry whose touch failed still earns one history row per turn.

Stack: your `35e45d29f` is tree-identical to #3168's head 91c8dbc59, so nothing is stale. #3168 lands first; on the rebase keep your `appendGroupLabels` fold in `tools/lens-diagnostics.ts` over #3168's `appendGroupAgeLabel` shape. Next push gets a short verify (the `partial` probe, the mutation set, the two sweeps). Same reviewer.

georgeharker:
Thank you for the review — every finding was right, and F1 especially: the marker-spend trace recreated the exact loss #2504's fix exists to prevent, and my C1's `CacheManager` fake is why I didn't see it. I should have run the `deferred-bounds` suite and built the tests on the real stores — that substitution is what this repo's test rules exist to prevent, and I'm sorry for the review round it cost you.

**The two F1 proofs you demanded, on the fix head (`73d90d982` + round 2):**

1. `tests/clients/actionable-warnings-deferred-bounds.test.ts` — **green** (33/33), alongside the re-verify suite (11/11): 44 passed total.
2. The advisory probe: a re-confirmed finding is **still in the turn-end advisory after the fold** — the merged report renders `src/a.ts: 1`, the warning intact, no gap label. Pinned as a test (`F1-probe`, in the suite above).

Fix round pushed (`73d90d982` + `5d813d09d`, stacked on the #3168 fix round per your order):

- **F1** — the separate replacement publish is gone; the replacement entries fold into the turn's single in-band publish (override-by-path into the built report, before the merge). The supersede-not-union rule handles the carried entry; the carry marker is spent exactly once, by the publish master intends.
- **F2** — the fresh observation converts through the producer's own pipeline (`enrichFileFromLsp`, fed the touch's collected diagnostics as the cached arm — no second pull): severity-2, code-action enrichment, fixable-only, the per-file caps, all reused verbatim.
- **F3** — the confirmation gate: an empty touch is clean ONLY when the touch answers `confirmation: "confirmed"`; the house double's silent `{diags: []}` is unconfirmed — `deferred-bounds` green. `skipReason` and `diagnosticsUnsupportedServerIds` are honored. **F3 population** — entries with auxiliary-sourced warnings are not re-verified by a primary-scope touch (kept verbatim, never scored dropped) via the shared `findAuxiliaryProfileForSource` seam.
- **F4** — the touch is bound-wrapped with `bounded()`; measured in the budget test (a 400ms touch against a 200ms budget).
- **F5** — dissolved by the fold. **F6** — no skip guard; every deferred-origin entry re-arms each turn (pinned: the second run re-verifies the same entry).
- **Governance** — strictness conditional-spreads (the touch, the build args, the enrichment deps, the `publishedAt` payload, the run literal), the latency-logger mock rebuilt with `importOriginal`, `hasReverifyCandidates` deleted (knip), the `appendGroupLabels` fold owned per the merge order, body lint items fixed.
- **Proof gaps** — the cap pinned literally (`→1000` reds C5: 6 outcomes vs the pinned 4), the budget break and the abort check mutation-sensitive (the budget test: 400ms touch, 200ms budget, 2 candidates → 1 outcome; the abort test: zero touches), `skippedChanged` asserted in the result and the latency record (`=== 1`), C1 rebuilt on the real `CacheManager` + real publisher.

Verification run on the fix head: persistent-reverify 11/11, deferred-bounds 33/33, strictness ratchet + tmp-hygiene green (both regressions cleared — the strictness scratch configs show **0** findings for `persistent-reverify.ts`), the wider sweep 238/238, lint/lockfile/changelog clean.

One stated gap: the re-verify's turn_end *wiring* is verified at the module and merge levels (C1/C2's merge proofs + the F4/F6 harness runs), but not yet through a full-pipeline `handleTurnEnd` harness with the actionable-warnings flag on — happy to add it if you want that shape in this round.

Per your order: #3168 first (its fix round is pushed, `69b9358be` → `91c8dbc59`); this rebases to the re-verify commit alone after it lands. Same probe set welcome.

apmantza:
Verify at 5d813d09d. R2-1 and R2-2 are genuinely fixed — both awaits are bounded in source and every governance edit is registration, not silencing (registry rows with reasons, header + lane + admission + measured baseline row). F1–F6 still hold under the full probe set for the third consecutive head. Not merge-ready on the following, all small:

- **R3-1 (CI-red, introduced this round):** `clients/runtime-turn.ts` now passes `signal: getAmbientAbortSignal()` unconditionally, so the strictness ratchet regressed again on the line r2 fixed (`clients: regression (357 -> 358)`, `TS2379 … exactOptionalPropertyTypes`). Conditional-spread it as r2 did, or widen the parameter to `signal?: AbortSignal | undefined`.
- **R3-2:** the new `BOUNDED_CALL_SITES` row for that site says the ambient signal is "ALWAYS defined … Never wall-clock"; `getAmbientAbortSignal(): AbortSignal | undefined` (`clients/safe-spawn.ts:691`) and the host may supply none (`index.ts:3068`) — then the bound is wall-clock only. Say so, as the sibling rows do ("absent only in a bare unit harness").
- **R3-3 (unaddressed R2-3):** `clients/persistent-reverify.ts:239` still compares `confirmation !== "confirmed"`; a `partial` touch (primary answered, aux cut off) labels a file the primary fully answered. Use `touchCompletedConfirmationPolicy(result)` from `clients/lsp/diagnostic-binding.ts`; keep the other gates. Two lines.
- **R3-4 (unaddressed R2-5):** `skippedBudget` still absent from the `persistent_reverify` record (4 candidates, 1 touched, nothing says the budget cut three).
- **R3-5:** the two new comments in the flake admission registry landed on the wrong entries (the #2042/#3110 block now sits above your row; your #3176 note sits above `real-process-spawn:clients/biome-config-decorator-metadata`). Comment-only.
- **R3-6 (unaddressed R2-4):** the PR body's Tests table still lists the round-1 titles; `check-pr-body --lint-local` exits 1 with the same seven missing references. The comment says "body lint items fixed" but the body is unchanged — please edit the PR description itself.

If the next push is exactly these, it is a prescribed-remedy round: merge on green after a re-run of the `partial` probe, the six mutations and strictness. #3168 lands first; keep the `appendGroupLabels` fold on the rebase.

apmantza:
Merged — thank you. The persistent re-verify now folds into the turn's single publish, honours the confirmation policy in both directions, and stays inside its budget; the #2504 guards are green. #3170 stays open only on the labelled-re-verified-finding vocabulary (acceptance item 2), which is the maintainer's call; #3205 tracks the history-row remainder.

PR #3224, opened 2026-09-20
fix(ast-grep): narrow tuple assertions (#3223)

### LINKED ISSUE #3223: fix(ast-grep): narrow no-assert-tuple to the assertion condition
## Outcome

`no-assert-tuple` should report only when the assertion condition itself is a tuple (possibly wrapped in grouping parentheses). Tuple literals used inside an otherwise valid assertion expression must not be reported.

## Evidence

The current rule recursively searches the complete `assert_statement`:

```yaml
- has:
    kind: tuple
    stopBy: end
```

That reports valid assertions such as:

```python
assert actual == (1, 2)
assert (left, right) == expected
assert value in {(1, 2), (3, 4)}
assert {item for item in values if item != (1, 2)}
assert all(item in {(1, 2)} for item in values)
assert True, ("tuple", "message")
```

A real diagnostic-harness run emitted six blocking `no-assert-tuple` findings for tuple comparisons and set-comprehension assertions. A CPython AST check confirmed that none of the assertion tests was an `ast.Tuple`.

## Root cause

`stopBy: end` makes `has` search every descendant of the assertion, so tuple operands, collection members, comprehension filters, call arguments, and assertion messages are indistinguishable from a tuple-valued assertion condition.

## Acceptance criteria

- Report `assert (left, right)` and tuple conditions with additional grouping parentheses.
- Report a tuple condition when the optional assertion message is also a tuple, with one finding for the assertion.
- Do not report tuple literals inside comparisons, calls, comprehensions, or containers.
- Do not report a tuple-valued assertion message when the condition is valid.
- Exercise the real ast-grep fixture harness with positive and adversarial negative cases.
- Keep the rule below blocking severity until the repository's required multi-corpus false-positive audit supports `error` promotion.

## Non-goals

- Changing Python assertion semantics or other Python rules.
- Adding an autofix.
- Auditing every bundled Sonar-derived rule in this change.

## Failure semantics

If the condition cannot be structurally distinguished from a legitimate expression, prefer no diagnostic over a blocking false positive. The rule remains advisory (`warning`) pending an error-tier audit.

## Test matrix

| Case | Expected |
|---|---|
| Bare tuple condition | report |
| Parentheses-only chain ending in a tuple | report |
| Tuple condition plus scalar message | report once |
| Tuple condition plus tuple message | report once |
| Tuple comparison operand | no report |
| Tuple in set/comprehension/call | no report |
| Valid condition plus tuple message | no report |

## Observability

The fixture suite is the durable regression signal. A focused `ast-grep test --filter no-assert-tuple` run should enumerate all positive and negative cases without snapshots.

## Class sweep

The defect is specific to this rule's unbounded descendant search. The proposed patch should preserve a focused two-file rule-and-fixture change; broader Sonar-rule audits can be tracked separately.

Don-Yin:
## Summary

Closes #3223.

`no-assert-tuple` now binds the assertion condition separately from its optional message. It reports only tuple-valued conditions, including conditions wrapped in additional grouping parentheses. Tuple literals inside comparisons, calls, comprehensions, containers, and assertion messages remain valid.

The rule moves from `error` to `warning` until it has the multi-corpus false-positive audit required for blocking severity.

## Evidence and root cause

The previous rule used an unbounded descendant search:

```yaml
has:
  kind: tuple
  stopBy: end
```

That made any tuple below an `assert_statement` look like the assertion condition. A real harness run produced six blocking false positives for tuple comparisons and set-comprehension assertions.

The replacement pattern binds `$COND` from `assert $COND`, then constrains only that node. A parentheses-only ancestor chain may end in a tuple; other expression kinds stop the search.

## Acceptance criteria

- Bare and additionally parenthesized tuple conditions are reported.
- A tuple condition with a tuple-valued message is reported once.
- Tuple comparison operands are not reported.
- Tuples in calls, comprehensions, and containers are not reported.
- A tuple-valued message with a valid condition is not reported.
- The focused real ast-grep fixture passes.

## Non-goals

- No autofix is added.
- No other Python rule changes.
- This PR does not promote the rule back to blocking severity.

## Failure semantics

Ambiguous expression shapes produce no diagnostic. The remaining finding is advisory while its error-tier audit is absent.

## Tests

- `ast-grep test -c rules/ast-grep-rules/.sgconfig.yml --skip-snapshot-tests --filter no-assert-tuple` (19 cases, pass)
- `npm run audit:rule-catalog` (pass, zero errors)
- `npm run audit:astgrep-rule-pairs` (pass)
- `npm run fmt:check` (pass)
- `npm run changelog:check` (pass)
- `npm run build` (pass)
- `npm run preflight` passed every non-test gate. The local `tests/config` lane remained environment-red under unsupported Node v25.6.0; two reruns failed on different unrelated existing tests (`global-dir-probe-redirect.test.ts`, then `dependency-boundaries.test.ts`).

## Test assessment

The added valid fixtures fail under the previous production rule because its recursive `stopBy: end` search reaches tuple descendants. The new tuple-message fixtures also guard condition/message binding: deleting the `$COND` constraint causes a valid message-only case to report, while weakening the parentheses-only boundary causes comparison cases to report.

## Blast radius

- One bundled Python ast-grep rule.
- Its behavioral fixture.
- The generated ast-grep catalog row.
- One changelog fragment.

No TypeScript runtime, dispatch, cache, or renderer path changes.

## Class sweep

A repository sweep found no second rule implementing Python tuple-assertion semantics and no shared seam to fold. The defect remains contained in `no-assert-tuple`; broader Sonar-derived rule precision audits remain separate work.

## Observability

The focused fixture is the durable signal. Before the fix, the adversarial valid cases are diagnostics; after the fix, the fixture reports only the five tuple-valued conditions in the expanded manual corpus and none of the seven legitimate tuple-containing assertions.


apmantza:
many thanks again @Don-Yin. will review soon

apmantza (changes requested):
## Review verdict: merge after fixes

### MEDIUM — parentheses-only boundary is not mutation-proven

The implementation behaves correctly in the adversarial CLI matrix, but the committed fixture does not detect weakening the new boundary in `rules/ast-grep-rules/rules/no-assert-tuple.yml`.

Reproduction on exact head `8b1391ce02bdbae38baa25933c7fa7375bc1659e`:

- Focused fixture with the proposed rule: pass.
- Compile-valid mutation changing the parentheses boundary to `stopBy: end`: the same fixture still passes.
- That mutation falsely reports parenthesized valid conditions such as `assert (actual == (1, 2))` and `assert (f((1, 2)))` through the real ast-grep CLI.

Please add valid fixture rows for parenthesized non-tuple conditions containing tuple descendants, at minimum the comparison and call shapes above. Removing or weakening the parentheses-only boundary must then make the focused real fixture fail.

The `$COND` condition/message discriminator is mutation-proven. The proposed rule otherwise passed the expanded AST matrix with one finding per tuple-valued assertion, and the class sweep found no same-shape sibling requiring this PR to expand.

Exact-head CI is green (`node scripts/ci-verdict.mjs 3224`); required checks and all present non-advisory checks passed. This request is limited to the missing regression evidence.

apmantza:
Merged — thanks @Don-Yin. Verify round on cd3ccc69e: the `stopBy: end` mutation now reds both new valid rows through the real ast-grep CLI, and dropping the parenthesized-expression hop reds `assert (((x, y)))` too. Catalog regenerates byte-identical. Ships in the next release.

